PR #565, opened 2015-02-02
#564 Swap position of From and To on arrow click

shashank19gaurav:
I have implemented this request. Go through it and let me know if any change is required.


trang:
@shashank19gaurav, you will have to create a new pull request because there are a few things to review.

I see that one of the commits in your pull request says "Merge pull request #1 from Tatoeba/dev...". I suppose you tried to sync your fork via a pull request. It's not exactly how you do it. You can perhaps read about how to keep your fork up-to-date in one of these links (I'm not sure which one will explain it the best for you):
* https://help.github.com/articles/syncing-a-fork/
* https://gun.io/blog/how-to-github-fork-branch-and-pull-request/
* http://www.developer.com/open/making-pull-requests-in-github.html

Your second commit include changes that are not related to the issue, and that should not be done (you removed the line that includes the CSS `layouts/elements.css`).

And last, there is a simpler way to solve the issue than the way you implemented it. I'd suggest you search about how to read a value from a select and how to set a value on a select with jQuery.

We can discuss all of this more in details this weekend on IRC if you need to. Meanwhile I will close this pull request.

PR #583, opened 2015-02-09
Make the translations collapsible

wallflower1:
sentence lists are now collapsible
only 5 translations show by default
"Show Translations/Hide Translations" toggle added


app/webroot/js/sentences.collapse.js
```diff
@@ -0,0 +1,28 @@
+/*
+    Tatoeba Project, free collaborative creation of multilingual corpuses project
+    Copyright (C) 2009 Allan SIMON <allan.simon@supinfo.com>
```
allan-simon: I didn't created that file :)

app/webroot/js/sentences.collapse.js
```diff
@@ -0,0 +1,28 @@
...
+   if (document.getElementById('translation-show-'+id).style.display != 'none') {
+     document.getElementById('translation-show-'+id).style.display = 'none';
+     document.getElementById('translation-'+id).style.display = 'block';
+  }
+  else {
+     document.getElementById('translation-show-'+id).style.display = 'inline';
+     document.getElementById('translation-'+id).style.display = 'none';
+  }
+}
+
```
allan-simon: try to limit the scope of javascript by
```
var MyModule = (function() {

var plop = function(id) {
    // your code here 
};
return { plop: plop} 
}());
```
so that you can latter use it like this  MyModule.plop(id)

app/views/helpers/sentences.php
```diff
@@ -147,7 +193,7 @@ public function displayTranslations($id, $translations, $indirectTranslations, $
                     $langFilter
                 );
             }
-
+            echo '<p><a id="translation-hide-'.$id.'" class="hideLink" onclick="toggle_visibility('.$id.');">&#x2191; Less Translations</a></p></div>';
```
allan-simon: Less Translations is not part of translated strings

app/views/helpers/sentences.php
```diff
@@ -133,9 +138,50 @@ public function displayTranslations($id, $translations, $indirectTranslations, $
...
+                        $langFilter
+                    );
+                    $visible++;
+                }
+            }
+            echo '<a id="translation-show-'.$id.'" class="showLink" onclick="toggle_visibility('.$id.');">&#x2193; More Translations</a></div>';
+            //expanded list of translations
+            echo '<div id="translation-'.$id.'" class="more">';
+            $keys=array_keys($translations);
+            for ($i=5; $i < count($keys); $i++) {
```
allan-simon: put 5 as a constant so that you dont have to repeat this magic number everywhere

app/webroot/js/sentences.collapse.js
wallflower1: i created that so i could simply include it on all the required pages. I am getting used to the code structure, so i figured it should work. about the scope, do i change the js file and push the changes ?

wallflower1: :P i just realised. copy-paste mishaps.

trang:
@wallflower1, we use jQuery so you could use the [id selector](http://api.jquery.com/id-selector/) instead of `document.getElementById()`.
But if you could find a solution that doesn't involve passing on the id, it would be better :)

For hiding and showing elements, you have `.hide()` and `.show()`: http://api.jquery.com/category/effects/basics/.

wallflower1:
@trang There is another solution I found here : http://jsfiddle.net/cse_tushar/6FzSb/ 
Looks much cleaner, can be implemented without making changes to for loops in sentences.php, but without passing ids it will create issue on sentences/search page. Do you think its possible?

trang:
@wallflower1, what you did in 2ce38e5 is good. No need for now to break into several steps the display of the extra translations.

I have a new more feedbacks.

1. The link is displayed outside of the sentence's block on the homepage and the sentences details. For instance here: http://prntscr.com/65gtvl, you see that the link is below the line. It would be better if it was displayed above the line, to be part of the sentences block.

2. I'm wondering if it might be better to display the link on the left rather than on the right side. It feels a bit difficult to notice it on pages with several sentences: http://prntscr.com/65gys4

3. Last thing, more important: the link is displayed even when the sentence has less than 5 translations. Make sure to display it only when there are more than 5 translations.

wallflower1:
@trang  i tried placing the link to the left but it looked even more stuffed to me. here : http://i.imgur.com/KSMO7e6.png , what do you think?

trang:
I've posted a message on the Wall to ask users for feedback: http://tatoeba.org/eng/wall/show_message/21843#message_21843
We'll see what users have to say about it.

I'm personally equally unsatisfied with the left and right position. Maybe at the center... 
Just try several positions and just place it where you like it the most for now. It's usually hard to guess what works the best until a feature is actually being used in prod.

wallflower1:
here is the center : http://i.imgur.com/Okt397u.png , http://i.imgur.com/1PzGaI5.png
I personally find right position better of the lot, but tell me if you like this one. Or i could make it box-like , similar to "previous" ,"next", "random". Meanwhile, i could work on some other issue. Do assign me if you find anything suitable.

trang:
Sabretou reported that he didn't notice the link on the screenshot I posted (where the link is positioned on the right side). He suggested to make the link stand out more, and make it look like a button.
http://tatoeba.org/eng/wall/show_message/21847#message_21847

I'm not sure about using the same green color as the submit buttons. 
On the Wall I used a blue color for the "show/hide replies": http://prntscr.com/65ukao. 
Perhaps you can try using these colors for the "more/less translations" links.
It might stand out a bit too much though... But I guess it's better that it stands out rather than goes unoticed. We can tune out the color later on.

And another thing I thought of: it would be nice to have the number of translations hidden.
--> "More translations (7)" // when there are 7 translations hidden

wallflower1:
I tried several colors, (http://i.imgur.com/Cdk9jrx.png) this one looks rad.  Also added count of hidden sentences.

trang:
Okay, I think there is no way we can add this button without redesigning a bit more the sentences. The block needs borders, otherwise it looks like the buttons are floating in between groups of sentences. But that won't be in the scope of this issue.

-----------------

First, there are 2 bugs I have detected.

**Bug 1**
* Go to "Browse" > "Browse by language".
* Expand the translations for several blocks of sentences (http://prntscr.com/66i4mw)
* Then collapse one group of translations.

--> It's supposed to display back the "More translations" only for the block of sentences you've selected, but it displays the button for every block on the page (http://prntscr.com/66i4ya).

**Bug 2**
* Go to the details page of sentence `#615808`.
* You have only 3 translations displayed, and 8 hidden (http://prntscr.com/66i5sf).

--> The number of translations displayed should be 5 when there are more than 5 translations.

-----------------

Then there is a small thing you need to fix in the string inside your `__()` function. You need to use a placeholder for your variable.

Here's an example of how to do it: 
https://github.com/Tatoeba/tatoeba2/blob/dev/app/views/sentences/show.ctp#L167
This code display this: http://prntscr.com/66i9ru.

-----------------

And last, since this is something that I'd like users to start being able to test in real situations so that they can give more feedback, we're gonna make it an option at first. But that's probably something I'm going to do after I've pulled your code because I'd like to have this released this weekend.

So just take care of fixing the 2 bugs and the locale string.

jiru:
> Then there is a small thing you need to fix in the string inside your __() function. You need to use a placeholder for your variable.

We’d better take care of plurals with `__n()`. This would allow you to use a real sentence like “Show 1 more translation” or “Show 3 more translations”. Here is an example of usage:
https://github.com/Tatoeba/tatoeba2/blob/b60bfc22685d84471587ca4a9b18b1ca7f8a26eb/app/views/private_messages/write.ctp#L43-L50

wallflower1:
Fixed the bugs you asked me to. Let me know if there are any issues with the code.

jiru:
The style of your code, especially in `displayTranslations()`, is inconsistent with the rest of the file. Please add spaces around equal signs, space after commas, split long lines etc. to make the style consistent over the file.

Don’t include magic numbers nobody except you understand.
```php
/* This is bad because everyone who’s gonna read that code
 * needs to lookup 2193 in order to understand that code. */
echo "&#x2193; More Translations";

/* Anyone understand this right away */
$downArrow = "&#x2193;"
echo "$downArrow More Translations";

/* You may also do this, which is both easy to understand and simple */
echo "↓ More Translations";
```
Choose more meaningful names for your variables.
```php
/* This is hard to understand */
$num = 5; // Okay, giving the name of this variable, I know it contains a number.
          // But I have no idea what it’s about and why we assign it five…

/* This is easier to understand */
$initiallyDisplayedTranslations = 5; // Now I got it!
```

I think the string `'More Translation (1)'` is not correct because:
  1. You should keep the word `translations` plural regardless of the number, since it’s prefixed by “more”.
  2. Remove the upper cased T in “translation”.

Actually, I think it would be better with a real imperative sentence that includes a verb, like “Show x more translations”, since that link is an action, like a button. I’m not a native English speaker though.

Don’t echo raw HTML because it’s error-prone. Use the [Html helper](http://book.cakephp.org/1.3/en/The-Manual/Core-Helpers/HTML.html) functions like `link()` or `tag()`.
```php
/* Instead of this: */
echo '<div class="blah"><a class="fooClass" href="/there">' . __('Click here', true) . '</a></div>';
/* Write this: */
echo $this->Html->tag(
    'div',
    $this->Html->link(
        __('Click here', true),
        '/there',
        array('class' => 'fooClass')
    ),
    array(
        'class' => 'blah',
        'escape' => false,
    )
);
```

Apart from that, the code you added in `displayTranslations()` is overly complex, and thus hard to understand and hard to maintain. I think it’s possible to make it simpler using the following approach:
  1. Join $translations and $indirectTranslations into a single array, let’s call it $allTranslations.
  2. Split $allTranslations into two arrays: the initially displayed translations and the ones initially hidden.
  3. Loop over the first array to display the initially displayed translations.
  4. If there are translations left in the second array, display the link and the other translations inside the div.

Finally, I think you should handle the case where only a single translation is hidden. In that case, it would be better to display that translation anyway since the “More translations” link takes about the size of that hidden translation. In other words, never hide only one translation, only two or more.

wallflower1:
@jiru please take look at sentences.php , tell me if it still needs improvement. Although, i did change it according to your suggestions. (been sick, hence the delay)
I changed 'More translations (x)' to 'Show x more translations', it does seem more sensible, but I am not a native English speaker either. I guess users/devs can give their feedback on this.

wallflower1:
I used links earlier, but that is causing page reload even when i don't provide url or 'target', is there a way to disable that? I didn't find anything in the docs.

wallflower1:
Thanks for your feedback, I surely keep that in mind.

jiru:
> I used links earlier, but that is causing page reload even when i don't provide url or 'target', is there a way to disable that? I didn't find anything in the docs.

Yes. A few examples:
* [Put it in a div](https://github.com/Tatoeba/tatoeba2/blob/da489edb2dc4580dc8acc422beb5b2328431996c/app/views/helpers/sentence_buttons.php#L60) (I actually didn’t know we were doing this, hence my initial comment)
* [Make the onclick event return false](https://github.com/Tatoeba/tatoeba2/blob/ed7984524dff7acc80d23da5749340b99a60e0b2/app/views/elements/random_sentence_header.ctp#L58)
* Use a styled [`<button>`](http://css-tricks.com/use-button-element/)

wallflower1:
@jiru I have made the suitable changes like you suggested. About using 'link' helper for show/hide tags, i tried ur solutions , but neither will work.

1. putting it in a div doesnt prevent the page from reloading.
2. if i make the onclick event false, then the jquery wont work since its uses .click() function
3. button is not a htmlhelper, its a form helper in cakephp, so a form needs to be created to be able to use it.

Issue #580, opened 2015-02-08
Edit button to edit sentence


trang:
A user can currently edit a sentence by clicking on it (edit-in-place). We want to change this mechanism and have instead an edit button to open the form.

The problem with edit-in-place is that when a user simply wants to copy-paste the text of a sentence, if they have the permissions to edit the sentence, it will open the edit form as soon as they click on it. 

This affects especially corpus maintainers and admins because they can edit all the sentences.

* For the edit icon, we could use this one: https://www.iconfinder.com/icons/284087/edit_editor_pen_pencil_write_icon#size=128
* You can place the icon between the "translate" and "favorite" icon.
* The icon should not be displayed if the user is not the owner of the sentence, or is not a corpus maintainer, or is not an admin.
* In the cases where the button is displayed, it needs to be disabled if the sentence has audio.
* It's also a good occasion to improve the look of the edit form (currently looks like this: http://prntscr.com/62s447).

kwikadi:
I tried to replicate the problem, but I wasn't able to, with admin/123 as credentials. I'm on windows, with the 2010 database. Do I need some other credentials to replicate it?

jiru:
@kwikadi The password is 123456.

trang:
In my database the password is 123, but you have to log in with one of the
admins listed on the Members page (admin is not an existing user).

Actually with my username (trang) the password is 1234.

kwikadi:
I fixed most of the issue, but found out that apparently, (http://stackoverflow.com/questions/3771169/in-cakephp-why-would-this-paramsform-be-empty-if-ive-just-posted-a-form) if I use a form helper, data will be posted to $this->params['data'] while /sentences/edit_sentence checks $this->params ['form'].
(https://github.com/Tatoeba/tatoeba2/blob/e09aaa79298dbeee440be8e1e0cd1d9c26995c12/app/controllers/sentences_controller.php#L351). A quick grep reveals that this method is referenced in few other places, most of them from inside the 'tests' directory. https://github.com/Tatoeba/tatoeba2/search?utf8=%E2%9C%93&q=edit_sentence
I could simply change ['form'] to ['data'] inside edit_sentences(), but I'm not completely sure if it would break test builds/the database or not. I was hoping you could tell me where to go from here.

Also, is isTrusted() sufficient to display the edit button? ( I couldn't find any documentation on who passes the isTrusted() check)

jiru:
[...]

Issue #578, opened 2015-02-08
Message about noneditability of sentences with audio


liori:
The recently introduced blocking of editing sentences with audio is presented in a non-discoverable way: clicking on a sentence on a sentence details page just does nothing, instead of showing a text field. This made me think it's a bug and I started looking in the source code etc.

I propose to display a small message like "You cannot edit a sentence that has audio. Please contact admins if your intended edit will not change pronunciation of the sentece." instead, right below the sentence.

trang:
@kwikadi, I've assigned you this issue but we need to take care of #580 first because we want to remove the edit-in-place mechanism.

We can't solve this issue here simply by adding a warning message on click on the sentence because we cannot assume that the user clicked necessarily to edit. 
They could have clicked simply to copy-paste and it be confusing/annoying for the user to see a warning every time.

Once you've solved #580, we can discuss how we will display the information "You cannot edit a sentence that has audio...". It could be on click on the edit button, it could be on rollover.
Note that there's already something for the delete button: http://prntscr.com/62siwx. If we decide to implement it differently than how it's done for the delete button, we should also change the behavior on the delete button.

kwikadi:
I have, as of right now, added tooltips to the edit button that work identically to the Delete button ('Edit' when editing is possible, and 'You cannot edit this sentence because it has audio' in case of audio. The button is missing altogether if the user does not have permissions to edit the sentence)
Do you think a more visible form of communication is needed?

trang:
It's fine for now. I'm leaving this issue open until after the release though, in case users have some feedback on this. I've updated the dev website so that they can test the new features.

PR #586, opened 2015-02-11
Edit in place functionality changed. Fixes Issue #580

kwikadi:
I also added a tooltip to it, similar to that implemented in the delete button (with reference to #578 ), for the time being.


trang:
@kwikadi, I have tested your code and there are a few things to fix.

1. When clicking on "Submit" or when pressing enter, the edited sentence is not saved. I just get redirected to the homepage.

2. You don't want to reload the page on submit. You want to keep the same behavior as before:
    * show a loader while the sentence is being saved
    * once the sentence is saved, the load disappears and the new sentence is displayed in place of the old sentence.

3. Use a textarea rather than an input, to allow the sentence to be displayed in multiple lines. It makes it more comfortable to edit a long sentence. One thing to be careful about: a user can press enter in a textarea and it adds a new line, but we don't want that. We want that when the user presses "Enter" it submits the sentence. Cf. fb105909411f629de28544506a9aa5f666f6bbf6 where I had added a listener to submit the sentence on enter.

kwikadi:
Check now. I decided to modify the existing code rather than introducing all new code. (Also, I updated the branch, not sure if that makes the pull request unusable or something. If yes, reject the pull and let me know, I will send a fresh pull).

trang:
@kwikadi, yea you will need to submit a new pull request =/

For updating a branch that you are using as submission for a pull request you want to do a rebase, not a merge.

Issue #563, opened 2015-01-28
Enhance sentence translation form


jiru:
I can see quite a few problems with the current the sentence translation form (the one that appears when you click on the translate icon of a sentence).
* Its two fields (text and language) lack labels. I think it greatly confuses new users (who happen to translate one of the translations instead of the main sentence) and it’s a bad practise anyway since it prevent screenreaders to give information about the fields.
* It is in a very compact shape which makes it even more less usable, although I see no need to make it that small. Aligned fields and labels are way easier to read.
* The warning text is unlikely to be read (most users don’t read text anyway) and it blames users for our own inability to make a comprehensive form, which is very very bad in my opinion.
* We provide automatic language detection, but its lack of accuracy is a secret that is yet to be discovered by newcomers, by making mistakes. We should make it clearer to users what the actual autodetected (or preselected) language is, to allow them to fix it when it’s wrong. Some ideas:
 * put the language drop down first, so that it’s more visible
 * autodetect the language as you type and update the language field, unless manually selected by the user; maybe include the language name in the submit button: “Submit a X translation”
 * display flag along with language name

trang:
I agree with all your points.

trang:
#589 took care of most of the points. 

For the reference, feedbacks from users can be found here:
* http://tatoeba.org/eng/wall/show_message/21776#message_21776
* http://tatoeba.org/eng/wall/show_message/21863#message_21863

I'm not closing this issue yet because there are still a few things I want to do.

PR #589, opened 2015-02-14
Fix Issue #563

shashank19gaurav:


app/webroot/js/generic_functions.js
```diff
@@ -85,3 +91,20 @@ $(document).ready(function() {
...
+        if($(this).val()!='auto'){
+            var lang_flag_url = '/img/flags/'+$(this).val()+'.png';
+            $(this).next().attr('src',lang_flag_url);
+            $(this).next().attr('alt',$(this).val());
+        }else{
+            $(this).next().attr('src','/img/flags/unknown.png');
+            $(this).next().attr('alt','Unknown');
+        }
+    });
+});
```
jiru: I think this Javascript code should actually go into `app/webroot/js/sentences.add_translation.js` because it isn’t used in all the pages, only in pages that include translatable sentences.

jiru:
@shashank19gaurav Thank you for taking care of this! :-)

trang:
@shashank19gaurav I will take care of fixing the things that gillux mentioned. There are a few other things I want to do on this translation form and I prefer you to work on the other task I assigned you so that you can get familiar about other aspects of the code.

Issue #592, opened 2015-02-17
Wrong label on button for hidden comments on sentences


trang:
Cf. [CK's message](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/tatoebaproject/lYMZ0Ha95U0/YpT2newXrpUJ).

The label should be 'unhide', not 'hide'.

PR #603, opened 2015-03-02
corrected label on buttons for hidden comments #592

wallflower1:
Button now reads 'unhide' when comment is hidden.


wallflower1:
@jiru the label wont be translatable in that case. I am just trying to adhere to the coding style that is used elsewhere in the project. otherwise,  __($hiddenLinkText, true) was good to go too.

Issue #453, opened 2014-10-08
PHP notices on “translate sentences of” page


jiru:
[This page](http://localhost:8080/eng/activities/translate_sentences_of/inexistantusername) produces PHP notices:
```
Notice (8): Undefined index: count [CORE/cake/libs/view/helpers/paginator.php, line 543]
Notice (8): Undefined index: options [CORE/cake/libs/view/helpers/paginator.php, line 546]
Notice (8): Undefined index: count [CORE/cake/libs/view/helpers/paginator.php, line 547]
Notice (8): Undefined index: count [CORE/cake/libs/view/helpers/paginator.php, line 548]
Notice (8): Undefined index: page [CORE/cake/libs/view/helpers/paginator.php, line 564]
Notice (8): Undefined index: current [CORE/cake/libs/view/helpers/paginator.php, line 566]
Notice (8): Undefined index: count [CORE/cake/libs/view/helpers/paginator.php, line 567]
```

jiru:
The page now redirects to `/users/all` with a flash message.

PR #604, opened 2015-03-02
fixed PHP notices on "translate sentence of" page #453

wallflower1:
Page after fixing the bug: http://i.imgur.com/nWgU7rZ.png

### DIFF
--- a/app/views/activities/translate_sentences_of.ctp
+++ b/app/views/activities/translate_sentences_of.ctp
@@ -28,7 +28,7 @@ $username = Sanitize::paranoid($username, array("_"));
 
 if ($results == null) {
     $title = format(
-        __("This user doesn't exist: {username}", true),
+        __("This user doesn't exist: {username} (0 results)", true),
         array('username' => $username)
     );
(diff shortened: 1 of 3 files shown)


jiru:
Sorry, but we should refrain from modifying CakePHP as much as possible. The reason is that every modification needs to be manually reintroduced after each CakePHP update. Not that we’re actually updating it frequently, but it’s bad practise in general and should be avoided unless there is no other solution at all.

wallflower1:
I get your point. I do have another fix for this which involves modifying the pages controller. I just thought this one was neater.

trang:
@wallflower1, I think a better way to deal with this problem is to redirect the user to the "Members" page with a flash message "No user with this username: {username}".
That's same mechanism as when trying to view the profile of a non existant user (http://tatoeba.org/eng/user/profile/someonewhodoesntexist).

Even if we fix the PHP warnings/errors, the page still contains several links that are invalid: http://prntscr.com/6ec24o
So we'd better just not go to that page at all.

I will close this pull request, when/if you have time feel free to open a new one.

Issue #151, opened 2014-03-30
names of countries in the drop-down list in the profile are not internationalized


alanfgh:
tommy_san pointed out that the names of the countries in the drop-down list in a user's profile appear only in English.

lagleki:
Why not add these strings to Transifex so that we can quickly localize them?

jiru:
First, the list of countries is stored in database instead of the source code, so it needs a little bit of work to add them to Transifex. Second, the list is out of date: it includes countries that doesn’t exist anymore. And finally, countries names have already been translated into many languages by e.g. CLDR, so we’d better use their data instead.

PR #623, opened 2015-03-26
Localization of countries names

jiru:
This adds localization for countries names displayed on the UI (on the profile page) (#151).
* Got rid of the Country model (it’s now a helper)
* Updated the country list (“Serbia and Montenegro” removed, Canary Islands, Montenegro, South Sudan, Kosovo etc. added)
* Made countries names translatable (in a separate domain (separate pot file))
* Imported translations from CLDR (Esperanto and Belarusian are partially translated; Interlingua, Lojban, Latin, Low German, Occitan and Tagalog are missing; the rest is fully translated)
* The script used to import translations can be reused to import updated translations from CLDR, or we can update the translations ourselves by editing the PO files. We will need to merge things by hand however.

Note that 7 users have country set as “Serbia and Montenegro” in their profile: Dimitrije, Amicus, Ninche, tepsa, neron, ban17 and kojot. We will need to fix this somehow.


jiru:
I agree about the naming.

I initially wanted to make a model, but I changed my mind giving the following:
* We need to have the English country list somewhere in the code, so that the strings are translatable.
* I wasn’t able to make it work as a read-only model with just a PHP array as data. A simple ->find('all') (to generate the list) worked, but joining with other tables (to grab informations about a user) did not.

To be honest I’m always having trouble with MVC when it comes to handling choices lists, and I’ve never been able to find a proper solution that doesn’t duplicate the list somehow. Usually you need the list both in the model (for data validation) and in the view (to generate the list), while you both want the values to be translatable (written in the source code) and stored in the database. That’s why I’d prefer not to duplicate the list while we don’t need to.

trang:
Sorry I was clear enough. I meant to have the `countries_list.php` in `app/models` instead of `app/views/helpers`. And then instead of doing `App::import('Helper', 'CountriesList')`, you would do `App::import('Model', 'CountriesList')`.

But thinking again about it now, I feel we should perhaps put it in `app/vendors`, in order to keep some consistency: the classes in `app/models` should extend `AppModel`, the classes in `app/controllers` should extend `AppController`, the classes in `app/views/helper` should extend `AppHelper`, etc.

As for simulating the array is if it was read from the database, I was reading about datasources in CakePHP. If you feel like investing more time on this issue you could try to check if this could work:
http://stackoverflow.com/questions/6924574/how-to-use-models-without-database-on-cakephp-and-have-associations

Otherwise I'm fine with current solution.

jiru:
Alright, I put it in `vendors`.

Thank you for the link you mentioned. But I wonder if CakePHP would be able to join tables between different datasources.

PR #629, opened 2015-04-06
Simplify language codes

jiru:
This pull request is an attempt to simplify things regarding translation files, UI language codes and corpus language code lists.

* Use the same language codes for both UI and corpus (this changes a few codes like chi to cmn, pt_BR to por etc. providing redirection for the previous codes).
* Move the 3-letters language lists to a unique place in `vendors` (LanguagesLib), along with a few tools.
* Get rid of the 2-letters-codes map in `core.php`, using the map I introduced recently (204a25e).
* Get rid of the language list in `generate_sphinx_conf.php` using LanguagesLib.
* Get rid of the language list in the `Sentence` model using LanguagesLib.
* Get rid of the language list in `update-translations.py` using the one from `.tx/config`.
* Get rid of useless MO files. This simplifies the process of updating translations from Transifex. We should ultimately not need `update-translations.py` at all since the Transifex command line client is enough to pull the translations.
* Misc fixes


trang:
There are a few languages for which the translation of the interface doesn't work: Dutch, Greek, French (perhaps more, I haven't checked everything).

The problem seems to come from CakePHP's I10n lib:
https://github.com/Tatoeba/tatoeba2/blob/simplify-language-codes/cake/libs/l10n.php#L101

There is a `$__l10nMap` array where these languages are present twice, but with different 3 letter codes.
<pre>
/* Dutch (Standard) */ 'dut' => 'nl',
/* Dutch (Standard) */ 'nld' => 'nl',
...
/* Greek */ 'gre' => 'el',
/* Greek */ 'ell' => 'el',
...
/* French (Standard) */ 'fre' => 'fr',
/* French (Standard) */ 'fra' => 'fr',
</pre>

And there is a `$__l10nCatalog` array where the locale code is defined.
<pre>
'nl' => array('language' => 'Dutch (Standard)', 'locale' => 'dut', 'localeFallback' => 'dut', 'charset' => 'utf-8', 'direction' => 'ltr'),
'el' => array('language' => 'Greek', 'locale' => 'gre', 'localeFallback' => 'gre', 'charset' => 'utf-8', 'direction' => 'ltr'),
'fr' => array('language' => 'French (Standard)', 'locale' => 'fre', 'localeFallback' => 'fre', 'charset' => 'utf-8', 'direction' => 'ltr'),
</pre>

It seems that when we try to display the interface in `lang = 'nld'`, 'nld' is found in `$__l10nMap`, and then cake looks for 'nl' in the `$__l10nCatalog` which then tells that the locale code is 'dut'. Since there are no folder 'dut', it doesn't find the translations and the interface remains in English.

This problem can happen for other future languages that we may add, since there are a bunch of languages that are defined twice in `$__l10nMap` and may not have the ISO 639-3 code that we want for the locale in `$__l10nCatalog`.

So... we either have to modify the I10n lib, or to change the names of the locale folders to stick to what is defined in `$__l10nCatalog` (nld -> dut, fra -> fre, etc)... 
Or perhaps you see a better solution?

jiru:
Well spotted, it was working on my side thanks to the cache. I fixed it by modifying `$__l10nCatalog` at run time. It has 143 elements, I think the overhead is minor. It’s not great to have to do this, but it’s the best I can think of regarding future maintenance.

Issue #630, opened 2015-04-20
Handle null language level correctly


jiru:
Currently, it’s possible to [insert languages](https://dev.tatoeba.org/eng/user/language) without specifying a level in the profile, by not clicking on any of the level radio buttons. It stores a null value as level. We should make it mandatory, and fix the existing null levels in the database (we may switch them to level zero).

alanfgh:
raggione mentioned the following on the Wall:

:::
In the members list that you get when you activate “Languages of members” for English in the drop-down menu, a ghost member appears as the very last item, i.e. without a name and without a profile picture other than the outline of a head.

Some of us are intrigued by the fact that if you click on it, you land on your own profile page.

It’s one of the Tatoeba mysteries, it seems.
:::

I think that the behavior that raggione describes is a consequence of this inconsistency in the database.

alanfgh:
Note to self: Add this line:

'value' => 0

at line 90 in app/views/user/language.ctp.

alanfgh:
Later on, we may also want to set that field to non-null in the database schema. If you agree, I'll write that up as a separate ticket.

trang:
I prefer to keep the possibility to have the value null.

At some point I would like to have a script that will scan the sentences and add automatically for each user the languages in which they have contributed, with a level null (which would carry the meaning of "level unknown"), if that language was not already listed.

In each user profile, there will be the number of sentences owned by the user in each language.

jiru:
Then, we should handle null values correctly.

trang:
Languages that are null are displayed with a question mark.
http://prntscr.com/70pucx

jiru:
Nice, thank you! :-)

trang:
I'm re-opening this issue.

I realized just earlier while doing some tests that the level was set by default to 0 even when you are editing a language for which it wasn't initially at 0. 

This can be problematic when a user wants to add some details to one of their languages. They will unintentionally set the level to 0.

We can probably revert the `'value' => 0` and give the option for users to choose "unknown". I think Guybrush88 requested this possibility.

jiru:
I agree. Maybe word it like “unspecified”.

PR #638, opened 2015-05-02
address #630; set default lang level to 0

alanfgh:
Set the default language (skill) level to 0. After this, we can change
the existing null values for this field to 0 in the db.


jiru:
I’m sure you figured it out already, but just to make things clear: don’t change the existing null values to 0, we want to keep the null values.

alanfgh:
Right.

PR #640, opened 2015-05-06
Kill ultimate query

jiru:
This PR replaces the infamous SQL query of the Translation model with a CakePHP find(). Since it’s a very critical part of the code, I’d like people to review it carefully.

As explained in the commit message:
> According to personal tests, the new query itself is rather faster than the old one. I can’t tell how much because there is so much noise, but at least, the worst benchmark of the new outran the best benchmark of the old query.

But this is to be balanced with the overhead CakePHP’s find(), which I don’t know, but I’m confident it won’t hurt much.

### DIFF
--- a/app/models/translation.php
+++ b/app/models/translation.php
@@ -44,22 +44,22 @@ class Translation extends AppModel
     {
         $translations = $this->_getTranslationsOf($sentenceId, $languages);
 
-        // Calling manually the trigger for afterFind of Transcriptable behavior
-        // because the query to retrieve the translations is a custom query
-        $results['Translation'] = $this->Behaviors->trigger(
-            $this,
-            'afterFind',
-            array($translations['Translation'], false), 
-            array('modParams' => true)
+        $orderedTranslations = array(
+            'Translation' => array(),
+            'IndirectTranslation' => array()
         );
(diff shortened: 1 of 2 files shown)


allan-simon:
I totally support a tentative to replace it by something less handcrafted

things to consider:

  * is it "cache efficient" (mysql one)  when running lot of queries  (something you may not see when running alone on your computer)
  * how does the query time increase with the number of sentences compared to current one (well we're already quite high in term of number of sentences,  and even doubling is taking years) ?

jiru:
Thank you for your support. I must admit I have very little knowledge about benchmarking nor SQL query deep analysis. I just crafted an SQL script that runs the query 10000 times on [sentence 1](http://dev.tatoeba.org/sentences/show/1) with SQL_NO_CACHE on the dev database of the server, along with an equivalent script that runs the old query for comparison. I have no idea about how much it’s cache-friendly nor how to measure this.

allan-simon:
I have to admit i don't know too how to measure that in a rigorous way,  but maybe get the "benchmark" script to run the query on a range of sentences rather  than just sentence 1 , this way you will get a good average of the size of sentences  (because maybe result vary depending on the number of interconnected sentences)

PR #641, opened 2015-05-06
Users languages stats

trang:
Related Wall thread: http://tatoeba.org/eng/wall/show_message/22462#message_22462

​There are 2 new stats pages.

1. The number of members in each language for each level. 
  * This page can be accessed from the menu: "Members" > "Languages of members". Previously, this link was redirecting to a specific language (the one with the highest number of members). Now we can have global stats.
2. The number of native speakers in each language for each user status:
  * This page can be accessed from the menu: "Members" > "Native speakers".
  * The user status icons should be improved to work for color blind people, but I don't have any good idea right now of how to improve them.

The "Sentences by language" page was updated, but only graphically. It still contains the same information than before.

There were requests to add a sort feature. I won't do it in this pull request, but I will create an issue for it, and we can take care of it later.

On the technical side:
* The table `languages` has new columns, to store each of the stats.
* There is a script to update the stats: `docs/database/scripts/update_languages_stats.sql`.


jiru:
> The user status icons should be improved to work for color blind people, but I don't have any good idea right now of how to improve them.

How about adding a representative symbol along with each icon. Some ideas:
* Admin: a crown or a key, to represent power or full access
* Corpus maintainer: the pen icon, to represent full edition privilege
* Advanced contributor: the link icon, to represent linking privilege
* Regular contributor: nothing, to represent it’s regular

Issue #647, opened 2015-05-09
Order sentences in lists by date of addition to the list


trang:
https://tatoeba.org/eng/wall/show_message/16285#message_16285

Issue #602, opened 2015-03-01
For a given sentence, offer a way to display the lists to which it belongs


alanfgh:
Here's my use case: When I come across a sentence that has vocabulary I want to learn, I add it to a list: list 1, list 2, etc. I keep each list limited to 100 words so that it can be downloaded. Sometimes I come across a sentence that seems familiar, but I can't remember whether I added it to one of the lists. Browsing by list and performing a sequential linear search for the sentence across all the lists is not practical.

I'm most interested in whether a sentence belongs to one of my own lists. But seeing whether it's used in any list belonging to any user doesn't hurt, and might be helpful in other use cases.

ckjpn:
> Sometimes I come across a sentence that seems familiar, but I can't remember whether I added it to one of the lists. 

Note that it doesn't hurt to add a sentence to the same list again.
It only gets listed once, so you're not creating duplicates on a list.

When I'm going through lists of sentences, I often just go ahead and add a sentence to any appropriate list.  When viewing a list of sentences, it's faster to do this than to go to a sentence's page to see if it's already been added, or not.

Issue #594, opened 2015-02-18
The username of the creator of a list should be display on the list page


trang:
Cf. http://tatoeba.org/eng/wall/show_message/21858#message_21858

Issue #406, opened 2014-08-15
The way to rename a list is not clear


alanfgh:
To edit a list title, you first need to notice this in the right sidebar:

"You can change the name of the list by clicking on it."

There's nothing about the display of the title that suggests that it's editable. I would prefer a link under "View this list" and "Delete this list" that says "Rename this list", either instead of or in addition to the "click on the title" mechanism.

PR #653, opened 2015-05-14
Lists improvements

trang:
This pull request addresses various issues related to the lists feature.
* #647 Order sentences in lists by date of addition to the list 
* #602 For a given sentence, offer a way to display the lists to which it belongs
* #594 The username of the creator of a list should be display on the list page
* #406 The way to rename a list is not clear
* #117 add option of translating sentences in private lists

In addition to that:
* There is pagination, sort and search on the lists pages.
* Rather than displaying the lists split into 3 sections (my lists, collaborative lists, other personal lists) on one single page, the lists have been split into different pages.
  1. The sentences_lists/index page (the one that is accessed from the menu "Browse" > "Browse lists") contains all the lists.
  2. There is a sentences_lists/collaborative page, that lists all the collaborative lists.
  3. And there is a page that lists all the lists of a specific user. This link can be found on the profile of each user. You also have a "My lists" link on your username menu and on the index/collaborative lists pages.


jiru:
Good job!

I feel like the sorting feature isn’t very intuitive. If you click two times on a sort link, it reverses the order. Which is nice but you can’t figure it out unless you look at the URL or figure out the results are reverse-ordered. Besides, there is no visual indication of the current sort order overall. How about adding an up or down arrow next to the sort link, and make the sort link non-clickable when it’s active?

In other words:
If you click on "list name", you get ([] means it’s a link)
> Sort by: list name ↓ [↑] | [date created] | [number of sentences]

Then if you click on "↑", you get
> Sort by: list name [↓] ↑ | [date created] | [number of sentences]

Then if you click on "date created", you get
> Sort by: [list name] | date created ↓ [↑] | [number of sentences]

jiru:
In /sentences_lists/show/*, I’d rename “Information” to “About this list” and insert bullets in front of each item. Maybe merge the first two items into “Created by x on the y”.

I’d also rename “Menu” to “Actions” or something. The layout of this part is rather inconsistent, but well, I’ll let you improve it if you feel so. Otherwise it’s not a big deal.

In /sentences/show/*, I’d rename “Lists” to “Lists this sentence belongs to” or something. In this area, personal lists of other users are displayed too. Maybe we want to avoid this?

trang:
I've renamed "Information" to "About this list".

I just removed the "Menu" title. It was actually called "Actions" before, if I'm not mistaken, but I wasn't satisfied with the wording, nor was I completely satisfied with Menu. And I was considering merging this block with the "About this list" block because the layout is indeed terrible.

The pagination is also in my todo list. It needs to be improved on the tags page, as well as on the members page. Since it was more general, I didn't want to take care of it in this update.

As for the visibility of lists, that's a deeper issue we'll need to change about the lists feature in a future update. 
We need to distinguish between lists that the user wants to be public (visible by everyone), and lists that are private (only visible to the creator). Today, the "private" level doesn't exist. Everything is public.
And then it's either personal (only the creator can edit) or collaborative (everyone can edit). It would be nice to also have something in between (a certain group of people can edit, but not everyone).

I'm going to merge now so that people can start translating on Transifex before the weekend.

PR #690, opened 2015-06-20
Optimization of "not translated into" filter

trang:
The query that is executed when users displayed sentences not translated into XXX contains a several joins.
<pre>
SELECT DISTINCT s.id FROM sentences s
  JOIN sentences_translations st ON ( s.id = st.sentence_id )
  JOIN sentences t on ( st.translation_id = t.id )
WHERE s.lang = '$source' AND t.lang = '$target'
</pre>

We could do without these JOINs because we have the required fields already in the `sentences_translations` table, in the `sentence_lang` and `translation_lang` fields.

<pre>
SELECT sentence_id FROM sentences_translations
WHERE sentence_lang = '$source'
AND translation_lang = '$target'
</pre>

For a comparison, on my machine, searching for
- English sentences not translated into French: takes 5 minutes... vs 1.49 s.
- Spanish sentences not translated into English: takes 40 seconds vs. 1.13 s. 
https://gist.github.com/trang/928652c1a56fd9b8bdb8

Now the problem when selecting from the `sentences_translations` table is that we need to make sure the `sentence_lang` and `translation_lang` fields are in sync with the language in the `sentences` table.

When a user changes the language of a sentence, it doesn't update the `sentences_translations` table accordingly. I didn't take care of this yet.

I have only added a file containing the queries to fix the `sentences_translations` table but the last queries are really, really slow to execute. I don't know if there's a solution to make them faster, or if it's better to just retrieve the id's to delete and generate a script with a bunch of `DELETE FROM sentences_translations WHERE sentence_id = XXX`.

So there are at least 2 things to do before this pull request can be merged:
1) Update the `sentences_translations` when the language of a sentence changes.
2) Find a faster way to sync the languages in `sentences_translations` with the languages in `sentences`.

### DIFF
--- a/app/models/link.php
+++ b/app/models/link.php
@@ -252,5 +252,18 @@ class Link extends AppModel
         ));
         return Set::classicExtract($links, '{n}.0.translation_id');
     }
+
+    public function updateLanguage($sentenceId, $lang)
+    {
+        $this->updateAll(
+            array('sentence_lang' => "'$lang'"),
+            array('sentence_id' => $sentenceId)
+        );
+
+        $this->updateAll(
+            array('translation_lang' => "'$lang'"),
+            array('translation_id' => $sentenceId)
+        );
+    }
 }
 ?>
--- a/app/models/sentence.php
+++ b/app/models/sentence.php
@@ -802,6 +802,7 @@ class Sentence extends AppModel
             $this->id = $sentenceId;
             $this->save($data);
 
+            $this->Link->updateLanguage($sentenceId, $newLang);
             $this->Contribution->updateLanguage($sentenceId, $newLang);
             $this->Language->incrementCountForLanguage($newLang);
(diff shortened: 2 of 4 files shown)


jiru:
> I have only added a file containing the queries to fix the sentences_translations table but the last queries are really, really slow to execute. I don't know if there's a solution to make them faster, or if it's better to just retrieve the id's to delete and generate a script with a bunch of DELETE FROM sentences_translations WHERE sentence_id = XXX.

I don’t think so. I tried to run these two queries on the dev database (and rolled back after). Each removed 2331 rows. The first one (join on `st.sentence_id = s.id`) took about 25 seconds to execute, but the second one (join on `st.translation_id = s.id`) took about 2 minutes to complete. It’s still reasonable for a one-time use, but that difference is rather strange. `EXPLAIN` doesn’t show any difference.

Issue #53, opened 2014-02-17
Provide advanced search w/any combination of tags, terms, owner, etc.


loolmeh:
Provide an advanced search with any combination (including exclusion) of the following:
- tags
- search terms (probably Sphinx-type)
- owner
- timestamp (probably last modification time)
- language
- presence/absence of audio

AlanF_US, Wall, later extended on wiki

PR #697, opened 2015-06-29
Advanced search

jiru:
This addresses #53. There are still some little things to consider before merging this:
* Add a reset button? There are [reasons not to use them](http://www.nngroup.com/articles/reset-and-cancel-buttons/). Maybe provide a link to a fresh page instead, so that one can press the browser back button if clicked by mistake.
* [Put the fields on the top of the results, or on the right side (like they are currently)?](https://tatoeba.org/jpn/wall/show_message/23255#message_23255) To put them back on the top-left, revert commit 675da35a20874c5ed4d7407c140fbd2871147259.
    * Right side pros: looks nicer, no need to scroll to get to the first result.
    * Right side cons: more scrolling needed to get to frequently changed fields and to the submit button, more packed, uglify the page when there are no results.
* Add a sort option to prioritize sentences with audio?


trang:
> Maybe provide a link to a fresh page instead, so that one can press the browser back button if clicked by mistake.

Yes, I think that would be best.

> Put the fields on the top of the results, or on the right side (like they are currently)?

Leave them on the right side.

> Add a link accessible from all pages that brings you directly to the advanced search

In the search bar, we should have a link "Advanced search". This can maybe serve as a reset as well.
I'll take care of this, since I've wanted for a long time to review a bit the design of the search bar.

> Add a sort option to prioritize sentences with audio?

It would be a plus, but we can do without it for a start. Users can filter sentences with or without audio.

> Add back a link to the text-search wiki page.

I'll take care of this. I'm thinking of putting a (?) icon in the title --> "More search criteria (?)".

jiru:
Thank you!

PR #738, opened 2015-08-02
Users corpus

trang:
Wall thread: http://tatoeba.org/eng/wall/show_message/23719#message_23719

This pull request includes feature that allows users to mark sentences as correct, incorrect or unsure. By doing so, each user will be building their own corpus within Tatoeba. The corpus of a user will basically represent their "linguistic world".

This is the first step for a more long term goal: once we have collected enough data about who thinks what is correct or incorrect, we can try to calculate a "trust" score for each sentence on which users can rely to decide how much they can trust that a sentence is correct.

### What the feature looks like so far

1. You can mark sentences as "correct", "incorrect" or "unsure". http://prntscr.com/7zprx9
2. When you mark a sentence, it is added to your corpus. http://prntscr.com/7zps39
3. On the sentence's details page, you will be able to see who marked which sentence. http://prntscr.com/7zps9d

### Things to consider

Note: I don't expect to have an answer to all the things listed below before the feature actually gets released and used in real conditions.

* Are the icons clear enough? Currently we have: 
  * check --> correct
  * question mark --> unsure
  * exclamation mark --> incorrect
* What text to use for the tooltips of the icons? Currently we have:
  * correct --> "Correct"
  * unsure --> "Not sure if correct"
  * incorrect --> "Incorrect"
* Where should we display the buttons? Is it fine to have them in the sentence menu after all the other buttons? Or should we move them closer to the owner name?
* Should we display these buttons all the time? Or should we make an option to activate the display of the buttons?
* Is it going to be confusing for users to have in their menu "My sentences" followed by "My corpus"? Can we re-organize this more clearly?
* Where should we display the new information on the sentence's page? What title to use for section on the sentence's page that lists the correctness chosen by each user? Currently it says "Reviewed by", and it is placed at the top in the right side bar.
* There is currently no instructions besides of the tooltips on the buttons. Do we need to add a help link, for instance in the "Reviewed by" section, to explain what each of the icons mean? If yes, what do we write exactly in this help section?

### Things to do next

1. When a sentence is modified, all the reviews need to be marked as deprecated.
  * Users need to have a page where they can view all their deprecated reviews.
  * On the sentence's page, we need to display the deprecated reviews in grey, or maybe even hide them.

2. We'll need to migrate the `OK` tags into the new system, with the permission of the contributor of course. But I'm sure that users like CK will prefer not to have to re-review their OK sentences, and instead have all the sentences that they tagged `OK` be marked as correct.

3. It may be useful to display a timeline on the homepage for sentences that were marked "incorrect" or "unsure", the same way that we have a timeline for the last sentences added/edited.


trang:
Some changes:

1. Replaced the wording "corpus" by "collection". [Suggested by tommy_san](http://tatoeba.org/eng/wall/show_message/23787#message_23787).

2. Instead of "correct" / "incorrect" / "not sure", we're going to use "OK" / "not OK" / "unsure". Since we already have the tag "OK" that is part of Tatoeba's vocabulary, we might as well reuse it.

3. Fixed a bug that prevented from marking a new sentence. [Reported by Silja](http://tatoeba.org/eng/wall/show_message/23752#message_23752).

ckjpn:
I suggest keeping the word "corpus".

define:corpus
https://www.google.com/search?q=define%3Acorpus

trang:
The fact that we defined Tatoeba as "a collection of sentences and translations" makes me feel that we should keep the same wording here. Mostly because it sounds less technical.

ckjpn:
If this "collection" refers to a member's "reviewed sentences', then "reviewed sentences' might be less confusing.  Reviewing sentences and collecting sentences are 2 difference things.  Members can use the lists to collect sentences.

PaulPeer:
I dont know if this is the right place to mention  it, but Horus should also merge the reviews, when combining two identical sentences.

PaulPeer:
img_2676 - 2015-08-11 at 17-58-55
This is how the icons look like in iPhone. Any relation with the fact that they cannot be selected?

trang:
@PaulPeer, you can simply report issues on Tatoeba's wall. We will do the necessary to log the issue in GitHub at the right place.

I'm aware of Horus not merging the reviews. This will be implemented in a later update.

Regarding the icons, this is another issue that I think you will also see on a Mac, not just on iPhone. It is not related to the new feature. You will probably notice other icons that are cut of (for instance I think the user icon next to the owner name is cut off).

PaulPeer:
No. On the Mac it is perfect.

ckjpn:
I think the word "collection" is not a good choice.
Why would a person want to collect bad sentences and sentences they are unsure of?

One of the following words would be better.  Perhaps others can suggest other possibilities.

rating / rated
quality
proofreading
reviews / reviewed
sorted / sorting 
opinion

I think like either the word "rating" or "quality" best, since what we are actually doing is rating the quality of an item.

"Proofreading" is close, but usually also implies making corrections and not just reading.


I'd still like to see some way to indicate that a sentence has been read and but not rated.
See https://tatoeba.org/eng/wall/show_message/23879#message_23879

I also think it would be very useful for people studying languages to know the difference between the following.

1. Good and I'd use it.
2. Good, but I wouldn't use it. -- maybe there's a more natural way to say the same thing, but there could be other reasons.

Issue #925, opened 2015-11-14
Annoucement system


trang:
The goal here is to be able to spread important information or an important message to all Tatoeba users (we'll call it "announcement").

The announcement is displayed below the search bar and can be closed by the user once they've read it and do not want to see the message again.

jiru:
I found a bug. When `Announcement.enabled` is disabled in the config, `jquery.cookie.js` is not loaded and the browser Javascript console shows a fatal error in `generic_functions.js`: `TypeError: $.cookie is not a function`

trang:
Yes, thank you. I noticed it when I updated Tatoeba. Didn't have time yet to push the fix but should have time tonight.

PR #661, opened 2015-05-18
Editable transcripts

jiru:
Addresses #77.

## How to test it locally
```sh
git checkout editable-transcripts

alias cake='./cake/console/cake -app ./app/' # or something
cake acl create aco controllers Transcriptions
cake acl create aco Transcriptions save
cake acl create aco Transcriptions view
cake acl create aco Transcriptions reset
cake acl grant Group.2 Transcriptions all
cake acl grant Group.3 Transcriptions all
cake acl grant Group.4 Transcriptions all

mysql ... < ./docs/database/updates/2015-05-02.sql # takes about 5h30
mysql ... < ./docs/database/triggers/last_contributions_triggers.sql

cake transcriptions script sentences # takes about 15m
```

## Unsettled questions
* Do we want to keep transcription for Shanghainese (it’s [much failing](https://www.reddit.com/r/shanghainese/comments/2xgb8t/tatoeba_convert_chinese_characters_to/) and using IPA that only a few people can read), [for Georgian](https://tatoeba.org/eng/wall/show_message/21466)?

## Related threads

* [Google group - Editable transcriptions progress](https://groups.google.com/forum/#!msg/tatoebaproject/gauMnAf4G4M/oFVzWxfl4oMJ)
* https://tatoeba.org/eng/wall/show_message/22679
* https://tatoeba.org/eng/wall/show_message/22870
* https://tatoeba.org/eng/wall/show_message/24956
* https://tatoeba.org/eng/wall/show_message/24982

PR #926, opened 2015-11-15
 added Ge'ez transcription for Tigrinya / Amharic. Changed Georgian t…

oyd11:
…ranslitaration to match ISO-9984 and Wikipedia/Wiktionary

### DIFF
--- a/app/vendors/autotranscription.php
+++ b/app/vendors/autotranscription.php
@@ -48,10 +48,123 @@ define('CMN_SCRIPT', 12);
 define('JPN_FURIGANA', 20);
 define('JPN_ROMAJI', 21);
 
+
+	// some Unicode routines for parsing / constructing Ge'ez chars:
+	//  Oyd11,  2014 , MIT license , based on JS code :: http://jsfiddle.net/oyd11/Z5Wws/
+	
+// Concider moving to utf8 lib:
+ // 	http://pageconfig.com/post/portable-utf8
+// (if tatoeba stays on php)	
+	/**
+	* Helper function: unicode safe version of 'str_split'
+	* Since: str_split - does not handle Unicode arrays:
+	 * str_split_unicode
+	 * 
+     * @param string $unicode_str
+     *
+     * @return array of unicode chars
+	 *
+	 * from http://php.net/manual/en/function.mb-split.php
+	 */
+	  function unicode_str_split($unicode_str) {
+		 return preg_split('//u', $unicode_str, null, PREG_SPLIT_NO_EMPTY); 
+	}
+	
+ /**
+ * @param string $unicode_str - a one-character string
+ *
+ * @return unicode representation as a number
+	*/
+	function unicode_str_ord($unicode_str) {	
+		$le_str = mb_convert_encoding($unicode_str, "UTF-32LE");	
+		$le_arr = unpack('C*', $le_str);
+		$acc = 0;
+		$factor = 1;
+		foreach ($le_arr as $b) {
+			$acc += $b*$factor;
+			$factor <<= 8;
+		}
+		return $acc;
+	}
+	
+     function unicode_chr_to_utf8($num) {
+		$ret = iconv('UCS-4LE', 'UTF-8', pack('V', $num));
+		return $ret;
+	}
+	
+    $geezVowels = array( 'ä' , 'u' , 'i' ,'a' ,'e' , "" ,'o' , 'wa' );
+	
+ $geezConsBase = array(
+		'ሀ' => 'h',		'ለ' => 'l',	'ሐ' => 'ħ',		'መ' => 'm',
+		'ሠ' => 'ś',		'ረ' => 'r',	'ሰ' => 's',		'ሸ' =>'š',
+		'ቀ' => 'q', 	'ቈ' => 'qʷ',	'ቐ' => 'Q',	'ቘ' => 'Qʷ',
+		'በ' => 'b',		'ቨ' => 'v',		'ተ' => 't',		'ቸ' => 'č',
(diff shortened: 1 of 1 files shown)


jiru:
Thank you, oyd11, but I don’t think we can merge your pull request.

First, [it’s not clear to me wether we should continue adding transcriptions, since their usefulness is controvesed](https://tatoeba.org/wall/show_message/21405#message_21405). To sum up the problem, there are situations where native speakers use several scripts concurrently, such as simplified and traditional characters for Chinese. In this case, transcribing from one script to another is perfectly fine because it helps native speakers writing sentences in their mother tongue, which is one of the primary goals of Tatoeba. However, when a transcription is not needed by native speakers, but only to learners (or worse: only intelligible to learners), one can consider it attracts very-low-level contributors since learning the writing system is one of the very first steps a learner should do. Some transcriptions lack a proper standard or have other problems. So whatever we decide needs to be debated on the Wall first.

Second, your code needs to be improved before we can merge it. I did not review it thoroughly, but here a few things that would need to be fixed:
* Remove usage of `global`. You can declare your arrays `$geezConsBase`, `geezPuncTable` etc. as private attributes of `Autotranscription` and access them from within functions using `$this`.
* `geezNum2Latin()` is a bit too obscure, in particular the bitwise operations. Add comments or break it down into subfunctions with meaningful names.
* I strongly suggest you try to avoid calculating unicode code points and using them directly because it makes your code obscure and error-prone. It’s certainly easier and safer to use [unicode properties](http://www.regular-expressions.info/unicode.html) instead. There are properties like `\p{Ethiopic}` and `\p{Punctuation}`.
* You’re using your conversion tables without prior checking of key existence (using `isset()` or `array_key_exists()`).
* The indentation is somewhat inconsistent. Never use tabs and always follows the indentation style of the existing code.

Finally, I’m in the process of totally rewriting the transcription code in #661, so it would be better to base any new transcription system on that branch instead.

trang:
> First, it’s not clear to me wether we should continue adding transcriptions, since their usefulness is controvesed.

I think we can add transcriptions as long as they are made optional. I'm personally in favor of hiding by default all transcriptions/transliterations, and display them only if the user activates the option.

This would of course imply warning our users about the change, which we could do with the announcement system (#925).

One thing to consider though is whether it's fine if this option would be available only for registered users or if we should implement an options system for non-registered users. I'd guess that it's fine to "force" users to register to have access to transliterations and transcriptions, but maybe I'm wrong.

> Finally, I’m in the process of totally rewriting the transcription code in #661 

Do you have any rough idea of how much work you still have to do until this branch will be ready to be merged? Just to know if it's a matter of a couple of weeks, or a couple of months, or half a year?

jiru:
> Do you have any rough idea of how much work you still have to do until this branch will be ready to be merged? Just to know if it's a matter of a couple of weeks, or a couple of months, or half a year?

There is not much work remaining, mainly fix a few bugs and find an icon. We may go with a new round of tests on dev.tatoeba.org in a week or two and eventually merge it. But there are also some unsettled questions (see the PR page) I’m not sure how to deal with.

oyd11:
Hello guy, sorry for the late reply::
So @jiru :: I agree with the coding style comments, and apologise - I have very little coding experience in PHP - and thus found the solutions I could for issues (eg, I couldn't made the code work without the _global_ variables), I would appreciate any revision of the code. As you say your branch will be merged soon - and I do see the surrounding code on your branch is much cleaner anyhow - I could try to re-implement this after you merge back into the 'master' branch.
 As for the _bitwise_ operations - that's part of how Ge'ez Consonant/Vowels are encoded in Unicode, yes, it's way better to self-document it in the code somehow.

As for adding or not adding transliterations - it seems these discussions on the harmfulness of transliteration are all ignoring the case of Digraphia [ https://en.wikipedia.org/wiki/Digraphia ], specifically - Serbian and Serbo-Croatian - were (as implemented in Wikipedia) some readers of the same language - prefer to read it in a different alphabet, and the transliteration is 1-1.

It also seems to me like the whole argument is centred around some social issues with Japanese learners who confuse the site with "lang-8.com".

Another important case (which is the case with transliterating Ge'ez) - is having close sister-languages with different alphabets, in the case of Ge'ez - other Semitic language speakers (speakers of Arabic, Hebrew and Maltese) - may want to see how something is pronounced in Amharic or Tigrinya and might recognise most words, without learning the alphabet. Also - a lot of diaspora speakers of Amharic and Tigrinya speak fluently and do not read the alphabet (but are familiar with the Latin alphabet), from the social context I know with these languages - these might be the majority of prospective learners (who want to improve their home language, and learn the alphabet on the way), I see the alphabet is a big obstacle for them.
[this is not unique to Semitic again, a similar situation exists with even closer Turkic languages written in Latin vs Cyrillic and I'm sure many others]

jiru:
I’m okay to add new transcriptions as long as they are justified, and your arguments are convincing.

Another thing that would be very much appreciated would be tests (or data to make tests). A test is a simple code that executes the transcription algorithm against a given string and verify that the output is as expected. This requires a list of source→transcription sentences that are considered correct. It’s important because it allows developers (like me) to work on the transcription code without being native or linguists. If you can provide me with such a list, I will be able to help you out.

oyd11:
Sure, I have test sentences, and there should be a test of the whole alphabet blit'ed in.
As Trang suggests - I'll make the "revised Georgian" and "Ge'ez" - separate commits. For now I'm waiting for your merge then.

jiru:
My branch is now merged. If you’re still interested in adding Ge’ez transcription, please follow [this guide](http://en.wiki.tatoeba.org/articles/show/new-transcription-request).

PR #930, opened 2015-11-21
Adding a link to users' sentences

lagleki:
I'm not an expert in the code so please add localization strings as needed.

### DIFF
--- a/app/views/pages/downloads.ctp
+++ b/app/views/pages/downloads.ctp
@@ -85,6 +85,7 @@ $this->set('title_for_layout', $pages->formatTitle(__('Download sentences', true
         $field_struct_str = __('Fields and structure', true);
         $file_desc_str = __('File description', true);
         $sent_id_str = __('Sentence id', true);
+        $mark_id_str = __('Sentence id', true);
         $text_str = __('Text', true);
         $lang_str = __('Lang', true);
         $username_str = __('Username', true);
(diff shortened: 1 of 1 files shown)


app/views/pages/downloads.ctp
```diff
@@ -85,6 +85,7 @@ $this->set('title_for_layout', $pages->formatTitle(__('Download sentences', true
         $field_struct_str = __('Fields and structure', true);
         $file_desc_str = __('File description', true);
         $sent_id_str = __('Sentence id', true);
+        $mark_id_str = __('Sentence id', true);
```
jiru: The name of the variable is not appropriate because it’s not an id. Use e.g. `$mark_str` instead.
The string is incorrect, it should be something like 'Mark'.

app/views/pages/downloads.ctp
```diff
@@ -412,7 +413,38 @@ $this->set('title_for_layout', $pages->formatTitle(__('Download sentences', true
...
+                <span class="symbol">[<?php echo $tab_str; ?>]</span>
+                <span class="param"><?php echo $date_added_str; ?></span>
+                <span class="symbol">[<?php echo $tab_str; ?>]</span>
+                <span class="param"><?php echo $date_last_mod_str; ?></span>
+            </dd>
+
+            <dt><?php echo $file_desc_str; ?></dt>
+            <dd>
+            <?php 
+             __('Contains sentences\' marked by users.'); ?>
```
jiru: Maybe you could elaborate: what marks is it about?
Maybe you could give an example. For instance, explain the possible values of the mark field.

jiru:
Thank you! Localized strings are fine. I left a few comments, otherwise it’s good.

trang:
I'll merge this and do the fixes.

PR #994, opened 2016-02-02
sql export

loolmeh:
Notes:
- This script builds an sql query from a config dictionary and can run an optimized mysqldump command
- should be mergable, I need a fresh set of eyes to take a look at the config dict in case I missed removing something that shouldn't be there
- the config dictionary is a list of key/value maps with:
  - table: being the table name
  - fields_whitelist: having a list of fields to copy to the copy db
  - fields_blacklist: having a list of defaults to scrub the given fields with {c} would insert a counter
  - append_updates: narrows down on missing fields instead of copying all rows
  - force_copy: tells the script to copy a table even if it's all whitelisted instead of exporting directly from the parent table
- I moved the models to a separate package so this will need to be kept in mind for server updates 

TODO:
- make the script a bit more readable with comments
- add tests

### DIFF
--- a/docs/tatoeba2-django/tatoeba2-django/settings.py.template
+++ b/docs/tatoeba2-django/tatoeba2-django/settings.py.template
@@ -38,6 +38,7 @@ INSTALLED_APPS = (
     'django.contrib.staticfiles',
     'tatoeba2',
     'dedup',
+    'xplisqlus'
 )
 
 MIDDLEWARE_CLASSES = (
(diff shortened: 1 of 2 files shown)


jiru:
> should be mergable, I need a fresh set of eyes to take a look at the config dict in case I missed removing something that shouldn't be there

At first, I though anything that’s publicly visible on the website was okay to export, but I’m considering it may not. By providing dumps, you’re not only allowing access to already accessible data, you’re also allowing large-scale treatments on it, while some users might not be okay with that. For instance, one could generate stats from profile’s contents, favourite sentences or sentences comments, to allow more efficient targeting of PM spam. Even if the ultimate motive of a given large-scale treatment is not evil, some users might not be okay with the principle (I’m not very happy with it personally).

That said, it’s not a problem if we restrict access to the dumps to trusted contributors, like people having root access to the server.

Apart from that, I think this needs to be anonymized or removed:
* `sentence_comments` where hidden is true
* `wall` where hidden is true (removing the post might break the thread tree, so you probably need to empty the message)
* private lists

trang:
Closed because PR is too old and not mergeable anymore.

We also discussed it during our weekly call with the developers and we established that such exports are not really needed. They are useful in cases where we want to test performance/scalability. But even then, developers can make some scripts to generate data.

Issue #970, opened 2016-01-03
Pagination for favorites


trang:
The favorites page does not have pagination, all the favorites are displayed on the same page.

This can result in very long loading time for users who have a lot of favorites.

ckjpn:
Here's a Favorites list I stumbled across that has over 1,000 sentences.
You could use it for testing the pagination, once you get it done.

https://tatoeba.org/eng/favorites/of_user/rabblerouzer

kwikadi:
Could I be assigned this issue?

jiru:
Yes. But keep in mind we don’t use the Github assignation feature seriously. Anyone is free to start working on any issue.

jiru:
Actually I can’t assign you, probably because you’re not a part of the organization or something. Just don’t mind.

trang:
Glad to see you back @kwikadi!

As jiru said, you can simply post a comment to say that you want to work on an issue and start working on it right away.

@jiru, the reason why you couldn't assign an issue to kwikadi is because I've been reorganizing the teams and permissions in GitHub. I'm still not finished, but I moved kwikadi to a team that I called "Inactive" where I've put people that haven't shown any sign of activity for a long time and that I should contact back someday to ask if they are still interested to contribute or if I should just remove them definitely. The "Inactive" team doesn't have access to any repo.

PR #996, opened 2016-02-04
Favorites pagination

kwikadi:
This pull introduces pagination in a user's favorites, as requested in Issue #970. It also fixes a bug where if a non existent user's favorites were requested, it would say that the user had no favorites instead of saying that the user does not exist.

### DIFF
--- a/app/views/favorites/of_user.ctp
+++ b/app/views/favorites/of_user.ctp
@@ -105,7 +105,6 @@ $this->set('title_for_layout', $pages->formatTitle($title));
 
         } else {
             __('This user does not have any favorites.');
-            echo $html->link(__('Go back to previous page', true), $backLink);
         }
     }
     ?>
--- a/app/views/favorites/of_user.ctp
+++ b/app/views/favorites/of_user.ctp
@@ -40,7 +40,7 @@ if ($userExists === true) {
 }
 
 $this->set('title_for_layout', $pages->formatTitle($title));
-$numberOfSentences = count($data);
+$numberOfSentences = $data;
 ?>
 
 <div id="annexe_content">
--- a/app/views/favorites/of_user.ctp
+++ b/app/views/favorites/of_user.ctp
@@ -40,7 +40,6 @@ if ($userExists === true) {
 }
 
 $this->set('title_for_layout', $pages->formatTitle($title));
-$numberOfSentences = $data;
 ?>
 
 <div id="annexe_content">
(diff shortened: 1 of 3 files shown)


app/views/favorites/of_user.ctp
```diff
@@ -26,9 +26,21 @@
...
+    );
+
+    $title = format(__("{user}'s favorite sentences", true), array('user' => $username));
+} else {
+    $title = format(__("There's no user called {username}", true), array('username' => $username));
+}
+
 $this->set('title_for_layout', $pages->formatTitle($title));
-$numberOfSentences = count($favorites);
+$numberOfSentences = $data;
```
jiru: `$numberOfSentences` already contains the number of sentences, it is fetched from `$paginator` above (although not initialized to zero). You don’t need to use `$data` at all. You may remove it from the controller as well.

app/views/favorites/of_user.ctp
```diff
@@ -43,45 +55,56 @@ $numberOfSentences = count($favorites);
...
-                        'controller' => 'sentences',
-                        'action' => 'show',
-                        $sentenceId
-                    )
-                );
+    if ($userExists === false) {
+        $commonModules->displayNoSuchUser($username, $backLink);
+    } else {
+        echo '<h2>';
+        echo $title . ' ('. $numberOfSentences.')';
```
jiru: We have a more or less standard way to set titles of paginated results. Take other paginated pages as example, such as [that one](https://github.com/Tatoeba/tatoeba2/blob/dev/app/views/transcriptions/of.ctp#L50-L54).

jiru: I just figured it was the actually the original code. Well, feel free to fix the title while you’re at it then. :-)

app/views/favorites/of_user.ctp
```diff
@@ -43,45 +55,56 @@ $numberOfSentences = count($favorites);
...
+                        $parentId,
+                        $withAudio
+                    );
+                }
             }
+            $pagination->display($paginationUrl);
+
+        } else {
+            __('This user does not have any favorites.');
+            echo $html->link(__('Go back to previous page', true), $backLink);
```
jiru: I don’t think such “go back” links are very useful. People already have a “back” button on their browser and they use it all the time.

kwikadi: I was actually just making sure all pages are similar. I picked this from the 'sentences' view.

kwikadi: In order to remove backlinks completely, I will need to edit displayNoSuchUser as well. Do you think I should create another pull where I remove backlinks from the codebase? (Removing them in this pull will make it messy, I think).

jiru: Yes, backlinks are out of the scope of this PR. You did right by imitating existing files, it’s just that that code happens to be old.

If you want to remove all the backlinks from the codebase, you can do it, but it’s not very fun and there are maybe more important things to do (although code maintenance is important too). It’s up to you.

@trang any comment about backlinks?

trang: Yep, I agree it's not useful here. No need to try to keep it consistent with the other pages that have a back link. It's better if the favorites page doesn't have a back link.

For the back links in the other pages, we can remove them later. It's indeed not so urgent, nor so important. But yea, if you are in the mood for cleaning up old code, feel free to make another pull request.

trang:
Thanks a lot for the pull request @kwikadi. 

I'm gonna hopefully test and merge it tomorrow, and we can deploy it to the prod next week (probably Feb 10th).

PR #1006, opened 2016-02-13
Print error message when search is disabled

### LINKED ISSUE #998: Proper handling of error for random sentences
Currently, when the search is disabled or Sphinx returns an error, the sentence which id is 1 is returned as random sentence. Instead, we should display a proper error message like in #23.

For the random sentence of the [main page](https://tatoeba.org/), maybe it’s wiser to just not display the random sentence instead of an error. For the [several random sentences](https://tatoeba.org/sentences/several_random_sentences) and the [show random sentence feature](https://tatoeba.org/sentences/show/random), display a proper error message.

kwikadi:
Attempts to fix #998

### DIFF
--- a/app/controllers/pages_controller.php
+++ b/app/controllers/pages_controller.php
@@ -188,12 +188,7 @@ class PagesController extends AppController
         $lang = $this->Session->read('random_lang_selected');
         $randomId = $this->Sentence->getRandomId($lang);
         if (is_bool($randomId) && !$randomId) {
-            $searchDisabled = !Configure::read('Search.enabled');
-            if ($searchDisabled) {
-                $this->set("searchProblem", 'disabled');
-            } else {
-                $this->set("searchProblem", 'error');
-            }
+            $this->set('searchProblem', true);
 
         } else {
             $randomSentence = $this->Sentence->getSentenceWithId($randomId);
--- a/app/controllers/pages_controller.php
+++ b/app/controllers/pages_controller.php
@@ -187,17 +187,27 @@ class PagesController extends AppController
         $this->loadModel('Sentence');
         $lang = $this->Session->read('random_lang_selected');
         $randomId = $this->Sentence->getRandomId($lang);
-        $randomSentence = $this->Sentence->getSentenceWithId($randomId);
-        $alltranslations = $this->Sentence->getTranslationsOf($randomId);
-        $translations = $alltranslations['Translation'];
-        $indirectTranslations = $alltranslations['IndirectTranslation'];
+        if (is_bool($randomId) && !$randomId) {
+            $searchDisabled = !Configure::read('Search.enabled');
+            if ($searchDisabled) {
+                $this->set("searchProblem", 'disabled');
+            } else {
+                $this->set("searchProblem", 'error');
+            }
 
-        $this->set('random', $randomSentence);
-        $this->set('translations', $translations);
-        $this->set('indirectTranslations', $indirectTranslations);
+        } else {
+            $randomSentence = $this->Sentence->getSentenceWithId($randomId);
+            $alltranslations = $this->Sentence->getTranslationsOf($randomId);
+            $translations = $alltranslations['Translation'];
+            $indirectTranslations = $alltranslations['IndirectTranslation'];
 
-        if (isset($randomSentence['Sentence']['script'])) {
-            $this->set('sentenceScript', $randomSentence['Sentence']['script']);
+            $this->set('random', $randomSentence);
+            $this->set('translations', $translations);
+            $this->set('indirectTranslations', $indirectTranslations);
+
+            if (isset($randomSentence['Sentence']['script'])) {
+                $this->set('sentenceScript', $randomSentence['Sentence']['script']);
+            }
         }
     }
 
--- a/app/controllers/sentences_controller.php
+++ b/app/controllers/sentences_controller.php
@@ -225,10 +225,11 @@ class SentencesController extends AppController
 
             if (is_bool($randomId) && !$randomId) {
                 $searchDisabled = !Configure::read('Search.enabled');
+                $this->set('searchProblem', true);
                 if ($searchDisabled) {
-                    $this->set('searchProblem', 'disabled');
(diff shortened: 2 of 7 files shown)


jiru:
You’re mixing the “search error” and “search disabled” cases. One can enable or disable the search in core.php, but when search is enabled, it may error. We need to know whether there are no results because the search is turned off or because of some problem. For instance, we sometimes need to disable the search during the site updates. In this case, we need to tell the users that the search is disabled for some reason instead of showing them an error, so that they don’t report the search is broken.

If you want to test the “search error” case, you can stop Sphinx
```
sudo /etc/init.d/sphinxsearch stop
```
and to remove the cached random ids
```
rm app/tmp/cache/cake_rand_array_*
```

If you do so, you’ll see there are some warnings with your current code. It’s okay to have the `Search query failed` notice however because that’s the error we need to know.

kwikadi:
I think I have it all covered now. There are some issues with the JS part though (it removes the spinner, and uses alert() to notify users of sphinx being down) please go through that and see if it is OK. Also, a standard error message across pages is needed (I have a placeholder for now).

app/models/sentence.php
```diff
@@ -382,7 +382,7 @@ public function getRandomId($lang = 'und')
     public function getSeveralRandomIds($lang = 'und',  $numberOfIdWanted = 10)
     {
         if(Configure::read('Search.enabled') == false) {
-            return array(1);
+            return false;
```
trang: It would make more sense to return `null` rather than `false`, I think. Then everywhere where you used `is_bool()`, you can check instead with `is_null()`.

app/controllers/sentences_controller.php
```diff
@@ -1033,14 +1054,24 @@ public function several_random_sentences()
 
         $randomIds = $this->Sentence->getSeveralRandomIds($lang, $number);
 
-        $this->Session->write('random_lang_selected', $lang);
+        if(is_bool($randomIds) && !$randomIds) {
+            $searchDisabled = !Configure::read('Search.enabled');
+            $this->set('searchProblem', true);
+            if ($searchDisabled) {
```
trang: Your if's can be ordered like this:

- if $searchDisable ...
- else if $searchError ...
- else ...

Instead of:

- if $searchError
  - if $searchDisabled ...
  - else ...
- else ...

app/controllers/sentences_controller.php
```diff
@@ -1033,14 +1054,24 @@ public function several_random_sentences()
...
+                $this->Session->setFlash(__('Search was disabled. Please try again later.', true));
+            } else {
+                $this->Session->setFlash(__('An error occurred while performing this function.', true));
```
trang: It would be good if users would let us know when there's an error with Sphinx, so this text could be changed into:
`An error occurred. If the error persists, please <a href="{}">contact us</a>.`

Something similar to: https://github.com/Tatoeba/tatoeba2/blob/dev/app/views/sentences/search.ctp#L87

app/controllers/pages_controller.php
```diff
@@ -187,17 +187,22 @@ private function _random_sentence() {
...
-        $translations = $alltranslations['Translation'];
-        $indirectTranslations = $alltranslations['IndirectTranslation'];
-
-        $this->set('random', $randomSentence);
-        $this->set('translations', $translations);
-        $this->set('indirectTranslations', $indirectTranslations);
-
-        if (isset($randomSentence['Sentence']['script'])) {
-            $this->set('sentenceScript', $randomSentence['Sentence']['script']);
+        if (is_bool($randomId) && !$randomId) {
```
jiru: Sentence ids can’t be zero, so you might as well just do `if (!$randomId)`.

app/webroot/js/sentences.show_another.js
```diff
@@ -30,6 +30,13 @@ $(document).ready(function(){
 });
 
 function loadRandom(lang){
-    $(".random_sentences_set").html("<div class='block-loader loader'></div>");
-    $(".random_sentences_set").load("/sentences/random/" + lang);
+
+    //$(".random_sentences_set").html("<div class='block-loader loader'></div>");
+    $.get( "/sentences/random/" + lang, function( data ) {  
+        if($(data).find(".mainSentence > .content > .sentenceContent").text() == "") {
+            alert("There was an error in performing this function. Please try later.")
```
jiru: There is no need to use `alert()` here. Besides, raw strings in javascript files can’t be localized so it’s a no-go. I think you can just return the error as HTML, just like the random sentence is returned.

app/controllers/sentences_controller.php
kwikadi: If I set it to the structure you suggest, I will need to check if search is disabled every single time, whether or not getSeveralRandomIds returns a null. (Since search is almost never disabled, I think this will be worse for performance)

trang: Performa​n​ce is not really an issue ​in this case but I see the problem.

​There's an ambiguity in `getSeveralRandomIds()`. The function returns the same value whether the search is disabled or whether there's an error.

One possibility would be to return a ​certain value in case of disabled search, and another in case of search error.

But I would rather suggest to remove the `Configure::read('Search.enabled');` from `getSeveralRandomIds()` because it shouldn't be the job of the model to deal with the config.
Then in the controllers, wherever we retrieve random sentences, you would first check ​in the config ​if the search is disabled, and only when the search is not disabled, you would try to retrieve random sentences.

kwikadi: I DID return 'error' or 'disabled' in an earlier commit (I've squashed it now), but gillux recommended I do it this way.

trang: Well I will leave it to you and @jiru then :)

My comment was initially triggered by the fact that it felt strange to have `$searchDisabled` encapsulated inside `if (!$randomIds)` because if the search is disabled in the first place, you wouldn't even search for random ids.

The way things are now, you have a bit of redundancy since you're checking for Search.enable both in the model and in the controller.

But this is far from being a blocking issue to me. In the end the code does what's expected.

PR #1007, opened 2016-02-13
Keyword highlight in search results and refactoring

jiru:
This PR adds keyword highlighting to search results, along with some refactoring.

The previous way of displaying a list of sentences and their translations was:
```php
// from a controller
$ids = $this->Sentence->getSomeSentenceIds($someConditions);
$results = $this->CommonSentence->getAllNeededForSentences($ids);
```
This resulted into `1 + number_of_sentences * 6` SQL requests: one to get the ids, and 6 to get sentence, favorites, users, lists, transcriptions and transcriptions authors data for each sentence. And it wasn’t very CakePHPish.

This PR makes it this way:
```php
// from a controller
$results = $this->Sentence->find('all', array(
    'conditions' => $someConditions,
    'contain' => array('Translation', 'User', 'SentencesList', ...)
));
```

Which results into 1 + 6 SQL requests, regardless of the number of sentences. CakePHP runs one request on each table and joins everything. In other words, the Translation model can now be added to the `contain` array just like other models.

Direct and indirect translations are no more returned separated into `Translations` and `IndirectTranslations` subarrays. Instead, all the translations are returned grouped, and each translation have an additional `type` field (0 or 1) that says whether it’s direct or indirect. Separating direct and indirect translations is done at the very end, in the view.

Filtering the returned translations by language isn’t very straightforward, but… It works that way:
```php
$this->Sentence->linkTranslationModel(array(
    'Translation.lang' => $lang
));
$results = $this->Sentence->find('all', ...);
```
You need to change the conditions under which the Translation model links with the Sentence one beforehand. This is because linking is performed with a precomputed SQL query that is used as-is by `find()`.

I also converted some of the raw SQL queries of `sentence_not_translated_into.php` to proper `find()` calls in the process.

Other than that, I replaced all the `contain => array()` or `'recursive' => -1` parameters of `find()` calls by a single `public $recursive = -1;` in `app_model.php`. This way, `find()` just returns no attached model unless explicitly specified in the `contain` parameter.

Everything seems to be working fine, but since this PR messes a lot with the core code, please test it thoroughly.

### DIFF
--- a/app/app_model.php
+++ b/app/app_model.php
@@ -17,6 +17,9 @@
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 class AppModel extends Model {
+
+    public $recursive = -1;
+
     /**
      * Function to be used in beforeSave().
      * Checks if the save() is about to update any of the provided $fields.
--- a/app/controllers/activities_controller.php
+++ b/app/controllers/activities_controller.php
@@ -196,7 +196,6 @@ class ActivitiesController extends AppController
         );
 
         $results = $this->paginate('Sentence');
-        $this->Sentence->addTranslationsToSentences($results);
 
         $this->set('results', $results);
         $this->set('lang', $lang);
--- a/app/controllers/activities_controller.php
+++ b/app/controllers/activities_controller.php
@@ -187,27 +187,16 @@ class ActivitiesController extends AppController
 
         $this->paginate = array(
             'Sentence' => array(
-                'fields' => array(
-                    'id',
-                ),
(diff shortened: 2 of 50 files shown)


RyckRichards:
Sorry, I changed by accident

trang:
I have tested the major features, it looks OK to me.

@jiru, I've put this PR in today's milestone but you can do the Tatoeba update whenever you have time (today, tomorrow, Friday). If you'd prefer that I'm around for possible hotfixes, then better do it on Saturday evening.

PR #1019, opened 2016-02-25
Sphinx correct output when disabled/down

kwikadi:
I had to close the other PR (#1006) because I ended up making up a new branch (because of merge conflicts and whatnot).


app/controllers/sentences_controller.php
```diff
@@ -216,8 +217,21 @@ public function show($id = null)
...
-            $this->Session->write('random_lang_selected', $lang);
-            $this->redirect(array("action"=>"show", $randomId));
+            if (is_null($randomId)) {
+                $searchDisabled = !Configure::read('Search.enabled');
+                $this->set('searchProblem', true);
+                if ($searchDisabled) {
+                    $this->Session->setFlash(__('Search was disabled. Please try again later.', true));
+                } else {
+                    $html = new HtmlHelper();
+                    $this->Session->setFlash(format(__('An error occurred while performing this function.
```
jiru: Please use a more specific message, like `An error occurred while retrieving a random sentence.`

app/controllers/sentences_controller.php
```diff
@@ -216,8 +217,21 @@ public function show($id = null)
...
+            if (is_null($randomId)) {
+                $searchDisabled = !Configure::read('Search.enabled');
+                $this->set('searchProblem', true);
+                if ($searchDisabled) {
+                    $this->Session->setFlash(__('Search was disabled. Please try again later.', true));
+                } else {
+                    $html = new HtmlHelper();
+                    $this->Session->setFlash(format(__('An error occurred while performing this function. 
+                            If the error persists, please <a href="{}">let us know</a>.', true),
+                        $html->url(array("controller"=>"pages", "action" => "contact"))));
```
jiru: You’re messing with MVC hard here. :-)

One way to avoid this is to print the message directly from the view, instead of using flash messages. See this [example](https://github.com/Tatoeba/tatoeba2/blob/dev/app/views/transcriptions/view.ctp).

app/controllers/sentences_controller.php
```diff
@@ -960,10 +974,36 @@ public function random($lang = null)
...
+                                        (
+                                            "id" => 0,
+                                            "username" => "Tatoeba",
+                                            "level" => 0
+                                        ),
+
+                                    "Translation" => array(),
+
+                                    "Transcription" => array()
+                                );
```
jiru: LOL. Did you really mean to commit that?

app/views/helpers/sentences.php
```diff
@@ -444,9 +444,11 @@ public function displayMainSentence(
...
             $script = $sentence['script'];
         }
-        $this->Menu->displayMenu(
-            $sentenceId, $user, $script, $canTranslate, $langFilter, $hasAudio
-        );
+        if ($user['id']!=0) {
+            $this->Menu->displayMenu(
+                $sentenceId, $user, $script, $canTranslate, $langFilter, $hasAudio
+            );
+        }
```
jiru: I don’t think you meant to commit this part.

jiru:
Thank you, kwikadi. Except the few things I mentioned, it’s pretty good. :-)

kwikadi:
I didn't quite understand the part about the transcriptions that you mentioned, but I removed the message from the controller to the view. Fixed the rest too. If there are still issues, please let me know :)

app/views/sentences/several_random_sentences.ctp
```diff
@@ -102,9 +102,28 @@ $this->set('title_for_layout', $pages->formatTitle(__('Random sentences', true))
...
+} else {
+?>
+    <div id="main_content">
+        <div class="module">
+            <?php
+            if($searchProblem == 'disabled') {
+                echo $html->tag('h2', __('Search is currently disabled, please try again later.', true));
+            } else if ($searchProblem == 'error') {
+                echo $html->tag('h2', format(__('An error occurred while fetching random sentences. 
+                        If this persists, please <a href="{}">let us know</a>.', true),
```
jiru: This string spans on two lines, which unnecessarily adds a newline and a lot of spaces between the two sentences. Please use concatenation instead:
```php
'First sentence. '. # note the leading space inside the string here
'And this is the '. # and here too
'second sentence'
```

app/views/sentences/show.ctp
```diff
@@ -233,4 +234,22 @@ $navigation->displaySentenceNavigation(
...
+} else {
+?>
+    <div id="main_content">
+        <div class="module">
+            <?php
+            if($searchProblem == 'disabled') {
+                echo $html->tag('h2', __('Search is currently disabled, please try again later.', true));
+            } else if ($searchProblem == 'error') {
+                echo $html->tag('h2', format(__('An error occurred while fetching random sentences. 
+                        If this persists, please <a href="{}">let us know</a>.', true),
```
jiru: Same comment here.

app/views/sentences/several_random_sentences.ctp
```diff
@@ -26,7 +26,7 @@
  */
 
 $this->set('title_for_layout', $pages->formatTitle(__('Random sentences', true)));
-
+if(!isset($searchProblem)) {
 ?>
 <div id="annexe_content">
     <div class="module">
```
jiru: The annexe content should be left visible regardless of the error, so that users can click the button to try again.

kwikadi:
Done :)

jiru:
Another minor problem: the h2 tag is not appropriate everywhere you used it. A simple p should be enough. If you want to make it bigger, style it with a CSS class.

app/views/sentences/several_random_sentences.ctp
```diff
@@ -90,21 +89,43 @@ $this->set('title_for_layout', $pages->formatTitle(__('Random sentences', true))
...
-                $sentence['Translation'],
-                $sentence['User']
-            );
+<?php
+if(!isset($searchProblem)) {
+?>  
+    <div id="main_content">
+        <div class="module">
+        <?php  
+            $this->Html->tag('h2', 'Random sentences', null);
```
jiru: Up to here, the code can be taken out of the `if()` block to avoid duplication with the corresponding `else`.

app/controllers/sentences_controller.php
```diff
@@ -1008,10 +1022,19 @@ public function several_random_sentences()
...
         ));
 
-        $this->set("allSentences", $allSentences);
+        if (empty($allSentences)) {
+                $searchDisabled = !Configure::read('Search.enabled');
+                if ($searchDisabled) {
+                    $this->set('searchProblem', 'disabled');
+                } else {
+                    $this->set('searchProblem', 'error');
+                }
```
jiru: Strange indentation here

app/views/sentences/several_random_sentences.ctp
```diff
@@ -89,22 +88,30 @@ $this->set('title_for_layout', $pages->formatTitle(__('Random sentences', true))
...
+                    $sentence['Translation'],
+                    $sentence['User']
+                );
+            }
+        } else if($searchProblem == 'disabled') {
+                echo $html->tag('p', __('The random sentence feature is currently disabled, please try again later.', true));
+        } else if ($searchProblem == 'error') {
+            echo $html->tag('p', format(__('An error occurred while fetching random sentences. '.
+                    'If this persists, please <a href="{}">let us know</a>.', true),
+                $html->url(array("controller"=>"pages", "action" => "contact"))
```
jiru: Please fix indentation here too.

app/views/sentences/show.ctp
```diff
@@ -233,4 +234,23 @@ $navigation->displaySentenceNavigation(
...
+    <div id="main_content">
+        <div class="module">
+            <?php
+            echo $html->tag('h2', __('Random Sentence',true));
+            if($searchProblem == 'disabled') {
+                echo $html->tag('p', __('The random sentence feature is currently disabled, please try again later.', true));
+            } else if ($searchProblem == 'error') {
+                echo $html->tag('p', format(__('An error occurred while fetching random sentences. '. 
+                        'If this persists, please <a href="{}">let us know</a>.', true),
+                    $html->url(array("controller"=>"pages", "action" => "contact"))
```
jiru: And here too.

PR #1022, opened 2016-02-25
New languages

RyckRichards:
> Changed 'Assyrian' by 'Assyrian Neo-Aramaic"
> Changed the UKR flag
> Added Morisyen


jiru:
It looks like the correct code for Morisyen is mfe, while the flag file name is mve.png.

jiru:
I fixed the flag name so that we can have Morisyen in today’s update.

Issue #947, opened 2015-12-04
Create a home page that is more useful for the first-time visitor.


ckjpn:
Create a home page that is more useful for the first-time visitor.

What follows are just a few ideas and not necessarily the best ideas.
However, if we start brainstorming, perhaps we can come up with something better.
Currently, if I send the website URL to someone and they arrive at the site, I don't think they have enough guidance to really understand what the website is all about.

** The Problem **

The current home page for non-logged-in members ...
1. ... makes it look like registration is necessary.
2. ... always requires a lot of scrolling to see the main content for people not logged in.
3. ... has a link that says "How it works", but really only links to a quick-start guide for people who have already become members.
4. ... has an "About us" link that goes to an old video that doesn't really explain the Tatoeba Project very clearly.
5. ... doesn't really explain that much to the first-time visitor, even if they take the time to click the links on the right.

** Some Suggested Solutions **

ONE: Include one sentence that concisely tells what the Tatoeba Project does.

The Tatoeba Project is a group of people working on creating a collection of sentences and their translations that can used by various projects under the Creative Commons - Attribution 2.0 license.

(Perhaps this could be re-worded.)

TWO: Clearly state what can be done without registering, and explain concisely what can be done after registering.

You can browse, search or download our data without becoming a member, or you can register and help us by translating from a foreign language you know into your own native language or by creating natural-sounding sentences in your own native language for others to translate into their native languages.

(Perhaps this could be re-worded.)

THREE: Possibly have this message dismissible like the blue "announcement" thing that has recently been introduced.  This would mean a non-logged-in visitor is not constantly irritated by this message.

FOUR:  Create new content for the "how it works" and "about us" links that is more useful for first-time visitors.

** Sample Layouts **

Example 1 - same size buttons.png
![example 1 - same size buttons](https://cloud.githubusercontent.com/assets/7934273/11579368/d9c82cda-9a70-11e5-8b9c-1939cf851134.png)

Example 2 - smaller buttons - smaller fonts.png
![example 2 - smaller buttons - smaller fonts](https://cloud.githubusercontent.com/assets/7934273/11579373/dfed30d8-9a70-11e5-9915-e59e7e5f1129.png)

Example 3 - text order reversed.png
![example 3 - text order reversed](https://cloud.githubusercontent.com/assets/7934273/11579378/eb0469d2-9a70-11e5-9e54-3cd1ed3eebed.png)

Original for comparison.png
[...]

PR #1053, opened 2016-03-09
New homepage, layout, menu and footer

trang:
This pull request addresses issue #947.

The homepage has been redesigned based on [@billoliver and @kamillel's work](https://github.com/Tatoeba/tatoeba2/issues/947#issuecomment-192708682). This redesign implied changing the general layout and the organization of the menu and footer.

-----

Some comments about the implementation:

(1) I skipped certain things because of lack of time:
  * the Twitter and Facebook icons --> @kamillel said it's fine to leave them as text links.
  * the number of sentences with audio --> nobody complained so I'm not going to add this.

(2) I skipped some other things because I didn't want redesign core design elements in the scope of this issue: the design of the dropdowns, the design of the modules in the content (there's no bottom border), the fonts.
--> @kamillel agreed it's better to leave them as they are, to align with the rest of the site.

(3) The search bar on the rest of the pages has been slightly redesigned to match the new search bar on the homepage. It was just a matter of changing some colors, as [illustrated by billoliver](https://github.com/Tatoeba/tatoeba2/issues/947#issuecomment-193024810).

(4) The item "Any" in the language dropdown has been modified into "Any language" as suggested by CK.

(5) CK made various [other suggestions](https://github.com/Tatoeba/tatoeba2/issues/947#issuecomment-193021043) that I did not take into account because I didn't feel they were a problem and nobody expressed they were in favor applying of those changes.

-----

I'll bump this topic on Tatoeba's Wall one last time and if there's no complaint I'll merge this so it can go into this Saturday's update.

### DIFF
--- a/app/controllers/pages_controller.php
+++ b/app/controllers/pages_controller.php
@@ -91,6 +91,13 @@ class PagesController extends AppController
         // Random sentence part
         $this->_random_sentence();
 
+        // Stats
+        $stats = ClassRegistry::init('Language')->getSentencesStatistics(5);
+        $numSentences = ClassRegistry::init('Sentence')->getTotalNumberOfSentences();
+
+        $this->set('stats', $stats);
+        $this->set('numSentences', $numSentences);
+
         if ($isLogged) {
             $this->_homepageForMembers();
         } else {
@@ -100,6 +107,9 @@ class PagesController extends AppController
 
 
     private function _homepageForGuests() {
+        $contribToday = ClassRegistry::init('Contribution')->getTodayContributions();
+
+        $this->set('contribToday', $contribToday);
         $this->render('index_for_guests');
     }
 
--- a/app/controllers/pages_controller.php
+++ b/app/controllers/pages_controller.php
@@ -83,12 +83,29 @@ class PagesController extends AppController
      */
     public function index()
     {
-        $this->helpers[] = 'Wall';
         $this->helpers[] = 'Sentences';
-        $this->helpers[] = 'Messages';
 
         $userId = $this->Auth->user('id');
         $isLogged = !empty($userId);
+
+        // Random sentence part
+        $this->_random_sentence();
+
+        if ($isLogged) {
+            $this->_homepageForMembers();
+        } else {
+            $this->_homepageForGuests();
+        }
+    }
+
+
+    private function _homepageForGuests() {
+        $this->render('index_for_guests');
+    }
+
+    private function _homepageForMembers() {
+        $this->helpers[] = 'Wall';
+        $this->helpers[] = 'Messages';
 
         /*latest comments part */
         $this->loadModel('SentenceComment');
@@ -105,11 +122,7 @@ class PagesController extends AppController
         $this->loadModel('Wall');
         $latestMessages = $this->Wall->getLastMessages(5);
 
-        $this->set('isLogged', $isLogged);
         $this->set('latestMessages', $latestMessages);
-
-        // Random sentence part
-        $this->_random_sentence();
     }
 
     /**
--- a/app/locale/default.pot
+++ b/app/locale/default.pot
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2016-03-07 21:29+0000\n"
+"POT-Creation-Date: 2016-03-09 15:19+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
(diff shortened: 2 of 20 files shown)


jiru:
I’m probably late to the party, but I feel this new homepage is not green enough. Today, cueyayotl [wrote](https://tatoeba.org/wall/show_message/25771#message_25771):
> The flags we have now are very pretty; anyone who walks behind me at work and sees all the flags (and characteristic green header) does a double-take.

The previous green header made Tatoeba different from any other website. There is nothing wrong with changing colors, but now, it looks too much like any other website to me.

Guybrush88:
I'm reporting also here what I just wrote on the wall:
personally I'd like to see the color around the search bar a bit darker, at first I had a little bit of difficulty in seeing where to click on the search bar to type anything to search

billoliver:
> The previous green header made Tatoeba different from any other website. There is nothing wrong with changing colors, but now, it looks too much like any other website to me.

Do you have any suggestions? We can change the color of the menu to green, but the problem is that the logo is green too.

> personally I'd like to see the color around the search bar a bit darker, at first I had a little bit of difficulty in seeing where to click on the search bar to type anything to search

@Guybrush88 Are  you using a Mac?

We can the color of the search bar section to light green. What do you think @trang?

trang:
@jiru, I understand perfectly, I had the same feeling while implementing the new homepage. The way Tatoeba currently looks (with the green header) certainly makes it more unique. But if we keep the search bar green, we would need to consider having the green background on the new homepage as well, because otherwise the transition from the new homepage to the search result feels odd.

But we're entering here the realm of graphics design. As much as I'd love Tatoeba to have its own GUI identity that would distinguish it from the other regular bootstrap-ish websites, I think that's quite a luxury. I hope we can catch the interest of graphics designers in the future but for now I'm fine with going for a regular but still neat looking design.

@Guybrush88, that's true. The border is actually the same color as the background so it's not visible. I'll make it a bit darker.

@billoliver, I agree that we can't make the menu green. It's definitely a problem because the logo is green. 
As for having the search bar section in light green, to be honest if you want to discuss colors with me, we're gonna be talking for months because that's how difficult it is for me to figure out these things.
I think the current grey background for the search bar is fine because it has the advantage that you can distinguish links (green) from labels (black). In the current search bar, both the links and the labels are white. I didn't like it, but I couldn't find a different color for links and labels. So I ended up adding little triangles in from the links, to make it look different from labels. 
Designing on a green background is very hard... If there's any good suggestion by Saturday I'll take it but I don't think we should delay the release for this.

billoliver:
I found a little bug on this page http://dev.tatoeba.org/eng/users/all
bug1

@trang  Can you change the font size of the Stats section to 14px? I'm talking about the links. :+1:

jiru:
@trang I see. If we turn the search bar into green, how about making the description text area green on the homepage:

green-homepage

trang:
That could work. I still feel an inconsistency because the initial search bar is not on a green background, but it definitely feels less intense than the inconsistency of switching from grey background to green background.

I'm not sure if the tagline text should then be in black or in white.
It feels less readable in black because the green color is more of a dark color.
But it might feel like a big button if the text is in white. I'd probably still pick white.

kamillel:
Why not keep the search bar background green?
I think the homepage layout was bit of mixed of design and wireframes and it would be a shame to lost some of the 'playful' look and feel Tatoeba has, even if later one it would make sense to make a more in-depth review of the visual design.

I made a super quick test with a green background for the search and I think it works because it also adds some visual hierachy (so the tagline is separate from the search) and it makes the search bar easier to see (so we won't have the light grey on white issue)
test1

trang:
> Why not keep the search bar background green?

It sounds more logical, but right now I can't visualize completely how it feels in practice. Maybe it give an awkward feeling to have a displacement of the green area. I'll have to try it out.

But if we go for this, I would need to know how to represent the links on the green background. By default I'll go with white color + little triangle on the left-side, like in our current green search bar.

kamillel:
Good point, if it's the case, one solution could be to include the tagline in the green area as well, so it will just feel like the area in reducing (sorry the only online example I could think of: http://www.just-eat.co.uk/) 

test2

jiru:
@kamillel Yes, all-green sounds good too, although I find it a little bit too provocative, too much “look at me!”.

I pushed my tentative changes to dev: green background behind the description only and previous green search bar on other pages. It’s now easier to modify the homepage backgrounds, to assign green to either the description, the search form or both. Feel free to experiment.

@trang To me, black text reads better than white over a green background, which is why I initially set it black. (But maybe that’s because of my screen.) Anyway, I realized it doesn’t fit with the rest of the website, so I made it white. I don’t think it can be seen as a big button because it’s spans horizontally, it’s more like a bar.

billoliver:
We can also change the color of the menu to white. What do you guys think?
white-menu

jiru:
@billoliver I think I prefer the the black menu because it makes it clearly different from the other black-on-white section (random sentence etc.). Your screenshot makes me feel the green area in just inserted in the middle of some text.

RyckRichards:
I'd prefer the menu using a darker green color (Maybe #006C00 ,
#148014 or #006200). They would display the Tatoeba logo and wouldn't
change the interface too much.

-- 
Ricardo Vernaut Jr

billoliver:
I found another small bug on the [search page ](http://dev.tatoeba.org/eng/sentences/search?query=Search&from=eng&to=und)
search-page-bug

Issue #1072, opened 2016-03-19
Misalignment of "Help", "Advanced Search" with "From" and "To"


alanfgh:
As this screenshot shows, the "Help" and "Advanced Search" captions are slightly misaligned with the "From" and "To" captions. I understand that they're conceptually different, but I still think it would look better if they were aligned. It would also be nice to move the search button below the fields so that it's not right up against them. Naturally, when the display is wide enough to show the search button to the right, this is not a problem.

This screenshot is from Firefox, but it looks the same on Chrome. I'm using a Windows 7 machine.

![misalignment](https://cloud.githubusercontent.com/assets/1213027/13900202/c2754234-edd6-11e5-931e-0444d8fceba2.png)

billoliver:
These changes is going to solve the issue:

Someone just need to add this div in the HTML:
![print](https://cloud.githubusercontent.com/assets/17387379/13974212/8ede0640-f088-11e5-91aa-6548e0fd3865.png)



Changes in the CSS:
```
html, body {
 box-sizing: border-box;
}

#top_menu_container {
    background-color: #414141;
    margin: 0 auto;
}

#top_menu {
    background: #414141;
    width: 960px;
    margin: 0 auto;
    overflow: hidden;
    padding: 10px;
}

#search_container {
    background-color: #6A0;
    margin: 0 auto;
}

.search_bar {
    color: #FFF;
    background-color: #6A0;
    position: relative;
    width: 960px;
    padding: 0 10px;
    margin: 0 auto;
}
```

PR #1075, opened 2016-03-23
Misalignment of the search bar on Chrome

billoliver:
This pull request addresses issue #1072

### DIFF
--- a/app/views/elements/search_bar.ctp
+++ b/app/views/elements/search_bar.ctp
@@ -30,6 +30,7 @@ if (isset($this->params['lang'])) {
 }
 ?>
 
+<div class="search_container">
 <div class="search_bar">
 <?php
 if ($selectedLanguageFrom == null) {
@@ -130,4 +131,5 @@ echo $form->create(
 <?php
 echo $form->end();
 ?>
+</div>
 </div>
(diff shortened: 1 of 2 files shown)


trang:
Thanks for the pull request @billoliver. Unfortunately I won't be able to accept it as is.

1. The problem of misalignment can be solved with just 1 line of code. You have made several modifications that seem to be solving something else.
2. If you have found other problems while trying to solve the initial problem and these other problems are not blocking for solving the initial problem, then it is better to create another pull request for these other problems.
3. If you feel it's not worth to create another pull request, at least split your changes into several commits: one commit that solve the problem of misalignment, and other commits that solve the other problems. Then explain in the pull request description what other fixes you've made.
4. For code indentation, we don't use tabs, but 4 spaces. To keep your code consistent with the rest of the code, please always make sure that you don't have tabs in there, but 4 spaces.

Note: you don't need to close this pull request (unless you want to give up on solving the issue). You can start things over simply by [resetting your branch](http://stackoverflow.com/questions/28774194/git-reset-revert-a-whole-branch-to-another-branches-state) to be like the `dev` branch.

trang:
@billoliver as agreed on IRC, I'm closing this pull request from you and will unassign you the related issue to make it available for other new developers :)

Issue #1084, opened 2016-03-29
Arabic (Gulf) (afb)


RyckRichards:
* Language name: Arabic (Gulf)
* Language info: https://en.wikipedia.org/wiki/Gulf_Arabic
* ISO 639-3 code: afb
* Public List: https://tatoeba.org/eng/sentences_lists/show/5877
* Icon: http://imgur.com/iWJGWZQ
* Requested by: https://tatoeba.org/eng/user/profile/cueyayotl

=========================================

[for dev.tatoeba.org]
CALL add_new_language('afb', 4235);

[for tatoeba.org]
CALL add_new_language('afb', 5877);

http://dev.tatoeba.org/eng/sentences_lists/show/4235

damascene:
To which delicate in Gulf this newly language belongs? is the next step is to create a British, Australian, South African and Canadian English to cover all the dialects?

RyckRichards:
As far Gulf Arabic has an ISO 639-3 code and some members were posting
sentences in this language and requested this language, it was added.
British/Canadian  English dont have it

damascene:
well it's slang. Arabic (ara) is the official language for all the Arabic countries and the language used in the Arab. there is nothing called Gulf Arabic there. I've lived in the Gulf.

And the highest Arabic institution is the Arab League where leaders from all Arab countries gather has the official language as Arabic (ara) https://en.wikipedia.org/wiki/Arab_League

the sentences you added as I can see is slang. and there is different slang words in every Arabic country. Could you please provide more solid reason to have this language? could you invite those users who requested this language somewhere here?

trang:
@damascene, as of today, we don't take responsibility for language classification. We rely on the ISO 639-3 classification. As long as a language is present in the ISO 639-3 classification, we consider that it is fine to add it as a language in Tatoeba. 
Since Gulf Arabic is recognized as a language in this classification (represented by the code `afb`), by default we won't oppose to add it to Tatoeba.

If you think that Gulf Arabic should not be considered as a language, we would need some academic reference.

If you think that the sentences in the list are not representative of the Gulf Arabic language, that's another problem.

SafaAlfulaij:
After looking at the current statistics of Arabic "dialects" in Tatoeba, I can say that adding new dialects are useless, as these "languages" contains only few sentences. 
[...]

PR #1095, opened 2016-04-13
Ara rename

loolmeh:
While I'm still disillusioned by the discussion on #1084 which amounts to a projection of a disgraceful and malicious victimhood mentality on none other than a technical issue tracker, I think something can be done to more explicitly indicate that Tatoeba isn't in the business of making political decisions.

This pull request has a few sql update statements to relabel ara and relabels arabic corpora with the pattern Arabic Dialect (...). Please check if I missed a table that should be updated and if active contributors on these corpora object to this change. Also I'm not sure what the impact on urls is gonna be from these changes for seo purposes.

@trang I think you should reconsider the current wave of migration of nontechnical discussions to github instead of confining them to the wall.

### DIFF
--- a/app/vendors/languages_lib.php
+++ b/app/vendors/languages_lib.php
@@ -50,7 +50,7 @@ class LanguagesLib
             'abk' => 'ab', // Abkhaz
             'afr' => 'af', // Afrikaans
             'amh' => 'am', // Amharic
-            'ara' => 'ar', // Arabic
+            'arb' => 'ar', // Arabic
             'arg' => 'an', // Aragonese
             'asm' => 'as', // Assamese
             'aym' => 'ay', // Aymara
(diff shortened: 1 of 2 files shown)


cueyayotl:
Right, Tatoeba is definitely not in the business of making political decisions, rather we follow ISO 639-3 where the SIL (Summer Institute of Linguistics) have made those decisions for us.

Being from Mexico, I must say how much I despise the word "dialect" as it is used in our country to mean inferior or undesirable language and usually is used to refer to Amerindian languages exclusively. The distinction between what constitutes a language and a dialect varies from source to source, which is why we have decided to use the ISO 639-3 classification.

Now, as for what we shall then call `afb`, I am EXTREMELY against using the actual word "dialect" in the actual name of a language, but I fear it may be our best choice in this case. "Arabic Dialect (Gulf)" may work for `afb`, however we cannot simply change all Arabic languages to this format, as attitudes differ in different regions that speak Arabic languages. In my experience, Egyptians call their language "Egyptian", and not even "Egyptian Arabic". I wonder how much renaming `arz` to "Arabic Dialect (Egyptian)" could hurt our chances of having native speakers; could it be that they hate the word "dialect" as much as I do when referring to their language? We should keep the name as either "Egyptian Arabic" or "Arabic (Egyptian)" The renaming of Arabic languages has to be on a language by language basis.

I will post this discussion on the Wall.

loolmeh:
seems like the response from contributors to the arabic corpora on this issue has been quite underwhelming, closing this for now

Issue #1106, opened 2016-04-22
afb and arb are not right-to-left


RyckRichards:
Arabic (Gulf) -afb -  on Dev Tatoeba - 
http://dev.tatoeba.org/eng/sentences_lists/show/4235

Arabic (Modern Standard) arb -
http://dev.tatoeba.org/eng/sentences_lists/show/4231

trang:
@RyckRichards, 

1) When an issue is about something that has not been released yet, don't create a new issue. If you have to make a fix, simply reference the related issues that are still open.

2) We didn't implement `arb` yet. As I've said in https://github.com/Tatoeba/tatoeba2/issues/1079#issuecomment-207817059, I have reverted your commit because we won't add `afb` as a new language, but simply replace the language code `ara` by `arb`.

PR #1107, opened 2016-04-22
Corrected afb and arb to be right to left

RyckRichards:
This pull request addresses issue #1106.


trang:
As mentioned in https://github.com/Tatoeba/tatoeba2/issues/1106#issuecomment-213754170, we don't have `arb` yet, therefore this pull request is not entirely valid.

Issue #736, opened 2015-08-02
Search button (with magnifying glass) doesn't look clickable


alanfgh:
The search button with the magnifying glass looks very flat. If I weren't already familiar with the interface, I wouldn't be sure that I had to click it. By contrast, the buttons for the pull-down lists next to it have a graded wash, so they look more three-dimensional and clickable.

![flat_button](https://cloud.githubusercontent.com/assets/1213027/9024445/62573658-389d-11e5-90ff-35b4829ad2d5.png)

billoliver:
New button.
```
.search_bar
.search-submit-button {
  color: #fff;
  cursor: pointer;
  background: #263238;
  width: 70px;
  height: 30px;
  border: none;
  border-radius: 2px;
  padding: 2px 0 4px 0;
}

button.search-submit-button:hover {
  background: #37474f;
}
```

alanfgh:
Thanks, Bill! When it's on the dev site, I'll take a look.

billoliver:
@alanfgh No problem. Here it is:
![search-button](https://cloud.githubusercontent.com/assets/17387379/13907373/66378f22-eecb-11e5-9d8c-5ab1c753eeb6.png)

trang:
@billoliver rather than changing the color, how about simply adding a border or a shadow, or both? I don't think that changing the color will make the button feel more like a button.

billoliver:
> @billoliver rather than changing the color, how about simply adding a border or a shadow, or both? I don't think that changing the color will make the button feel more like a button.

I saw this coming :smile:

I didn't add a shadow because as I said in my previous comment, it doesn't work on every browser. But I'm going to make a better one.

trang:
You talked about gradient, not shadow. 

> I could have used a linear-gradient, but it doesn't work on every browser.

Shadow seems to be a bit more supported than gradient:
- http://caniuse.com/#search=box-shadow (works in IE 9)
- http://caniuse.com/#search=css%20gradient (doesn't work in IE 9)

billoliver:
Okay, this one looks more "clickable".
![button](https://cloud.githubusercontent.com/assets/17387379/13919659/9039f364-ef4d-11e5-8118-e31617a3b0c7.png)


```
.search_bar .search-submit-button {
  cursor: pointer;
  width: 70px;
  height: 30px;
  padding: 2px 0 4px 0;
  background: rgb(255, 255, 255);
[...]

PR #1115, opened 2016-04-24
make search button more look clickable

fadhelafif:
This pull request addresses issue #736

### DIFF
--- a/app/webroot/css/layouts/default.css
+++ b/app/webroot/css/layouts/default.css
@@ -6,7 +6,7 @@
  * it under the terms of the GNU Affero General Public License as published by
  * the Free Software Foundation, either version 3 of the License, or
  * (at your option) any later version.
- *
+ * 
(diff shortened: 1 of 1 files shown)


trang:
Thanks for your pull request @fadhelafif. Here's my feedback:

1) The changes you've made don't follow the same color as the screenshot in https://github.com/Tatoeba/tatoeba2/issues/736#issuecomment-199268460. Your button has a red-ish gradient (http://prntscr.com/awd0rw). If you believe the solution chosen in #736 is not appropriate you are free to comment on the issue and we can discuss to find another solution. But otherwise your pull request should stick to the same colors suggested in that issue.

2) You commit includes the removal of trailing spaces. Nothing wrong with removing trailing spaces, but it is better to make have this type of modification in a separate commit, otherwise they make the commit less readable. I assume this was made automatically by your text editor or IDE. Try to find the option to disable it.

Issue #1012, opened 2016-02-21
Put a link to https://tatoeba.org/stats/sentences_by_language on the "settings" page.


ckjpn:
Someone on the IRC today was asking about the ISO 639-3 codes needed for the "Languages" setting.
It might be useful to have a link there, so people can easily find the codes we use.

![screen shot 2016-02-21 at 17 06 34](https://cloud.githubusercontent.com/assets/7934273/13201638/edc47a64-d8bd-11e5-97e8-dac309c2cb5d.png)

https://tatoeba.org/user/settings
https://tatoeba.org/stats/sentences_by_language

damascene:
exactly thank you for taking a screen shot and explaining it

PR #1118, opened 2016-04-26
put a link to sentences_by_language on the user setting page

fadhelafif:
This pull request addresses issue #1012

### DIFF
--- a/app/views/user/settings.ctp
+++ b/app/views/user/settings.ctp
@@ -85,11 +85,18 @@ $this->set('title_for_layout', $pages->formatTitle(__('Settings', true)));
             <div>
                 <?php
                 $tip = __(
-                    'Enter ISO 639-3 codes, separated with a comma (e.g.: jpn,epo,ara,deu). '.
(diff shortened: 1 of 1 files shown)


trang:
Thanks again for your pull request. Your solution is not incorrect, but it is not optimal for the UI translators.

In your solution, the text is divided into 2 string:
1) `Enter %s, separated with a comma (e.g.: jpn,epo,ara,deu)...`
2) `ISO 639-3 codes`

The problem is that UI translators will not necessarily be aware that these two strings are connected, i.e. that `ISO 639-3 codes` is used in `Enter %s, separated with a comma (e.g.: jpn,epo,ara,deu)...`. Therefore they may translate `ISO 639-3 codes` incorrectly.

Also, it's not `ISO 639-3 codes` that is variable. It's only the URL that is variable.

You can find here an example of a solution that would keep the whole text into one string:
https://github.com/Tatoeba/tatoeba2/blob/dev/app/views/activities/adopt_sentences.ctp#L60

Issue #993, opened 2016-02-02
Sometimes it would be useful to have exact dates and times rather than times listed as "6 hours ago".


ckjpn:
I wonder if it would be possible to include exact times in places like this, since there are often cases where people reading comments and comparing them to the logs, tags, or ratings may find it useful to know which came first.    This is not the first time that I've wished I knew the exact time sometime was added.

Would it add too much clutter to the page to include something like this?

6 hours ago (2016-02-02 09:33:58)

.... or maybe this way.

2016-02-02 09:33:58 (6 hours ago)

... or maybe this way.

2016-02-02 09:33:58 GMT

... or maybe this way.

2016-02-02 09:33:58 GMT (6 hours ago)

My assumption is that this could be labeled "effort: low", but I'm not sure.

The question is this, I think.
Would the added benefit outweigh the clutter such text adds to the page?

Possibly the part in parentheses could also be a mouseover.  This way it wouldn't really clutter the surface of the page.

2016-02-02 09:33:58 GMT (6 hours ago)

![screen shot 2016-02-02 at 22 59 45](https://cloud.githubusercontent.com/assets/7934273/12751683/eef543d4-ca01-11e5-9888-ca7858d0cbae.png)

jiru:
If you look at how labels, milestones and assignations are displayed in Github issues, you’ll see that just don’t need to compare dates in the first place because the information is already displayed on a timeline. I think that’s be the proper way to solve this problem.

ckjpn:
> If you look at how labels, milestones and assignations are displayed in Github issues, ....

I'm not sure what you mean by this. Could you include a URL?

Does this mean you think comments, tagging, etc. should be added to the log on the right side of the page?

jiru:
> I'm not sure what you mean by this. Could you include a URL?

#993. You have a timeline that mixes comments, adding of a label, adding/removing of a milestone, and assignation.

> Does this mean you think comments, tagging, etc. should be added to the log on the right side of the page?

Currently there are two timelines on sentence pages. The comments on the left and the sentence log on the right. As a result, you need to go back and forth if you want to know what links the sentence had at the time a particular comment was posted, which is troublesome and unpractical, especially if the date is formatted like “X ago”. This would be much easier if there was a unique timeline combining comments and sentence log.

[...]

PR #1123, opened 2016-05-01
show the exact time in a tooltip

fadhelafif:
This pull request addresses issue #993

### DIFF
--- a/app/views/helpers/messages.php
+++ b/app/views/helpers/messages.php
@@ -182,7 +182,9 @@ class MessagesHelper extends AppHelper
 
                 <div class="date">
                 <?php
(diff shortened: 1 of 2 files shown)


trang:
I like the `cursor: help`. I didn't even know such thing existed.

There's just one small thing missing (which I'll take of, but just so you know): the timezone should be specified, because the time is not displayed in the user's timezone but in UTC.

It would be nice if the time could be displayed in the user's timezone, but that would be another issue.

fadhelafif:
I realized it when I create example comment for this issue. at that time i got confused when I see this screenshot

a35

trang:
I created this issue: #1125

Issue #1050, opened 2016-03-08
Sentences with slashes appear empty in search results


jiru:
[Original post](https://tatoeba.org/jpn/wall/show_message/25753#message_25753).

> I reported the bug below as a comment for sentence #4202165 but it might have been missed by the developers. Here it is again:

> This Turkish sentence #4202165 is obviously not empty. But if you search: https://tatoeba.org/eng/sentenc...tur&to=eng it shows an empty string for this sentence in the results. Similarly the Portuguese sentence #4795676 is not empty. But if you search:https://tatoeba.org/eng/sentenc...por&to=eng it shows an empty string for the said Portuguese sentence in the results.
 My guess is that there's a bug in the code which doesn't like the presence of the slash ("/") in the sentence.

ckjpn:
This should be reopened, perhaps.

https://dev.tatoeba.org/eng/sentences/search?query=24%2F7

![screen shot 2016-03-12 at 01 22 15](https://cloud.githubusercontent.com/assets/7934273/13708262/f1c91a4c-e7f0-11e5-9844-f8d8231cb2d4.png)

It does however, work on the dev site, if the item is quoted, even though it doesn't always give an exact match.

https://dev.tatoeba.org/eng/sentences/search?query="24%2F7"&

![screen shot 2016-03-12 at 01 24 56](https://cloud.githubusercontent.com/assets/7934273/13708320/4b901fda-e7f1-11e5-854f-85c8c1ebfd3f.png)

jiru:
I suspect this is because the slash, when not surrounded by spaces, is interpreted as part of the [NEAR operator](http://sphinxsearch.com/docs/current.html#extended-syntax).

ventice:
When / without quotes is searched, the following error occurs:

> Notice (1024): Search query failed: index abk_delta_index,..., ,zza_main_index: syntax error, unexpected '/' near '/7' [APP\models\behaviors\sphinx.php, line 123]

So, definitely this bug is related to translation of search query to sphinx query.
Taking it.

RyckRichards:
Working pretty well :D Thanks **a lot** for that!

ventice:
You are welcome (:

PR #1124, opened 2016-05-01
Escaped slashes in sphinx query and removed some code duplication

ventice:
This PR addresses issue #1050 
De facto I did the following:
1. Added `escapeQuery` function adding escaping backslash in front of slashes. Note that even if the slash is put in quotes, sphinx is okay with escapement, so complex parsing of the query in order to escape slashes only out of quotes is not necessary
2. Called `escapeQuery` before making search query and computation of excerpts
3. Removed some duplication of code by replacing `$this->runtime[$model->alias]['sphinx']` with temporary local variable `$sphinx`.
That's it.

**NB**: looks like the file has been written by different coding conventions by different people at different times so it was a little bit difficult to follow single style conforming with the file. I will refactor this file later to make its style consistent and remove duplication it currently contains.

### DIFF
--- a/app/models/behaviors/sphinx.php
+++ b/app/models/behaviors/sphinx.php
@@ -68,16 +68,16 @@ class SphinxBehavior extends ModelBehavior
             $query['page'] = 1;
         }
 
+        $sphinx = $this->runtime[$model->alias]['sphinx'];
         foreach ($query['sphinx'] as $key => $setting)
         {
-
             switch ($key)
             {
                 case 'filter':
                     foreach ($setting as $arg)
                     {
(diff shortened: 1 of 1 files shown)


trang:
Thanks for this PR. Looks good to me.

Regarding the coding style, if you wonder where to put brackets and spaces for functions and if/else, in most of the rest of the code it should look like this:

```
function someFunction($param1, $param2)
{
    if ($condition1) {
        //some code
    } else if ($condition2) {
        // some code
    } else {
        // some code
    }
}
```

I'll try to document this some time next week. For now don't worry too much about it :)

Issue #741, opened 2015-08-14
Remove suspended users from "Languages of members" and "Native speakers"


alanfgh:

trang:
**Clarification of this issue**

* The "Languages of members" page can be accessed from the menu "Community" > "Languages of members"
* The "Native speakers" page can be access from Community" > "Native speakers".

These pages display information about the languages of the members in Tatoeba.

At the moment, these pages include as well members who have been suspended (they can no longer log in with their account). It would be preferable not to display these members, because they are no longer part of the community and they may also have false information about the languages that they know.

The original issue was also mentioning to remove "blocked users", i.e. users who can still log into Tatoeba but are temporarily blocked from adding new sentences or translations. Since it's possible that these members become normal contributors again, I think there is no need to exclude them from these pages (hence I changed the title).

PR #1127, opened 2016-05-01
remove suspended users from languages of members

fadhelafif:
This pull request addresses issue #741

### DIFF
--- a/app/models/users_languages.php
+++ b/app/models/users_languages.php
@@ -161,7 +161,7 @@ class UsersLanguages extends AppModel
         $result = array(
             'conditions' => array(
                 'language_code' => $lang,
-                'User.group_id != 6'
+                'User.group_id NOT' => array(5,6)
             ),
             'fields' => array(
                 'of_user_id',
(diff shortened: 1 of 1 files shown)


app/models/users_languages.php
```diff
@@ -160,7 +160,8 @@ public function getUsersForLanguage($lang)
     {
         $result = array(
             'conditions' => array(
-                'language_code' => $lang
+                'language_code' => $lang,
+                'User.group_id != 6'
```
trang: Actually I realize there is something missing in the original issue: we want to exclude as well `group_id = 5`.

Functionally speaking, we use `6` for people who are suspended because they did something wrong, and we use `5` for people who ask to have their account deleted but have done nothing wrong (they just don't want to contribute anymore).

app/models/users_languages.php
```diff
@@ -190,7 +191,15 @@ public function getNumberOfUsersForEachLanguage()
             array(
                 'fields' => array('language_code', 'COUNT(*) as total'),
                 'group' => 'language_code',
-                'order' => 'total DESC'
+                'conditions' => array(
+                    'User.group_id != 6'
+                ),
+                'order' => 'total DESC',
+                'contain' => array(
+                    'User' => array(
```
trang: Here you can just do: `'contain' => array('User')`. There's no need to add the an empty `fields` array.

fadhelafif: Thanks for your advice.

app/models/users_languages.php
fadhelafif: Ok, I will exclude deleted account too.

Issue #307, opened 2014-06-08
Unclear "Click to edit" message when user tries to delete all sentence content


alanfgh:
When a user tries to delete all the content of a sentence (or all the content other than spaces), the phrase "Click to edit" is displayed where the content was. In fact, no content has been removed, but the user can't tell that unless s/he refreshes the screen. We could definitely have a more informative display.

theblindspring:
What kind of error/alert would we like to give? Why not just put the original sentence back into the content if the user tries to edit and save it as empty?

trang:
Yes, we just put the original sentence back.

PR #1136, opened 2016-05-06
Unclear "Click to edit" message when user tries to delete all sentence content

theblindspring:
This pull request addresses issue #307.

I still am not sure entirely of the spacing format, so I am sure it is incorrect at the moment. Would be happy to fix it if I know what to fix!

### DIFF
--- a/app/webroot/js/sentences.edit_in_place.js
+++ b/app/webroot/js/sentences.edit_in_place.js
@@ -28,6 +28,13 @@ $(document).ready(function() {
         div.editable(rootUrl + '/sentences/edit_sentence', {
             type      : 'textarea',
             submit    : div.attr('data-submit'),
+            onsubmit  : function () {
+                if($(this).find('textarea').val().trim().length == 0){
+                    return false;
+                } else {
+                    return true;   
+                }
+            },
             cancel    : div.attr('data-cancel'),
             event     : 'edit_sentence',
             data : function(value, settings) {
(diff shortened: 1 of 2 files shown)


trang:
@theblindspring I've added an article in the wiki regarding the code style: https://github.com/Tatoeba/tatoeba2/wiki/Code-style

Although your solution works, it's not the most optimal. There's no need to change the controller. Instead, you need to work on the Javascript side so that submitting an empty value is equivalent to clicking on "Cancel".

But there's actually another solution, which I didn't think of until I reviewed your pull request: we could disable the "OK" button if the input is empty.

I think disabling the "OK" button would feel more "standard" UX-wise. It gives a clear hint to the user that they can't submit an empty value. While allowing them to click "OK" but not erasing the sentence might give the impression that there's a bug.

theblindspring:
OK cool I should be able to make that work.

theblindspring:
So, in addition to disabling the "OK" button for submit when the input is empty, we still must handle the submit event, correct? At the moment the user can submit the edit by pressing the enter key also, so disabling the OK button only prevents them from submitting the form with a button, but there are still other ways to submit the change.

trang:
> So, in addition to disabling the "OK" button for submit when the input is empty, we still must handle the submit event, correct?

It's already handled. In the controller that you initially modified, the behavior is that if the user submits an empty value, nothing is saved. So there's nothing else to do there.

I'll check your pull request tomorrow :)

app/webroot/js/sentences.edit_in_place.js
```diff
@@ -53,9 +60,23 @@ $(document).ready(function() {
             cssclass  : 'editInPlaceForm',
             onblur    : 'ignore'
         }).bind('edit_sentence', function(e) {
+            $(this).find('textarea').keyup(function(event) {
+                var submitBtn = $(this).parent().find('button[type=submit]');
+                if($(this).val().trim().length == 0) {
+                    submitBtn.prop('disabled', true);
+                    submitBtn.css("background-color","gray");
```
trang: It would be better to define the CSS in the CSS files rather than set it in the Javascript.

You can add this in the `elements.css`:
````
.editableSentence button[disabled] {
    background: #ccc;
    border-color: #aaa;
}
````

trang:
@theblindspring, there's just a small fix to do with the CSS, and also you would need to fix the comment of your commit (which currently says `Trying to force a push anow`), then I can merge your pull request :)

Issue #751, opened 2015-08-27
Add "daily average" to the https://tatoeba.org/eng/contributions/activity_timeline/


ckjpn:
As I was doing a screenshot for "Daily Contribution Stats for 2014 & 2015 at a Glance" (http://goo.gl/3PUMKG), I thought it would be very easy for you to add in a "daily average" on each of the pages.

It would be a simple formula, calculated on all "complete days."

I'm referring to these pages.
https://tatoeba.org/eng/contributions/activity_timeline/

RyckRichards:
I'm pretty sure that it would help people (new members) see how important and "being used" is Tatoeba. It's not a "static" page, but a page that everyone add;correct, tag, link sentences everytime. Is it too hard to implement, @Tatoeba/tatoeba-team ?
I'm going to post it on the Wall since some members complain that they haven't been notified and so, Tatoeba is not "democratic"...

jiru:
> It's not a "static" page, but a page that everyone add;correct, tag, link sentences everytime.

I don’t understand this sentence.

RyckRichards:
It's not like many pages such as Duolingo that always displays the same thing all the time or LingQ, Memrise, etc

trang:
This is low severity, but if someone wants to implement it: the daily average can be displayed at the bottom, below the chart.

The string would be:
- `Daily average: {number} sentences` (if more than 1)
- `Daily average: {number} sentence` (if less or equal to 1)

theblindspring:
I have to confess, I don't use Tatoeba myself, so I am not sure what is being asked for. Do we want the daily average of sentences added by anyone to Tatoeba? Or just sentences added by the user? If we want the users daily average, do we count only days they logged in or do we count every day since they have been a member?

trang:
We want the daily average of sentences added by anyone in Tatoeba. You already have all data from the activity timeline page. You just have to make the sum of everything, divided by the number of days.

ckjpn:
To make it more clear, what I was suggesting is an average of contributions each month, similar to what's on this page each month on the gray line.

http://a4esl.org/temporary/tatoeba/stats/timeline/

I've included a partial screenshot since sometimes the above website is overloaded with traffic and is slow.

![screen shot 2016-05-10 at 22 53 56](https://cloud.githubusercontent.com/assets/7934273/15148872/56d044e8-1702-11e6-9776-6f3d171c2be5.png)

PR #1139, opened 2016-05-10
Add "daily average" to the https://tatoeba.org/eng/contributions/activity_timeline/

theblindspring:
This pull request addresses issue #751.

### DIFF
--- a/app/views/contributions/activity_timeline.ctp
+++ b/app/views/contributions/activity_timeline.ctp
@@ -114,9 +114,7 @@ foreach ($stats as $stat) {
             ),
             array('n' => $dailyAverage)
         );
-        
         echo $html->div("daily-average", $averageString);
-        
     }
     
     ?>
--- a/app/views/contributions/activity_timeline.ctp
+++ b/app/views/contributions/activity_timeline.ctp
@@ -104,12 +104,19 @@ foreach ($stats as $stat) {
     }
     
     if($numberOfDays > 0){
-        $dailyAverage = $totalSentences / $numberOfDays;
-        if($dailyAverage == 1){
-            echo '<div class="daily-average">Daily Average: ' . $dailyAverage . ' sentence</div>';         
-        } else {
-            echo '<div class="daily-average">Daily Average: ' . $dailyAverage . ' sentences</div>';         
-        }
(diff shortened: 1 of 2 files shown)


app/views/contributions/activity_timeline.ctp
```diff
@@ -93,8 +93,25 @@ foreach ($stats as $stat) {
...
+        $numberOfDays = date('d');
+    } else if (($year < date('Y')) || ($year == date('Y') && $month < date('m'))){
+        $numberOfDays = cal_days_in_month(CAL_GREGORIAN, $month, $year);
+    }
+    
+    if($numberOfDays > 0){
+        $dailyAverage = $totalSentences / $numberOfDays;
+        if($dailyAverage == 1){
+            echo '<div class="daily-average">Daily Average: ' . $dailyAverage . ' sentence</div>';         
+        } else {
```
trang: You can use `$html->div($className, $content)` instead of writing the HTML code.
Cf. http://book.cakephp.org/1.3/en/The-Manual/Core-Helpers/HTML.html#div

And you will need to use the `__()` function, or more precisely the `__n()` function, to output the `Daily average` string. This is needed to make the string translatable.
* Related doc: http://book.cakephp.org/1.3/en/The-Manual/Common-Tasks-With-CakePHP/Internationalization-Localization.html
* Example in the code: https://github.com/Tatoeba/tatoeba2/blob/dev/app/views/sentences_lists/show.ctp#L52

app/views/contributions/activity_timeline.ctp
```diff
@@ -93,8 +93,26 @@ foreach ($stats as $stat) {
...
+    
+    
+    if( $month == date('m') && $year == date('Y')) {
+        $numberOfDays = date('d');
+    } else if (($year < date('Y')) || ($year == date('Y') && $month < date('m'))){
+        $numberOfDays = cal_days_in_month(CAL_GREGORIAN, $month, $year);
+    }
+    
+    if($numberOfDays > 0){
+        $dailyAverage = $totalSentences / $numberOfDays;
```
trang: We probably don't need more than 1 decimal.

theblindspring:
I think this is better.

Issue #986, opened 2016-01-21
The title on "browse" pages are not correct when "only with audio" option is selected.


ckjpn:
https://tatoeba.org/eng/sentences/show_all_in/eng/und/none/only-with-audio

Perhaps the priority for fixing this could be set to "low", but it is something that should be eventually corrected.

![screen shot 2016-01-21 at 12 20 07](https://cloud.githubusercontent.com/assets/7934273/12470752/41a35822-c03a-11e5-9034-2866e980eeac.png)
![screen shot 2016-01-21 at 12 19 18](https://cloud.githubusercontent.com/assets/7934273/12470753/41a8fe1c-c03a-11e5-8d98-2e0c90c3a0e7.png)

pakhandibaba:
I want to work on this issue.

trang:
Hi @pakhandibaba, welcome to our GitHub! 
I've sent you an invitation to join our "New members" team :) 
I'll assign you the issue once you've joined.

trang:
@pakhandibaba I have unassigned this issue from you. I'm assuming you don't have time anymore to take care of this, so I'm freeing it for other developers who might want to fix it.

PR #1141, opened 2016-05-11
Fix title on show_all_in pages when 'only sentences with audio' selected

zachleigh:
This pull request addresses issue #986.


trang:
Thanks for your pull request @zachleigh :)

Just a small thing: could you reformulate the strings to be `Sentences in {language}` instead of `{language} sentences`. It's just to make the formulation more consistent throughout the website.

Issue #969, opened 2016-01-03
Add possibility to remove a favorite from favorites page


trang:
Users can add sentences to their favorites, and can view the list of their favorites from the top menu: {username} > "My favorites".
For instance this is the list of my favorites: https://tatoeba.org/eng/favorites/of_user/TRANG

At the moment, if a user wants to remove a sentence from their favorites, they need to go to the sentence's page, i.e. they need to:
- click on the favorited sentence from the "My favorites" page
- then click on the icon to remove the favorite

It would more be practical if they could remove a favorite directly from the favorites page.

This would also allow users to remove deleted sentences from their favorites (cf. #965).

kwikadi:
What sort of a solution are you looking for? The simplest would be to switch displayGenericSentence to  displaySentencesGroup, but it will basically put the entire Menu bar on top of every sentence. If that is acceptable, let me know.

trang:
Adding the whole menu would be too overwhelming. 

I'm rather looking for a solution that looks like the "My collections" page. For instance: http://dev.tatoeba.org/eng/collections/of/TRANG
You see there are icons at the end of each sentence. Except they are not clickable, but for the favorites, the icon should be clickable.

* The icon would be a filled heart initially (=favorited). 
* When the user clicks on it, it becomes empty (=unfavorited).
* The user should be able to click on it again in case they made a mistake, so that they can re-favorite the sentence. That's the same behavior as the icon in the sentence's menu.
* If the user leaves the page and comes back, the unfavorited sentence is no more listed on the page.

kamillel:
I made some mock-ups to show the behaviour described by Trang:

**Initial list**
![fav1_1-list](https://cloud.githubusercontent.com/assets/6699709/14121242/86fb90bc-f5ee-11e5-9735-ef0a14d7b8cd.jpg)

**Mouse hover favourite**
![fav1_2-hover](https://cloud.githubusercontent.com/assets/6699709/14121251/90c9935a-f5ee-11e5-8f23-6e437d422a4a.jpg)

**Favourite removed**
![fav1_3-unfav](https://cloud.githubusercontent.com/assets/6699709/14121280/a7f00a46-f5ee-11e5-876f-ce5f6b3c9ad6.jpg)
(grey-ing the text is a suggestion to give an additional clue to the user that something has happen but it's not a must have)

[...]

PR #1143, opened 2016-05-12
Add possibility to remove a favorite from favorites page

theblindspring:
This pull request addresses issue #969.


app/views/favorites/of_user.ctp
```diff
@@ -52,7 +52,7 @@ $this->set('title_for_layout', $pages->formatTitle($title));
 </div>
 
 <div id="main_content">
-    <div class="module">
+    <div class="module" id="favorites-list" data-success="<?php echo __("Successfully Removed Favorite", true); ?>" >
```
alanfgh: The second and third words in the message string should not be capitalized.

trang:
It's looking good :D Still a few fixes before I can merge this:

* The size of the icons should be bigger (24x24 instead of 16x16).
* The color of the icon should be grey (`#ccc`) instead of green, and black (`#000`) on hover.
* The string `Successfully removed favorite` should be `Favorite successfully removed.` (with the punctuation).

alanfgh:
I agree: "Favorite successfully removed." is better.

theblindspring:
Hey @trang,

What browser are you running on? My computer shows the icons as black, not green.

trang:
Found the problem: if you're logged in you will see the icons are green. If you're not logged in they will be black.

Your icons are displayed all the time but they shouldn't. So there's one more thing to work on:

* The "remove" and "undo" icons should be displayed only for the user who owns the favorites. Other people shouldn't see these icons, since they shouldn't be able to delete favorites that don't belong to themselves.

theblindspring:
Hmmm, I was logged in and they were black. I will work on getting them to only appear if you are logged in.

trang:
Actually you're right, it's not a matter of being logged in, it's a matter of browser. They appear black on Chrome, green on Firefox.

Issue #932, opened 2015-11-23
@username can produce multiple notification emails


jiru:
[Reported by tommy_san](https://tatoeba.org/jpn/wall/show_message/24958#message_24958).

> When someone writes my username in a comment to my sentence or a sentence where I also posted a comment, I get two email notifications. When someone mentions me there twice, I even get one more notification. This needs to be changed.

wellsyw:
I think this patch (untested) might fix it.
 
http://pastebin.com/9CqALHEW

jiru:
wellsyw, thank you for your patch but it would be better if you test it before sending. Another problem that comes to my mind is that we’re using different email templates for each notification type. With you approach, only the first notification would get through and the email template wouldn’t convey correct information about the origin of the notification.

Looking at how Github handles this, it just has one very simple email template with the title of the issue as subject, the comment as content and a link to it as footer. We should use a similar approach.

wellsyw:
I fail to see the problem about templates.

There are two notification templates. One goes to people who either own the sentence or have commented on it. One goes to people who have been @mentioned

This is suppressing the latter e-mail if the former has been sent. Which is the whole point of this bug, is it not?

wellsyw:
Or did you mean that the "why was this e-mail sent" should be stripped out of the templates? Or should an e-mail list all the reasons it was sent?

jiru:
> Or did you mean that the "why was this e-mail sent" should be stripped out of the templates? Or should an e-mail list all the reasons it was sent?

Yes. If someone mentions me in a comment of a sentence of my own, what is the reason of the notification? Is it because I have been mentioned, or is it because someone commented my sentence? It’s both, so neither of the current email templates fit.

Having said that, it’s not a serious problem if we just use the “commented a sentence” template in that case, like you did.

Having a unique wordless template would also work around the problem that notifications are not localized.

That said, we have private messages while Github doesn’t. A unique template may add confusion between private and non-private messages. It is often that people exchange private messages along with a public discussion.

wellsyw:
There are two or three templates used.

a) you own the sentence -- you get an e-mail
`"xyz has posted a comment on one of your sentences."`
b) you commented on the sentence -- you get an e-mail
`"xyz has posted a comment on a sentence where you also posted a comment."`
c) you were mentioned -- you get an e-mail
`"xyz has mentioned you in a comment."`

[...]

PR #1146, opened 2016-05-13
@username can produce multiple notification emails

theblindspring:
This pull request addresses issue #932.

It seems like this is probably some code we can get rid of (i.e. views/elements/email/html/user_mentioned_in_comment.ctp I think can be removed completely), but I wasn't sure if this was a different issue or if we want to try and eliminate any unneeded code now.


trang:
@theblindspring I'll take care of the clean up :)

theblindspring:
Hey Thanks!

Issue #1126, opened 2016-05-01
Searching for "$" generates an error message


RyckRichards:
When I tried to search for "$", it just returned this message:

> Search error

> An error occured while performing the search. If the problem persists, please let us know.

https://dev.tatoeba.org/eng/sentences/search?query=%24&from=eng&to=und
https://dev.tatoeba.org/eng/sentences/search?query=%24&from=und&to=und

I
![screenshot_1](https://cloud.githubusercontent.com/assets/7658430/14944089/1d7e7082-0fc0-11e6-8790-60a57205de81.png)

RyckRichards:
Maybe it's somewhat related to #1119

ventice:
Taking.

ventice:
Actually, there are two different issues here:
1. `$` isn't escaped which leads to error mentioned in the description (can be checked by escaping it in query itself by putting backslash in front of it, i.e. looking for `\$`)
2. When searching for escaped `$` (or `/`) only, no results are returned even though no error occurs. This is in my opinion because language indices do not include punctuation.

ventice:
After fixing 1. it appeared that my guess concerning 2. was true. Only underbars (`_`) are currently indexed. When I added `$, /,` to `charset_table` in `common_index` index in my local *sphinx.conf* file I was able to search for standalone `/` and `$` symbols (not those included in strings like **$100** because this fix makes punctuation part of indexed word and star syntax is not enabled in sphinx by default to search for substrings using `*` operator).
In any case, I am not sure I can update sphinx_conf to include dollars and slashes as part of charset without approval of some empowered tatoeba representative (:

trang:
#### Regarding escaping `$`

Since `$` can be used after a word to find sentences ending with that word, escaping `$` would make this feature unavailable, wouldn't it?

####  Regarding indexing `$` and `/`

From a strictly UX point of view, I'm in favor of it. I think it should be possible to search for `$` alone and have all the sentences containing `$`. I also think that if I'm searching for `$100`, I shouldn't get sentences with `100%`. Same if I search for `1/4`, I don't want results with `1.4`.
So indexing these symbols would make sense to me, but we would need to evaluate the impacts.

[...]

PR #1150, opened 2016-05-17
Made human friendly error message on invalid sphinx queries and made $ searchable.

ventice:
This PR addresses the issue #1126.
Note that this fix does not return $100 for the query "\$" since $ sign is considered as part of word and one has to search it as a beginning of word to be found. On the other hand, since we have a limitation that the search string used in wildcard searches should be at least 3 characters long, "\$\*" doesn't return anything either. On the other hand "\$10\*" will return $100000 and anything else.
Standalone $ signs such as that in the sentence "In English the shorthand for buck or dollar is "$"." are found by my fix.

### DIFF
--- a/app/views/sentences/search.ctp
+++ b/app/views/sentences/search.ctp
@@ -90,6 +90,7 @@ if ($search_disabled) {
                      'inconvenience. Please try again later.'); ?></p>
 <?
 } else if (!is_array($results)) {
+  if (!isset($sphinx_markers)) {
 ?>
     <h2><?php echo __('Search error'); ?></h2>
     <p><?php
@@ -102,6 +103,20 @@ if ($search_disabled) {
         );
     ?></p>
 <?
+  } else {
+?>
+    <h2><?php echo __('Search error'); ?></h2>
+    <p><?php
+        echo format(
+            __(
+                'Invalid query. '.
+                'Please refer to '.
+                '<a href="{}">text search wiki</a> for more details.', true),
+            'http://en.wiki.tatoeba.org/articles/show/text-search'
+        );
+    ?></p>
+<?
+  }
 } elseif (!empty($results)) {
 
     if (!$is_advanced_search && !empty($query)) {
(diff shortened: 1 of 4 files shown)


ventice:
Technically speaking, if Sphinx on production is installed with [RE2](https://github.com/google/re2) support I can include the following directive in config to strip `$` signs off their number:
`regexp_filter           = \$(\w+) => "$ \1"`

trang:
Alright so according to the [Sphinx documentation](http://sphinxsearch.com/docs/current/conf-regexp-filter.html):

> **12.2.64. regexp_filter**
> [...]
> Added in 2.1.1-beta. 
> [...]
> Binary packages should come with RE2 builtin. 

Currently we have in production: `Sphinx 2.2.9-id64-release (rel22-r5006)`.
And I believe it was installed from the binaries (not sure how I can verify that but I don't see any reason why we would have built it from the source), so it should have RE2 support.

app/views/sentences/search.ctp
```diff
@@ -102,6 +103,20 @@ if ($search_disabled) {
...
 <?
+  } else {
+?>
+    <h2><?php echo __('Search error'); ?></h2>
+    <p><?php
+        echo format(
+            __(
+                'Invalid query. '.
+                'Please refer to '.
+                '<a href="{}">text search wiki</a> for more details.', true),
```
trang: `Please refer to text search wiki...` sounds a bit weird.
Maybe `Please refer to the search documentation...` instead.

ventice: Will fix.

ventice:
Okay, I will add splitting of dollar signs off the sums in sphinx then. Looks like in this case they are not highlighted in search, so I will try to tweak it as well.

ventice:
I failed to find a way to make sphinx highlight `$` signs adjacent to the words - `BuildExcerpts` just does not insert corresponding markers around the sign.

trang:
> I failed to find a way to make sphinx highlight $

That's fine, it's a really minor inconvenience. We can figure out a solution later, someday.

Issue #1109, opened 2016-04-23
Translating from a sentence often generates duplicates


RyckRichards:
Reported by GrizaLeono, Wezel and soweli_Elepanto on the [Wall](https://tatoeba.org/eng/wall/show_message/26253#message_26253):

>Il arrive parfois que j'envoie une phrase, qui ensuite apparaît avec deux numéros consécutifs, par exemple les phrases #5075133 et #5075134.

> Cela arrive quand vous pressez le bouton « traduire » deux fois par mégarde. Une fois, j’en ai eu quatre, ces doubles. 

> I met such a problem too:
https://tatoeba.org/rus/sentences/show/5078678
https://tatoeba.org/rus/sentences/show/5078679
> It might be an bug in the software of the site.

ventice:
I've seen this myself several times during batch translations (100 sentences per session). Frequency of this bug appears to be 1 or 2 out of 100 translations. So, I guess it has some race condition issue behind it which means that the bug should be fixed more by elaborate analysis rather than by reproduction.
I'll try to do it.

ventice:
Looks like some echo from bug that `unbind()` on `app/webroot/js/sentences.add_translation.js:35`. When I commented it down, I got 6 instances of the sentence added. Looks like the second click sometimes occurs before unbind is run, which causes double request. I think unbinding isn't a stable solution - we have to hold registry of currently active requests and if second request arrives while another one is still active, discard it.

ventice:
Yes the reason was related to 'unbind()' but in a little different way. Actually, the original bug description provided a way to reproduce this bug with 100% guarantee:

> Cela arrive quand vous pressez le bouton «traduire» deux fois par mégarde. Une fois, j’en ai eu quatre, ces doubles.

Initially I thought that the author intended the button "Submit translation" with «traduire» but now I see that he/she meant the button that opens the translation controls - if you click the translate button while the controls are already open, you get second instances of event handlers bound to buttons and textarea so, when pressing the button you get two calls (that are execute both even though unbind is called in the first one). This can be worked around by the registry I mentioned, but it's easier to unbind event handlers immediately before binding them.

PR #1152, opened 2016-05-18
Unbound event handlers before binding and not on processing requests

ventice:
This PR addresses the issue #1109.
I am sorry but I refactored within the same commit as the fix itself since the changes are really straightforward and the diff is easily read. I can split to two changes if you want. The following is done:
1. I removed unbind and calls to it and moved unbinding before binding itself. Thus it is guaranteed that only single event handler is bound to the button and textarea.
2. I created temporary variables for calls to `$(...)` to improve performance and code readability.
3. Just in order to exclude any possibility to post two translation requests, I introduced the `translating` variable which discards second request, if there's one active.


trang:
Looks good :) 

No need to split into two changes, I'm not _that_ picky.

Issue #1100, opened 2016-04-20
Add Sort by Last Updated for Lists


ckjpn:
This would be useful, but perhaps not worth doing if it is difficult to do.

## The Feature

* Have the option to sort lists in the order that the most recent addition was added.

## The Reasons

* This would allow people to access this URL, for example, to see which members have been most recently adding audio files.

https://tatoeba.org/eng/sentences_lists/of_user/CK/audio

* This would also help members see which public lists are being updated and push the old public lists not being updated to the bottom of the list.

* Perhaps, this would also help members find their own recently-used private lists, and with a reverse sort, be able to find which lists of theirs that they may want to consider deleting.

dispyfree:
I see two different solutions to this problem:

1) explicitly keeping track of this information and storing it in a special field in the database. This will be more cumbersome as it must be kept up to date manually.
 
2) using virtual fields with CakePHP (introduced in CakePHP 1.3). This is very easy to implement; virtual fields can even be created on the fly. As it is computationally expensive to query the information each time, it is probably advisable to only include this information if the user actually sorts by it. 
Having said that, using the on-the-fly method, sorting will actually influence the fields which are shown in the table. 

What would you prefer?

trang:
I think virtual fields won't work there. If I understand properly, virtual fields are based on fields from the same model. In our case, we have the data for the lists in `sentences_lists`, and the date of when sentences are added in `sentences_sentences_lists`.

As for the update of the new field, it could be handled by a MySQL trigger, no?

dispyfree:
Virtual fields are basically arbitrary SQL expressions - that includes subqueries.

MySQL triggers are an option too - so adding a new field for that information and updating it using triggers is the preferred option? As long as this field is not indexed, the trigger should virtually have no overhead.

trang:
> so adding a new field for that information and updating it using triggers is the preferred option?

Yes. I have no preferences regarding the trigger, but we for sure want to avoid subqueries and joins as much as possible, for performance reasons.

trang:
Changes to do on the UI:

![lists](https://cloud.githubusercontent.com/assets/221850/15226698/17dbe11a-1884-11e6-8e1c-48be3c3ffdae.png)


[...]

PR #1153, opened 2016-05-18
Add Sort by Last Updated for Lists

dispyfree:
This pull request addresses issue #1100 .

+ made sentences lists sortable by the last time a sentence was added…or deleted

+ added respective triggers


trang:
Thank you for your pull request, @dispyfree! It looks pretty good overall :D
There are just a couple of things to fix, which I'll comment in your commits.

Just one thing, please take a moment to read our guidelines for pull requests:
https://github.com/Tatoeba/tatoeba2/wiki/Guidelines-for-submitting-pull-requests
And also our code style:
https://github.com/Tatoeba/tatoeba2/wiki/Code-style

Note that you don't have to close this pull request and open a new one that would follow entirely the guidelines. Your code works and it's all that really matters. It's just so you're aware that there are guidelines and can submit your next pull requests in harmony :)

app/views/helpers/lists.php
```diff
@@ -144,15 +160,24 @@ public function displayRow(
...
             echo format(__('created by {listAuthor}', true),
                         array('listAuthor' => $link));
+            
+            echo $this->Html->tag(
+                'span',
+                ' '.$this->Date->ago($createdDate),
+                array(
+                    'class' => 'date'
+                )
+            );
```
trang: It feels a bit strange to have the list author and date only separated by a space. I'd suggest to include the date in the "created by" string and have it separated by a comma: `created by {listAuthor}, {date}`.

docs/database/triggers/update_modified_sentences_list.txt
```diff
@@ -0,0 +1,27 @@
...
+CREATE TRIGGER extend_sentences_lists_delete AFTER DELETE ON `sentences_sentences_lists`
+FOR EACH ROW 
+	UPDATE `sentences_lists` AS sl 
+      SET sl.modified = NOW()
+      WHERE sl.id = OLD.sentences_list_id
+      LIMIT 1;
+|
+
+
+delimiter ;
```
trang: For consistency, could you rename the file into *.sql instead of *.txt.

app/views/helpers/lists.php
dispyfree: I was thinking about that too, but finally dismissed the idea. The dates are enclosed by spans which in turn are specially formatted. I could either drop the special formatting (which I think is a bad idea) or I could include HTML into a __()-Expression which is even worse ;-)
I added a comma, however due to the aforementioned reasons I refrained from using an atomic expression.

dispyfree:
Sorry, I really missed the guidelines on github ;) 
I adjusted the pull request as best as I could, but I can't find a way to change the target branch. However, there is now a branch issue_1100.

app/views/helpers/lists.php
trang: You wouldn't include HTML in the `__()` function itself, only a placeholder (`{date}`). Then the `format()` function will replace `{date}` by the corresponding HTML. 

It's the same mechanism as `{listAuthor}`, which is replaced by the HTML defined in the `$link` variable.

trang:
No need to change the branch, you can finish this pull request on your `dev` branch. 

Once I merge this pull request, you will have to sync your `dev` branch with Tatoeba's `dev` branch. And from there you can carry on by creating new branches for each issue (and remember to re-sync your `dev` branch every time before you create a new branch).

The only reason why you want to create branches is to make sure that you have the flexibility to work on several issues in parallel. Right now you're only working on one issue so it's fine if it's on `dev`.

app/views/helpers/lists.php
dispyfree: Still, I
a) do not use a separate td for the date. Then the dates will not be aligned any more and the more the lengths of the user names differ, the uglier the layout will be
b) close the previous td and open another td within the __() expression. Then the structure of the code will not reflect the structure of the HTML any more. 

Do you still prefer to include it directly, i.e. by using (a)?

dispyfree: Still, I
a) do not use a separate td for the date. Then the dates will not be aligned any more and the more the lengths of the user names differ, the uglier the layout will be
b) close the previous td and open another td within the __() expression. Then the structure of the code will not reflect the structure of the HTML any more. 

Do you still prefer to include it directly, i.e. by using (a)?

dispyfree: Still, I
a) do not use a separate td for the date. Then the dates will not be aligned any more and the more the lengths of the user names differ, the uglier the layout will be
b) close the previous td and open another td within the __() expression. Then the structure of the code will not reflect the structure of the HTML any more. 

Do you still prefer to include it directly, i.e. by using (a)?

trang: Hmm, I'm not following anymore... Maybe I was not very clear in my previous comments. 

In terms of code, what I meant was that instead of:
```
$link = $this->Html->link(
    $listCreatorName,
    array(
        "controller"=>"user",
        "action"=>"profile",
        $listCreatorName
    )
);
echo format(__('created by {listAuthor}', true),
            array('listAuthor' => $link));

echo $this->Html->tag(
    'span',
    ' '.$this->Date->ago($createdDate),
    array(
        'class' => 'date'
    )
);
```

to have:
```
$link = $this->Html->link(
    $listCreatorName,
    array(
        "controller"=>"user",
        "action"=>"profile",
        $listCreatorName
    )
);
$creationDate = $this->Html->tag(
    'span',
    ' '.$this->Date->ago($createdDate),
    array(
        'class' => 'date'
    )
);
echo format(
    __('created by {listAuthor}, {date}', true),
    array(
        'listAuthor' => $link,
        'date' => $creationDate
    )
);
```

Your adding a comma after `{listAuthor}` works too, but it's just a bit less "translator-friendly" because it could happen that in this context, the translator needs to add some words before or after `{date}`, and they wouldn't be able to do it if the string was only `created by {listAuthor},`.

dispyfree: I got that; my point was that the date is currently aligned separately from the author. That means that all dates appear to be aligned in one col, no matter how long the author's name on its left side is. 
The names of the authors will differ and if the dates are not aligned separately, they will have different indentations - depending on the the length of the author's name. 
(I currently use td for alignment).

trang: In the code you've submitted, only the "last updated" date is aligned in a separate column. The "created" date is not.

lists-date

Your initial code was showing like above, except without a comma between the author and creation date. Then I pointed out it looks strange with only a space separating the author and the date, so you added a comma (screenshot above).

If we agree to keep this layout, then what I was saying is that you should change `__('created by {listAuthor}', true)` into `__('created by {listAuthor}, {date}', true)` for the reason I explained in my previous comment.

If what you meant to say (but that I failed to understand) is that you'd rather have the created date aligned in its own column, just like the "last updated" date, then I'm okay with it too. It gives a bit less space for the title, but it's not too big of a deal. In this case you can remove the comma, put the created date in its own `<td>` and add a title `created` on the column.

I hope we've managed to come to a common understanding ^^;

dispyfree: Sorry, I somehow missed your answer until just now. 

aligned
I wanted to have it aligned and did it almost as you suggested. 
By doing that I had a tinker at the original design. Please take a look at that, I'm pretty sure you can think of a better solution. 
I want to have the created-col left-aligned to the author, on the other hand I want the spacer to take up as much space as possible. On the other hand, I do want to prevent unnecessary word-wraps within the sentences' titles. 
All of that does currently work but I changed the original design to achieve that.

trang: It seems to me that no matter where we place the date it will feel kinda awkward. 

I wouldn't spend too much time on this right now, since there are plans to redesign of the UI to make it responsive, and we'll be reviewing every elements in every pages. When the time comes we'll rework completely the display of the lists.

So for now feel free to position the date where it feels best for you, even if it's not perfect. I'll make your changes available on dev.tatoeba.org so that the community can try out and see that's good enough for them. If not, I'll take care myself of re-adjusting the position of the date.

trang:
Merged #1208. Closing this one since it's no longer relevant.

Issue #654, opened 2015-05-14
offer ability to mark a list inactive


alanfgh:
It would be nice to be able to mark a list inactive. Inactive lists would not show up in the list of list names that opens above a given sentence when the list icon is clicked. They could also be hidden in the "Browse by list" view unless the user chose to view all lists. The idea is that someone who has created a bunch of lists could mark the old ones inactive in order to shorten the the list of list names.

trang:
I'm planning to close this issue because it is a bit outdated. Now that it is possible to set the visibility of a list to ["unlisted"](https://github.com/Tatoeba/tatoeba2/issues/677#issuecomment-170373110), do we still need to have this "inactive" status?

alanfgh:
Setting the visibility of a list to "unlisted" means that **other** people don't see your list, but it does nothing to help you look at a trimmed-down list of your own lists. 

To simplify, I have basically two kinds of lists: 
(1) Sentences that already existed in Tatoeba in Hebrew that I want to study, which I collect under the title "Heb (x): worth learning: private", where x is a number from 1 to (now) 24.
(2) Sentences that I added in English that I requested a Hebrew translation for, which I collect under the title "Eng (x): translation to Hebrew requested", where x is a number from 1 to (now) 8.

When a list reaches 100 sentences, the maximum number that can be downloaded, I stop and go on to a new one. (I realize that the Tatoeba maximum was chosen somewhat arbitrarily, but it's a convenient size for a list: not too short and not too long for printing and studying. It's also a nice round number that makes it easy to calculate how many sentences are contained in the lists.)

As you can see, the list of lists is too long to fit on the screen. I'd like to have a way to say that a particular list is inactive, so I don't want to see it in the list of lists displayed when I click the list icon for a sentence, but I don't want to get rid of the list. As it is, I need to rename the old lists with a prefix of "z~" just to get them out of the way. I'd rather not have to do that.

![possible_lists_for_sentence](https://cloud.githubusercontent.com/assets/1213027/14037053/255a50b8-f216-11e5-9857-ba2500234a3a.png)

trang:
Oh I see. I thought you were talking about the index list. Makes sense now.

trang:
The best solution I can think of would be to have 3 options in the permissions to edit:

1. editable by anyone
2. editable by only myself
3. editable by no one

When a list is editable by no one, it won't appear in the dropdown.

RyckRichards:
[...]

PR #1158, opened 2016-05-22
Add ability to make lists uneditable.

zachleigh:
This pull request addresses issue #654 

As mentioned in the issue comment thread, this fix requires a database update in order to add a third enum option to the sentences_lists editable_by column. I ran this mysql command locally:       
SQL Query: ALTER TABLE sentences_lists MODIFY COLUMN editable_by enum('creator', 'anyone', 'no_one') NOT NULL DEFAULT 'creator';

### DIFF
--- a/app/views/helpers/lists.php
+++ b/app/views/helpers/lists.php
@@ -95,8 +95,13 @@ class ListsHelper extends AppHelper
         $visibility = 'private',
         $editableBy = 'creator'
     ) {
+        $listStatus = 'activeList';
+
+        if ($editableBy == 'no_one') {
+            $listStatus = 'inactiveList';
+        }
         ?>
-        <tr class="listSummary">
+        <tr class="listSummary <?php echo $listStatus; ?>">
 
         <td class="icon">
             <?php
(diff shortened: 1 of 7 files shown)


app/views/helpers/lists.php
```diff
@@ -305,36 +305,38 @@ public function displayVisibilityOption($listId, $value)
...
-            echo $this->Html->tag('label',
-                __('Allow anyone to add and delete sentences from the list', true),
-                array('for' => 'editableCheckbox')
-            );
+                echo $this->Form->radio(
+                    'editable_by',
+                    array(
+                        'anyone' => __('Anyone', true),
+                        'creator' => __('Only Me', true),
+                        'no_one' => __('No One', true),
```
trang: Small detail: only the first word should be capitalized.
* `Only Me` => `Only me`
* `No One` => `No one`

zachleigh: Changed it to 'No one (list inactive)'.

trang:
To be coherent, when the list is editable by no one, the form to add new sentences should not be displayed.

screenshot_1

zachleigh:
Fixed this. Is there anywhere else a person can add a sentence to a list?

trang:
> The change is only noticeable after a browser refresh. Would you like it to update instantly? 

No need to update it instantly. It would be nice but not a priority at all.

> Is there anywhere else a person can add a sentence to a list?

No.

Issue #1151, opened 2016-05-18
Warning on synchronous use of AJAX during posting of translations


ventice:
**Environment:** Chrome for Windows (any other Chrome will do).
**Prerequisites:**
Open developer console by entering **F12**.
**Steps to reproduce:**
1. Open any sentence by clicking on it
2. Click on translation button
3. Input any translation
4. Submit the translation
**Result:**
In **Console** view you can see the following message:

> Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience. For more help, check https://xhr.spec.whatwg.org/.

![image](https://cloud.githubusercontent.com/assets/3504880/15362789/84ea1d92-1d27-11e6-9498-243db4502cff.png)

**Expected result:**
No warning in console.
**Possible troubles:**
Because the API is deprecated, its support may be dropped in some future version.

ventice:
Taking this.

ventice:
Looks like this happens because of `<script>` tags returned as part of response from AJAX request. For instance, this happens in `SentencesController::save_translation()` + `app/views/sentences/save_translation.ctp`. As a side effect we get set of `<script>` tags bound to each translation. So, this is fixed by removing `<script>`s from AJAX response.

ventice:
This is the list of all JavaScript files I found in sources:

1. announcement.js
2. chosen.jquery.min.js
3. generic_functions.js
4. jquery.cookie.js
5. jquery.jeditable.js
6. jquery.scrollTo.min.js
7. sentences.change_language.js
8. sentences.contribute.js
9. sentences.copy.js
10. sentences.edit_in_place.js
11. sentences.logs.js
12. sentences.show_another.js
13. sentences_lists.edit_name.js
14. sentences_lists.remove_sentence_from_list.js
15. sinograms.search.js
16. svg4everybody.min.js
17. transcriptions.js
18. users.check_registration.js
19. wall.reply.js
20. wall.show_and_hide_replies.js

The files are sprayed all over view and controller files, so it is a little bit difficult to understand the final set of scripts sent with each page. Anyhow, for the case described in this issue, it looks like only transcriptions.js is to be included in upper level script set and scripts can be removed from views/sentences/save_translation.ctp

ventice:
[...]

Issue #692, opened 2015-06-23
Problems adding multiple sentences to a list from the "Add sentence" page


alanfgh:
User 123xyz reports on [https://tatoeba.org/eng/wall/show_message/23192#message_23192](the Wall):

Bug:

When I add several sentences at once using the "Add Sentence" option, they line up on the page one after the other, but when I start adding them to my lists (well, my single list), I can only add two. The first sentence I add can be any sentence, but afterwards, the "add to list" button becomes unresponsive for all of the sentences except the FIRST one (the top one). If I added the first one as my first sentence, then I can only add one sentence in total.

alanfgh:
I see that when I have three sentences on the page and add the first one to the list, I can then add the third but not the second.

RyckRichards:
I've tested this way:

1st step - added bunch of sentences http://prntscr.com/eczgpa  http://dev.tatoeba.org/eng/contributions/of_user/Ricardo14

2nd step - deleted most of them

they got deleted without any problem

Also, #1313  has been closed too.

jiru:
The problem looks solved now.

Issue #566, opened 2015-02-04
Multiple queries sent when changing a language multiple times


jiru:
Steps to reproduce:
1. Go to a sentence on which you can change the language.
2. Open your favourite dev tool/inspector to scan queries sent by the browser (ctrl-shift-C on Firefox)
3. Change the language of the sentence. You’ll notice one POST query.
4. Change the language of the sentence again. You’ll notice two POST queries.
5. Change the language of the sentence again. You’ll notice four POST queries.

The function executed on language change gets rebinded each time by the response itself.

RyckRichards:
Mentioned by Guybrush88 on [Tatoeba Day #9](https://tatoeba.org/eng/wall/show_message/24479#message_24479).

ventice:
Looks like this issue is already fixed since I can see `$(".languageFlag").unbind('click');` at the beginning of the click handler.

ventice:
Just reproduced it. Taking.

PR #1162, opened 2016-05-23
Got rid of inclusion of .js files in AJAX response for rerunning them

ventice:
This PR addresses issue #1151 
Here is what's done:
1. `watch.js` is developed. This is a small jQuery plug-in holding registry of rules that are to be ensured for the entire document. There are two families of actions in it: **registration** and **DOM modification**. The rules are registered on the startup by all the pages that run jQuery and are immediately applied to the scope in which they are called. After that the rule (a JS function doing jQuery calls) is stored in a list and all the rules are applied to the element being added to DOM by the second family of functions. See image for more details.
watch_js_workflow
2. The JS files related to this issue are changed to use `watch.js`. This effectively means putting all the code inside `$(document).ready(...)` callback into a new function which is registered with `watch.js`. Additionally, all the global calls to `$(...)` are replaced to calls with context to operate only under the node whose DOM has been modified. Note that initially I planned to operate on the node being added (before attaching to DOM) but it may result to error for queries using `parent()` calls for localizing siblings.
3. `watch.js` is shipped by default before other scripts are loaded.
4. `transcriptions.js` is added to be available without being embedded via AJAX call.
5. All the scripts initially shipped with AJAX response are removed.


ventice:
Please note that if you prefer to keep the initial hack with running JS by embedding it in HTML to avoid introduction of such "frameworks", I can fix the issue by simply adding `transcription.js` to default set of scripts. In this case it isn't downloaded upon receiving the response, so the warning described in the issue isn't written.
What are the drawbacks of keeping the "hack". Since the code inside `$(document).ready` is unexpectedly run multiple times, it may result in registration of the same handler as many times, as many responses are received which may result in bugs like double sentences. So, the code inside `$(document).ready` should make sure it is idempotent (running it multiple times make the same result is running it once), for instance by removing any handlers before registering them.
This code partially mimics AngularJS's approach to watching variables, so, if you are planning to migrate to AngularJS, we can live without my reinvented wheel.

trang:
Thank you very much for the explanations :)

I do have plans to migrate to AngularJS, but at this stage I have no certainty that the plans won't be dropped for whatever reason, and I have no clear idea of how long it will take. So I see no problem in using your reinvented wheel in the meantime :p

To ensure that I understand properly, let's say I want to fix the synchronous XMLHttpRequest warning for `/sentences/add`, I would have to:

1. Remove the Javascript from [sentences/add_an_other_sentence.ctp](https://github.com/Tatoeba/tatoeba2/blob/dev/app/views/sentences/add_an_other_sentence.ctp#L42).
2. Inject these scripts from [sentences/add.ctp](https://github.com/Tatoeba/tatoeba2/blob/dev/app/views/sentences/add.ctp) instead.
3. Use `$(document).watch("addrule", ...)` for each of the [scripts involved](https://github.com/Tatoeba/tatoeba2/blob/dev/app/views/helpers/sentences.php#L754).
4. Use `.watch("prepend", data);` instead of `.prepend(data);` in [sentences.contribute.js](https://github.com/Tatoeba/tatoeba2/blob/dev/app/webroot/js/sentences.contribute.js#L56).

Right?

If that's correct I can merge your pull request. 

I can also wait if you'd like to also take care of the warnings on `sentences/add` and `sentences_lists/show/{listId}` (there might be other places, but those are the 2 main ones).

ventice:
> let's say I want to fix the synchronous XMLHttpRequest warning for /sentences/add
To tell the truth in this refactoring the warning itself is very secondary issue. The warning occurs when a **new** script is returned from AJAX request and inserted into the document, so, since browser cannot render the added content before the script is downloaded (because the script may do some tweaking of the document which in our case it really does). This downloading of requested script is where the warning comes from since the downloading may take a while blocking the browser. So, if our sole goal is to get rid of the warning (and its symptoms), we have to make sure that no **new** scripts are returned from AJAX requests. This can be easily achieved by moving all such scripts to default set of scripts loaded with every page. In this case we won't have any warning even though the trick of running JavaScript inside `$(document).ready` by multiple insertion of scripts into the document will still be present.
Anyhow, such trick has potential problems with the maintainability of the code. When writing code any future tatoeba developer may rely on the fact that code inside `$(document).ready` handler should run only once. So, this code may be made non-idempotent (idempotent behavior is the one when multiple invocations have the same effect as single invocation) which may cause bugs. An example of bug is the one with multiple translations of the same sentence. The handler of click of 'translate" button wasn't idempotent, so clicking it multiple times resulted in registration of many handlers. Same kind of bugs are potentially hidden in code in which handlers are misused or used in non-traditional way. My fix tried to address this issue by moving the code from ready-handler to rule. I also tried to avoid processing of entire document at every modification but I failed to completely isolate the scope of the rule, because some rules require logic that only works in the context of the document (navigating by 'parent()'). There are also cases when insertion of rules by processing the entire would work (for instance, rules referring to `body`) and my "fix" would not. Currently it works because our rules don't do global queries but if you are ever planning to run such queries (namely: after completing AJAX request, change something in the header), this new infrastructure would be an obstacle for this.

So, if you still welcome such changes provided the restrictions, I can also fix the cases you mentioned, if not, I can discard it and simply get rid of the warnings in the minimal possible way (that fix will affect 10 lines of code at most).

ventice:
On the other hand, I can still keep the "watching" approach but run the rules not over modified piece of code, but over entire document. It will eliminate restrictions posed over scripts and still be better than multiple invocations of `$(document).ready`

trang:
> An example of bug is the one with multiple translations of the same sentence. The handler of click of 'translate" button wasn't idempotent, so clicking it multiple times resulted in registration of many handlers. Same kind of bugs are potentially hidden in code in which handlers are misused or used in non-traditional way.

Possibly these issues are related to what you're saying there:
* https://github.com/Tatoeba/tatoeba2/issues/566
* https://github.com/Tatoeba/tatoeba2/issues/692


> On the other hand, I can still keep the "watching" approach but run the rules not over modified piece of code, but over entire document. It will eliminate restrictions posed over scripts and still be better than multiple invocations of $(document).ready

Alright so in the end it's up to you :) 

Would you rather do minimal changes, because there are plans to migrate to AngularJS and you'd rather spend time on other issues? Or would you rather do the refactoring because it makes the code more robust and makes you sleep better at night, even if your framework will be in use only for a "short" time?

I'm personally fine either way so you should do what feels more fulfilling for you.

ventice:
Ok, I will keep watch.js (it's very tiny framework so we can happily drop without remorse when we move to AngularJS) but make the scope of queries the entire document. I will also check the issues you have cited.

Issue #1180, opened 2016-06-04
Responsive UI


trang:
## The problem

More and more users are browsing Tatoeba from a mobile device. In April 2016, the percentage of visits from mobile devices started to exceed the percentage of visits from desktops. Considering that mobile visitors are now the majority, it becomes urgent that we rework the UI to be mobile-friendly.

Below some stats from Google Analytics.

**April 2013**
![april-2013](https://cloud.githubusercontent.com/assets/221850/15801563/7a896996-2a99-11e6-925f-b8f521c6fe50.png)

**April 2014**
![april-2014](https://cloud.githubusercontent.com/assets/221850/15801569/83a1cbcc-2a99-11e6-8dea-8873235c8386.png)

**April 2015**
![april-2015](https://cloud.githubusercontent.com/assets/221850/15801570/8825a77c-2a99-11e6-9d94-ed19ae3c275f.png)

**April 2016**
![april-2016](https://cloud.githubusercontent.com/assets/221850/15801581/c614d3be-2a99-11e6-9fd6-f5791a64c74b.png)
## The solution

We basically have 3 solutions.
1. We develop a mobile app for Android/iPhones.
2. We develop a mobile version of the website.
3. We develop a responsive UI.

The solution we'll be trying is the 3rd one: responsive UI. We will be using [Angular Material](https://material.angularjs.org) for this.

**Why?**
- The first solutions requires that we have an API. We currently don't.
- The second solution means that we will have 2 websites to maintain. In other words, if we introduce a new feature, we have to code it twice. We can't afford it.
- As for Angular Material, it's a subjective choice. I like the [Material Design](https://material.google.com/) spec and Angular Material feels rather comfortable to work with.
## The plan

We want to make the transition as smooth as possible. There will be more or less 3 phases.
### Phase 1

Since we'll be following the Material Design spec, we need to prepare users to get used to the new look and feel. During the first phase, we don't touch the general layout of the website. We only change small elements (buttons, icons, fonts, colors, dropdowns, etc). These changes will be introduced progressively, over the course of a month, maybe two.

This will be a good time to start a styleguide.
### Phase 2

In this phase we start to tackle bigger elements: the menu, the search bar, the sentences, the comment, the logs... Again, we don't touch the layout of the page. We just want to redesign these elements in a way that makes them easy to migrate to a responsive UI.
### Phase 3

[...]

PR #1184, opened 2016-06-05
New buttons

trang:
This pull requests is part of phase 1 of the UI migration (cf. issue #1180).

In this pull request, most of the buttons have been migrated to use the buttons from the Angular Material framework.

The color used for the buttons corresponds to the `green` color in the Material Design palette.
https://www.google.com/design/spec/style/color.html

In the process, the background color of the search bar and the color of the links have been changed as well, to match the color of the buttons.

Overall, users should not feel a huge change.


Guybrush88:
the text within the button to submit a translation is still truncated

trang:
I don't see it truncated.

screenshot_4

billoliver:
I think we should change the wording to something shorter like "submit".

Guybrush88:
@trang yep, now it's fixed also for me. maybe i saw it still truncated for cache related things

Guybrush88:
I don't know if it's a known issue, but on dev, the button to copy a sentence into the clipboard stopped working again

RyckRichards:
+1

Also on the main Tatoba

Guybrush88:
on the main site, it works for me

trang:
I modified the code on the dev website for Ricardo to test certain things. Right now it's reverted back to normal so the copy feature should work again, but don't be surprised in the next couple of weeks if some features are not working on the dev website.

Guybrush88:
Is it normal that, if a sentence is edited and it takes a while to save the changes, no loader appears? This happens on both the normal and dev sites, and this didn't happen with the previous icons

trang:
No it's not normal.

By the way for issues related to loaders you should comment on  #1194.

In general, for issue related to the UI, you should check what has been done in #1180. You can look at issues that reference to #1180, and you can report problems on the corresponding "sub-issue". If you can't find a sub-issue that covers your problem, then you can post a comment on #1180.

Issue #960, opened 2015-12-30
Include an option to list tagged sentences with the newest addition first


ckjpn:
Include an option to list tags with the newest addition first, or perhaps just make this the default sort.

I often find that I am using the following link to jump to the last page of the OK tag.

https://tatoeba.org/eng/tags/show_sentences_with_tag/7/eng/page:99999998

Perhaps others, too, would find it more useful to see the most-recently tagged sentences first, rather than always getting the oldest tagged sentences first.

If you made it an option, then people could have it both ways.

theblindspring:
Okay, so what exactly would we like to sort by? Do we want to sort the sentences by most recently created or do we want to sort by the most recently tagged?

trang:
We want to sort by most recently tagged. I'm not completely sure about the label we could use for the sort. Maybe:

**Sort by:** date of tag

theblindspring:
okay cool I will work on that

PR #1189, opened 2016-06-06
Include an option to list tags with the newest addition first.

theblindspring:
This pull request addresses issue #960. 

I didn't change the default sort method to sort by 'date created' because I wasn't sure if that was really agreed upon (and I am not sure how to change the default sort method, so if anyone would like to give me some quick direction that would be great!).


trang:
@theblindspring it's better to leave the default sort as is. We currently don't have any clear evidence that it is more useful for users to see the latest tagged sentences first.

trang:
@theblindspring, you misunderstood the issue. But I have to say, the title was misleading (I changed it).

You added a sort function on this page: https://tatoeba.org/eng/tags/view_all

But the issue is about sorting the tagged sentences, like for instance on this page: https://tatoeba.org/eng/tags/show_sentences_with_tag/7

Issue #1134, opened 2016-05-05
Graphical redesign of pagination


trang:
Current pagination:

![pagination](https://cloud.githubusercontent.com/assets/221850/15056068/25d2940e-1310-11e6-991d-1578b7728761.png)

Draft of new pagination:

![a834fc9e-f258-11e5-8578-232903b79918](https://cloud.githubusercontent.com/assets/221850/15056084/32ff2e3a-1310-11e6-8f33-1374f83cff02.png)

This needs to be done before #24.

RyckRichards:
It would be great to have an "extra field" which would allow users to
type the page they want to go to and both a "Go" icon and an
"information" that he can also press Enter. It would save us a lot of
time when we search some sorts of sentences, for example.

On 5/5/16, Trang <notifications@github.com> wrote:
> Current pagination:
>
> ![pagination](https://cloud.githubusercontent.com/assets/221850/15056068/25d2940e-1310-11e6-991d-1578b7728761.png)
>
> Draft of new pagination:
>
> ![a834fc9e-f258-11e5-8578-232903b79918](https://cloud.githubusercontent.com/assets/221850/15056084/32ff2e3a-1310-11e6-8f33-1374f83cff02.png)
>
>
> ---
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/Tatoeba/tatoeba2/issues/1134


-- 
Ricardo Vernaut Jr

trang:
@RyckRichards, this is what #24 is about.

RyckRichards:
Oh sorry...

-- 
Ricardo Vernaut Jr

billoliver:
### The Redesign
![pagination-tatoeba](https://cloud.githubusercontent.com/assets/17387379/15274493/60f43cba-1a89-11e6-8987-ce9bf7ce7b5d.png)

- I increased both the size of clickable elements and the space between each page link so that users can't accidentally click on the wrong one.

- Only the current page link has a background color so that users can easily keep track of what page they are.

- The first and last page icon was removed because some users might think they will take them to the next page. If a user wants to go to those pages he or she just need to click the links.

**Colors:**
```
Chevron icons: #000
Current page background: #4caf50
Current page font size: 14px
Current page font weight: bold
Current page element size: 24px | 24px
Default color of the links: #757575
Hover: #eee
```

trang:
[...]

PR #1195, opened 2016-06-14
Graphical re-design of pagination

phtan:
This pull request addresses issue #1134 . It is relevant to issue #1180 

I've left out the [Angular Material icon](https://design.google.com/icons/#ic_more_horiz) for the ellipsis. 

The screenshots below show the pagination when the user is on the following pages:

- random page,
- first page, and
- last page.

screenshot from 2016-06-15 03-13-47
screenshot from 2016-06-15 03-13-59
screenshot from 2016-06-15 03-14-15

### DIFF
--- a/app/views/helpers/pagination.php
+++ b/app/views/helpers/pagination.php
@@ -75,7 +75,8 @@ class PaginationHelper extends AppHelper
         $this->Paginator->options(array('url' => $extramParams));
         // -----------------------------------------------------------
 
-
+        $numberFirstLinks = 1;
+        $numberLastLinks = 1;
         $numbersOptions = array(
             'separator' => '',
             // Print up to 6/2 = 3 numbered links in rectangles on each side
@@ -94,9 +95,8 @@ class PaginationHelper extends AppHelper
             // end of the list of English sentences).
             'modulus' => 6,
             'class' => 'pageNumber',
-            'first' => 1,
-            'last' => 1,
-            'ellipsis' => 'more_horiz'
+            'first' => $numberFirstLinks,
+            'last' => $numberLastLinks,
         );
         ?>
         <div class="paging">
--- a/app/views/helpers/pagination.php
+++ b/app/views/helpers/pagination.php
@@ -93,27 +93,25 @@ class PaginationHelper extends AppHelper
             // the largest numbers (for example, when we're near the
             // end of the list of English sentences).
             'modulus' => 6,
-            'class' => 'pageNumber'
+            'class' => 'pageNumber',
+            'first' => 1,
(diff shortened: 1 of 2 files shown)


trang:
Thanks for the pull request @phtan. Here's my feedback:

1. I forgot to update the issue for the colors: you'll have to change the background of selected page from #66AA00 to #4caf50, since #4caf50 is going to be our primary color.
2. Remove the grey background on hover over the arrows. There's no effects when hovering over the arrows. If you'd like to add an effect, it would be better to change the color rather than the background color.
3. There is a big space between `{lastpage}` and `>`, while there is almost no space between `<` and `1`. The alignment of the `>` arrow seems off as well (it's below the baseline). The spacing and alignment should be equal on both sides.

phtan:
Addressing @trang 's comment above, the pagination now looks like the following:

screenshot from 2016-06-16 00-37-34

app/views/helpers/pagination.php
```diff
@@ -93,27 +93,25 @@ public function display($extramParams = array())
             // the largest numbers (for example, when we're near the
             // end of the list of English sentences).
             'modulus' => 6,
-            'class' => 'pageNumber'
+            'class' => 'pageNumber',
+            'first' => 1,
+            'last' => 1,
+            'ellipsis' => 'more_horiz'
```
trang: The `ellipsis` property can be removed. It doesn't have any effect.

app/webroot/css/layouts/elements.css
```diff
@@ -736,42 +736,56 @@
...
-    color: white;
+.numbers a:hover {
+    background: #E0E0E0;
+    color: #757575;
 }
 
 .paging .disabled {
     display: inline;
     padding: 3px 5px;
     margin: 3px;
```
trang: The `display`, `padding` and `margin` can be deleted.
* `display: inline` doesn't have any effect
* the padding and the margin are causing misalignment with the disabled arrow icons

app/webroot/css/layouts/elements.css
```diff
@@ -736,42 +736,56 @@
...
+a.prev {
+    background: #FFFFFF;
 }
 
+a.next {
+   background: #FFFFFF;
+}
+
+md-icon.disabled.material-icons {
+   background: #FFFFFF;
```
trang: Should be `background: none`. We want a transparent background for the icons.

trang:
Almost there :) I've posted a few more comments on the code.

phtan:
Made changes as per [above comment](https://github.com/Tatoeba/tatoeba2/pull/1195#issuecomment-226298669).

Issue #1194, opened 2016-06-11
New loaders


trang:
This is part of phase 1 of the UI migration (cf. #1180).

We want to replace all the current loaders with the Angular Material progress circular.

Pages/elements affected:
* Homepage (loading a new random sentence)
* Add sentence page: `/sentences/add`
* List page (when list is editable by current user): `/sentences_lists/show/{listId}`
* Adding a reply to a message on the Wall
* Every option in the sentence's menu (translate, edit, favorite, list...)

PR #1204, opened 2016-06-16
New Loaders

theblindspring:
This pull request addresses issue #1194.

### DIFF
--- a/app/views/helpers/lists.php
+++ b/app/views/helpers/lists.php
@@ -289,7 +289,7 @@ class ListsHelper extends AppHelper
         <dl>
             <?php
             $title = __('List visibility', true);
-            $loader = "<div class='is-public loader-container'></div>";
+            $loader = "<md-progress-circular class='is-public loader-container' md-diameter='16' style='display: none'> </md-progress-circular>";
             echo $this->Html->tag('dt', $title . $loader);
 
             echo $this->Form->radio(
@@ -326,7 +326,7 @@ class ListsHelper extends AppHelper
         <dl>
             <?php
                 $title = __('Who can add/remove sentences', true);
-                $loader = "<div class='is-editable loader-container'></div>";
+                $loader = "<md-progress-circular class='is-editable loader-container' md-diameter='16' style='display: none'> </md-progress-circular>";
                 echo $this->Html->tag('dt', $title.$loader);
 
                 echo $this->Form->radio(
@@ -470,7 +470,7 @@ class ListsHelper extends AppHelper
                 'class' => 'submit'
             )
         );
-        echo "<div class='loader-container'></div>";
+        echo "<md-progress-circular id='list_add_loader' md-diameter='16' style='display: none;'></md-progress-circular>";
         ?>
 
         <p>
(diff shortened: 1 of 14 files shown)


app/views/helpers/lists.php
```diff
@@ -470,7 +470,7 @@ public function displayAddSentenceForm($listId)
                 'class' => 'submit'
             )
         );
-        echo "<div class='loader-container'></div>";
+        echo "<md-progress-circular id='list_add_loader' md-diameter='11px' style='display: none;'></md-progress-circular>";
```
trang: `md-diameter` can be increased to 16px. 11px is a bit too small.
Also note that you don't need to specify "px". `md-diameter="16"` works too.

app/views/helpers/menu.php
```diff
@@ -645,10 +647,7 @@ public function displayMenu(
 
         <li>
         <?php
-        echo $this->Html->div('loader-small loader', '', array(
-            'id' => '_'.$sentenceId.'_in_process',
-            'style' => 'display:none',
-        ));
+        echo "<md-progress-circular  md-diameter='11px' style='display: none;' id='_" . $sentenceId ."_in_process'></md-progress-circular>";
```
trang: The loaders up there for the favorites and correctness can be removed, and you can use this `#_{sentenceId}_in process` loader instead.

app/views/helpers/sentences.php
```diff
@@ -553,6 +550,7 @@ public function displayGenericSentence(
         if( $isFavoritePage && $this->params['pass'][0] == CurrentUser::get('username') ){
             echo '<div class="favorite-page column">';
             $this->Menu->favoriteButton($sentenceId, true, true, true);
+            echo '<li class="option favorite progress" style="display: none;"><md-progress-circular md-diameter="24px"></md-progress-circular></li>';
```
trang: This loader doesn't align very well, but I think to fix the alignment we'd need to review the HTML of the whole `displayGenericSentence()` function, so we can leave it like that for now.

app/webroot/js/sentences.show_another.js
```diff
@@ -16,8 +16,9 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-
+$("#random-progress").hide();
```
trang: This looks strange as it's outside of the `$(document).ready()`. What is for?

app/views/helpers/menu.php
theblindspring: I am not sure what code you are talking about? Are you saying I should remove the 'loader-small' class from 'ul' loader?

theblindspring: Okay, so you are saying reuse the loader used for adding a sentence to a list?

trang: Yes.

app/views/helpers/lists.php
theblindspring: Should I change just these instances or all instances of 11 to 16?

trang: All instances. Unless it breaks the UI, in which case, just make it as close as possible to 16px.

Issue #32, opened 2014-02-17
Support drafts of private messages


loolmeh:
In most e-mail clients, there's a way to save messages as drafts without sending them. This lets the user work on them until they're ready to send. This would be nice functionality to provide in our private message area as well.

----

Original ticket: https://www.assembla.com/spaces/tatoeba2/tickets/408

RyckRichards:
Requested on the 1st Tatoeba day (2015-06-06)
https://tatoeba.org/eng/wall/show_message/22959#message_22959

trang:
Writing down some notes about a possible implementation.

* Add the value `Draft` in the `folder` enum of the `private_messages` table.
* Add function `saveDraft()` in PrivateMessage model.
* Add function `save_draft` in PrivateMessagesController.
* Add button **Save as draft** in the private message form.
* Modify `send()` function in PrivateMessagesController to handle "Save as draft".
* Add a link **Drafts** in the list of folders in the sidebar.
* Modify `folder()` function in PrivateMessagesController to retrieve Drafts messages.
* Modify URL in `private_messages/folder.ctp` for Draft messages: they should redirect to `write` and not `show`.
* Modify `write()` function in PrivateMessagesController to take into account an extra param: `messageId`. When the messageId is set, the content of the form should be pre-filled with the content of the corresponding message.
* Modify `private_message/write.ctp` to take into account the content of the Draft message when displaying form.

RyckRichards:
According to Guybrush88, he wants something (pretty) similar as drafts in e-mail clients which would help us to decide to either to send it or save it in order to send it in a second moment

zachleigh:
While going through this issue, I found that in private_messages_controller, the write method takes $recipients as its first argument. According to the doc-block, its supposed to be a string, but the database column is of type int.

| Field     | Type                               | Null | Key  | Default | Extra          |
|-----------|--------------------------------|-----  |--------|----------|----------------|
| recpt     | int(11)                            | NO   | MUL | NULL    |                |

Setting the recipients simply puts a userid in the recipients field, without actually finding the user. is this the desired behavior?

trang:
I'd go for a shorter solution: 

* Don't save the recipient(s) if the user saves as draft.
* Add a tooltip on hover on the "Save as draft" button: `Only the title and content of your message will be saved. The recipient will not be saved.`
* The avatar used for drafts would then be the "unknown" avatar.

zachleigh:
[...]

PR #1206, opened 2016-06-17
Add draft feature to private messages.

zachleigh:
This pull request addresses issue #32.

### DIFF
--- a/app/controllers/private_messages_controller.php
+++ b/app/controllers/private_messages_controller.php
@@ -77,7 +77,7 @@ class PrivateMessagesController extends AppController
         $conditions = array('folder' => $folder);
         if ($folder == 'Inbox') {
             $conditions['recpt'] = $currentUserId;
-        } else if ($folder == 'Sent') {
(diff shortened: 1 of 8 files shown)


trang:
It's looking good :) Just a few things to fix:

1. You want to make sure that users cannot see other people's drafts. At the moment, if I save a message as draft, copy the URL, log in as another user, and go the draft URL I copied, I can see the message.

2. When viewing a draft message, the label above the form is always "Reply", even when I save a draft for a new message. You may want to make the label more general, like "Message".

3. It would be better to remove the `md-primary` for the "Save as draft" button, so that it looks less important than the "Send" button. in general, we don't want several primary buttons, only one should be primary.

zachleigh:
Fixed the issues and changed one other thing.     

1. Enforced draft privacy by checking to make sure the private message's sender id is equal to the current user id in the controller `write` method.

2. Set the label to 'Message'.

3. Removed the `md-primary` class. The button is now white, which looks fine to me and moves focus to the Send button as desired.

4. When saving a draft, the To field of the message is no longer required. Only the content field needs to be filled.

PR #1208, opened 2016-06-17
Add Sort by Last Updated for Lists

dispyfree:
This pull request addresses issue #1100 .

+ made sentences lists sortable by the last time a sentence was added…or deleted

+ added respective triggers

### DIFF
--- a/app/models/sentences_sentences_lists.php
+++ b/app/models/sentences_sentences_lists.php
@@ -81,8 +81,19 @@ class SentencesSentencesLists extends AppModel
             'sentence_id' => $sentenceId,
             'sentences_list_id' => $listId
         );
-
+        $tmp = $this->belongsTo;
+        /**
+         * We have to temporarily remove the relation at least to sentence_list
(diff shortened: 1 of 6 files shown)


trang:
When the list name is very long, the table goes off of its content. In my case I have a list named `All you need is love (a list of romantic, cheesy, lovey-dovey sentences)`.

screenshot_6

trang:
A few more things:
* I just noticed the SQL file for the triggers is missing. 
* There's an extra file that should be remove (`list.php~`).
* Feel free to use a `git rebase -i` to squash all your commits :)

And some tips about commit messages:
* I'm not sure if there's a reason why you add `+` at the beginning of your commit messages. In our case they are not really necessary, they only use up extra characters.
* If you have a lot to say in your commit message and it doesn't fit in ~50 characters, you can split it into a title and  a description: `git commit -m "{title}" -m "{description}"`. You can see for instance [here](https://github.com/Tatoeba/tatoeba2/commit/8f7774973c3837f07dceb4a585cb7a70e053cfe6) an example of an ideal commit message.

PR #1224, opened 2016-06-24
Redesigned the users settings page.

theblindspring:
https://files.gitter.im/theblindspring/jtiu/blob
image

image

### DIFF
--- a/app/views/user/settings.ctp
+++ b/app/views/user/settings.ctp
@@ -36,258 +36,286 @@ $this->set('title_for_layout', $pages->formatTitle(__('Settings', true)));
 </div>
 
 <div id="main_content">
-    <div class="module options">
+    <div class="options">
         <?php echo $form->create(null, array('action' => 'save_settings')); ?>
+        <div md-whiteframe="1" class="settings-form">
+            <h2><?php __('Options'); ?></h2>
+            <fieldset>
+            <md-list flex role="list" class="flex" >
+                <md-list-item>
+                    <div>
+                        <?php $sendNotifications = $this->data['User']['send_notifications']; ?>
+                        <md-checkbox 
+                            ng-false-value="0" 
+                            ng-true-value="1" 
+                            ng-model="sendNotifications" 
+                            ng-init="sendNotifications = <?= $sendNotifications ?>"
+                            class="md-primary">
+                        <label> <?= __('Email notifications') ?></label> 
+                        </md-checkbox>
+                        <?php
+                            echo $form->hidden(
+                                'send_notifications',
+                                array(
+                                'value' => '{{sendNotifications}}'
+                                )
+                            );
+                        ?>
+                    </div>
+                </md-list-item>
 
-        <h2><?php __('Options'); ?></h2>
-        <fieldset>
-            <div>
-                <?php $sendNotifications = $this->data['User']['send_notifications']; ?>
-                <md-checkbox 
-                    ng-false-value="0" 
-                    ng-true-value="1" 
-                    ng-model="sendNotifications" 
-                    ng-init="sendNotifications = <?= $sendNotifications ?>"
-                    class="md-primary">
-                <label> <?= __('Email notifications') ?></label> 
-                </md-checkbox>
+                <md-list-item>
+                    <div>
+                        <?php $isPublic = $this->data['User']['settings']['is_public']; ?>
+                        <md-checkbox 
+                            ng-false-value="0" 
+                            ng-true-value="1" 
+                            ng-model="isPublic" 
+                            ng-init="isPublic = <?= $isPublic ?>" 
+                            class="md-primary"> 
+                        <label><?= __('Set your profile public?') ?></label>
+                        </md-checkbox>
+                        <?php
+                            echo $form->hidden(
+                                'settings.is_public',
+                                array(
+                                'value' => '{{isPublic}}'
+                                )
+                            );
+                        ?>
+                    </div>
+                </md-list-item>
+
+                <md-list-item>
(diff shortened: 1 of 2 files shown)


trang:
Two things we need to work on:

* The HTML and CSS can be simplified.
* The fieldsets should be removed. We'll use `<md-subheader>` to separate be "main options" and "experimental options".

Cf. https://material.angularjs.org/latest/demo/list for how to structure items in `<md-list>`. I'm posting more specific comments on the code.

app/views/user/settings.ctp
```diff
@@ -36,282 +36,318 @@ $this->set('title_for_layout', $pages->formatTitle(__('Settings', true)));
 </div>
 
 <div id="main_content">
-    <div class="module options">
+    <div class="options">
         <?php echo $form->create(null, array('action' => 'save_settings')); ?>
+        <div md-whiteframe="1" class="settings-form">
```
trang: This div is unnecessary. The attributes can be moved into the div above (`<div class="options">`).

app/views/user/settings.ctp
```diff
@@ -36,282 +36,318 @@ $this->set('title_for_layout', $pages->formatTitle(__('Settings', true)));
...
+            <h2><?php __('Options'); ?></h2>
+            <fieldset>
+            <md-list flex role="list" class="flex" >
```
trang: Removed the `<fieldset>` and add in here a `<md-subheader>` with the text `Main options`.

app/views/user/settings.ctp
```diff
@@ -36,282 +36,318 @@ $this->set('title_for_layout', $pages->formatTitle(__('Settings', true)));
...
+                <md-list-item>
+                    <div>
```
trang: This div can be removed. Same for all the div's inside the other `<md-list-item>`.

app/views/user/settings.ctp
```diff
@@ -36,282 +36,318 @@ $this->set('title_for_layout', $pages->formatTitle(__('Settings', true)));
...
+                        <?php $sendNotifications = $this->data['User']['send_notifications']; ?>
+                        <md-checkbox 
+                            ng-false-value="0" 
+                            ng-true-value="1" 
+                            ng-model="sendNotifications" 
+                            ng-init="sendNotifications = <?= $sendNotifications ?>"
+                            class="md-primary">
+                        <label> <?= __('Email notifications') ?></label>
```
trang: Move this label outside of the `<md-checkbox>`. The structure should be:
```
<md-checkbox></md-checkbox>
<p><? _(....) ?></p>
```

Note: you should remove the `=` after `<?`. 
`__('Email notifications')` is equivalent to `echo "Email notifications"`.
`__('Email notifications', true)`  is equivalent to `return "Email notifications"`.

app/views/user/settings.ctp
```diff
@@ -36,282 +36,318 @@ $this->set('title_for_layout', $pages->formatTitle(__('Settings', true)));
...
+                                    array(
+                                    'value' => '{{copyButton}}'
+                                    )
+                                );
+                            ?>
+                        </div>
+                    </md-list-item>
+                </md-list>
+            </fieldset>
+            <div layout="column">
```
trang: Use `<div layout="row" layout-align="center center">`. This will align the button in the center and you can remove the `save-btn` class from the md-button.

app/webroot/css/user/settings.css
```diff
@@ -43,4 +33,27 @@
...
+}
+
+.settings-form{
+    padding: 20px;
+    margin: 20px auto;
+}
+
+.settings-form h2{
+    margin: 0 0 40px 0;
+    text-align: center;
```
trang: I think it would be better to not center the text here. For consistency with the rest of the pages.
On every div block that had a class "module", keep the title aligned left.

The title on the "login" and "password forgotten" pages can remain centered because the blocks were not "modules".

app/views/user/settings.ctp
theblindspring: Should everything that was nested in the fieldset also be nested in the 'md-subheader' or should the subheader strictly contain text?

app/views/user/settings.ctp
theblindspring: This is what happens when I remove the div.

image

theblindspring:
I made the changes, but I think a few other issues have arisen. The label for the checkboxes is not aligned and the save buttons now take 100% width of their container.

trang:
* The misalignment of the checkbox is caused by the CSS. You have to remove:
```
md-checkbox{
    margin-bottom: 0px;
}
```

* The `layout="row" layout-align="center center"` should be on the `<div>`, not the `<md-button>`. The reason why the button is stretched is because `layout="column"` has by default a "stretch" on horizontal layout (cf. https://material.angularjs.org/latest/layout/alignment). You could also use `layout="column" layout-align="center center"` but when it comes to the container of buttons, it makes more sense to have a `layout="row"`: in case you add extra buttons, you don't want them to be aligned on top of each other, but on a same row.

* It seems you have a closing `</div>` missing somewhere because the email and password blocks are inside the options block.

* Minor thing: the `<md-subheader>` should be inside the `<md-list>`, not outside. Let's say you want to increase the margin between the two lists in the options, it would make more sense to set the margin on the `<md-list>` than on the `<md-subheader>`.

trang:
There are still a few things to rearrange, but it'll be easier I take care of it since it's mostly graphical details that I have to experiment :)

trang:
Just for the record.

**Old settings page**

screenshot_25

**New settings page**

screenshot_26

Issue #1234, opened 2016-07-02
"birthday" field lets me set an invalid date: February 31, 1916


alanfgh:
On the dev site, I was able to set my birthday to February 31, 1916. It was then displayed as "0001-11-30".

![invalid_bday](https://cloud.githubusercontent.com/assets/1213027/16541856/e4a3d746-405e-11e6-98c7-6a668b43de23.png)

RyckRichards:
I set it to November 09,1999 and it worked fine for me 

http://prntscr.com/bnyr43

RyckRichards:
The same to November 9, **1916**

![screenshot_1](https://cloud.githubusercontent.com/assets/7658430/16541880/32c85164-4068-11e6-8563-3e8b964da898.png)

PR #1244, opened 2016-07-06
Validate birthday before saving user profile.

zachleigh:
This pull request addresses issue #1234 

I also extracted the unique email check to its own method to clean up the save_basic method and make it consistent with the birthday validation check. I wrapped both checks in conditionals checking for the set property  to avoid unnecessary errors being thrown.

### DIFF
--- a/app/controllers/user_controller.php
+++ b/app/controllers/user_controller.php
@@ -282,12 +282,7 @@ class UserController extends AppController
             $this->redirect('/');
         }
 
-        $isEmailUnique
-            = $this->User->isEmailUnique(
-                $this->data['User']['email'],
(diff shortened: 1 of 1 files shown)


app/controllers/user_controller.php
```diff
@@ -299,6 +294,18 @@ public function save_basic()
             );
         }
 
+        if (!$this->_isValidBirthday($this->data)) {
+            $this->Session->setFlash(
+                __("The entered birthday is an invalid date. Please try again", true)
```
trang: Small detail: missing period at the end of `Please try again`.

zachleigh:
Fixed

Issue #863, opened 2015-10-25
Diacritics are not displayed correctly.


RyckRichards:
[By wells](https://tatoeba.org/eng/wall/show_message/24413#message_24413)

> I'm having a minor problem with diacritics getting cut off in the sentence view. Here we have the word 'Ääniala' but it renders as 'Aäniala':
> 
> https://i.imgur.com/AQyDthv.png
> 
> This only applies to sentences I own, so I think the buttons bar is at fault. This is on Mozilla/*nix.
> 
> Here's a second related bug I experience:
> 
> http://i.imgur.com/6MVVHcq.png
> 
> When submitting a new sentence, the 'diacritics' get cut off in the same way. The newly-submitted sentence #4630768 gets chopped but the pre-existing sentence #4629634 displays correctly.
> 
> If I reload the page, both display correctly.
> 
> I need to double-check these sentences for typos whenever this rendering bug shows up.
> 
> And here's a CSS solution that works for me -- fixes both problems and one more:

`div.editableSentence, .addTranslationsTextInput { padding-top: 4px; }`

> (The text input was tight as well, so I added a padding there.)
> 
> I know languages such as Vietnamese use more complex diacritics that take more vertical space. If someone would produce a sentence for me to test with such a language, I'd be happy to test it.
> 
> Edit: It seems the input textboxes are all a bit tight vertically. The Å in Åke gets cut off in the topmost search bar, and in the 'Advanced search' form. This CSS rule fixes it:
> 
> input[type="text"] { padding-top: 5px !important; }
> 
> (It's an 'override' rule. Not something to check into the code.)

RyckRichards:
[wells](https://tatoeba.org/eng/user/profile/wells) can still reproduce this bug - https://tatoeba.org/eng/wall/show_message/26027#message_26027

halfdan:
I believe this was fixed with #1245 - I can't reproduce this anymore.

PR #1245, opened 2016-07-07
Change font to Noto

trang:
Ref. #1180 

This pull request modifies the font used in Tatoeba, from Trebuchet MS to Noto. The font sizes have also be readjusted.

Since Noto is not installed by default on most systems, this means that most users will end up seeing Tatoeba with the default sans-serif font of their browser (Arial, Helvetica... who knows), until they download and install Noto.

I hesitated for a long time between using Roboto vs. Noto, and ended up opting for Noto. And this is still not a stable decision.

Roboto is the font which Material Design is designed with, and if we want to be consistent with the spec, we should use Roboto. It offers a larger range of font weights than Noto. The difference is mostly noticeable for buttons and sub-headers, where the font weight is 500. With other fonts, the thickness of the characters appears thinner than with Roboto, which can make the labels for buttons and sub-headers a bit harder to read.

Roboto however doesn't cover all scripts, and Material Design relies on Noto for scripts that are not covered by Roboto.

Noto's base height is slightly bigger than Roboto, and Arial, for that matter. In other words, for the same font-size, the characters in Noto take more vertical space than Roboto. This can make designing the UI a bit more complicated... but this is something we need to deal with anyway for languages like Marathi, one of the languages in which the UI is translated.

With the current setup, we don't include the fonts in the HTML header. Users will have to download and install the fonts if they want to see Tatoeba in its new font. Later on, we can create options for users to have these fonts included. But right now, anyone who wants Tatoeba to look prettier (or more accurate) has to download the fonts and/or configure their browsers settings accordingly. In this context, it feels easier for users to go to only one website (the Noto website) in order to download the whole font set.

In any case, whatever Web font we choose, it won't affect the lambda user at this time.

Also, for now, only Noto Sans is included in the CSS. The other Noto fonts for other languages have not been included yet. They will be introduced slowly, when we can get confirmation from native speakers that the font is appropriate.

### DIFF
--- a/app/webroot/css/layouts/default.css
+++ b/app/webroot/css/layouts/default.css
@@ -56,8 +56,45 @@
 /*
  * Generic rules
  */
-body, select, input, textarea, button {
-    font-family: Trebuchet MS, sans-serif;
+body, select, input, textarea, button, option, optgroup {
+    font-family: "Noto Sans", sans-serif;
+
+    /* Fonts for other languages */
+    /*
+    "Noto Kufi Arabic", "Noto Naskh Arabic",
+    "Noto Nastaliq Urdu", "Noto Sans Armenian", "Noto Sans Avestan",
+    "Noto Sans Balinese", "Noto Sans Bamum", "Noto Sans Batak",
+    "Noto Sans Bengali", "Noto Sans Brahmi", "Noto Sans Buginese",
+    "Noto Sans Buhid", "Noto Sans CJK JP", "Noto Sans CJK KR",
+    "Noto Sans CJK SC", "Noto Sans CJK TC", "Noto Sans Canadian Aboriginal",
+    "Noto Sans Carian", "Noto Sans Cham", "Noto Sans Cherokee",
+    "Noto Sans Coptic", "Noto Sans Cuneiform", "Noto Sans Cypriot",
+    "Noto Sans Deseret", "Noto Sans Devanagari", "Noto Sans Egyptian Hieroglyphs",
+    "Noto Sans Ethiopic", "Noto Sans Georgian", "Noto Sans Glagolitic",
+    "Noto Sans Gothic", "Noto Sans Gujarati", "Noto Sans Gurmukhi",
+    "Noto Sans Hanunoo", "Noto Sans Hebrew", "Noto Sans Imperial Aramaic",
+    "Noto Sans Inscriptional Pahlavi", "Noto Sans Inscriptional Parthian",
+    "Noto Sans Javanese", "Noto Sans Kaithi", "Noto Sans Kannada",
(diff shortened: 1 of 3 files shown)


alanfgh:
In Firefox on Windows 7, the content was unreadable until I installed Noto Sans from the [Google Noto page](https://www.google.com/get/noto/). Here's what it looked like:

without_noto

PaulPeer:
The diacritics on the capital letters are not visible anymore in Firefox on Mac (but OK on Chrome and Safari).

trang:
@alanfgh, currently, what happens is that if you don't have Noto installed, your browser will use the "sans-serif" font that is defined in your browser's preferences. By default, on Windows, that would be Arial. 

My guess is that you have another font than Arial defined in your Firefox preferences for "Other Writing Systems" > "Sans-serif". For some reason this other font doesn't handle very well the spacing for Latin characters. I'm not sure if it's a bug from the font, or if that's intentional... But if you uninstall Noto and change your "Sans-serif" font settings to Arial (or any other standard sans-serif font), the comments should become readable again. 

However, I assume you've set this font in your preferences because of a certain language in Tatoeba that requires a special font. Which means you wouldn't be able to view the sentences in that language anymore. Do you remember which language it is? And could you also tell me which font you are using?

trang:
@PaulPeer

> The diacritics on the capital letters are not visible anymore in Firefox on Mac (but OK on Chrome and Safari).

This problem is not related to the new font configuration. It was reported already before: #863

alanfgh:
@trang , I think you're right. My default Firefox font (Options/Content/Fonts & Colors) is Padauk. I think I might have installed it in order to see Burmese or some related language back when we were adding it.

PaulPeer:
@trang 

> This problem is not related to the new font configuration. It was reported already before: #863

Sure? The problem #863 started years ago. The bug that I noticed started 3 days ago. That bug concerned only the sentences of the owner; this bug concerns all sentences with diacritics on capital letters. The accents disappear in the Italian "É", the german "Über", the Esperanto "Ĉu"...

trang:
@PaulPeer, to clarify, what I meant is that the bug you are experiencing already existed before the recent changes with the font. It will happen depending of the font your are using, but it's not a problem with the font itself. It's a problem with the layout of the sentences (height, margins, paddings...). The browser may also have an influence because the way browsers display a page can differ from a few pixels.

You were not experiencing this bug before because the previous font (Trebuchet MS) was installed on your computer, and with that font, diacritics on capital letters could be displayed properly within the sentence layout. Other people who did not have Trebuchet MS installed were not seeing the diacritics because their browser was using some other font, and the text that other font does not fit perfectly in the layout.

Issue #1249, opened 2016-07-10
I would suggest not requiring members to enter their birthdays when editing their profiles.


ckjpn:
Members are now required to enter a birthday in order to edit their profiles.  Was this intentional?

## Problem

This is an unneeded invasion of privacy, requiring members to lie if they don't want to give this information.

This is not information that we need.

![screen shot 2016-07-10 at 23 41 22](https://cloud.githubusercontent.com/assets/7934273/16714265/54b23308-46f8-11e6-848f-99849c4d4a3f.png)

trang:
No it wasn't intentional.

@zachleigh, I'm assign this to you. It's a regression that I failed to notice when merging your pull request for #1234.

zachleigh:
Ok. Ill get on it now.

trang:
@ckjpn this should be corrected on the main website now. I let you test it and close this issue if it's good.

ckjpn:
I was able to remove the date from my profile.
https://tatoeba.org/eng/user/profile/CK

Issue #1235, opened 2016-07-02
"birthday" value that lacks month, day, or year is not displayed at all


alanfgh:
On the dev site, I tried saving my birthday without the month, without the day, and without the year. Only when all three are saved is anything displayed. There are valid reasons for wanting to display partial dates:

- displaying the year lets people know my age, but keeps my birthday secret
- displaying the month and day lets people know my birthday, but keeps my age a secret

zachleigh:
Ill take this one.      
   
The reason why we cant save partial dates is that the database table requires a DATETIME and partial/invalid dates don't save properly using the current code. I think if we changed the unspecified values to 00 it would work.    
    
Also, this column should really be a DATE column rather than a DATETIME. I think changing the column type will simply cut the timestamps and leave the date.

zachleigh:
So I got this mostly working. It took a little more effort than I had expected...the real problem was with the php date function. When given a partial date, it falls back to a valid date rather than allowing an incomplete date. I wrote a little date parser which allows partial dates and displays them in the current format. However, the current date format (Y-m-d) makes the date strings awkward. Currently, they look like this:     
  - Full date (2005-06-15)
  - Only year (2005-00-00)
  - Month and day (0000-06-15)
  - Month and year (2005-06-00)
    
To me, these dates are strange and difficult to understand. I think dropping the zero strings makes them look much better:
  - Full date (2005-06-15)
  - Only year (2005)
  - Month and day (06-15)
  - Month and year (2005-06)     

But of course that breaks the convention a little. Thoughts?

trang:
Ideally we should format the date to be more human-friendly: #1219.

Having the year alone, or year and month look more or less fine as 2005 and 2005-06. But the month and day is not very intuitive as 06-15.

Until we have translatable dates, I would go for question marks, like this: 
* 2005-??-??
* ????-06-15
* 2005-06-??

trang:
@zachleigh, I've merged your PR with the printed zeros. Probably some users will report that it's confusing, but we'll see what they prefer to use instead of zeros. In the end this should be only a temporary state until we solve #1219.

PR #1250, opened 2016-07-10
Fix regression that forced users to enter birthday when saving profile.

zachleigh:
This pull request addresses issue #1249.      
     
This allows users to save profiles with an incomplete or unset birthday. The only problem I can see is that if a user tries to save an incomplete birthday (only the year, for example), the database won't save it and they won't get any error message explaining why. If this is a problem, we can add an additional check for incomplete birthdays and give users an appropriate error message.

### DIFF
--- a/app/controllers/user_controller.php
+++ b/app/controllers/user_controller.php
@@ -363,7 +363,7 @@ class UserController extends AppController
     }
 
     /**
-     * Return true if entered birthday is a valid date.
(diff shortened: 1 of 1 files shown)


trang:
Thank you for the quick fix :)

> The only problem I can see is that if a user tries to save an incomplete birthday (only the year, for example), the database won't save it

This will be handled as an enhancement: #1235 (I suppose I can assign that issue to you?).

Issue #1252, opened 2016-07-11
Pashto and Sindhi are listed as LTR languages, they should be RTL


sabretou:
This is my first issue here, so I hope I'm doing this right.

I noticed that Pashto and Sindhi sentences are aligned left and have their punctuation to the right of the sentence. Like other languages that use the Arabic script, Pashto and Sindhi are written from right-to-left, and have their punctuations to the left of the sentence. This needs to be fixed.

RyckRichards:
I am on it!

sabretou:
@RyckRichards Don't forget Sindhi!

PR #1253, opened 2016-07-11
Fixed Pashto and Sindhi script from LTR to RTL

RyckRichards:
This pull request addresses #1252 

It fixes Pashto script which, like other Arabic languages, should be written Right to Left. Currently, it is Left to Right.

### DIFF
--- a/app/vendors/languages_lib.php
+++ b/app/vendors/languages_lib.php
@@ -604,6 +604,7 @@ class LanguagesLib
             "aii",
             "afb",
             "pus"
+            "snd"
         );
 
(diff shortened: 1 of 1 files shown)


app/vendors/languages_lib.php
```diff
@@ -603,6 +603,8 @@ public static function getLanguageDirection($lang) {
             "ary",
             "aii",
             "afb",
+            "pus"
```
trang: There's a comma missing.

RyckRichards: Thanks, @trang . I've just fixed this.

-- 
Ricardo Vernaut Jr

PR #1264, opened 2016-07-18
Allow for partial birthdays.

zachleigh:
This pull request addresses issue #1235.    
     
The only thing that doesn't work is saving a leap year without a year value. This works fine: "February 29, 2011" but if you take the year off and try "February 29" Mysql refuses to save the date. This edge case probably isnt that big of a deal, but if you want it fixed, the only thing I can think to do is enter in an impossible birthday that is a valid leap year, like February 29, 1804. In the date formatter, we could then catch that impossible birthday and remove the year. Its hacky, but it would probably work.

### DIFF
--- a/app/controllers/user_controller.php
+++ b/app/controllers/user_controller.php
@@ -393,9 +393,8 @@ class UserController extends AppController
             if ($year && $month && $day) {
                 return checkdate($month, $day, $year);
             } elseif ($month && $day) {
-                // Use 2015 because its not a leap year. Leap year dates without
-                // year fail when saving in mysql.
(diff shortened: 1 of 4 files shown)


trang:
> the only thing I can think to do is enter in an impossible birthday that is a valid leap year, like February 29, 1804

@zachleigh, go for it. Apparently that's what people do, and I can't think of another solution either.

Cf. http://stackoverflow.com/questions/4998470/storing-day-and-month-without-year

zachleigh:
The impossible date is 1904. This date cant be selected from the year drop down, so it should be safe.

PR #1276, opened 2016-08-02
Refactor private messages

zachleigh:
Things accomplished by this refactor:
  - Added clear, readable documentation blocks to all methods.
  - Extracted logic from main methods (send, write, show etc.) to make them more readable and easier to maintain.
  - Moved some of the logic to the model.

This refactor made the controller a little bigger, but this is mostly documentation blocks, which is a positive thing in my mind.

### DIFF
--- a/app/controllers/private_messages_controller.php
+++ b/app/controllers/private_messages_controller.php
@@ -66,46 +66,10 @@ class PrivateMessagesController extends AppController
 
         $folder = Sanitize::paranoid($folder);
 
-        $currentUserId = $this->Auth->user('id');
-
-        $conditions = array('folder' => $folder);
-
-        if ($folder == 'Inbox') {
-            $conditions['recpt'] = $currentUserId;
-        } else if ($folder == 'Sent' || $folder == 'Drafts') {
-            $conditions['sender'] = $currentUserId;
-        } else if ($folder == 'Trash') {
-            $conditions['user_id'] = $currentUserId;
-        }
-
-        if ($status == 'read') {
-            $conditions['isnonread'] = 0;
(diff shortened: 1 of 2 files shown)


trang:
Thanks a lot for your refactoring :) Here's my feedback.

1) I noticed you added square brackets around the description of the parameters in the code documentation. Is this just to improve the readability (to make it easier to see where the description starts), or is there another reason? My main concern is that it becomes inconsistent with the rest of the code documentation, where there's no brackets around params description. I can undertand the readability issue, but I feel it is still better to remain conistent.

2) There are some bits of code you could move from the controller to the model.
  * https://github.com/zachleigh/tatoeba2/blob/bb0565cdd3c8b844f91c36a9e8b8706403fc3ac8/app/controllers/private_messages_controller.php#L87 => You can follow what's done for the [lists model](https://github.com/zachleigh/tatoeba2/blob/bb0565cdd3c8b844f91c36a9e8b8706403fc3ac8/app/models/sentences_list.php#L136) and [controller](https://github.com/zachleigh/tatoeba2/blob/bb0565cdd3c8b844f91c36a9e8b8706403fc3ac8/app/controllers/sentences_lists_controller.php#L95).
  * https://github.com/zachleigh/tatoeba2/blob/bb0565cdd3c8b844f91c36a9e8b8706403fc3ac8/app/controllers/private_messages_controller.php#L262 => I think code that is constructing data belogns to the model rather than the controller.

zachleigh:
1) Yes, I've always added brackets around my documentation descriptions to improve readability. But you are right, consistency is more important. I will remove them.    
     
2) Ok. I'll pull those two methods out and put them in the model.

PR #1286, opened 2016-08-13
Redesign "Add sentences" page

trang:
This pull request is part of #1180.

**Old "Add sentences" page**

screenshot_37

**New "Add sentences" page**

screenshot_38


RyckRichards:
How about this change?

From: Language

To: Language:

trang:
I don't think it makes a big difference and in general, I think we will try to avoid adding colons in labels, so I'll keep it without the colons.

Issue #1246, opened 2016-07-10
There is no error message when accessing /vocabulary/of/ a non-existing username.


ckjpn:
## Minor Problem

This URL goes to a page without an error message.
https://tatoeba.org/eng/vocabulary/of/there-is-no-member-by-this-name

## Solution

Perhaps you should put an error message for that, somewhat similar to how this URL does it.
https://tatoeba.org/eng/user/profile/NOT_A_REAL_USERNAME

This one doesn't get an error message.  I guess it's because of the hyphens.
https://tatoeba.org/eng/user/profile/there-is-no-member-by-this-name

## Note

Perhaps you could also just ignore this, but I thought you'd like to know.

PR #1290, opened 2016-08-20
Redirect to members page with error for vocabulary/of/ invalid user.

zachleigh:
This pull request addresses issue #1246.     

When an invalid username is given, users will be redirected to the members list with an error message.

### DIFF
--- a/app/controllers/vocabulary_controller.php
+++ b/app/controllers/vocabulary_controller.php
@@ -63,14 +63,27 @@ class VocabularyController extends AppController
     {
         $this->helpers[] = 'Pagination';
         $this->helpers[] = 'CommonModules';
+
+        $username = Sanitize::paranoid($username, array('_'));
         
         $userId = $this->User->getIdFromUsername($username);
-        $this->paginate = $this->UsersVocabulary->getPaginatedVocabularyOf(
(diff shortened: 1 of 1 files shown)


app/controllers/vocabulary_controller.php
```diff
@@ -63,14 +63,27 @@ public function of($username, $lang = null)
...
         $this->helpers[] = 'Pagination';
         $this->helpers[] = 'CommonModules';
+
+        $username = Sanitize::paranoid($username, array('_'));
         
         $userId = $this->User->getIdFromUsername($username);
+
+        if (!$userId) {
+            $this->Session->setFlash(
+                __('No user with this username: ', true).$username
```
jiru: `$username` should be included in the translated string as a placeholder, so that translators are not forced to put it at the end of the sentence. Besides, it’s also currently unclear for the translators that the string is to be followed by a username. You can find some code examples [here](https://github.com/Tatoeba/tatoeba2/search?utf8=%E2%9C%93&q=no+such+user).

trang: I'm taking care of this. There are a few other places where a placeholder should be used.

Issue #1278, opened 2016-08-04
Number of sentences in vocabulary list is not updated properly


trang:
As a contributor, I decide to add new sentences to illustrate vocabulary that do not have sentences yet. I go to "Contribute" > "Add sentences" then click on "Check out the vocabulary for which we need sentences". This leads me to a page where I can see which vocabulary items have less than 10 sentences.

I click on the `+` icon to add a sentence for a certain vocabulary that has 0 sentences. After the sentence is saved, the number of sentences is still 0. The count is only updated if I add another sentence a bit later.

How the count is updated (cf. [Wall message](http://tatoeba.org/eng/wall/show_message/26688#message_26688)):

> The count is updated whenever someone adds a new sentences for a vocabulary item, from the page "Vocabulary that needs sentences". But this number depends on whether or not the sentence has been indexed by the search engine. The sentences are indexed every 15 minutes I think.
> 
> So in other words, if 10 people add a sentence for "splice" in the same minute, the count will remain at "0 sentences" (because it counts only sentences that are already indexed by the search engine). And if 15 minutes later someone adds another sentence for "splice", it will jump to "10 sentences".

This makes it difficult for contributors to keep track of which vocabulary they already contributed for, and also makes them less motivated to add more sentences.

----

A quick fix would be to systematically increment the number of sentences for a vocabulary item whenever someone adds a sentence for it. It would give a better user experience than currently but the number of sentences can still be inconsistent.

Ideally, we would have to take into account the following cases:

1. We have to make sure that the new sentence contains the vocabulary item. If I'm adding a sentence for the vocabulary "apple", but my sentence is "I bought oranges", the count shouldn't be incremented.
2. If the count is incremented, users will expect to see their sentence in the list of results when they click on the `X sentences` link. But the search engine doesn't index sentences right away, which will probably make users feel like there's a bug.
3. It can happen that a sentence containing a certain vocabulary item gets deleted, or gets modified in a way that it no longer matches the vocabulary.

PR #1292, opened 2016-08-23
Improve vocabulary list number of sentences functionality

zachleigh:
This pull request address issue #1278.

### Behavior
  - When a user adds a new sentence to a vocabulary item, the numSentence value on the item is incremented by one if the sentence contains the vocabulary item.
  - In order to avoid data consistency issues, when a user visits the vocabulary/of/username page, the vocabulary numSentences values are synced with the Sphinx index count. This way, users who regularly use and keep track of their vocabulary will have more accurate numSentence values.
  - When searching for a word, if a vocabulary item exists and its numSentence count differs from the Sphinx count, this message will appear below the search title: 'Recently added sentences may not appear in search results.’ This should help avoid confusion when users dont see their sentences immediately after adding them.

## Notes
  - In the end, I decided to us a simple `mb_stripos` statement instead of a regex to determine if a sentence contains a word. The logic has been extracted to its own method so if we want to improve it now or in the future, it should be easy.
  - I wanted some way to keep the numSentence values and the actual Sphinx count synced. Doing a cron job would be another option, but it may be overkill. By doing it when a user loads their vocabulary page, we can push responsibility for this to the user so only vocabulary for active users will be regularly synced. Possible downside is that it could make loading the page a little slower but I havent noticed any significant slowdown on production.

### DIFF
--- a/app/controllers/sentences_controller.php
+++ b/app/controllers/sentences_controller.php
@@ -72,6 +72,7 @@ class SentencesController extends AppController
         'UsersLanguages',
         'Tag',
         'UsersSentences',
+        'Vocabulary'
     );
 
     private $defaultSearchCriteria = array(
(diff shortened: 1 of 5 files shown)


trang:
Looks good to me :) Let's see how it goes in production.

PR #1294, opened 2016-08-25
Redesign search bar

trang:
This pull requests is part of issue #1180. This is a small rework of the search bar. 

Visually, there is no big difference beside the arrow icon which has been replaced with the Material Design icon `swap_horiz`. This makes it clearer for the user that they can swap the languages by clicking on the icon. The icon also takes more space and should be a bit easier to tap on mobile devices.

The re-positioning of the elements based on the width of the window is also a bit different. Before, the elements would take as much space as possible, and the elements that did not have enough space would go to the next line. Now, if the width is lower than 960px, both language dropdowns are displayed on a second line, and the search button is displayed on a third line.

**Old search bar**

screenshot_44

screenshot_45

screenshot_46

**New search bar**

screenshot_42

screenshot_43

### DIFF
--- a/app/views/elements/search_bar.ctp
+++ b/app/views/elements/search_bar.ctp
@@ -47,82 +47,80 @@ echo $form->create(
     )
 );
 ?>
-<fieldset class="input text">
-    <label for="SentenceQuery">
-        <?php __('Search'); ?>
-    </label>
-    <?php
-    echo $html->div('search-bar-extra');
-    echo $html->link(
-        __('Help', true),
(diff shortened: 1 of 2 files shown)


jiru:
I think the background color of the search text field is not set. As a result, it depends on what the clients browser uses as default. On my browser, it’s black, which makes it inconsistent with the rest. Here is a screenshot of dev.tatoeba.org:
bug

When I browse tatoeba.org, it’s white.

Issue #1293, opened 2016-08-23
Duplicate vocabulary item causes sql error


zachleigh:
This issue was originally described here: https://github.com/Tatoeba/tatoeba2/issues/1148#issuecomment-241605056

## Problem
In the current dev branch, adding a duplicate vocabulary item causes a sql error and screen freeze, but no message is given to user. In model vocabulary.php
```
    public function addItem($lang, $text) {
        $text = trim($text);
        if (empty($text) || empty($lang)) {
            return null;
        }
        $numSentences = $this->_getNumberOfSentences($lang, $text);
        $hash = murmurhash3($lang.$text);
        $data = array(
            'id' => $hash,
            'lang' => $lang,
            'text' => $text,
            'numSentences' => $numSentences
        );

        $this->save($data);
        $this->UsersVocabulary->add($hash, CurrentUser::get('id'));

        return $data;
    }
```
`id` in data needs to be unique. 

## Possible solutions: 

- Wrap `$this->save($data);` in a try/catch block and $return an error if sql throws error
- Check for unique hash value before saving and return error

In the controller, check for 'error' index on $result:
```
        $result = $this->Vocabulary->addItem($lang, $text);
        if (!isset($result['error'])) {
                $hexValue = unpack('H*', $result['id']);
                $result['id'] = str_pad($hexValue[1], 32, '0');
                $numSentences = $result['numSentences'];
                $result['numSentencesLabel'] = format(
                    __n('{number} sentence', '{number} sentences', $numSentences, true),
                    array('number' => $numSentences)
                );
        }
        $this->set('result', $result);

        $this->layout = 'json';
```
And then pick up the error in add.ctrl.js.    
     
Note by @jiru:

> You can also use CakePHP’s data validation system to make CakePHP to check for uniqueness before saving. This way, the error is not reported by SQL but CakePHP, and thus easier to catch and handle.

trang:
As discussed, this issue happens only for developers who have their debug level set to something higher than 0. The response to the HTTP POST request contains additional warning message outputed by CakePHP.

```
Warning (512): SQL Error: 1062: Duplicate entry '1vd4fko' for key 'PRIMARY' [CORE\cake\libs\model\datasources\dbo_source.php, line 684]
Code | Context
DboSource::showQuery() - CORE\cake\libs\model\datasources\dbo_source.php, line 684
DboSource::execute() - CORE\cake\libs\model\datasources\dbo_source.php, line 266
[...]

PR #1299, opened 2016-09-02
Prevent mysql error when adding new vocabulary in development.

zachleigh:
This pull request address issue #1293.     

I also added comments to existing methods on the vocabulary and users_vocabulary models. If you want this in a separate pull request let me know.


trang:
@zachleigh as usual thanks for your pull request =)

No need to create a separate pull request when you add a few comments to the code or make some small refactoring on the fly while solving an issue. Separating it in a separate commit is enough. Having a separate pull request would be relevant if you're undertaking a refactoring that needs several days (or weeks, months) to complete.

Issue #1302, opened 2016-09-07
Mi'kmaq (mic)


cueyayotl:
* Language name: Mi'kmaq
* Language info: https://en.wikipedia.org/wiki/Mi%27kmaq_language
* ISO 639-3 code: mic
* Public List: https://tatoeba.org/eng/sentences_lists/show/6424
* Icon: https://s14.postimg.org/67aqj2cy9/MIC.png
* Requested by: https://tatoeba.org/eng/user/profile/milo

=========================================

[for dev.tatoeba.org]
CALL add_new_language('mic', 4264);

[for tatoeba.org]
CALL add_new_language('mic', 6424);

RyckRichards:
taking this.

PR #1305, opened 2016-09-08
Added language Mi'kmaq

RyckRichards:
This PR addresses #1302


app/vendors/languages_lib.php
```diff
@@ -544,6 +544,7 @@ function __($string) {
                 'rue' => __d('languages', 'Rusyn', true),
                 'tah' => __d('languages', 'Tahitian', true),
                 'tly' => __d('languages', 'Talysh', true),
+                'mic' => __d('languages', 'Mi'kmaq', true),
```
trang: @RyckRichards, you need to use double quotes (`"Mi'kmaq"`), similarly to some other languages like: https://github.com/Tatoeba/tatoeba2/blob/dev/app/vendors/languages_lib.php#L525

RyckRichards:
Done! Thanks a lot, @trang and sorry for this.

Issue #577, opened 2015-02-08
Detect if a sentence is a duplicate and don't add it if it is


trang:
We have a script that eliminates/merges duplicates from the database but it would be good to detect if a sentence is a duplicate before it gets added, and simply not add it if it's a duplicate.

Note that we need to distinguish 2 situations:
1. The duplicate sentence is detected while adding a **new sentence**. In this case we just don't add it.
2. The duplicate sentence is detected while adding a **translation**. In this case still don't add the sentence itself, but we probably want to add the links.

ckjpn:
Perhaps this is no longer needed now that the duplicate-merging script is being run so often.

trang:
@zachleigh, here's some more details about this issue.

There is currently no mechanism in Tatoeba to prevent users from adding a sentence that already exist. This has been the case since the beginning of Tatoeba. This is a problem because:

* It generates unnecessary work for contributors. Let's say I add a sentence today, and it gets translated into a bunch of languages. Someday you add the exact same sentence. People might translate it again into languages for which translations already exist.
* It generates unnecessary data to download and process for people who are using our [CSV files](https://tatoeba.org/eng/downloads).
* And it probably generates a bunch of other problems which don't cross my mind right now.

Over the years, the database accumulated a significant amount of duplicate sentences. To clean up the database from existing duplicates, a couple years ago, [a script has been developed](https://github.com/Tatoeba/tatoeba2/milestone/33?closed=1). It currently runs every 15 minutes, if I remember correctly. The actions performed by the script are stored via the user account [Horus](https://tatoeba.org/eng/users/show/63758) (all the comments and logs of this account were created by the deduplication script).

This script did solve the problem of having tons of duplicates in the database, but it's not the optimal way to handle duplicates. It generates additional data (comments and logs) which would not be there if the duplicate sentence was not added in the first place. It can create confusion for users who are unaware of the deduplication process: they contribute a lot of sentences, a significant amount of their contributions are duplicates, they get removed, they wonder why they lost sentences. And it has other problems which I won't detail here.

----

Possible scenarios and desired behaviors

**1) I'm adding a sentence that already exists.**

==> The sentence appears in the list of added sentence (just like the vocabulary feature), but with a warning that it already exists so that the user knows their sentence was not added, and who added it before them.

```
	(!) The sentence you tried to add already exists.
	
	(menu icons)				belogns to SomeUser
	------------------------------------------------
	# Duplicate sentence text
```

[...]

PR #1311, opened 2016-09-12
Detect duplicate sentences when adding/editing sentences

zachleigh:
This pull request address issue #577.     
     
Notes:     
  - I extracted the hashing logic out the the Hashable behavior to make it easier to reuse.
  - When a new duplicate sentence is added, the sentence appears in the 'Sentences added' section with a red warning icon and a message that reads: 'This sentence already exists.'
  - When a new duplicate sentence is added through the vocabulary section, the normal arrow icon is replaced with a grey warning icon and a tooltip says: 'This sentence already exists.' Not sure how I feel about this tooltip. If you want to approach this a different way, I can either do it myself or take out all the front end stuff and let you do it.
  - When a sentence is edited and it becomes a duplicate, it is simply not added and the original sentence is returned. There is currently no message which is going to be confusing for users. Im not sure how to go about adding this message...
  - When a duplicate sentence is added via a translation, the two sentences are simply linked.
  - There was a TODO comment in the sentences controller saying that the edit logic should be extracted to the model and sentences should not be edited if they include audio. This has been done and the comment removed.
  - The pull request adds a `hash` column to the sentences table, but does not contain a script to hash the sentences already in to the database.

### DIFF
--- a/app/models/sentence.php
+++ b/app/models/sentence.php
@@ -41,12 +41,11 @@
 App::import('Model', 'CurrentUser');
 App::import('Sanitize');
 App::import('Vendor', 'LanguagesLib');
-App::import('Vendor', 'murmurhash3');
 
 class Sentence extends AppModel
 {
     public $name = 'Sentence';
-    public $actsAs = array('Containable', 'Transcriptable');
+    public $actsAs = array('Containable', 'Transcriptable', 'Hashable');
 
     const MIN_CORRECTNESS = -1;
     const MAX_CORRECTNESS = 0;
@@ -765,9 +764,9 @@ class Sentence extends AppModel
      */
     public function saveNewSentence($text, $lang, $userId, $correctness = 0)
     {
-        $hash = murmurhash3($lang.$text);
+        $hash = $this->makeHash($lang, $text);
 
-        if ($sentence = $this->findByBinary($hash)) {
+        if ($sentence = $this->findByBinary($hash, 'hash')) {
             $this->id = $sentence['Sentence']['id'];
 
             $this->duplicate = true;
@@ -789,20 +788,6 @@ class Sentence extends AppModel
         }
 
         return $this->save($data);
-    }
-
-    /**
-     * Find a users_vocabulary record by a binary vocabulary_id value.
-     *
-     * @param  string $binary Binary vocabulary_id value.
-     *
-     * @return array
-     */
-    public function findByBinary($binary)
-    {
-        $binary = $this->_getPaddedBinary($binary);
-
-        return $this->find(['hash' => $binary]);
     }
 
     /**
@@ -1147,9 +1132,9 @@ class Sentence extends AppModel
             return false;
         }
 
-        $hash = murmurhash3($lang.$text);
+        $hash = $this->makeHash($lang, $text);
 
-        if ($duplicate = $this->findByBinary($hash)) {
+        if ($duplicate = $this->findByBinary($hash, 'hash')) {
             return false;
         }
 
--- a/app/models/sentence.php
+++ b/app/models/sentence.php
@@ -1157,7 +1157,7 @@ class Sentence extends AppModel
     {
         $sentence = $this->findById($id);
 
-        if ($sentence['Sentence']['hasaudio'] === 'yes') {
+        if ($sentence['Sentence']['hasaudio'] !== 'no') {
             return true;
         }
 
--- a/app/models/sentence.php
+++ b/app/models/sentence.php
@@ -743,6 +743,7 @@ class Sentence extends AppModel
             $userId,
             $translationCorrectness
         );
+
         // saving links
         if ($sentenceSaved) {
             $this->Link->add($sentenceId, $this->id, $sentenceLang, $translationLang);
--- a/app/models/sentence.php
+++ b/app/models/sentence.php
@@ -41,6 +41,7 @@
 App::import('Model', 'CurrentUser');
 App::import('Sanitize');
 App::import('Vendor', 'LanguagesLib');
+App::import('Vendor', 'murmurhash3');
 
 class Sentence extends AppModel
 {
@@ -763,16 +764,44 @@ class Sentence extends AppModel
      */
     public function saveNewSentence($text, $lang, $userId, $correctness = 0)
     {
+        $hash = murmurhash3($lang.$text);
+
+        if ($sentence = $this->findByBinary($hash)) {
+            $this->id = $sentence['Sentence']['id'];
+
+            $this->duplicate = true;
+
+            return true;
+        }
+
         $this->create();
 
+        $this->duplicate = false;
+
         $data['Sentence']['text'] = trim($text);
+        $data['Sentence']['user_id'] = $userId;
+        $data['Sentence']['correctness'] = $correctness;
+        $data['Sentence']['hash'] = $hash;
+
         if (!empty($lang)) {
             $data['Sentence']['lang'] = $lang;
         }
-        $data['Sentence']['user_id'] = $userId;
-        $data['Sentence']['correctness'] = $correctness;
 
         return $this->save($data);
+    }
+
+    /**
+     * Find a users_vocabulary record by a binary vocabulary_id value.
+     *
+     * @param  string $binary Binary vocabulary_id value.
+     *
+     * @return array
+     */
+    public function findByBinary($binary)
+    {
+        $binary = $this->_getPaddedBinary($binary);
+
+        return $this->find(['hash' => $binary]);
     }
 
     /**
--- a/app/models/sentence.php
+++ b/app/models/sentence.php
@@ -507,7 +507,7 @@ class Sentence extends AppModel
             return array_keys($results);
         }
     
-    return 1;
+        return 1;
     }
 
     /**
@@ -1128,6 +1128,55 @@ class Sentence extends AppModel
     {
         $this->id = $sentenceId;
         return $this->saveField('hasaudio', $hasaudio);
+    }
+
+    /**
+     * Edit the sentence.
+     *
+     * @param  int $id      Sentence id.
+     * @param  string $text New sentence text.
+     * @param  string $lang New sentnece lang.
+     *
+     * @return boolean
+     */
+    public function editSentence($id, $text, $lang)
+    {
+        $this->id = $id;
+
+        if ($this->hasAudio($id)) {
+            return false;
+        }
+
+        $hash = murmurhash3($lang.$text);
+
+        if ($duplicate = $this->findByBinary($hash)) {
+            return false;
+        }
+
+        $data['Sentence']['text'] = $text;
+        $data['Sentence']['hash'] = $hash;
+
+        if (!empty($lang)) {
+            $data['Sentence']['lang'] = $lang;
+        }
+        
+        return $this->save($data);
+    }
+
+    /**
+     * Return true if sentence has audio.
+     *
+     * @return boolean
+     */
+    public function hasAudio($id)
+    {
+        $sentence = $this->findById($id);
+
+        if ($sentence['Sentence']['hasaudio'] === 'yes') {
+            return true;
+        }
+
+        return false;
     }
 }
 ?>
--- a/app/models/sentence.php
+++ b/app/models/sentence.php
@@ -751,42 +751,41 @@ class Sentence extends AppModel
                                // Never mind the links.
     }
 
-
     /**
      * Add a new sentence in the database
      *
-     * @param string $text   The text of the sentence
-     * @param string $lang   The lang of the sentence
-     * @param int    $userId The id of the user who added this sentence
+     * @param string $text        The text of the sentence.
+     * @param string $lang        The lang of the sentence.
+     * @param int    $userId      The id of the user who added this sentence.
+     * @param int    $correctness Correctness level of sentence.
      *
      * @return bool
      */
     public function saveNewSentence($text, $lang, $userId, $correctness = 0)
     {
         $this->create();
+
         $data['Sentence']['text'] = trim($text);
         if (!empty($lang)) {
             $data['Sentence']['lang'] = $lang;
         }
         $data['Sentence']['user_id'] = $userId;
         $data['Sentence']['correctness'] = $correctness;
-        $sentenceSaved = $this->save($data);
 
-        return $sentenceSaved;
+        return $this->save($data);
     }
 
     /**
-     * Add a new sentence and a translation in the database
+     * Add a new sentence and a translation in the database.
      *
-     * @param string $sentenceText    The text of the sentence
-     * @param string $sentenceLang    The lang of the sentence
-     * @param string $translationText The text of the translation
-     * @param string $translationLang The lang of the translation
-     * @param int    $userId          The id of the user who added them
+     * @param string $sentenceText    The text of the sentence.
+     * @param string $sentenceLang    The lang of the sentence.
+     * @param string $translationText The text of the translation.
+     * @param string $translationLang The lang of the translation.
+     * @param int    $userId          The id of the user who added them.
      *
      * @return bool
      */
-
     public function saveNewSentenceWithTranslation(
         $sentenceText,
         $sentenceLang,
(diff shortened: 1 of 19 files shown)


trang:
@zachleigh, thanks for this PR, I think this will be a quite valuable improvement =)

I still need to review the code more in detail but there is one important thing missing: the SQL update script needs to handle as well the generation of the hash for existing sentences.

zachleigh:
I wasnt sure how to do that. How can I access the murmurhash3 function in a mysql query? I noticed that one of the former contributors Allan Simon has a Mysql version of the function here:   

http://allan-simon.github.io/blog/posts/murmur_hash_v3_in_mysql_utf8_compatible/    

Is that already in Tatoeba somewhere?

trang:
> http://allan-simon.github.io/blog/posts/murmur_hash_v3_in_mysql_utf8_compatible/
>
> Is that already in Tatoeba somewhere?

No, it's not in Tatoeba. You have to add it.

* Create a file in the `docs/database/procedures` folder where you'll re-use Allan's code. The script in this folder is used for when developers create a database from scratch.
* You'll also have to copy-paste the code into your update script, because this script will be used to update existing databases. And you have to add the query `UPDATE sentence SET hash = ...`, like in Allan's blog post.

zachleigh:
Ive put his function into docs/database/procedures, but mysql isnt finding it. 
```
ERROR 1305 (42000): FUNCTION tatoeba.murmur_hash_v3 does not exist
```
Am I missing a something?

zachleigh:
Ive put his function into docs/database/procedures, but mysql isnt finding it. 
```
ERROR 1305 (42000): FUNCTION tatoeba.murmur_hash_v3 does not exist
```
Am I missing a something?

zachleigh:
Ah, thats what I missed. I was under the impression that Cake would take care of creating the procedure in the database for me. Thanks @trang!

app/models/sentence.php
```diff
@@ -1096,5 +1114,54 @@ public function editAudio($sentenceId, $hasaudio)
...
+    /**
+     * Return true if sentence has audio.
+     *
+     * @return boolean
+     */
+    public function hasAudio($id)
+    {
+        $sentence = $this->findById($id);
+
+        if ($sentence['Sentence']['hasaudio'] === 'yes') {
```
trang: Should be `if ($sentence['Sentence']['hasaudio'] !== 'no')`

The value `yes` is not a valid value for `hasaudio`. Possible values are documented here: https://github.com/Tatoeba/tatoeba2/blob/dev/docs/database/tables/sentences.sql#L34

trang:
@zachleigh, I've reviewed more thoroughly your pull request. I've just found one mistake to fix. The rest looks good to me.

trang:
And regarding:

> When a sentence is edited and it becomes a duplicate, it is simply not added and the original sentence is returned. There is currently no message which is going to be confusing for users. Im not sure how to go about adding this message...

I'll be taking care of this when I redesign the sentences. The behavior you implemented is good enough for the time being since this case shouldn't create too much confusion since it doesn't happen very often.

zachleigh:
@trang Ill fix the hasAudio error after I take care of the murmurhash mysql stuff.    

Im still having a little trouble. I created the function in MySql but cant get it to match the key created in php. Right now, my query looks like this:
```
UPDATE sentences SET hash=CONV(murmur_hash_v3(CONCAT(lang, text),0),10,32);
```
Im trying to match this php function found here: https://github.com/Tatoeba/tatoeba2/blob/dev/app/vendors/murmurhash3.php#L55
```
function murmurhash3($key,$seed=0){
  return base_convert(murmurhash3_int($key,$seed),10,32);
}
```
Ive tried the mysql with and without the base convert but cant get the hashes to match up. Any ideas?

trang:
No idea right now. I assumed it would work simply with `UPDATE sentences SET hash=murmur_hash_v3(CONCAT(lang, text),0);`

Do you have some samples of what the PHP function returns compared to what the MySQL function returns?

zachleigh:
For the sentence 'test sentence', php is creating and inserting `174qqt3`.    The Mysql functions create the following:
```
UPDATE sentences SET hash=murmur_hash_v3(CONCAT(lang, text), 0);
// 1313696675

UPDATE sentences SET hash=CONV(murmur_hash_v3(CONCAT(lang, text), 0), 10, 32);
// 174QQT3
```
The second one matches, just the letters are all uppercase. However, its inconsistent. For the sentence 'test', php creates `15d15m` but the mysql function creates `3UQIUQA'.    

It seems like shorter sentences are more likely to not match...

zachleigh:
Another option would be to ditch Murmurhash, which also has some [vulnerabilities](http://emboss.github.io/blog/2012/12/14/breaking-murmur-hash-flooding-dos-reloaded/), and use something like sha256 which is included in the core of both php and MySql. It is possibly a little slower than Murmur, but since we are only hashing a single string at a time (for the users) it shouldnt be too much of a problem.

trang:
In our case I don't think we need to worry about the vulnerabilities, since our hash column has a unique constraint. Worst thing that can happen is that someone tries to add a completely valid sentence, that is not a duplicate, but will be detected as a duplicate because the hash collides with the hash of another sentence.

What we want is speed, and if we believe this thread http://programmers.stackexchange.com/questions/49550/which-hashing-algorithm-is-best-for-uniqueness-and-speed, Murmur hash v3 seems to provide the best performance.

If it's too complicated to get the MySQL procedure to produce the same result as the PHP function (or vice versa), we can also consider using a PHP script to update the `sentences` table, instead of doing it via a SQL script.

I don't mind either trying to use a hash that is implemented in both PHP and MySQL. I can't really evaluate how much impact this would have on performance, in the long term, compared to the murmur hash v3. We can experiment and see how it goes.

zachleigh:
@trang Figured it out. I was reading up on different hashing functions that php has and found this in the [crc32 documentation](http://php.net/manual/en/function.crc32.php):

> Warning
> 
> Because PHP's integer type is signed many crc32 checksums will result in negative integers on 32bit platforms. On 64bit installations all crc32() results will be positive integers though.
> 
> So you need to use the "%u" formatter of sprintf() or printf() to get the string representation of the unsigned crc32() checksum in decimal format.

I ran a bunch of tests and sure enough, for all the incorrect hash values, the php version of Murmur was producing a negative hash value. So ill be updating the Hashable trait and will also include a MySql script to update the hashes for vocabulary and user_vocabulary.    

That was a tough one! :weary:

trang:
Awesome, you rock! :)

I'm moving this to the next milestone because there won't be enough time to test this for this week update. Maybe the testers will find a few more things to fix, but anyway, good job ^^

trang:
One more thing missing: `docs/database/tables/sentences.sql` needs to be updated as well, with the `hash` column.

We should also add a UNIQUE constraint on the `hash` column, but this cannot be done for now. We should do this only when we have a proper workflow for handling sentences that are edited and become duplicates. 

There is still a scenario where users can still create duplicates:
1) Adding an existing sentence with the wrong language (`I like music.` as Italian).
2) Changing the language to the correct language (Italian -> English).

After experiencing this scenario, I'm thinking we should keep the possibility for users to edit a sentence into an existing one, then let Horus do its job for such sentences. So the duplicate verification should only happen on adding sentence, not editing. The edited sentences will be handled by Horus, like they are now. This will feel less buggy for users.

As for the case when the user adds a translation which turns out to be an existing sentence, I feel we need a message to warn users, otherwise it's not going to be transparent enough for them. Message can be: `This translation is a duplicate of an existing sentence. The translation itself was not added, only the translation link was added.`

Before you make the changes though, I'd like to get some feedback from the community. Maybe someone will suggest something better.

zachleigh:
Ok. I added the hash column to sentences.sql. Are you going to ask users about the issue on the wall?

loolmeh:
> As for the case when the user adds a translation which turns out to be an existing sentence, I feel we need a message to warn users, otherwise it's not going to be transparent enough for them. Message can be: This translation is a duplicate of an existing sentence. The translation itself was not added, only the translation link was added.

Yes. This is what I meant by asking for intent on that other ticket. I think I can spend sometime making the dedup script into a daemon with a simple http api, then we can cancel the cron job and let the php side display dialogs for editing and translating cases then decide to run the procedure. For the language setting case, I suppose no dialog would be needed.

trang:
@zachleigh yes I will ask users on the Wall. I deployed your changes on the dev website and will simply have users try out your implementation and give feedback.

I noticed a couple more things to fix in the update script:
* The SQL code to create the murmur hash procedure needs to be copy pasted into the update script. Someone who wants to update their database will not know that they have first to run the `murmur_hash_v3.sql` script before they run the `2016-09-11.sql` script.
* The `2016-09-11.sql` script doesn't really need to update the the vocabulary tables, because there is already a hash for all the rows in those tables.

@loolmeh yep it will definitely be more efficient if we could call Horus on demand rather than making him patrol every 15 minutes. I think we will still need a dialog for the case language modification. I think for modification in general, we'll need to sit down and think of all the possibilities and reasons why a user ends up creating a duplicate by modifying a sentence. In which case can we trigger a merge right away and in which case should we request intervention of a corpus maintainer. And we of course have to think of how to make it clear to the user what is happening behind the scenes.

zachleigh:
@trang I changed the php hash so we have to update the current vocabulary hashes to make correct. The old php version of the hash created negative hashes for some strings.     

https://github.com/Tatoeba/tatoeba2/pull/1311/commits/39f87b695c3665f3c75724c7f9e56beab85f0765     


Ill add a call to the procedure that creates the hash function in the update script.

alanfgh:
Your suggested wording was:

"This translation is a duplicate of an existing sentence. The translation itself was not added, only the translation link was added."

My suggestion of an alternative:

"This translation was a duplicate of an existing sentence, which is now linked to the main sentence."

or more explicitly:

"This translation was not added, since it is a duplicate of an existing sentence, which is now linked to the main sentence."

If you want to stay closer to the original, it should at least avoid repetition of the word "added":

"This translation is a duplicate of an existing sentence. The translation link, but not the translation itself, has been added."

alanfgh:
When the translation link to the existing sentence is added, the block below the main sentence will be updated to show the newly linked translation, right?

ckjpn:
An incorrect sentence cannot be edited to match a correct, existing sentence.

https://dev.tatoeba.org/eng/sentences/show/3843231

screen shot 2016-10-01 at 07 42 44

ckjpn:
https://dev.tatoeba.org/eng/sentences/show/252245
and 
https://dev.tatoeba.org/eng/sentences/show/3843232

These 2 are duplicates.

* I added "I like dogs."[#3843232] as Japanese, then changed the flag to English.
* Since changing flags doesn't happen so much, perhaps it's just a case where to leave it up to the duplicate-merging script to take care of.

screen shot 2016-10-01 at 07 59 36

ckjpn:
It's possible to submit duplicates right away.   Perhaps this is to be expected if it takes a while for the search engine to index new sentences.

screen shot 2016-10-01 at 08 10 23

I tested the "import sentences" function, too.  It worked well.
The 2 duplicate sentences I purposely added into the file were not imported, but the other 2 new sentences were.

ckjpn:
I tested importing 2 languages.

It worked great, and exactly what I was hoping for.

It properly imported the new English translations and attached them to the existing Japanese.
It also properly linked 2 existing sentences.

すごい！
https://dev.tatoeba.org/eng/sentences/show/214733
screen shot 2016-10-01 at 08 23 22

続けて。
https://dev.tatoeba.org/eng/sentences/show/2555964
screen shot 2016-10-01 at 08 25 15


== The imported data ===

すごい！	Wow!
すごい！	That's great!
すごい！	Wonderful!
すごい！	Cool!
すごい！	Example Translation1
すごい！	Example Translation2
続けて。	Please continue.
続けて。	Go ahead.
続けて。	Example Translation3

ckjpn:
https://dev.tatoeba.org/eng/sentences/show/3843243

It's possible to add a duplicate as a translation and the user doesn't get any alert that it was a duplicate.

However, it does properly link to the duplicate instead of adding the same sentence with a new number.

### After Adding
screen shot 2016-10-01 at 10 02 29

### After Reloading the Page
screen shot 2016-10-01 at 10 02 52

Perhaps some kind of alert would be useful.  I'm not sure.

zachleigh:
> An incorrect sentence cannot be edited to match a correct, existing sentence.

Yeah, this is a big problem. There are a lot of sentences that are essentially similar, but we can not catch them without a mindbogglingly complex algorithm. Maybe there should be a way to report a duplicate as well? Not sure who would be in charge of dealing with the list of reported duplicates though...

> It's possible to submit duplicates right away. Perhaps this is to be expected if it takes a while for the search engine to index new sentences.

Hmm... That shouldnt happen. It doesnt rely on search, it relies on a hash column in the database. How did you create the duplicates?

> It's possible to add a duplicate as a translation and the user doesn't get any alert that it was a duplicate.

Yes. I wasnt sure where to put the error message or what it should say, so this has not been done yet. I agree that it is needed though. Without a notification of some kind, it makes it feel broken.

ckjpn:
> How did you create the duplicates?

That's easy, just go to 

https://dev.tatoeba.org/eng/sentences/add

and add the same new sentence over and over again

For example, "This is a duplicate."

### Screenshots

screen shot 2016-10-01 at 18 10 38

screen shot 2016-10-01 at 18 11 07

zachleigh:
Weird. Looks like this on my local machine:
screenshot3

trang:
Okay there was a cache issue. The cache is usually turned off on the dev website, but I enabled it again lately to test a pull request.

The database schema is cached, so the newly created `hash` column of the `sentences` table was not detected, and Tatoeba acted as if that column didn't exist, because it was reading from the cache. So newly created sentences had an empty hash.

I disabled the cache again on the dev website, so now it should be working. It works for me:
http://prntscr.com/cow56k

trang:
@alanfgh, the warning message for translations would be like this:

screenshot_67

But we won't implement it for now because I figured there are actually several cases possible:

1) The user is copy-pasting the main sentence and is adding it as a translation. Not likely to happy, but could happen.
2) The user is adding a translation that is already a translation of the main sentence. It's possible that a user doesn't notice there is already the exact translation they want to add.
3) The user adds a translation which is an existing sentence but is not yet linked to the main sentence.

Ideally we should handle each case with a different message... I will create a different issue for this.


@zachleigh

* Regarding the issue of editing sentences that become then duplicates: allow the user to edit. We'll let Horus handle those sentences, this way it doesn't feel like a bug or a regression compared to the current behavior.

* Regarding the warning message for translations that are duplicates, we'll handle it in another issue because there are those additional cases to handle. Currently, if a user adds a translations that is a duplicate, they don't get any warning when Horus cleans up their duplicate. So we're not making things worse by not displaying any message. So let's consider this out of scope. I would prefer anyway to have the warning message implemented after the redesign of the sentences is done so that you don't implement stuff into a code that will be obsolete in a couple of months.

app/webroot/css/sentences/add.css
```diff
@@ -36,3 +36,13 @@
     margin: 40px 0;
 }
 
+.sentences_duplicate {
+    height: 20px;
+}
+
+.sentences_duplicate_icon {
+    display: inline-block;
+    width: 20px;
```
trang: You should add the `height: 20px;` here rather than in `.sentences_duplicate`.

trang:
@zachleigh 

> @trang I changed the php hash so we have to update the current vocabulary hashes to make correct. The old php version of the hash created negative hashes for some strings. 

You're right, I forgot about that ^^

trang:
I merged this by mistake -_-'

I canceled the merge on the dev branch, but you may have to re-open a pull request as it doesn't look like I can re-open this.

ckjpn:
Re: https://cloud.githubusercontent.com/assets/221850/19021407/bf5c8b9a-88c0-11e6-9c80-249e853d15ed.png

Perhaps a simpler message could be something shorted and like this.  Maybe Alan can suggest a better wording.

An existing duplicate has been linked instead of adding yours as a new sentence.

I imagine we can come up with something better, but more concise like this.

Issue #829, opened 2015-10-14
Instead of generating a page with sentences in all languages with a certain tag when clicking that tag on a sentence's page, have the link generate the page limited to sentences in that language.


ckjpn:
Instead of generating a page with sentences in all languages with a certain tag when clicking that tag on a sentence's page, have the link generate the page limited to sentences in that language.

For example, on this page...
https://tatoeba.org/eng/sentences/show/2172739

... instead of having the link for the "dialogue" tag to be this ...

https://tatoeba.org/eng/tags/show_sentences_with_tag/1452

... have it this way.

https://tatoeba.org/eng/tags/show_sentences_with_tag/1452/eng

I think it would be more useful this way.
Otherwise, most visitors are likely to load the former URL, then have to choose "English" and load the second URL.
This not only means it takes the user longer to get what he's after, but puts an extra "wasted" demand on the database

I labelled this "effort:low", since I think it is likely to be very easy to do.  However, I may be wrong.

ckjpn:
This is perhaps not a big problem and may not need to be corrected.

However, when first adding a tag, if you click the tag, it does not go to the page of tags limited to that language, but goes to the page showing all languages.

I just thought you would like to know.

![screen shot 2016-11-17 at 01 54 58](https://cloud.githubusercontent.com/assets/7934273/20356979/955e6ee0-ac69-11e6-9201-f6122f2f2f8e.png)

PR #1320, opened 2016-09-27
Click tag on sentence page shows sentences w/same lang

alexandresmirnov:
This pull request addresses issue #829.

When clicking on a tag in the tag cloud to the right of a sentence (in the single sentence view), that tag now leads to a page with sentences in the same language as the original language.

Modified Tags helper, now has optional "$sentenceLang" parameter that
allows the tag link to point to generated page where all sentences are of
the same lang that the original sentence (where tag was clicked) was.

Modified sentences/show.ctp to pass $sentenceLang to
$tags->displayTagsModule

### DIFF
--- a/app/views/helpers/tags.php
+++ b/app/views/helpers/tags.php
@@ -53,7 +53,7 @@ class TagsHelper extends AppHelper
      *
      * @return void
      */
-    public function displayTagsModule($tagsArray, $sentenceId = null)
+    public function displayTagsModule($tagsArray, $sentenceId = null, $sentenceLang = null)
     {
         $this->Javascript->link(JS_PATH . 'autocompletion.js', false);
         ?>
@@ -72,7 +72,7 @@ class TagsHelper extends AppHelper
                     $date = $tagArray['TagsSentences']['added_time'];
 
                     $this->displayTag(
-                        $tagName, $tagId, $sentenceId, $userId, $username, $date
+                        $tagName, $tagId, $sentenceId, $userId, $username, $date, $sentenceLang
                     );
 
                 }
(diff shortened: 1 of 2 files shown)


app/views/helpers/tags.php
```diff
@@ -102,14 +102,14 @@ public function displayTagsModule($tagsArray, $sentenceId = null)
      * @param null $date
      */
     public function displayTag(
-        $tagName, $tagId, $sentenceId, $userId, $username = null, $date = null
+        $tagName, $tagId, $sentenceId, $userId, $sentenceLang = null, $username = null, $date = null
```
trang: When you modify the signature of a function to add a new parameter, it is generally better to add the new parameter at the very end, especially if it is an optional parameter. This way there is less risk to break something. If you choose to add it somewhere in the middle, you have to make sure to fix all the usages of the function.

In our case, you will need to fix the params in `app/views/tags/add_tag_post.ctp`.
Cf. https://github.com/Tatoeba/tatoeba2/search?utf8=%E2%9C%93&q=displayTag

trang:
Thanks for your PR @alexandresmirnov =)

Besides of my comment above, there's one more thing that would be nice: if you could do a `rebase -i` to squash your 2 last commits into one commit (https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase-i).

Also, if it happens to you again that right after you made a commit, you realize you forgot something in the last commit, you can use `git commit --amend` (https://www.atlassian.com/git/tutorials/rewriting-history/git-commit--amend).

alexandresmirnov:
Thank you for the suggestions!

Squashed the commits and changed the order of the parameters.

alexandresmirnov:
Thank you for your help!

Squashed the commits and changed the order of the parameters.

trang:
It seems like the commit with the indentation fix disappeared. Maybe you didn't push it?

alexandresmirnov:
No, I just rebased and squashed all of the commits (including the indentation issues) into one.  This commit contains all the changes I implemented, up to date :)

trang:
Oh I see, I thought that you were fixing the indentation for the whole file, in which case you should make it a separate commit. But you were just fixing it for your own changes, in which case it makes more sense indeed that you squash everything into one commit.

PR #1321, opened 2016-09-27
Cache the audio stats query

jiru:
While investigating on the server slowdown, I found that the audio stats query is taking up to 11 seconds (averaging on 4~5 seconds). This query is triggered by https://tatoeba.org/stats/sentences_by_language and https://tatoeba.org/sentences/with_audio.

This probably won’t solve the whole slowdown problem, but it will surely improve performance.


trang:
Thanks @jiru =)

PR #1327, opened 2016-10-02
Detect duplicate sentences when adding/editing sentences

trang:
Re-opening a request as a continuation of #1311 because I messed up and merged the branch by mistake. Sorry!

This pull request address issue #577.

### DIFF
--- a/.Trash-1000/files/.directory
+++ b/.Trash-1000/files/.directory
@@ -1,0 +1,4 @@
+[Dolphin]
+PreviewsShown=true
+Timestamp=2016,9,11,17,20,50
+Version=3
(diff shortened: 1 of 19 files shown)


zachleigh:
@trang Added the murmur_hash_v3 procedure to the database update script and fixed the css issue.    

So just to confirm, if a user tries to edit a sentence and that sentence then becomes a duplicate, we will allow the edit and let Horus deal with it later?   

Was there anything else that I missed?

trang:
>So just to confirm, if a user tries to edit a sentence and that sentence then becomes a duplicate, we will allow the edit and let Horus deal with it later? 

Yes, we will allow the edit.

> Was there anything else that I missed?

No, that's it.

zachleigh:
Users can now edit sentences even if they are duplicates.

Issue #1323, opened 2016-09-28
Unnecessary JOIN for sentence count query in /sentences/show_all_in


trang:
For this page: http://dev.tatoeba.org/eng/sentences/show_all_in/eng/none/none

The following query is executed:
```
SELECT COUNT(*) AS `count` FROM `sentences` AS `Sentence` 
LEFT JOIN `users` AS `User` ON (`Sentence`.`user_id` = `User`.`id`) 
WHERE `Sentence`.`lang` = 'eng' 

// From my test on the dev website: took 5755 ms
```

The query should simply be:
```
SELECT COUNT(*) AS `count` FROM `sentences` AS `Sentence` 
WHERE `Sentence`.`lang` = 'eng'
```

PR #1329, opened 2016-10-04
Remove unnecessary join from sentence count query

zachleigh:
This pull request address issue #1323.    

I overrode the standard count used for pagination by creating a paginateCount method on the sentence model. I simply hardcoded the sql query everything, at least for now. Right now, Im covering the following sentence pagination possibilities:
  - Paginate sentences by `lang`
  - Paginate sentences by `hasaudio`
  - Paginate sentences by `user_id`

Am I missing anything?     

Also, I dont have the audio database set up on my dev machine, so Im not 100% sure if that one works or not.    

Before merging, Id like to refactor this. I want to make sure Im covering all the possibilities before doing the refactor though.

### DIFF
--- a/app/models/sentence.php
+++ b/app/models/sentence.php
@@ -578,28 +578,13 @@ class Sentence extends AppModel
         $recursive = 0,
         $extra = array()
     ) {
-        if (isset($conditions['hasaudio'])) {
-            $sql = "SELECT COUNT(*) AS `count` FROM `sentences` AS `Sentence`
-                    WHERE `Sentence`.`hasaudio` != 'no'";
-        } else if (isset($conditions['Sentence.lang'])) {
-            $lang = $conditions['Sentence.lang'];
-
-            $lang = Sanitize::paranoid($lang);
-
-            $sql = "SELECT COUNT(*) AS `count` FROM `sentences` AS `Sentence`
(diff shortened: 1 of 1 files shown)


trang:
On the `show_all_in` page, the conditions `lang` and `hasaudio` can be combined. The way your conditions are defined doesn't handle such case.

I found a place where the `paginateCount` is overriden for optimization purpose, you can probably get inspiration from there: https://github.com/Tatoeba/tatoeba2/blob/dev/app/models/sentence_comment.php#L201

With this method you won't need to try and find out what are all the possible cases.

And probably we can simply force `contain` to be empty. I can't think of a situation where we'll ever want the count to make a JOIN on the `sentences` table. 

The only case where it currently happens is if the user choose to show all sentences in langA not translated into langB, but this feature needs to be removed from the show_all_in page. And the count for this query is handled in the `SentenceNotTranslatedInto` model, not in the `Sentence` model.

trang:
It turns out it won't be that simple :( 

I deployed this PR on the prod instance after merging and users reported a problem with the number of results in the search (#1334).

By overriding the `paginateCount()`, we also override the count performed by the SphinxBehavior.

So I've reverted this. We're gonna have to dig deeper...

PR #1335, opened 2016-10-09
Override paginator count only when conditions are present.

zachleigh:
This pull request addresses issue #1323.    

I played around with this for a while and found that when Sphinx uses paginateCount, $conditions is an empty array. So if $conditions is empty, we will use the default behavior as defined here: https://github.com/Tatoeba/tatoeba2/blob/dev/cake/libs/controller/controller.php#L1205    

I havent found any problems with it, but maybe we should deploy it on the dev site before putting it on production to make sure it doesnt break Sphinx again.

### DIFF
--- a/app/models/sentence.php
+++ b/app/models/sentence.php
@@ -565,6 +565,40 @@ class Sentence extends AppModel
     }
 
     /**
+     * Override standard paginateCount method to eliminate unnecessary joins.
+     * If $conditions is empty, as in Sphinx search, return default behavior.
+     *
+     * @param  array   $conditions
+     * @param  integer $recursive
+     * @param  array   $extra
+     *
+     * @return integer
+     */
+    public function paginateCount(
+        $conditions = null,
+        $recursive = 0,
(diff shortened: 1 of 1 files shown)


trang:
I experimented a bit further upon your code and this would be simpler:

```
    public function paginateCount(
        $conditions = null,
        $recursive = 0,
        $extra = array()
    ) {
        $parameters = compact('conditions');
        $extra['contain'] = [];

        return $this->find('count', array_merge($parameters, $extra));
    }
```

Basically, the `contain` data is carried in the `$extra` param, and so is the `sphinx` data. So we need to make sure that the `find('count')` takes into account the `$extra` data (for sphinx), but we override `contain` to be always empty.

And we don't care about `$recursive`. If anything we should force it to `-1` but that's redundant with `contain = []`.

You can try this code on your side, then push the changes if you see no other issue. I will then deploy it dev website and ask users to test.

zachleigh:
Seems to work great. Hopefully we got it this time :)

Issue #1336, opened 2016-10-11
Wrong sentence is linked instead of adding a new translation


Guybrush88:
As reported by odexed on the wall:

> Bug.
> When I'm trying to translate #3307344 into Russian, my translation disappears but #5483277 appears as a link. I belive it's the only sentence with this problem so far.

https://tatoeba.org/ita/wall/show_message/27428#message_27428

Guybrush88:
I just reproduced this with another sentence. As I wrote on the wall:

I wanted to add a translation to #5448863 (a translation that isn't present in the database, I checked), but #2990793 gets linked instead

trang:
1. What is the text of the sentence you tried to add.
2. Were you on "Auto-detect" for the language?

trang:
I tested with odexed's sentence. 

It turns out that there's a collision between:
* `rus` + `Я начинаю понимать, почему тебе так нравится эта игра.`
* `ell` + `Ποιο είναι το καλύτερο πανεπιστήμιο στη Γερμανία;`

They are both converted into: `3tsnqit`.

```
mysql> select lang, text, hash, LOWER(CONV(murmur_hash_v3(CONCAT(lang, text), 0), 10, 32)) as murmurhash from sentences where id in (3843344, 3843345);
+------+----------------------------------------------------------------------------------------------------+------------------+------------+
| lang | text                                                                                               | hash             | murmurhash |
+------+----------------------------------------------------------------------------------------------------+------------------+------------+
| rus  | Я начинаю понимать, почему тебе так нравится эта игра.                                             | 3tsnqit          | 3tsnqit    |
| ell  | Ποιο είναι το καλύτερο πανεπιστήμιο στη Γερμανία;                                                  | 3tsnqit          | 3tsnqit    |
+------+----------------------------------------------------------------------------------------------------+------------------+------------+
```

Since Guybrush also ran into the problem, it seems collision can happen fairly often, from human generated sentences.

-----

I think the best approach here is that when duplicate candidates are detected, we make an additional check on the text of the sentences before determining if it's a real duplicate. So we use the hash to quickly find duplicate candidates, but we always need in the end to check against the actual text to decide if there's really a duplicate.

The only consequence I can think of is that we can never put a UNIQUE constraint on the `hash` column, but I don't think it is a problem.

We could also look for another hash function that ensures no collision, but that means time investing which hash we could use, and it would require to shut down again Tatoeba to update the hash of the sentences. It wouldn't be worth it.

[...]

PR #1337, opened 2016-10-12
Add additional check for duplicate sentences.

zachleigh:
This pull request address issue #1336.   

I added checks for both the text and language of the new sentence. If these two conditions are met, it will be marked as a true duplicate.

### DIFF
--- a/app/models/behaviors/hashable.php
+++ b/app/models/behaviors/hashable.php
@@ -57,6 +57,24 @@ class HashableBehavior extends ModelBehavior
     }
 
     /**
+     * Find all records by binary hash value.
+     *
+     * @param  object $model  Model
+     * @param  string $binary Binary id value.
+     * @param  string $column Column on table to search.
+     *
+     * @return array
+     */
+    public function findAllByBinary($model, $binary, $column)
+    {
+        $binary = $this->padHashBinary($model, $binary);
+
+        $method = 'findAllBy'.$column;
+
+        return $model->{$method}($binary);
+    }
+
+    /**
      * Convert a binary id to a padded binary id.
      *
      * @param  object $model  Model.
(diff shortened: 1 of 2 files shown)


ckjpn:
I don't know if this should be added as a new issue, but I wonder how duplicates of "orphan" sentences are being handled.

1.
Ideally, if a member contributes a duplicate of an existing sentence without an owner, the person contributing the duplicate should automatically become the owner of the existing sentence.

2.
Another option would be to automatically delete the old sentence and link all the translations to the newly-contributed sentence, sort of like what the duplicate-merging script does.

Personally, I think suggestion 1 is probably better.  However, I can understand why others might prefer suggestion 2.

trang:
@zachleigh, one more thing to fix.

[Odexed reported an issue](https://tatoeba.org/eng/wall/index#message_27452) with the sentence `Постарайся успокоиться.`.

I tried to add this sentence several times on the dev website (this change is deployed on the dev website), and strangely it was not detected as a duplicate.

While debugging locally `_confirmDuplicate()`, after I added several times the sentence `Постарайся успокоиться.`, I found out that:

* $text: `Постарайся успокоиться.`
* $lang: `rus`
* $sentence:
```
Array
(
    [Sentence] => Array
        (
            [id] => 540152
            [lang] => epo
            [text] => Ŝia deziro iĝi kuracistino spronis ŝin.
            [correctness] => 0
            [user_id] => 3138
            [created] => 2010-09-30 18:35:48
            [modified] => 2010-09-30 18:35:48
            [dico_id] => 
            [hasaudio] => no
            [script] => 
            [hash] => 3ms1niv
        )

)
```

Since several senences can have the same hash. We have to retrieve all the sentences with the corresponding hash and loop through these sentences.

trang:
@ckjpn 

> I don't know if this should be added as a new issue, but I wonder how duplicates of "orphan" sentences are being handled.

This is another issue.

zachleigh:
@trang Ill finish this one up tomorrow.

zachleigh:
@trang It now gets all sentences that match the hash and loops through them looking for a exact match.

Issue #1324, opened 2016-09-28
Unnecessary queries for non authenticated users on pages where sentences are displayed


trang:
For this page: http://dev.tatoeba.org/eng/sentences/show_all_in/eng/none/none

The following query is executed:
```
SELECT `SentencesList`.`id`, `SentencesSentencesList`.`id`, 
  `SentencesSentencesList`.`sentences_list_id`, 
  `SentencesSentencesList`.`sentence_id`, 
  `SentencesSentencesList`.`created` 
FROM `sentences_lists` AS `SentencesList` 
JOIN `sentences_sentences_lists` AS `SentencesSentencesList` 
ON (`SentencesSentencesList`.`sentence_id` IN (
  3843196, 3843195, 3843194, 3843193, 3843192,
  3843191, 3843190, 3843189, 3843188, 3843187) 
AND `SentencesSentencesList`.`sentences_list_id` = `SentencesList`.`id`) 

// From my test on the dev website: took 1781 ms
```

This query is executed on every page that has sentences with the sentence menu (pages like the sentences lists, search results, tags...), not just on /sentences/show_all_in.

This query is unnecessary when the user is not logged in. It is only used to generate the list for the "Add to list" option in the sentence menu, for registered members to add sentences to lists.

Similarly, there are queries related to the favorites which are not needed either. But they don't consume as much as the query for the sentences lists.

trang:
There were missing indexes for `sentences_list_id` and `sentence_id` in `sentences_sentences_lists` table. That's why the query took so much time.

Issue #888, opened 2015-10-28
Southern Altai (alt)


RyckRichards:
•Language name: Southern Altai
•Language info: https://en.wikipedia.org/wiki/Altai_language
•ISO 639-3 code: alt
•Public List: https://tatoeba.org/eng/sentences_lists/show/4856
•Icon: https://drive.google.com/open?id=0B6VpA-j8A4f9djVjSHNvMHNIWEdFdElRYmNFT0ZicS1pb1BB
•Requested by: https://tatoeba.org/eng/user/profile/cueyayotl

RyckRichards:
This language has been added successfully. Closing this ticket.

http://prntscr.com/8yxdby

PR #1339, opened 2016-10-13
Redesign sentences

trang:
This pull request is part of #1180. It also addresses #1324.

The sentences have been redesigned for non-authenticated users.

* The icons to translate, add to favorite, add to list have been removed. For non-authenticated users, they just redirect to a page to log in. Although they do give a hint to users what are the features of Tatoeba, I think they mostly pollute the UI which is why they were removed.
* There are now a explicit labels for the sentence (`Sentence {number} - belongs to {user}`), the translations (`Translations`) and the indirect translations (`Translations of translations`). Hopefully users will understand better why some translations are in grey.
* Users no longer navigate to the sentence by clicking on the `#` or arrow icons, but on the info icon that is at the end of each sentence/translation.
* An arrow icon has been added in front of each translation in order to help distinguish between direct and indirect translations when there are a lot of translations. Direct translations have a black arrow. Indirect translations have a grey arrow.
* The text of the indirect translations are kept in black for better readability. There is now a label that says explicitly "Translations of translations", so there is less need to have it grey.
* Unapproved sentences are displayed the same as before (grey + warning icon), with a tooltip that says `This sentence is not reliable.` when hovering over the sentence. I'm not very satisfied with this tooltip text, but I'd like to find something else than "unapproved".
* The language icon is kept on the left side. We cannot have it on the right side because the language icon is a pillar in the structure of the sentences. It helps users find more easily the translations in the lanugages they are more specifically interested. And for sentences that have several translations in a same language, it helps users identify more quickly which blocks of translations are in which language.
* The sentence in the comments will be re-adapted later on to be consistent with the new sentence design. It's not in the scope of this pull request.

Authenticated users still have by default the old design, but can activate the new one from their settings if they wish to try to it out. Some pages will still keep the old design no matter what the setting is:

* The random sentence on the homepage for authenticated users, because the feature to load another random sentence doesn't work yet with the new design. Non-authenticated users have the new design of the random sentence, but do not have the reload option.
* The sentence details page, because the new sentence design does not have the sentence menu (translate, edit, etc), and users should still have access to a page where they can access every sentence feature without switching back and forth between old and new design. Also, the audio icon was not re-added (yet?) in the new design, so the sentence details would allow to listen to the audio of a specific sentence.

The new design is notably faster to render than the old one. For instance someone who has their settings set to 100 sentences per page will have some noticeable lag while searching sentences (the page will stay grey for a few seconds, or even freeze). The new design doesn't have this lag. So for users who mostly browse Tatoeba and use the sentence menu only punctually, it can be more advantageous to switch to the new design.

The features of the sentence menu will be introduced into the new design progressively. This will probably take until end of December. Once every feature has been implemented in the new design, the old design will be removed for everyone.

### DIFF
--- a/app/controllers/activities_controller.php
+++ b/app/controllers/activities_controller.php
@@ -185,11 +185,16 @@ class ActivitiesController extends AppController
             $conditions['Sentence.lang'] = $lang;
         }
 
+        if (CurrentUser::isMember()) {
+            $contain = $this->Sentence->contain();
+        } else {
+            $contain = $this->Sentence->minimalContain();
+        }
         $this->paginate = array(
             'Sentence' => array(
                 'fields' => $this->Sentence->fields(),
                 'conditions' => $conditions,
-                'contain' => $this->Sentence->contain(),
+                'contain' => $contain,
                 'limit' => CurrentUser::getSetting('sentences_per_page'),
                 'order' => 'created DESC'
             )
--- a/app/controllers/sentences_controller.php
+++ b/app/controllers/sentences_controller.php
@@ -910,10 +910,16 @@ class SentencesController extends AppController
         $this->addLastUsedLang($translationLang);
         $this->addLastUsedLang($notTranslatedInto);
 
+        if (CurrentUser::isMember()) {
+            $contain = $this->Sentence->contain();
+        } else {
+            $contain = $this->Sentence->minimalContain();
+        }
+
         $pagination = array(
             'Sentence' => array(
                 'fields' => $this->Sentence->fields(),
-                'contain' => $this->Sentence->contain(),
+                'contain' => $contain,
                 'conditions' => array(
                     'Sentence.lang' => $lang,
                 ),
@@ -931,7 +937,6 @@ class SentencesController extends AppController
 
 
         if (!empty($notTranslatedInto) && $notTranslatedInto != 'none') {
-
             $model = 'SentenceNotTranslatedInto';
             $pagination = array(
                 'SentenceNotTranslatedInto' => array(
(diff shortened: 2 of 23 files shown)


ckjpn:
Perhaps also changing the indirect translation’s text to gray might be a good idea -- something light enough that it's obviously not the same as direct translations, but still dark enough to be easy to read.
screen shot 2016-10-14 at 07 09 16

(perhaps between #888 ~ #aaa)

ckjpn:
> Once every feature has been implemented in the new design, the old design will be removed for everyone.

Perhaps, at least temporarily, at that time you could make the old way an option, just in case some members would still prefer the old way, and/or just in case the new way doesn't work exactly like expected.

I often add sentences to lists from the "browse" page, and I also find it very useful to see which sentences I have rated OK and can also add ratings while browsing.  Usually I only jump to the sentence's detailed page if I need to add a tag.

If you could also make it possible to add tags from the "browse" page it would save me a lot of time.  However, that might add too much clutter for most people.  Maybe you could make it an option in the settings.

Perhaps in the settings, you could make it an option to only use the "cut-down version" that you are currently using.

Maybe, you could also only use this "cut-down / light" version for non-logged in members, since they are only browsing and don't need the full set of "tools".  This might help the pages load faster for those who don't use the tools.  If a high percentage of page views were using a "light" version, perhaps it would also make the website faster for everybody else, too.

trang:
> Perhaps also changing the indirect translation’s text to gray might be a good idea

I'll keep it with the same color as the translations for now. I'll first have to see if the label `Translations of translations` and the color of the arrow are enough or not.

> Perhaps, at least temporarily, at that time you could make the old way an option, just in case some members would still prefer the old way, and/or just in case the new way doesn't work exactly like expected.

There will most likely be one or two weeks where users will still have by default the old UI even if the new UI is completely ready. But once we've switched, there will be no option to go back to the old sentence. If something doesn't work as expected by then, it will be fixed in the new UI.

> Maybe, you could also only use this "cut-down / light" version for non-logged in members.

As I've said in the description, the sentences have been redesigned for **non-authenticated users** (= non-logged in). It was only towards the end that I implemented the option for logged in users to optionally use the new design, if they wish to. But the new design is aimed at first for non-logged in users, and is part of the optimization mentioned in #1324.

Issue #1088, opened 2016-04-04
make Latin sentence search macron- and breve-insensitive


alanfgh:
alexmarcelo, on the Wall ( https://tatoeba.org/eng/wall/show_message/26122#message_26122 ):

> Is there a way to make Tatoeba treat breve-marked vowels (ă, ĕ, ĭ, ŏ, ŭ) as unmarked vowels for indexation purposes? I'd like to use them to indicate accent. The problem is that "edere" and "edĕre" are treated as different words, and one wouldn't be able to find "edĕre" by typing "edere" in the search field.

I told him that this was certainly possible, since I had made a similar change for Hebrew, but asked him to make sure that other Latin contributors/searchers would want that. His reply:

> I'll try to contact some users who used to write in Latin. Currently, CarpeLanam and I are the only ones translating into Latin. It would be great if we could standardize all Latin sentences, but this is nearly impossible. Out of the 15950 sentences we have in this language, 11745 are owned by me. Even if some users decide not to use diacritics, I think it's worth programming Tatoeba to take both breves and macrons ((ă, ĕ, ĭ, ŏ, ŭ, y̆); (ā, ē, ī, ō, ū, ȳ)), cause some people do use them by default.

As I wrote on the Wall, it's possible to make such a change globally or just for specific languages. I'm wondering whether there are any languages for which we wouldn't want breves and macrons to be ignored during searches.

alanfgh:
This is what I found:

(1) There is only 1 existing sentence that contains "combining macron" (U+0304), and it's not Latin.
(2) There are only 5 existing sentences that contain "combining breve" (U+0306), none of them Latin.
(3) There are some existing sentences (in Latin and other languages) that used precomposed characters with macron and breve.
(4) Combining macron and combining breve are in the "combining diacritical mark" range that we accept, and do not ignore, for all languages.
(5) Here are the relevant Unicode values for the composed characters with macron and breve:
'LATIN CAPITAL LETTER A WITH BREVE' (U+0102)
'LATIN SMALL LETTER A WITH BREVE' (U+0103)
'LATIN CAPITAL LETTER E WITH BREVE' (U+0114)
'LATIN SMALL LETTER E WITH BREVE' (U+0115)
'LATIN CAPITAL LETTER I WITH BREVE'  (U+012C)
'LATIN SMALL LETTER I WITH BREVE' (U+012D)
'LATIN CAPITAL LETTER O WITH BREVE' (U+014E)
[...]

PR #1355, opened 2016-11-03
Change Sphinx search conf for Latin

trang:
This pull requests addresses issue #1088.

### DIFF
--- a/app/vendors/shells/sphinx_conf.php
+++ b/app/vendors/shells/sphinx_conf.php
@@ -220,6 +220,67 @@ class SphinxConfShell extends Shell {
             )
         );
 
+        $this->indexExtraOptions['lat'] =
+            "
+        charset_table = ".implode(', ', array_merge(
+            array(
+                'A..H->a..h', 'a..h',
+                'I->i', 'J->i', 'j->i', 'i',
+                'K..T->k..t', 'k..t',
+                'U->v', 'u->v', 'V->v', 'W->v', 'w->v', 'v',
+                'X..Z->x..z', 'x..z',
+                'U+100->a',       # cap a + macron -> small a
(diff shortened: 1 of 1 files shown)


app/vendors/shells/sphinx_conf.php
```diff
@@ -220,6 +220,67 @@ public function __construct(&$dispatch) {
...
+                'U+115->e',
+                'U+C9->e',
+                'U+E9->e',
+                'U+12A->i',       # same pattern as for a
+                'U+12B->i',
+                'U+12C->i',
+                'U+12D->i',
+                'U+CD->i',
+                'U+ED->i',
+                'U+14C->o',       # same pattern as for a
```
jiru: s/for a/for o/

app/vendors/shells/sphinx_conf.php
```diff
@@ -220,6 +220,67 @@ public function __construct(&$dispatch) {
...
+                'U+103->a',       # small a + breve -> small a
+                'U+C1->a',        # cap a + acute -> small a
+                'U+E1->a',        # small a + acute -> small a
+                'U+112->e',       # same pattern as for a
+                'U+113->e',
+                'U+114->e',
+                'U+115->e',
+                'U+C9->e',
+                'U+E9->e',
+                'U+12A->i',       # same pattern as for a
```
alanfgh: What I meant here is that we follow the same pattern that we followed for "a", except that here we're following it for "e", for "i", and so on. Thus, the comment on line 239 would say "# cap e + macron -> small e", the comment on line 244 would say "# cap i + macron -> small i", and so on. Likewise, the comment on line 240 would say "# small e + macron -> small e", the comment on line 245 would say "# small i + macron -> small i", and so on.

Issue #1296, opened 2016-08-31
Search problem from Turkish to English


Guybrush88:
As reported by DostKaplan on the wall:

> [Bug] Search from Turkish to English...
> 
> "iyi olur" returns 948 results.
> "iyi Olur" returns 948 results. (So clearly upper or lower case doesn't matter.)
> 
> But...
> 
> "İyi olur" (note the capital İ Turkish character) returns 0 results:
> 
> No results found for: "İyi olur"

https://tatoeba.org/ita/wall/show_message/27068#message_27068

alanfgh:
This can be fixed via the index. In that Wall thread, I asked Turkish speakers to specify all sets of characters that should be treated identically for the purpose of search.

alanfgh:
See new Wall thread: 

https://tatoeba.org/eng/wall/show_message/27292#message_27292

alanfgh:
I tried fixing this, but it's not working correctly. Not sure whether this is due to my testing procedure or my code change.

Here's the code I added at line 255 of app/vendors/shells/sphinx_conf.php:

```
        /* Turkish
         *   Vowels:
         *     Fold dotless capital I into lowercase dotless i ('I->U+131')
         *     Fold dotted capital I into lowercase dotted i ('U+130->i')
         *     Fold {A,a,U,u} + ^ (U+C2, U+E2, U+DB, U+FB) into letters without ^
         *     Fold {I,i} + ^ (U+CE, U+EE) into lowercase dotless i without ^
         *     Rely on Latin-1 Supplement rules for O, o, U, u with/without umlaut
         *   Consonants:
         *     Fold G + breve (U+11E) into g + breve (U+11F)
         *     Fold S + cedilla (U+15E) into s + cedilla (U+15F)
         *     Rely on Latin-1 Supplement rules for C, c with/without cedilla
         *   Combining marks:
         *     Ignore combining circumflex (U+302)
         */
        $this->indexExtraOptions['tur'] = 
            "
        charset_table = ".implode(', ', array_merge(
            array(
                'A->a', 'U+00C2->a', 'U+00E2->a', 'a',
                'B..H->b..h', 'b..h',
                'I->U+131', 'U+00CE->U+131', 'U+00EE->U+131', 'U+131',
                'U+130->i', 'i',
                'J..T->j..t', 'j..t',
                'U->u', 'U+00DB->u', 'U+00FB->u', 'u',
                'V..Z->v..z', 'v..z',
[...]

PR #1356, opened 2016-11-03
Change Sphinx search conf for Turkish

trang:
This pull request addresses issue #1296.

### DIFF
--- a/app/vendors/shells/sphinx_conf.php
+++ b/app/vendors/shells/sphinx_conf.php
@@ -263,8 +263,6 @@ class SphinxConfShell extends Shell {
          *     Fold C + cedilla (U+C7) into c + cedilla (U+E7)
          *     Fold G + breve (U+11E) into g + breve (U+11F)
          *     Fold S + cedilla (U+15E) into s + cedilla (U+15F)
-         *   Combining marks:
-         *     Ignore combining circumflex (U+302)
          */
         $this->indexExtraOptions['tur'] = 
             "
--- a/app/vendors/shells/sphinx_conf.php
+++ b/app/vendors/shells/sphinx_conf.php
@@ -240,7 +240,7 @@ class SphinxConfShell extends Shell {
          * to search if they are ignored. Since all the Russian diacritics
          * are not single characters but combining characters (e.g. и + ´ = и́)
          * we simply ignore the *´* combining char (U+301) so that characters
-         * are considered not having a diacritic. */
+         * are considered as not having a diacritic. We also ignore soft hyphen. */
         $this->indexExtraOptions['rus'] =
             "
         charset_table = ".implode(', ', array_merge(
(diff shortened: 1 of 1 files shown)


app/vendors/shells/sphinx_conf.php
```diff
@@ -240,7 +240,7 @@ function($v) { return $v != 'A..Z->a..z' && $v != 'a..z'; }
          * to search if they are ignored. Since all the Russian diacritics
          * are not single characters but combining characters (e.g. и + ´ = и́)
          * we simply ignore the *´* combining char (U+301) so that characters
-         * are considered not having a diacritic. */
+         * are considered as not having a diacritic. We also ignore soft hyphen. */
```
jiru: Note that we ignore soft hyphen here only because we ignore them in every language: the parent config block `common_index` says `ignore_chars = U+AD`. Because of the way Sphinx config is constructed, whenever you need to append other values to a config variable in a child block like here, you need to repeat the whole value of the parent. In other words
```php
index parent {
  foo = bar
}
index child : parent {
  foo = baz # This would make the value to be assigned 'baz' only.
  foo = bar, baz # This is what we actually want.
}
```

Now I see that the `ignore_chars` option of `common_index` has some other values, namely `U+5B0..U+5C5, U+5C7`. So you want to put them here too, or do something to avoid code duplication.

app/vendors/shells/sphinx_conf.php
```diff
@@ -251,6 +251,55 @@ function($v) { return $v != 'U+300..U+36F'; }
...
+         *     Fold dotted capital I into lowercase dotted i ('U+130->i')
+         *     Fold {A,a,U,u} + ^ (U+C2, U+E2, U+DB, U+FB) into letters without ^
+         *     Fold {I,i} + ^ (U+CE, U+EE) into lowercase dotless i without ^
+         *     Fold O + diaeresis (U+D6) into o + diaeresis (U+F6)
+         *   Consonants:
+         *     Fold C + cedilla (U+C7) into c + cedilla (U+E7)
+         *     Fold G + breve (U+11E) into g + breve (U+11F)
+         *     Fold S + cedilla (U+15E) into s + cedilla (U+15F)
+         *   Combining marks:
+         *     Ignore combining circumflex (U+302)
```
jiru: The following code does not ignore U+302. Either it needs to be added to the `ignore_chars` option, or that comment needs to be removed.

app/vendors/shells/sphinx_conf.php
alanfgh: Those are Hebrew/Yiddish combining characters (vowels). No one should ever be using them with Russian characters, or searching for them in Russian, so even if we don't explicitly ignore them in the Russian search, we shouldn't run into problems. However, if you can tell me a way to do so without code duplication, I could make that change in a later pull request.

app/vendors/shells/sphinx_conf.php
alanfgh: It's best to remove the comment for now. I wrote it when I was thinking of ignoring combining marks, but I decided not to implement that, at least in the first version. I'll check with @trang to see about removing that line from the pull request.

trang: I removed the comment.

Issue #767, opened 2015-09-12
Allow people to limit searches to sentences on certain lists


ckjpn:
Make it so people can limit searches to a certain list or lists, similar to the way it can now be done with tags.

For example, members who would like to limit their searches to sentences that I've personally proofread could enter the following 2 list numbers.

907, 996

*\* Related URLS **

Here are the full URLs to those lists:
http://tatoeba.org/sentences_lists/show/907/und
http://tatoeba.org/sentences_lists/show/996/und  *note 2015-12-29: what was previously here was moved to 907.

Here is the link to our advanced search:
https://tatoeba.org/eng/sentences/advanced_search

RyckRichards:
Requested by CK on [Tatoeba Day 7](https://tatoeba.org/eng/wall/show_message/24192#message_24192)

ckjpn:
I would like to ask that this be set to "priority high."

This would allow members to search for sentences limited to list 907 which currently has 379,286 sentences.  This would mean that people could be fairly certain that the sentences they found were correct and natural-sounding.  I think this would enhance the user experience, and would help encourage members to translate "good" sentences, which would help improve the quality of the Tatoeba Corpus.

To put this in perspective, 379,286 proofread English sentences is more than the 4th language Turkish which has 378,021 sentences.



This is somewhat related to the following issue.
in list view allow showing sentences "not directly translated into" #611
https://github.com/Tatoeba/tatoeba2/issues/611

#611 would be a nice feature, but wouldn't really be necessary if this could be done with "advanced search", since the advance searching allows excluding sentences with direct links to a selected language.

ckjpn:
Related Wall Post:  https://tatoeba.org/eng/wall/index#message_25413


> I also wonder if we could develop a set of good sentences that contributors of every language could consider translating. 

This could be done right now with lists. A collaborative list would be possible, but that opens up the possibility accidental or malicious additions and deletions.

Instead, I would suggest that each member interested in this idea should create a list and add sentences to it as they find such sentences.

Here as an example list that I have.
https://tatoeba.org/eng/sentences_lists/show/4003

Hopefully, we will eventually have the possibility to search lists. (See https://github.com/Tatoeba/tatoeba2/issues/767). We could then search for sentences on these lists that are not yet translated into our own native languages. If we are also allowed to search more than one list at a time, we could choose which members' lists we felt we could trust, and then do our searches on more than one list at a time.

ckjpn:
[...]

PR #1368, opened 2016-12-03
Search by list

jiru:
This aims to solve #767 with a single dropdown (only one list at a time can be specified as search criterion).
* For guests, the dropdown only shows lists marked as public. For logged-in users, it shows public lists and lists owned by the member separately (similarly to the *Add to list* sentence button).
* Unlisted lists can be used as search criterion by anyone by forging the correct URL (using the `list=<id>` GET parameter) or using a URL provided by the owner of the list. Unlisted lists are hidden from the dropdown unless you’re the owner (in which case they will appear under the 'My lists' category) or you’re currently searching on that list (in which case it appears at the very bottom of the dropdown).
* Private lists can only be searched by their owner and are only displayed to them (under the 'My lists' category). Attempting to forge the URL of a private list that you don’t own will result into the same error message as when the list is not found.
* Whenever a list is modified, Sphinx search results are instantly updated accordingly, without needing to wait for reindexation (just like tags).

Deployment instructions:
* Make sure the Sphinx daily merge will not start in the middle of the update (it starts at 5:25 GMT+0).
* Regenerate Sphinx config using the new code but don’t update the code running on production yet
* Update Sphinx main indexes with `cake sphinx_indexes update main`
* Update Sphinx delta indexes with `cake sphinx_indexes update delta`
* Restart Sphinx daemon
* Update the code running on production with `git pull`

### DIFF
--- a/app/controllers/sentences_controller.php
+++ b/app/controllers/sentences_controller.php
@@ -777,10 +777,21 @@ class SentencesController extends AppController
         }
 
         // filter by list
+        $searchableLists = $this->SentencesList->getSearchableLists();
         if (!empty($list)) {
             $isSearchable = $this->SentencesList->isSearchableList($list);
             if ($isSearchable) {
                 $sphinx['filter'][] = array('lists_id', $list);
+                $found = false;
+                foreach ($searchableLists as $rec) {
+                    if ($list == $rec['SentencesList']['id']) {
+                        $found = true;
+                        break;
+                    }
+                }
+                if (!$found) {
+                    $searchableLists[] = $isSearchable;
+                }
             } else {
                 $ignored[] = format(
                     /* @translators: This string will be preceded by
@@ -877,7 +888,7 @@ class SentencesController extends AppController
         $vocabulary = $this->Vocabulary->findByText($strippedQuery);
 
         $this->set('vocabulary', $vocabulary);
-        $this->set('searchableLists', $this->SentencesList->getSearchableLists());
+        $this->set('searchableLists', $searchableLists);
         $this->set(compact(array_keys($this->defaultSearchCriteria)));
         $this->set(compact('real_total', 'search_disabled', 'ignored', 'results'));
         $this->set(
--- a/app/controllers/sentences_controller.php
+++ b/app/controllers/sentences_controller.php
@@ -797,8 +797,8 @@ class SentencesController extends AppController
                     /* @translators: This string will be preceded by
                        “Warning: the following criteria have been
                        ignored:” */
-                    __("“belongs to list number {listId}”, because this list does ".
-                       "not exist or cannot be searched", true),
+                    __("“belongs to list number {listId}”, because list ".
+                       "{listId} is private or does not exist", true),
                     array('listId' => $list)
                 );
                 $list = '';
--- a/app/controllers/sentences_controller.php
+++ b/app/controllers/sentences_controller.php
@@ -776,6 +776,24 @@ class SentencesController extends AppController
             $tags = implode(',', $tagsArray);
         }
 
+        // filter by list
+        if (!empty($list)) {
+            $isSearchable = $this->SentencesList->isSearchableList($list);
+            if ($isSearchable) {
+                $sphinx['filter'][] = array('lists_id', $list);
+            } else {
+                $ignored[] = format(
+                    /* @translators: This string will be preceded by
+                       “Warning: the following criteria have been
+                       ignored:” */
+                    __("“belongs to list number {listId}”, because this list does ".
+                       "not exist or cannot be searched", true),
+                    array('listId' => $list)
+                );
+                $list = '';
+            }
+        }
+
         // filter orphans
         if (!empty($orphans) && empty($user)) {
             $exclude_orphans = $orphans == 'no';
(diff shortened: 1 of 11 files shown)


jiru:
@alanfgh (or any other native English speaker), could you check the wordings of the new strings introduced by this? Look for the string `__` in the code. In particular, I’m not very confident with `Unspecified` (dropdown item selected when no list is selected) and `Your lists` (maybe `My lists` is more appropriate?). Thank you.

app/controllers/sentences_controller.php
```diff
@@ -773,6 +776,35 @@ public function search()
...
+                }
+                if (!$found) {
+                    $searchableLists[] = $isSearchable;
+                }
+            } else {
+                $ignored[] = format(
+                    /* @translators: This string will be preceded by
+                       “Warning: the following criteria have been
+                       ignored:” */
+                    __("“belongs to list number {listId}”, because this list does ".
```
alanfgh: I think what you're saying is that the criterion "phrase belongs to list number XXX" is being ignored because list XXX does not exist or cannot be searched. If I'm right, I would express it a bit more explicitly:

"belongs to list number {listId}" (ignored because list {listId} does not exist or cannot be searched) 

I see that there are curly quotes in the string. Do we use curly quotes elsewhere?

app/views/helpers/lists.php
```diff
@@ -662,5 +662,38 @@ public function displayListsLinks()
...
+            $sortedLists = array(0 => array(), 1 => array());
+            $currentUserId = CurrentUser::get('id');
+            foreach ($lists as $list) {
+                $where = (int)($list['SentencesList']['user_id'] != $currentUserId);
+                $listId   = $list['SentencesList']['id'];
+                $listName = $list['SentencesList']['name'];
+                $sortedLists[$where][$listId] = $listName;
+            }
+
+            $listsOfCurrentUser = __('Your lists', true);
```
alanfgh: We use "My lists" elsewhere in the GUI, so we should use it here, too.

app/views/helpers/languages.php
```diff
@@ -487,7 +487,7 @@ public function getLevelsLabels($index = null)
     {
         if (!isset($__languagesLevels)) {
             $__languagesLevels = array(
-                -1 => __('Unspecified', true),
+                -1 => __p('level', 'Unspecified', true),
```
alanfgh: What effect will this have on the displayed string?

alanfgh:
I added comments to the code display in GitHub. If there are any issues
with that, just let me know.

Thanks,
Alan

app/controllers/sentences_controller.php
jiru: It is displayed that way because there may be more than one ignored criterion. See [this example](https://tatoeba.org/jpn/sentences/search?query=always&from=und&to=und&orphans=no&unapproved=no&native=yes&user=nonexsitant+user&tags=nonexistant+tag&has_audio=&trans_filter=limit&trans_to=und&trans_link=&trans_user=&trans_orphan=&trans_unapproved=&trans_has_audio=&sort=words).

app/views/helpers/languages.php
jiru: I’m adding the translatable string `Unspecified` while it already exists somewhere else. So this will allow translators to translate the string `Unspecified` as a language level and `Unspecified` in the list dropdown separately. The first parameter `level` is a context indicator that gets displayed in Transifex (this is explained on the [wiki](http://en.wiki.tatoeba.org/articles/show/interface-translation)).

jiru:
Thank you very much, Alan.

app/controllers/sentences_controller.php
alanfgh: That's what I thought, but it was good to see the actual display with some sample lines. Now that I've seen it, I would recommend this wording:

"belongs to list number {listId}" (because list {listId} does not exist or cannot be searched)

By "cannot be searched", I assume you mean that it's a private list, right? In that case, you could alternatively say "or is a private list".

jiru: By “cannot be searched” I mean that we prevent the user from using that list as a search criterion. The reason is that the list is private (more precisely, it is private while not owned by the user), but it may change in the future (or not; maybe I’m just over-abstracting things).

That said, when a member hands over a search link filtering on a private list to someone, a warning message talking about private list would be way more useful, as it gives him a hint about the need to switch the privacy setting of the list.

ckjpn:
Perhaps you might want to exclude any list from the drop-down menu that has fewer than 10 sentences,  since one can easily browse 10 sentences on one page by looking at the list.  This would help keep the number of items in the drop-down menu to fewer items, so it would be easier for users to look through the "select options."

If you don't like that idea, perhaps at least consider not putting lists on the drop-down menu that have no sentences.

I left other comments on the Wall.

trang:
Although I agree it is not very useful to provide search for lists that have less than 10 sentences, it is not a problem either (besides the fact that it bloats the dropdown).

As for removing unlisted lists from the dropdown, I cannot see why we should prevent users from searching unlisted lists.

The various feedback you provided all point out to the problem that the dropdown has too many items. It's definitely a problem, but to me it's not in scope of this pull request and should not interfere with the decision of making a list searchable or not.

It's a UI issue that should be addressed separately: if one day we have thousands of lists, all of them having hundreds of sentences or more, all of them being useful in some way or another, what can we do for users not to feel overwhelmed when selecting a list.

ckjpn:
It's your site, so, of course, you can do what you want.

I would assume, though, that whatever reason you had for providing the possibility of have "unlisted" lists, would also be a valid reason in this case.

Issue #547, opened 2014-12-24
Indicate the contributors of audio


tommy-3:
Currently, there's no way to know who recorded which sentence. By indicating the contributors of audio, we'll be able to help learners interested in a specific accent and, above all, encourage more people to contribute audio.

This involves basically three things.
1. Sentence -> username
   On sentence pages, indicate the contributors of audio. The easiest way would be to change the alt text of the speaker icon to e.g. "Play audio recorded by TRANG".
2. Username -> sentences
   Make a page that lists the sentences recorded by a certain member and add a link to it on the profile of him/her.
3. Download
   Change the structure of the export file sentences_with_audio. It will look like "12345[tab]TRANG".

We could consider solving at the same time another issue #183 "allow multiple audio files per sentence". In that case:
1. Sentence -> username
   1.1. For a sentence with audio, show a number of recordings on the right of the speaker icon.
   1.2. When you click the speaker icon, one of the recordings (chosen randomly) are played.
   1.3. The alt text of the icon will be e.g. "Play audio #2 recorded by TRANG".
   1.4. When you click the number next to the speaker icon, a list of recordings appear below the sentence:
   **#1 2011-01-01 01:01:01 recorded by sysko [play icon]
   #2 2022-02-02 02:02:02 recorded by TRANG [play icon]**
2. Username -> sentences
   2.1. Make a page e.g. "Audio recorded by TRANG" and add a link to it on the profile of the member.
   2.2. The page will basically look like 1.4 above, but only the audio recorded by the user (TRANG in this case) are displayed:
   **C'était seulement il y a un an. [flag] [speaker icon] 2
   #2 2022-02-02 02:02:02 recorded by TRANG [play icon]**
   
   2.3. When you click the number next to the speaker icon, the other recordings will be displayed under those by TRANG.
3. Download
   The export file sentences_with_audio will look like "12345[tab]2[tab]TRANG", which means the audio #2 of the sentence #12345 was recorded by TRANG.
4. Filenames
   The filename of the audio #1 of the sentence #12345 will remain 12345.mp3. The filename of the audio #2 will be 12345_2.mp3.

I tried suggesting a concrete plan, but this is of course only one of many possibilities. I hope this will be the basis for further discussions.

trang:
@tommy-3, I've just sent you the emails.

jiru:
@trang can you make the mails available to us too?

@tommy-3 can you use like a shared google document or similar so that we may help out contacting people and gathering information?

[...]

PR #1378, opened 2016-12-21
Indicate audio contributors

jiru:
This PR solves #547.

### What’s new
* Page `/audio/of/<user>` that displays all the sentences with audio that belongs to `user`. It also shows licensing information clearly (this will hopefully encourage users who haven’t chosen a license to do so), and allows to change the license and attribution URL of all the audios. One user must use the same license for all his audios. The attribution URL is where the username link points when referred as audio author.
* Block `Audio` on the right of `/sentences/show/<id>`. It used to be displayed only to admins. Now, it is displayed to everyone for sentences with audio, to show brief information about the audio. Admins can change the audio ownership from there too. They may choose a username that is not a Tatoeba user, in which case it is considered as an external author. External authors can have a license and an attribution URL too, on an audio-basis (not user-basis, because they are not proper users), but it’s not possible for admins to edit that information yet. For the moment, this feature will be only used for Barack Obama’s voice, and proper licensing information will be inserted with the SQL update script.
* The tooltip of the speaker icon shows the audio author.
* User profile page now mentions the number of audios, along with number of sentences, comments etc.
* Page `/sentences/with_audio` renamed to `/audio/index`
* Page `/recordings/import` renamed to `/audio/import`. Admins must assign any newly imported audios to a user.
* `sentences_with_audio.csv` has more fields to indicate author, license and attribution URL, as described on the Downloads page.

### Deployment instructions
* Run these ACL commands
```
cake acl delete aco Recordings
cake acl create aco controllers Audio
cake acl create aco Audio import
cake acl create aco Audio save_settings

cake acl grant Group.2 Audio all
cake acl grant Group.3 Audio all
cake acl grant Group.4 Audio all
cake acl deny  Group.2 Audio/import all
cake acl deny  Group.3 Audio/import all
cake acl deny  Group.4 Audio/import all
```
* Comment out Sphinx delta reindexation cron job (`sudo crontab -e`)
* Regenerate Sphinx conf with new code
* Set maintenance mode ON (`touch maintenance.mode`)
* Run 2016-12-22_1-schema.sql
* Update production code (`git pull`)
* Set maintenance mode OFF (`rm maintenance.mode`)
* Uncomment cron job
* Run 2016-12-22_2-data.sql (Only on production. On dev, you rather want to try out things manually with the procedures in that file.)

### TODO
* Update the deduplication script to work with the new database schema (@loolmeh could you take care of that?)
* Native check of new translatable strings


alanfgh:
For what it's worth, the English word "audio" is treated as a non-count noun, so you don't see "audios". (I realize that this is different from "video"/"videos".) Furthermore, in a noun phrase consisting of multiple nouns in succession, all but the last are usually in the singular (for example, "audio list selections"). I have the impression that CakePHP might require a final "s" in these situations, but the result is a little weird. The effect could be mitigated by sticking with "recording", but then CakePHP might still require you to use "recordings controller". However, as long as "audios" is confined to a URL, and not to the UI text, it's probably okay.

jiru:
@alanfgh Thank you for your insightful comment. CakePHP doesn’t requires a final *s*, it’s only a convention that, if followed, allows you to write simpler code. I’ll change the controller name to *audio*. Generally speaking, and as a project promoting languages, I think we should not feel limited in our wording by the technology and keep high standards.

trang:
When going to the page `/audio/import`, there is this error:
```
Warning (512): SQL Error: 1054: Unknown column 'Sentence.hasaudio' in 'field list' [CORE\cake\libs\model\datasources\dbo_source.php, line 684]
Code | Context

DboSource::showQuery() - CORE\cake\libs\model\datasources\dbo_source.php, line 684
DboSource::execute() - CORE\cake\libs\model\datasources\dbo_source.php, line 266
DboSource::fetchAll() - CORE\cake\libs\model\datasources\dbo_source.php, line 410
DboSource::read() - CORE\cake\libs\model\datasources\dbo_source.php, line 821
Model::find() - CORE\cake\libs\model\model.php, line 2142
Audio::getFilesToImport() - APP\models\audio.php, line 230
AudioController::import() - APP\controllers\audio_controller.php, line 63
Dispatcher::_invoke() - CORE\cake\dispatcher.php, line 204
Dispatcher::dispatch() - CORE\cake\dispatcher.php, line 171
[main] - APP\webroot\index.php, line 86

Query: SELECT `Sentence`.`id`, `Sentence`.`lang`, `Sentence`.`hasaudio` FROM `sentences` AS `Sentence`   WHERE `Sentence`.`id` IS NULL    
```

trang:
Another error, but not sure if this one is just on my side. On the sentence's page, in the block Audio, I tried to add "Trang HO" in the owner name, but it doesn't save. Got this error:

```
Warning (2): array_merge() [function.array-merge]: Argument #1 is not an array [APP\models\audio.php, line 89]
Code | Context

array_merge - [internal], line ??
Audio::encodeExternal() - APP\models\audio.php, line 89
Audio::beforeSave() - APP\models\audio.php, line 97
Model::save() - CORE\cake\libs\model\model.php, line 1282
Audio::assignAudioTo() - APP\models\audio.php, line 203
SentencesController::edit_audio() - APP\controllers\sentences_controller.php, line 1321
Dispatcher::_invoke() - CORE\cake\dispatcher.php, line 204
Dispatcher::dispatch() - CORE\cake\dispatcher.php, line 171
[main] - APP\webroot\index.php, line 86
```

docs/database/updates/2016-12-22_1-schema.sql
```diff
@@ -0,0 +1,24 @@
+CREATE TABLE `audios` (
+  `id` int(11) NOT NULL AUTO_INCREMENT,
+  `sentence_id` int(11) NOT NULL,
+  `user_id` int(11) DEFAULT NULL,
+  `external` varchar(500) DEFAULT NULL,
+  `created` datetime NOT NULL,
+  `modified` datetime NOT NULL,
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```
trang: We need an index on the column `sentence_id`.

I started deploying this PR on the dev website yesterday and launched the Sphinx indexation. It was still not done today. After some investigation I figured out that the new joins for audio in the sphinx query made it really slow. The query to retrieve and index the English sentences was running after 10 hours and still running.

I did a `ALTER TABLE audios ADD KEY `sentence_id` (`sentence_id`);` and now it's indexing at a normal speed.

ckjpn:
## Suggested Improvements.


### On the sentence's page, ...

Recorded by: fucongcong
Licensed under: unknown

Instead of "unknown", put "no license" or "no offsite use"
(My gut-feeling is "no offsite use" may be better.)

Maybe it would be better to put "License of Offsite Use"
If you did that, then instead of "unknown", you could put "none".

Instead of linking to 
https://dev.tatoeba.org/eng/user/profile/fucongcong
link to 
https://dev.tatoeba.org/eng/audio/of/fucongcong

The reason would be that that is likely what people want to see, and it would save them an extra click.


### On the Browse Audio Pages
https://dev.tatoeba.org/eng/audio/of/fucongcong

Just above the message put the license name.

Something like this.

License of Offsite Use: None
You may not reuse the following audio recordings outside the Tatoeba project, because fucongcong did not chose any license for them yet.

License of Offsite Use: CC BY-NC-ND 3.0
(Link directly to https://creativecommons.org/licenses/by-nc-nd/3.0/)
Attribution URL: www.manythings.org/tatoeba


I would also suggest re-wording this message.

FROM:
You may not reuse the following audio recordings outside the Tatoeba project, because fucongcong did not chose any license for them yet.
TO:
fucongcong has not chosen a license that allows you to reuse the following audio recordings outside the Tatoeba project,

REASON:
It's shorter.
It also does not imply that we expect the member to choose any license in the future.

ckjpn:
### Another suggestion

I think we also need a URL to allow showing only recordings in a given language.

This will only likely effect members who also record sentences in constructed languages or dead languages.

For example, add these possibilities.

https://dev.tatoeba.org/eng/audio/of/fucongcong/wuu
https://dev.tatoeba.org/eng/audio/of/fucongcong/cmn

https://dev.tatoeba.org/eng/audio/of/marcelostockle/por
https://dev.tatoeba.org/eng/audio/of/marcelostockle/epo

https://dev.tatoeba.org/eng/audio/of/alexmarcelo/por
https://dev.tatoeba.org/eng/audio/of/alexmarcelo/lat

ckjpn:
## This message needs to be changed.

FROM:
marcelostockle does not have contributed any audio
TO:
marcelostockle has not contributed any audio

ckjpn:
## Notes on https://dev.tatoeba.org/eng/audio/of/CK

This message should be changed.

FROM:
The following audio recordings, attributed to CK, are licensed under the CC BY-NC-ND 3.0 license.
TO:
The following audio recordings by CK are licensed under the CC BY-NC-ND 3.0 license.

Also, an attribution URL should be displayed.   I entered it on the "My audio" form, but it doesn't show except when logged in as CK.


### Screenshots - Logged in as CK

screen shot 2017-01-06 at 07 26 23


### Screenshots - Logged in as CM

screen shot 2017-01-06 at 07 21 31


### Related to a previous comment about licenses.

The wording that is used on this select menu would be OK for "no license."

No license for offsite use.

screen shot 2017-01-06 at 07 22 26

ckjpn:
## To make the User Interface as expected, ...

You should probably also add a "My audio" item in the drop-down menu.

screen shot 2017-01-06 at 07 58 06

ckjpn:
I now noticed that clicking the username on the page jumps to the attribution URL.
Maybe my earlier comments can be disregarded.  However, I wonder if when someone clicks this link if this is really where they expect to go.  It doesn't really much matter to me, though.

screen shot 2017-01-06 at 08 17 07

jiru:
@trang Remember it’s not yet ready for production because it breaks the deduplication script.

trang:
@jiru yes indeed. Thanks for reminding me. I forgot the dedup script was relying on `hasaudio` to prioritize which sentence to keep.

Who will be taking care of updating script? @loolmeh or you? Or someone else should take care of it?

app/views/audio/of.ctp
```diff
@@ -0,0 +1,121 @@
...
+                $withAudio,
+                $parentId
+            );
+        }
+
+        $pagination->display($paginationUrl);
+    }
+} else {
+    echo $html->tag('h2', format(
+        __("There's no user called {username}", true),
```
zachleigh: "There is no user named {username}" sounds better to me.

ckjpn:
> "There is no user named {username}" sounds better to me.

+1

zachleigh:
The page /audio/of/user throws a 404 error for the missing css file `/css/audio/of.css`

docs/database/updates/2016-12-22_1-schema.sql
```diff
@@ -0,0 +1,25 @@
...
+INSERT INTO audios
+(sentence_id, created, modified)
+SELECT
+  id as sentence_id,
+  0 as created,
+  0 as modified
+FROM sentences
+WHERE hasaudio != 'no';
+
+ALTER TABLE sentences DROP hasaudio;
```
trang: I've been thinking maybe we shouldn't drop the `hasaudio` field yet. We should make it deprecated, and keep it there for long enough so that:
- we have time to re-adapt the deduplication script
- we have time to make sure there's nothing else of our scripts/tools that will break

This would mean of course re-adapting the implementation in this PR so that `hasaudio` still gets updated accordingly when adding/removing audio.

trang:
@jiru, I'm gonna merge this PR now, but it won't be deployed on prod yet.

@halfdan mentioned that he will make an attempt Saturday (tomorrow) to upgrade to CakePHP 2.x. This PR works good overall, so better have this merged before he starts anything. Once the CakePHP upgrade is done, we can spend some time fixing a few more things for the audio feature. And we'll deploy this on prod along with the CakePHP upgrade.

There's just a couple of things I'll change once I've merged:
- I'll remove the `DROP hasaudio` from the update script. We can drop the column later, in a couple of months or something. There's no harm in keeping it for now.
- I'll move the script `2016-12-22_1-schema.sql` outside of the `updates` folder, probably into the `scripts` folder. This script doesn't make changes in the database structure and we don't want Imouto to run this script when people are installing Tatoeba.

jiru:
> I'll move the script 2016-12-22_1-schema.sql outside of the updates folder, probably into the scripts folder. This script doesn't make changes in the database structure and we don't want Imouto to run this script when people are installing Tatoeba.

This doesn’t make any sense to me. 1-schema.sql does change the database structure. Did you mean 2-data.sql? Also, Imouto isn’t running any of the scripts from the updates folder when it installs Tatoeba. It [initializes a new database from the `tables` folder](/Tatoeba/imouto/blob/master/ansible/roles/setup_database/tasks/create_db.yml) that we’re supposed to keep up to date. Updates scripts are meant to be ran manually after one pulls new code.

docs/database/updates/2016-12-22_1-schema.sql
jiru: I don’t think it’s a good idea, because adapting this PR like you said sounds like more work than updating the deduplication script. Besides, keeping the hasaudio field will make it much harder for us to notice that something broke. And I’m pretty confident we’ll be set once the deduplication script is updated. I can take care it of it before the next update. I was just hoping loolmeh would do it because he would certainly do a better and faster job. We don’t need to postpone this forever.

trang: > adapting this PR like you said sounds like more work than updating the deduplication script

I guess it depends from which perspective :) From my perspective, I much feel less confident touching the deduplication script because I haven't worked on it. So I would spend more time overall, because of the learning curve and because I would be more paranoid to mess up something.

> keeping the hasaudio field will make it much harder for us to notice that something broke

That's true. I guess we can find a compromise with:
```
ALTER TABLE sentences CHANGE COLUMN `hasaudio` `hasaudio_old`
enum('no','from_users','shtooka') NOT NULL DEFAULT 'no';
```
It's easier/faster to revert this, if we ever need to revert. It happened to me on the dev instance, that I after I deployed your PR, I wanted to test another PR and needed to revert back but the column was deleted.

trang:
You're right, I got things mixed up.

docs/database/updates/2016-12-22_1-schema.sql
jiru: > I can take care it of it before the next update.

It turns out I couldn’t because I’ve been very busy with other stuff. Feel free to put this code on production the way you want.

PR #1379, opened 2016-12-24
Searchable transcriptions

jiru:
This PR allows to find sentences by their transcriptions. More specifically:
* Chinese (Mandarin) sentences can now be found regardless of being written in simplified or traditional script. It’s also possible to search using pinyin, but one needs to use the numbered notation like ni3hao3 for nǐhǎo (你好) because that’s how we store it in the database. Which is rather counterintuitive, but we may improve that in the future. I just don’t know what would be the best for users. We could ignore the tone numbers so that *nihao* would find ni3hao3, but this would lead to being unable to differentiate words that only vary in tones. I’d like to hear @allan-simon’s opinion on that.
* Uzbek sentences can be now found regardless of being written in Latin or Cyrillic.
* Japanese sentences can be now found using kana, but there’s a catch explained in 9d8515720ecabf2ae81da579e75f118529919394. I [reported the bug](http://sphinxsearch.com/bugs/view.php?id=2586) to the Sphinx project.
* Cantonese sentences can now also be found using jyutping but it’s not very useful because of the way we store these: we use number exponents for tones, which are ignored by Sphinx so they act as word separators. For example one need to type *zung ji* in order to find zung¹ji³. I guess we want to use regular numbers instead of exponents, and then we will have the problem as for Mandarin. [Glancing over Cantonese sentences](https://tatoeba.org/jpn/sentences/show_all_in/yue/none/none/indifferent/page:2) shows a high number of one-syllable words, which are likely to “conflict” with many Latin languages. For example if you search *sin* or *go* on “Any language”, a number of Cantonese sentences are going to come up. Which is not new, but the problem here is that since jyutping is hidden by default, it will certainly confuse users.
* Sphinx makes it possible to search only in the sentence or only the transcription by respectively using the `@text` or `@transcription` field specifier (we may change their names).

### Deployment instructions
* Regenerate Sphinx config.
* Regenerate indexes.

### DIFF
--- a/app/models/behaviors/sphinx.php
+++ b/app/models/behaviors/sphinx.php
@@ -150,13 +150,23 @@ class SphinxBehavior extends ModelBehavior
         // Sort the results by lang, i.e. by index
         $docsByLang = array();
         $size = count($results);
-        foreach ($results as $i => $result) {
+        foreach ($results as $result) {
+            $size += count($result['Transcription']);
+        }
+        $i = 0;
+        foreach ($results as $result) {
             $lang = $result[$model->name]['lang'];
             $text = $result[$model->name]['text'];
             if (!array_key_exists($lang, $docsByLang)) {
                 $docsByLang[$lang] = array_fill(0, $size, '');
             }
-            $docsByLang[$lang][$i] = $text;
+            $docsByLang[$lang][$i++] = $text;
+        }
+        foreach ($results as $result) {
+            $lang = $result[$model->name]['lang'];
+            foreach ($result['Transcription'] as $transcResult) {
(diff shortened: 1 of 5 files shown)


allan-simon:
@jiru for chinese most of people will not think the system is able to handle tones, so you're right to want to skip the numbers in the future. Because either the one searching is a native on a computer that can't type pinyin, and tones for them are so "natural" that they need to pause to consciously think "which tone it is already ?" ,  or it's a learner searching, and is very likely to struggle already to get the pinyin right, let alone the tones :P

allan-simon:
anyway great that it's finally possible :+1:

PR #1381, opened 2016-12-27
Fix busted vocabulary.sql

halfdan:
- Remove duplicate primary key
- Fix syntax error
- Make hash a unique field

Just tried setting up tatoeba and ran into a syntax error and a duplicate primary key error. This fixes the issue.


zachleigh:
Good catch. I always forget to clean up the database files...

trang:
Oh nice! I was about to push a fix for this. I noticed just yesterday there was a problem with the vocabulary table script.

Just one thing: the hash is not unique. Two strings can have the same hash because we don't use a function that ensure uniqueness. It should simply be `NOT NULL`.

halfdan:
@trang Fixed.

PR #1382, opened 2016-12-27
Add Legend to "Languages of Members" screen

halfdan:
screen shot 2016-12-27 at 17 30 37

It's been a while since I last entered PHP land so this might not be the best solution. Happy to make changes. I'm a long time Piwik (piwik.org) and Ghost (ghost.org) contributor looking for some new software to hack on. Tatoeba interests me both from a linguistic and engineering point of view - so I think this might be a good match :)


trang:
Looks good to me :)

Happy to have you onboard.

Issue #1317, opened 2016-09-19
Some submissions are being added multiple times.


ckjpn:
https://tatoeba.org/eng/sentences/show/5450937

![screen shot 2016-09-19 at 12 43 33](https://cloud.githubusercontent.com/assets/7934273/18622113/dd06f464-7e66-11e6-8589-e24a75fd1971.png)

https://tatoeba.org/eng/sentence_comments/of_user/Dejo

![screen shot 2016-09-19 at 12 45 23](https://cloud.githubusercontent.com/assets/7934273/18622185/fab9338c-7e66-11e6-931a-03bfdac4a4f6.png)

jiru:
I am pretty confident this problem has been fixed in #1284. One can reopen a ticket if this happens again.

PR #1383, opened 2016-12-27
Move click handlers out of watch to prevent re-adding.

### LINKED ISSUE #1284: after adding a translation to a page with multiple sentences, the "Add to list" buttons on that page don't work
I'm using this search, looking at page 3 of the results:

https://tatoeba.org/eng/sentences/search/page:3?query=&from=heb&to=und&orphans=no&unapproved=no&native=yes&user=&tags=&has_audio=&trans_filter=exclude&trans_to=eng&trans_link=direct&trans_user=&trans_orphan=&trans_unapproved=&trans_has_audio=&sort=random

All the elements have loaded, and when I hover over the "Add to list" button for any sentence, it turns black, but clicking it doesn't do anything. If I open an individual sentence in another page, the "Add to list" button works normally there.

When I go to page 4 of the results, the "Add to list" buttons work normally. When I copy the URL into another tab and change the 4 to a 3, every thing works fine, too.

I was using Firefox on Windows 7 when I saw this problem. I didn't see it in Chrome on the same machine when I went to the same URL, but that difference in behavior may not have been related to the browser.

halfdan:
closes #1284

The issue here is that adding a new translations triggers the "addrule" event and hence causes the event handler to be added on top of the other, ending up doing a double toggle on the "Add to List" element.

There are a lot of other places where similar things can happen. This fixes the original issue, it does however not address all the other click handlers in other JS files where similar problems will surely occur. If desired I can do a manual audit of the Javascript. 

It's best to remember to always unbind / [off](http://api.jquery.com/off/) events before adding them. This might seem verbose but it will prevent these things from happening.

### DIFF
--- a/app/webroot/js/sentences_lists.menu.js
+++ b/app/webroot/js/sentences_lists.menu.js
@@ -18,127 +18,112 @@
 
 
 $(document).ready(function() {
+    // Clicking on "Add to list" displays the list selection.
+    // Reclicking on it hides the list selection.
+    $(".addToList").click(function(){
+        sentenceId = $(this).attr("data-sentence-id");
+        $("#addToList"+sentenceId).toggle();
+    });    
+
+    // The sentence is added to the list after user has clicked
+    // the button. I was hesitating between this solution and
+    // having the sentence added to the list onChange, that is,
+    // directly after the selection in the <select>.
+    $(".validateButton").click(function(){
+        var listId = $("#listSelection"+sentenceId).val();
+        var rootUrl = get_tatoeba_root_url();
+
+        // Add sentence to selected list
+        if(listId > 0){
+            $("#_"+sentenceId+"_in_process").show();
+
+            $.post(
+                rootUrl
+                + "/sentences_lists/add_sentence_to_list/"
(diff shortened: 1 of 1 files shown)


app/webroot/js/sentences_lists.menu.js
```diff
@@ -18,127 +18,112 @@
...
+    $(".addToList").click(function(){
+        sentenceId = $(this).attr("data-sentence-id");
+        $("#addToList"+sentenceId).toggle();
+    });    
+
+    // The sentence is added to the list after user has clicked
+    // the button. I was hesitating between this solution and
+    // having the sentence added to the list onChange, that is,
+    // directly after the selection in the <select>.
+    $(".validateButton").click(function(){
```
halfdan: Had the above .addToList worked this would've resulted in two calls to add_sentence_to_list.

halfdan:
https://github.com/Tatoeba/tatoeba2/issues/1317 is very likely related and of the same cause.

trang:
> If desired I can do a manual audit of the Javascript.

If you feel like it, go ahead =)

Issue #1343, opened 2016-10-16
Make what shows on Google search results better (Enhancement)


ckjpn:
## Here is an example of what is currently seen.

![screen shot 2016-10-16 at 09 34 48](https://cloud.githubusercontent.com/assets/7934273/19414336/4cf138d4-9384-11e6-9e6a-835a04ff6603.png)

Here's the URL to the tatoeba.org page.
https://tatoeba.org/eng/sentences/show/5517424

### Use the googleoff  googleon pair to block certain parts of the sentence details pages.

Remember that it's not always the meta description that gets shown.

```
<!--googleoff: index-->
This is the text to be blocked.
<!--googleon: index>
```

Perhaps the only part of the page that is really important for google to index is the sentence and its translations.  Perhaps the comments, too, but then maybe not.

### Perhaps change the meta description to be something more useful.

```
	<meta name="description" content="Browse translated example sentences. This page shows translations and information about the sentence: The colors aren&#039;t adequate." />
```

I personally would start with the sentence, since in cases where sentences are very long, people would more likely see the complete sentence in the google results.

### Change the order of the title tag, so the sentence is show first.

This would help show the complete sentence in cases where the sentence is very long.

So, instead of this.

```
<title>
        English example sentence: The colors aren't adequate. - Tatoeba    </title>
```

Have something like this.
```
<title>
        The colors aren't adequate. - Tatoeba  English example sentence:</title>
```

Or maybe, something like this.
```
<title>
        The colors aren't adequate. - English sentence example:</title>
```

## Note

While this may not help a great deal, it should be a relatively simple thing to do to improve the usefulness of what's displayed by google in their search results.

halfdan:
@ckjpn Thanks for this! I think there's a lot that can be done in terms of SEO friendliness on Tatoeba. I agree that the current website isn't very supportive of letting Google know what a page is displaying.

- Use the googleoff googleon pair to block certain parts of the sentence details pages.

I don't think this is going to help much. The comments can be a valuable addition to the search result and may provide additional clues as to how and why a sentence was translated a certain way (things a search engine user might find useful). Google is quite smart in picking the main content and ignoring navigation elements.

- Perhaps change the meta description to be something more useful.

[...]

PR #1385, opened 2016-12-29
Improve search results

halfdan:
This is a solution for #1343. Putting the sentence first in both the title and the meta description should help show people what they actually searched for. 

This sadly invalidates a couple of translation strings, but I think the benefits from improved search results should pay off.

### DIFF
--- a/app/views/sentences/show.ctp
+++ b/app/views/sentences/show.ctp
@@ -45,11 +45,12 @@ if (isset($sentence)) {
         'description',
         format(
             __(
-                "Browse translated example sentences. ".
-                "This page shows translations and information about the sentence: {sentenceText}"
+                "\"{sentenceText}\" is part of the Tatoeba language corpus " .
+                "and has {translationsCount} translations available. " .
+                "Browse this and other translated example sentences. "
                 , true
             ),
-            compact('sentenceText')
+            array('sentenceText' => $sentenceText, 'translationsCount' => sizeof($sentence['Translation']))
         ),
         array('inline' => false)
     );
--- a/app/views/sentences/show.ctp
+++ b/app/views/sentences/show.ctp
@@ -35,14 +35,14 @@ if (isset($sentence)) {
     $sentenceText = $sentence['Sentence']['text'];
     $sentenceCorrectness = $sentence['Sentence']['correctness'];
     $sentenceHasAudio = $sentence['Sentence']['hasaudio'];
-    
+
     $languageName = $languages->codeToNameToFormat($sentenceLang);
-    $title = format(__('{language} example sentence: {sentence}', true),
+    $title = format(__('{sentence} - {language} example sentence', true),
                     array('language' => $languageName, 'sentence' => $sentenceText));
     $this->set('title_for_layout', $pages->formatTitle($title));
 
     $html->meta(
-        'description', 
+        'description',
         format(
             __(
                 "Browse translated example sentences. ".
@@ -50,7 +50,7 @@ if (isset($sentence)) {
                 , true
             ),
             compact('sentenceText')
-        ), 
+        ),
         array('inline' => false)
     );
 } else {
@@ -102,7 +102,7 @@ $navigation->displaySentenceNavigation(
     }
     ?>
 
-    <?php 
+    <?php
     if (isset($sentence)){
         $tags->displayTagsModule($tagsArray, $sentenceId, $sentenceLang);
 
@@ -115,7 +115,7 @@ $navigation->displaySentenceNavigation(
                     'sentenceId' => $sentenceId,
                     'hasaudio' => $sentenceHasAudio
                 )
-            ); 
+            );
 
             // TODO For the beginning we'll restrict this to admins.
             // Later we'll want CurrentUser::isModerator();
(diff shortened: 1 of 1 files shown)


trang:
* I'm considering dropping the "example" in the title, and just have: `{sentence} - {language} sentence`
  * Many of our sentences are not typical example sentences, and the difference between a sentence and an example sentence is not so important that it needs to be specified in the title.

* I'm also considering not mentioning the language in the title: `{sentence} - Sentence #{id}`
  * Some sentences may be assigned to the wrong language.
  * Some sentences have no language assigned. 
  * Duplicates can happen (although they are deleted quite fast), which could result in duplicate titles for different pages.
  * The id itself is not something the user needs to know, but it conveys the idea that this sentence is an item in a (very big) list, and putting an emphasis on that may be more important than indicating the language. The language would still be in the description. I don't feel strongly about this point though. Whether we emphasize the language or the fact that it's an item in a corpus, either way it's fine.

* I don't think there's a need to repeat the sentence in the description if we already have it in the title. We don't have a lot of characters for the description either, so we should try to be as efficient as possible. It can also be a problem to repeat the title in the description, for very long sentences. Let's imagine there's a very long sentence about global warming or something, the Google results would display just the beginning of the sentence and trunctae the rest, both in the title and description. It could mislead the user into thinking the page is talking about global warming.

------

Concretely, my suggestions would look like this:

(1)
```
The colors aren't adequate - Sentence #5517424 - Tatoeba

English sentence in the Tatoeba corpus. 2 translations available.
Browse this and other translated sentences.
```
If no language, description would be: `Sentence in the Tatoeba corpus...`

(2)
```
The colors aren't adequate - English sentence - Tatoeba

Sentence #5517424 of the Tatoeba corpus. 2 translations available.
Browse this and other translated sentences.
```
If no language, title would be: `The colors aren't adequate - Tatoeba`

For comparison, this is your current version:
```
The colors aren't adequate - English example sentence - Tatoeba

"The colors aren't adequate" is part of the Tatoeba language corpus and 
has 2 translations available. Browse this and other translated example sentences.
```

@alanfgh, we'll need your English knowledge here again :) Can you confirm if the wording of my suggestions is okay? Feel free to suggest other possibilities if you have more ideas.

alanfgh:
@trang, thanks for asking. The wording is fine, but I have several questions and comments:

(A) In all the versions you provided, there are no punctuation marks. Is it possible to have this, for instance?

"The colors aren't adequate." - Tatoeba - Sentence \#5517424 

(B) As you might have guessed from my previous example, I would rather see "Tatoeba" in second place rather than last place in the title. As an end user, I might be searching a variety of sites, and the fact that I found this hit on Tatoeba is more important to me than the sentence number. It's also more important than the language, which the person is likely to already know or be able to remember by seeing the sentence text.

(C) It's better to avoid starting a sentence with a number. Thus, I'm not crazy about "2 translations available" as a standalone sentence. I would rather see it as part of the previous sentence, introduced by words (like "and has" in the current version) or a semicolon.

(D) Does "Browse this and other [translated example] sentences" constitute a clickable link? If not, I don't see the point of that sentence.

(E) I agree that it's better not to repeat the sentence text in the note below it.

(F) I prefer removing both "translated" and "example" from "Browse this and other translated example sentences". Not all sentences on Tatoeba are translated, and untranslated sentences can still address a user's needs.

halfdan:
Closing this for the time being - will reopen a new PR once the CakePHP upgrade is merged. 

Thanks for all the really good feedback @trang / @alanfgh.

alanfgh:
Glad to add my opinion. Thanks for all the improvements you've been making to the codebase, @halfdan.

trang:
I left some questions unanswered there. My answers below.

> (A) In all the versions you provided, there are no punctuation marks. Is it possible to have this, for instance?

It's possible, yes. My only concern if we do this is that some of our sentences are mini-dialogs and contain quotes. For these sentences the title would end up with double quotes. It can also look strange if the sentence contains a word in quotes.

> (B) As you might have guessed from my previous example, I would rather see "Tatoeba" in second place rather than last place in the title.

This would mean we have to break our pattern `{title} - Tatoeba`. I have nothing against breaking this pattern, but in our case, I feel it's maybe not necessary. Search engine will usually present search results the following way:
- title
- URL
- description

From the URL you can see that the result is from tatoeba.org.

> (C) It's better to avoid starting a sentence with a number. Thus, I'm not crazy about "2 translations available" as a standalone sentence. I would rather see it as part of the previous sentence, introduced by words (like "and has" in the current version) or a semicolon.

"...and has 2 translations" sounds good to me.

> (D) Does "Browse this and other [translated example] sentences" constitute a clickable link? If not, I don't see the point of that sentence.

No, it's not clickable. You're right, this sentence doesn't bring any additional information. I think we can do without it.


`

ckjpn:
> "The colors aren't adequate." - Tatoeba - Sentence #5517424

Perhaps a little shorter title would be good.  I don't think the quotes are needed.

The colors aren't adequate. - Tatoeba #5517424
... or maybe a little more exact ...
The colors aren't adequate. - Tatoeba Corpus #5517424
...or this is the same number of characters
The colors aren't adequate. (Tatoeba Corpus #5517424)

The last one is perhaps the way people might include the sentence in a research paper.


> I don't think there's a need to repeat the sentence in the description if we already have it in the title. We don't have a lot of characters for the description either, ...

Descriptions can be longer than titles, so having it in the description lets people see more of a long sentence than they can see in the title.

Perhaps you could truncate long sentences before putting them into the description to whatever the length that Google actually shows.

Perhaps you don't even need a description for shorter sentences that get completely displayed in the title.

From an SEO standpoint, the "description" tag is mostly ignored, I think, so changes to the description tag would be mainly at improving the usefulness for people viewing Google (and maybe other search engines) search results.

PR #1387, opened 2016-12-29
Correctly unbind click events before rebinding

halfdan:
refs #1284

### DIFF
--- a/app/webroot/js/sentences_lists.menu.js
+++ b/app/webroot/js/sentences_lists.menu.js
@@ -18,96 +18,8 @@
 
 
 $(document).ready(function() {
-    // Clicking on "Add to list" displays the list selection.
-    // Reclicking on it hides the list selection.
-    $(".addToList").click(function(){
-        sentenceId = $(this).attr("data-sentence-id");
-        $("#addToList"+sentenceId).toggle();
-    });    
-
-    // The sentence is added to the list after user has clicked
-    // the button. I was hesitating between this solution and
-    // having the sentence added to the list onChange, that is,
-    // directly after the selection in the <select>.
-    $(".validateButton").click(function(){
-        var listId = $("#listSelection"+sentenceId).val();
-        var rootUrl = get_tatoeba_root_url();
-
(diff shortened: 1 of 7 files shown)


app/webroot/js/sentences_lists.menu.js
```diff
@@ -125,5 +37,94 @@ $(document).ready(function() {
...
+            sentenceId = $(this).attr("data-sentence-id");
+            $("#addToList"+sentenceId).toggle();
+        });
+
+        // The sentence is added to the list after user has clicked
+        // the button. I was hesitating between this solution and
+        // having the sentence added to the list onChange, that is,
+        // directly after the selection in the <select>.
+        $(".validateButton").off();
+        $(".validateButton").click(function(){
```
halfdan: Had to move these back as they depend on `feedbackValid` being defined. Instead I simply `off` existing events and rebind them.

app/webroot/js/wall.reply.js
```diff
@@ -84,73 +84,64 @@ $(document).ready(function(){
...
-            // that way we're sure replyFormInMessageId will not be set to -1
-            // before the end of saveMessage
-    }); 
-
+    $('.ajaxSubmit').click(function(){
+        saveMessage() ;
+        // this line is not in save message due to ajax's asynchronousity
+        // that way we're sure replyFormInMessageId will not be set to -1
+        // before the end of saveMessage
+    });
```
trang: For some reason the "Save" button (in reply form on the Wall) doesn't work anymore with this change.

Same issue with the "Cancel" button.

halfdan:
Interesting, @trang, I'll check it out and make sure it works. It might be worth spending some time on proper unit/acceptance tests after we upgrade to a newer CakePHP version.

trang:
@halfdan I'll merge your PR since it fixes #1389 and I'd like to deploy that fix on prod. I'll just revert the changes in `wall.reply.js`.

>  It might be worth spending some time on proper unit/acceptance tests after we upgrade to a newer CakePHP version.

Definitely.

Issue #1279, opened 2016-08-06
"remember me" doesn't seem to work for everyone.


ckjpn:
http://tatoeba.org/eng/users/login

A new member wrote and mentioned this to team@tatoeba.org.

Also, it doesn't "remember me" on Firefox on my MacBook Pro, running MacOS 10.9.5.

ckjpn:
For a long time, recently, this hasn't been a problem for me, but today, again, I had to sign in twice using Firefox.  Chrome doesn't seem to have this problem.  (I often use both of these browsers.)

NM-30:
+1

Windows 7 - Mozilla Firefox 49
Windows 8 - Chrome

ckjpn:
I just got logged out of Chrome, too.

ckjpn:
I've had to log in twice already today with Firefox, so this is definitely a reoccurring problem.

ckjpn:
I've had to log in several times today, too, using Firefox.

Is there some additional setting I can make in Firefox to prevent this from happening, or is this a problem with tatoeba.org.

trang:
It's unclear whether it's a problem with your settings or with tatoeba.org. We haven't changed anything regarding the login and "remember me" feature. The problem hasn't occured to me. Until we identify more precisely in which context the "remember me" fails, there is not much we can do.

halfdan:
Nope, this problem is Angular related. It is broken in production - only however when logging in through the central login form (the one on the top right works). The rememberMe field is not correctly set even though it is displayed as checked. 

I will investigate this after the CakePHP upgrade.

halfdan:
As a side note: The cookie is currently not encrypted which is terrifying. The password/user are just stored in plain text on the client side.

halfdan:
Figured out what the issue was. Checkboxes are weird! The `value=1` is only submitted if at the same time the checkbox is also `checked="checked"`.  Fix in my cakephp29 branch.

NM-30:
Hi there! 
I wanted to tell you that I'm really enjoying using italki for my online language learning.  I want to invite you to learn (or even teach) a language as well!

You can easily meet other language learners, find a teacher, and schedule lessons, all on italki.  AND, once you join, we can follow each other's progress!


Hope to see you soon. Please click on my referral link and sign up and connect with me:
https://www.italki.com/?ref=3984727

Issue #1122, opened 2016-05-01
Update to the current version of CakePHP and adapt the code to work with it.


ckjpn:
Perhaps it might be a good idea to update from Version 1.3 to Version 3.2.

I know it would involve a lot of time, but in the long run, it may help save total man hours, assuming that we'll be getting more and more developers volunteering their time to help.

I think potential developers for the Tatoeba Project would likely be more interested in spending time to learn how to use the current version of CakePHP rather than an old version, since they could apply what the learn to possible other projects that they may want to develop for themselves in the future.

I know that I personally would be more interested in helping if we were working with the current version of CakePHP.

http://cakephp.org

I'm not sure, but additionally very likely the newer version of CakePHP is more efficient.

halfdan:
CakePHP 3.3 has since been released. This is certainly a good idea to also take advantage of all the new developments in cake.

trang:
There's a few things to consider:

1) How much effort would this take?
2) What exactly would be the benefits?
3) Do we want to continue with CakePHP at all, or move to another framework?

I don't have clear answers to these questions.

I don't reject at all the idea of upgrading, but I would like not to do it for purely "marketing" reasons, because working on the latest version just sounds more attractive. We need to look at the project's needs, and choose according to that.

I'll be more than happy though, if someone wants to try plugging the newest Cake version into Tatoeba's code, and see how things run, get an overall idea of how much time it'd take to migrate, what would be better, what would be worse.

@zachleigh said to me he'd be interested to have a look at this. But he's also interested to work building an API for Tatoeba, which (to me at least) would be a higher priority. Probably he'll be focusing on that.

On my side, for the next six months or so, my focus is on the UI, on finishing #1180. So this upgrade would definitely not be a sub-project I could take on.

halfdan:
These are all very important questions.

- How much effort would this take?

I just attempted an upgrade to CakePHP 2.x, from there the upgrade path becomes easier. The upgrade will certainly require a significant amount of work but will also help structure the app more. A number of parts in the current app could be componentized and hence better testes / reused. 

- What exactly would be the benefits?

[...]

PR #1395, opened 2017-01-15
CakePHP 2.9.4

halfdan:
This PR upgrades Tatoeba to the latest CakePHP 2.x version.

**NOTES: ** 
- Highly experimental. Not yet tested at all. Still has bugs.
- nginx.conf needs a change to work with the new CakePHP version
- more to be documented


halfdan:
Known Issues:

- Viewing sentences is broken
- Registration screen does not work
- There seems to be an issue with authorization when submitting forms

18 hours is enough for today - continuing tomorrow.

halfdan:
A couple things here:

- This now fixes #1279 and properly initializes the remember me feature.
- The remember me cookie is now encrypted since we don't want to store user/pass in plain text on the client side
- We update angular to the latest 1.x and angular-material to its latest version.

trang:
Alright, time to merge this. It's stable enough :) I'll be fixing whatever is left to fix on the dev branch.

There are a couple of things that have been left out intentionally:
- the mass import of sentences doesn't work at all
- some of the tools are not working

These will not be fixed for the incoming Tatoeba update. I'll create issues for these topics later on.

Just for the reference, this PR addresses #1122.

jiru:
Good job! :-)

PR #1403, opened 2017-02-01
Deduplication script update for audio contributors

jiru:
This updates the deduplication script to make it work with the changes in database schema introduced by #1378. This adds no new functionality, the script should work like before. It needs Tatoeba/dj-tatoeba2#1 to work.


trang:
Thanks @jiru :)

There's just one thing to fix in Tatoeba/dj-tatoeba2#1 and I can merge.

Issue #1414, opened 2017-02-13
Can't import audio


RyckRichards:
https://tatoeba.org/eng/wall/show_message/28036#message_28036

CK  

"Bug: Can't import audio
I tried to import some new audio today and go this error message.
Not Found
Error: The requested address '....' was not found on this server."

ckjpn:
This issue can be closed.
I wasn't informed that the URL had changed.  That was the only problem.

PR #1416, opened 2017-02-13
Fix audio import

jiru:
This fixes audio import. Note that the issue #1414 mentions an error in the address, likely because I changed the address in #1378. But I found another bug while investigating, which this PR fixes.


ckjpn:
Maybe this just hasn't been added yet, and you have already corrected the problem.

https://tatoeba.org/eng/audio/import

screen shot 2017-02-14 at 03 14 02

jiru:
> Maybe this just hasn't been added yet, and you have already corrected the problem.

Yes! This PR fixes this problem.

PR #1461, opened 2017-03-23
Gdrive backup

loolmeh:
Please check if I missed any paths or files that should be included. The cron job is active on the server, runs an hour after the dumps.

### DIFF
--- a/docs/cron/backup.sh
+++ b/docs/cron/backup.sh
@@ -15,6 +15,7 @@ GDRIVE(){
     $GDRIVE_BIN --config "$CONFIG_DIR" --service-account "$AUTH_JSON" "$@"
 }
 
+GDRIVE_FOLDER="0B3fWO5pH-xU1bFQ0aEFSQmNsMzA"
 BACKUP_DIR="backupdir"
 mkdir -p $BACKUP_DIR
 
@@ -61,7 +62,7 @@ for i in "${BACKUP_LIST[@]}"; do
 
         f="$f""_""$DATE$ext"
         echo "Uploading file $f ..."
-        GDRIVE upload --name "$f" --chunksize "$CHUNK_SIZE" "$i"
+        GDRIVE upload --name "$f" --chunksize "$CHUNK_SIZE" "$i" -p "$GDRIVE_FOLDER"
     elif [[ -d "$i" ]]; then
         f=$(echo "$i" | sed 's/^\///g; s/\//-/g')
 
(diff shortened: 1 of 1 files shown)


trang:
Thanks @loolmeh.

The images uploaded to the wiki are missing: `/home/debian/tatowiki/tatowiki/upload`.

I think it's not too necessary to backup the `profiles_36` since the images in there are just a reduced size version of `profiles_128`. But no problem keeping it, doesn't cost much.

loolmeh:
> The images uploaded to the wiki are missing

That's inside ../app actually. It's covered.

trang:
Indeed, my bad. 
So we're good. On my side I can't think of other data to back up.

docs/cron/backup.sh
```diff
@@ -0,0 +1,72 @@
...
+)
+
+GDRIVE="/home/debian/gdrive/gdrive"
+
+BACKUP_DIR="backupdir"
+mkdir -p $BACKUP_DIR
+
+# in Mbs
+CHUNK_SIZE=32
+CHUNK_SIZE=$(echo "$CHUNK_SIZE*1024*1024" | bc)
```
jiru: How about `CHUNK_SIZE=$(( $CHUNK_SIZE*1024*1024 ))`?

docs/cron/backup.sh
```diff
@@ -0,0 +1,72 @@
...
+    flst=$($GDRIVE list --query "name contains '$search_name'" --order "createdTime desc" | tail -n +2);
+    curr_num=$(echo "$flst" | wc -l)
+    if [[ $curr_num -gt $keep_num ]]; then
+        dlst=$(diff <(echo "$flst") <(echo "$flst" | head -n $keep_num) |\
+         tail -n +2 | sed 's/< //g' | awk '{print $1}'
+        )
+
+        echo "Pruning $(echo "$dlst" | wc -l) files related to $search_name ..."
+        IFS=$'\n' dlst=($dlst)
+        for d in ${dlst[@]}; do
```
jiru: I think you should quote this: `"${dlst[@]}"`. Otherwise it will mess up with IFS chars contained in array elements.

docs/cron/backup.sh
```diff
@@ -0,0 +1,72 @@
...
+            $GDRIVE delete "$d"
+        done
+    fi
+}
+
+echo "Starting gdrive backup script at $DATE ..."
+
+for i in "${BACKUP_LIST[@]}"; do
+    # check for file or dir
+    if [[ "$(echo $i | grep -o '\.' | uniq)" == "." ]]; then
```
jiru: Is this supposed to check if $i is a file? If yes, I suggest using `[[ -f "$i" ]]` instead (and `-d` for directory check).

loolmeh:
@jiru @ventice Thanks for the review.

ventice:
Just found out that after switching to service account the script started to upload the files to service account's Google drive which is something that I don't understand.
Actually, to work in service account logic the following is to be done:
1. A folder is to be created in tatoeba.backup's Google Drive.
2. Write permission to service account's user is to be set up on that folder.
3. That folder is to be added with `-p <folder-id>` parameter to `GDRIVE upload`'s arguments.

trang:
Backup script is working now. Thanks all :)
It's currently running with a gap of 48 hrs.

Issue #1476, opened 2017-07-20
Adding vocabulary items is not possible


Guybrush88:
As reported by sundown (https://tatoeba.org/ita/user/profile/sundown) on the Wall:

> I'm unable to add vocabulary (endlessly spinning page-load icon). Anyone else, just me?

https://tatoeba.org/ita/wall/show_message/28284#message_28284

I was able to reproduce this.

This is the related URL: https://tatoeba.org/ita/vocabulary/add

trang:
Can't reproduce this anymore.

Guybrush88:
this feature now works

PR #1490, opened 2017-07-31
Fill hash value when adding an UsersVocabulary entity (#1476).

mycroft:
When adding a new Vocabulary item, an hash is created and filled in both
Vocabulary & UsersVocabulary entity. For any reason, it was not passed in the
last one, leading to the SQL error: "General error: 1364 Field 'hash' doesn't
have a default value".


mycroft:
By the way, I was able to reproduce this by setuping an httpd, picking the code, the old (2010) db dump, & applying patched on it.
I am not being able to reproduce the same issue using the vagrant stuff. My guess is that there is some differences between the production db showing the error & schemas. So I guess my patch should be thrown away & debug to do on production instance.

PR #1565, opened 2018-06-09
New password hashing algorithm

### LINKED ISSUE #1396: Use better password hashing algorithm
Tatoeba currently uses md5 password hashing: https://github.com/Tatoeba/tatoeba2/blob/260358cfb080ba3976c442a5bd70b0f3c81a76b6/app/app_controller.php#L120

This is widely considered very insecure. We should update to bcrypt as soon as possible.

It's obviously not possible to change the password hashing without some kind of migration path as we can't rehash existing passwords (since we don't have them in plain text). The solution is as follows:

- We change the default hashing to bcrypt
- When a user logs in
  - we compute the bcrypt hash of their PW and compare it with the one in the DB. If it doesn't match...
  - we compute the MD5 hash of their PW and compare it with the one in the DB. If it matches
  - we log the user in and store the bcrypt hash in the DB. Which will then be used by default the next time the user logs in.

The password field is currently a VARCHAR(50) and will likely have to be changed to a CHAR(60). See [for reference](http://stackoverflow.com/questions/5881169/what-column-type-length-should-i-use-for-storing-a-bcrypt-hashed-password-in-a-d)

jiru:
Fixes #1396. Reviews welcome.

Installation instructions:
1. Switch to maintenance mode
2. `git pull`
3. Uncomment backuping instructions in `docs/database/updates/2018-06-09.sql` if you want
4. `mysql tatoeba < docs/database/updates/2018-06-09.sql`
5. `cake update_password_version` (it won’t break things even if you run this several times)
6. Remove maintenance mode
7. Remove backup table `users_bak` if you uncommented code in 3.

### DIFF
--- a/docs/database/updates/2018-06-09.sql
+++ b/docs/database/updates/2018-06-09.sql
@@ -1,0 +1,4 @@
+-- Maybe backup login credentials if anything goes wrong
+-- CREATE TABLE users_bak LIKE users;
+-- INSERT users_bak SELECT * FROM users;
+ALTER TABLE users CHANGE password password VARCHAR(62);
--- a/app/Console/Command/AppShell.php
+++ b/app/Console/Command/AppShell.php
@@ -26,5 +26,24 @@ App::uses('Shell', 'Console');
  * @package       app.Console.Command
  */
 class AppShell extends Shell {
+    protected function batchOperation($model, $operation, $options) {
+        $batchSize = 1000;
+        $proceeded = 0;
+        $options = array_merge(
+            array(
+                'contain' => array(),
+                'limit' => $batchSize,
+                'offset' => 0,
+            ),
+            $options
+        );
 
+        do {
+            $data = $this->{$model}->find('all', $options);
+            $proceeded += $this->{$operation}($data, $model);
+            echo ".";
+            $options['offset'] += $batchSize;
+        } while ($data);
+        return $proceeded;
+    }
 }
--- a/app/Console/Command/TranscriptionsShell.php
+++ b/app/Console/Command/TranscriptionsShell.php
@@ -20,7 +20,7 @@
 App::import('Model', 'Sentence');
 App::import('Model', 'Transcription');
 
-class TranscriptionsShell extends Shell {
+class TranscriptionsShell extends AppShell {
 
     public $uses = array('Sentence', 'Transcription', 'Contribution');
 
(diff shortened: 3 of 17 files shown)


trang:
I'm deploying this on the dev website. Tried it locally it seems to be fine, but I didn't look into the code itself yet. Will do that while users test on dev.

docs/database/updates/2018-06-09.sql
```diff
@@ -0,0 +1,4 @@
+-- Maybe backup login credentials if anything goes wrong
+-- CREATE TABLE users_bak LIKE users;
+-- INSERT users_bak SELECT * FROM users;
```
trang: Hmm so I actually forgot to uncomment this while deploying on the dev website... I still made a back up before running the shell command, with the altered password column. It's perhaps best to uncomment it by default? It doesn't cost much to have this backup. We just have to remember to clean it up later.

trang: Okay I see. The process of uncommenting is meant to avoid having the table being created without being noticed.

trang:
Okay so I tested further than just logging in. It looks like changing password or resetting password doesn't work.

jiru:
Okay, I fixed the change and resetting of password, please check again. Users now also get their password hash updated with the new hashing method upon log in, for better security.

I also improved the security of the "remember me" cookie. It used to contain the password in plaintext, and now we use the password hash stored in database instead. This could be [improved even more](https://stackoverflow.com/questions/244882/what-is-the-best-way-to-implement-remember-me-for-a-website) but that’s out of the scope of this PR.

I incidentally found out that the method User::getUsersWithSamePassword() won’t be working any more. This function is used to fill the [email that is sent to admins about blocked or suspended users](https://github.com/Tatoeba/tatoeba2/blob/bcf9790d9e14e8abe71fa9e70281ea13a93c2db6/app/View/Emails/html/blocked_or_suspended_user.ctp). With this PR, there is just no way of finding out who’s sharing the same password. Should I remove this information from the email?

trang:
> With this PR, there is just no way of finding out who’s sharing the same password.

Hmm, indeed.

> Should I remove this information from the email?

It'll need to be removed at some point yes, but doesn't have to be in this PR. First I'll warn the admins about this change and check with them if we can replace this information with something else. They've been using it to identify problematic users who create new accounts after being suspended (as they likely use the same password for their new accounts).

trang:
I think this is ready for merge. Let's merge...

Thanks again for taking care of this, @jiru.

Issue #936, opened 2015-11-26
Reviewing​ names of languages


trang:
There has been requests to rename some languages such as ["Moroccan Arabic" into "Arabic (Moroccan)"](https://tatoeba.org/eng/wall/show_message/24920#message_24920), or ["Literary Chinese" into "Chinese (literary)"](https://github.com/Tatoeba/tatoeba2/issues/750)

There are many cases like these​ ​where it may make sense to rename the language so that it can be found more easily by users. It is however not always obvious.

​There is also the case of a language (Odia) which name given by the SIL is outdated, and we decided to use the more updated name, leaving into parenthesis the outdated one.

In order to make the naming consistent, we need to review the names of all the languages.​

---

Below is the list of languages for which we altered the name.

Chinese (Mandarin)
Norwegian (Bokmål)
Norwegian (Nynorsk)
Punjabi (Eastern)
Punjabi (Western)
Juhuri (Judeo-Tat)
Odia (Oriya)

---

​Below is the list of the languages which names are a combinations of words.

**Which of these languages should we rename?**

Low Saxon
Literary Chinese
Old East Slavic
Central Dusun
Old Prussian
Upper Sorbian
Lower Sorbian
Algerian Arabic
Scottish Gaelic
Egyptian Arabic
North Moluccan Malay
Coastal Kadazan
Crimean Tatar
Ancient Greek
Ottoman Turkish
Old English
Hill Mari
Tok Pisin
Haitian Creole
Guerrero Nahuatl
Nigerian Fulfulde
Fiji Hindi
Lingua Franca Nova
North Levantine Arabic
Pennsylvania German
Northern Sami
Orizaba Nahuatl
Middle English
Middle French
Min Nan Chinese
Southern Altai
Palatine German
Old Norse
Tagal Murut
Old Aramaic
Hakka Chinese
Southern Sotho
Swiss German
Louisiana Creole
Moroccan Arabic
Old Saxon
Seychellois Creole

---

​There may be languages that are not in the list above but which should still be renamed, for reasons similar to Odia.

jiru:
[...]

PR #1590, opened 2018-07-01
Reviewed names of Chinese languages

RyckRichards:
This PR solves part of the issue #936 by standardizing names of Chinese languages

Min Nan Chinese' ► Chinese (Min Nan)
Literary Chinese ► Chinese (Literary)
Hakka Chinese ► Chinese (Hakka)
Xiang Chinese ► Chinese (Xiang)
Gan Chinese ► Chinese (Gan)

@loolmeh already reviewed Arabic languages - https://github.com/Tatoeba/tatoeba2/commit/bbb8fc0d95167cfbd1d0b3421f98cceb053d288a


trang:
@RyckRichards I will not merge this pull request because it gives only a short term solution.

The commit that you reference was from a pull request to change the name of Arabic languages that was never merged. You can check on the website: for instance "Egyptian Arabic" is still there and there's no "Arabic dialect (Egyptian)" in the list.

The proper solution would be to implement a searchable/autocomplete dropdown. We have anyway more and more languages and the traditional dropdown simply doesn't do the job anymore.

Issue #1589, opened 2018-06-29
Segregate sentences created as translations from originals


jiru:
We want to have a way to tell wether a sentence has been created as a translation of another, or as an original sentence (from the /sentences/add page for example).

The license of original sentences is simpler to handle because they are presumably not derived from any other work.

Related issue: #73.

trang:
We had a meeting where we discussed various cases of when is a sentence original and when it is derivative work. I'm documenting the cases here.

We would need confirmation from a legal expert whether or not the assumptions we are making are correct. There may also be other cases we need to consider.

### Case 1
User A creates `Sentence A`.
User B creates `Sentence B` as a translation of `Sentence A`.
→ `Sentence B` is a derivative work of `Sentence A`.

### Case 2
User A creates `Sentence A`.
User B creates `Sentence B`.
User C links `Sentence A` to `Sentence B`.
→ None are derivative work. They are both original sentences.

### Case 3
User A creates `Sentence A`.
User B creates `Sentence B`.
User A links `Sentence A` to `Sentence B`.
→ None are derivative work. They are both original sentences.

### Case 4
User A creates `Sentence A`.
User B creates `Sentence B`.
User B links `Sentence A` to `Sentence B`.
→ None are derivative work. They are both original sentences. Technically, User B could be hiding the fact that `Sentence B` is not an original sentence, but we cannot take responsibility for that.

### Case 5
User A creates `Sentence A`.
User B creates `Sentence B` as a translation of `Sentence A`.
User B unlinks `Sentence B` from `Sentence A` because they realized they made a mistake.
→ `Sentence B` is still a derivative work of `Sentence A`.

### Case 5
User A creates `Sentence A`.
User B creates `Sentence B` as a translation of `Sentence A`.
User C unlinks `Sentence A` from `Sentence B` because they are not correct translations of each other, but each sentence on its own is a correct sentence.
→ `Sentence B` is still a derivative work of `Sentence A`.

jiru:
I am not sure how to deal with sentences that were added at the very beginning of the project. The log table indicates that sentences 1 to 330929 were first inserted without linking information, and then all their links have been added at once, in the order of sentence IDs. I believe that’s because Tatoeba doesn’t use to have the concept of links at the beginning. It looked like the sentences in a given group were all linked to each other.

[...]

Issue #1555, opened 2018-06-03
Show license of a sentence


trang:
On the sentence details page, for people who are not the owner of the sentence, the block "License" will not show a dropdown (#1554) but just the name of the license as a link. Clicking on the link redirects to the page where the license is defined.

<img width="737" alt="show-license" src="https://user-images.githubusercontent.com/221850/40888882-0f9a935e-675e-11e8-91f5-4dfc58347f82.png">

trang:
This is implemented but currently only visible by admins. I'll keep this issue open till we make it visible to everyone.

Issue #1553, opened 2018-06-03
License selection for new sentences


trang:
On the page "Add sentences", in addition to the "Language" and "Sentence" fields, we will need a 3rd field as dropdown where people can select a license.

<img width="517" alt="license-new-sentences" src="https://user-images.githubusercontent.com/221850/40888835-70318d40-675d-11e8-9807-a0ee152030da.png">

jiru:
I wonder if it's a good idea to put a license dropdown here, because:
1. I think most people won't need to switch license when adding new sentences. Most people are likely to stick to a single license for all their sentences.
2. It will confuse people who don't know what a license is. What should I select here? What’s the meaning of CC-BY? That could be especially negative for new contributors.

How about having a text like "Your sentence will be published under the CC-BY license ([learn more](https://tatoeba.org/somewhere))" close to the submit button? The "learn more" link could bring the user to a page that explains licensing, allows to change the default license under which sentences are published etc.

trang:
I wouldn't make early assumptions as for how many people will be switching licenses. Besides, the way we implement the feature will have an impact. If we don't manage to make it intuitive and clear enough, then people will less likely switch between licenses.

But you're correct, people who don't know about licenses and don't know about CC-BY will be confused if they just see a dropdown listing some accronyms they never saw before.


For your suggestion, I think that with only "learn more" some people may not even realize they can choose another license.

So we could have:
Your sentence will be published under the [CC-BY]() license. [[Learn more]()] [[Change license]()]

Or:
Your sentence will be published under the [CC-BY]() license. [[Change license]()]


We can also just keep the dropdown and have a question mark icon next to it. Clicking on this question mark icon would lead to a page explaining the licenses.

jiru:
Thanks for your comment. I eventually kept your initial design, because I’m not sure about what to do, and we can rearrange things later on. This is functional already.

Currently, when adding sentences through other pages (like /vocabulary/add_sentences),  the value of the license is read from the "default_license" user setting, which is yet to be made editable and defaults to 'CC BY 2.0 FR'.

trang:
Tricky situation reported by [Ricardo on the Wall](https://tatoeba.org/eng/wall/show_message/29537#message_29537):
* User A adds `Sentence A` with CC-BY.
* User B later adds `Sentence A` with CC0. 

With our current system, user B wouldn't be able to add their sentence at all and `Sentence A` will not be reusable under CC0.

PR #1623, opened 2018-07-26
Sentence licensing (first iteration)

jiru:
This brings the following changes:
1. Each sentence now has a license, and all the existing sentences are under CC BY 2.0 FR (to match our current ToS).
2. Only for admins:
   1. The page /sentences/add now shows a license dropdown that allows to set the license of sentences added from this page.
   2. Sentence pages (/sentences/show/xxx) now have a license block on the right that shows the license of the sentence. For owners of the sentence, it is shown as a dropdown that does nothing yet. For other people, it is shown as a text.
3. Sentence pages (/sentences/show/xxx) now have a little text in the "Logs" block that says if the sentence was initially created as a translation, or is original, or "unknown". Sentences with id < 330930 are all "unknown" for this first iteration.

Implementation note:
* Each user has a (currently hidden) "default_license" setting with a default value of "CC BY 2.0 FR". This setting is used to set the preselected value of the license dropdown in /sentences/add, and to set the license when adding sentences from other pages (adding a translation, adding sentences to one of the lists you own etc.)

Installation instructions:
1. Run SQL update script: `mysql tatoeba < /docs/database/updates/2018-06-19.sql`
2. Run CakePHP migration script (took about 4h30 on dev.tatoeba.org): `cake sentence_derivation`

Questions:
1. I am not sure where is the best place for the "translation/original" text. It still looks a bit awkward to me there. What do you think?

This PR covers the following issues: #1553, #1555, #1589


trang:
> I am not sure where is the best place for the "translation/original" text. It still looks a bit awkward to me there. What do you think?

The other options I can think of:

1. Somewhere still in the sidebar but in a separate block. We already have many blocks in the sidebar so I'm not too fond of adding yet another one.
2. Next to the sentence id (above the sentence block). We would have to find shorter wordings, or use icons with tooltips.
3. Below the sentence id and above the sentence block. Maybe not an optimal use of vertical space. The information is not *that* important to be above the main subject of the page.
4. Below the sentence block. I'm afraid some people might think that the information refers to the last translation in the sentence block. Maybe not a big deal.
5. In the sentence block itself. It's a quite difficult one, but there's surely a way to fit the information in there. We could then mark the "original" sentence even when we're not on the page of that sentence.

In any case, this is not blocking. I wouldn't mind leaving the info text in the logs for now. We can take a bit more time to ask opinion from the community and have @alanfgh help us out with the final wording.

jiru:
Okay. Note that I’m still working on improving the algorithm used to figure out which sentence is original or translation. It currently works for the vast majority of the sentences of the "link-era", but there are still some edge cases where it is unable to find the answer and results into the "unknown" message.

trang:
I'll move this PR to be merged for next week. Even though I'm very tempted to merge this ASAP, it feels a bit rushed to merge it for today's update.

I'll make a post on the Wall to ask the community to check if they notice anything wrong with the detection of "original" sentences.

jiru:
I’m running an ultimate pass of the sentence_derivation script on dev.tatoeba.org, in should be completed in a few hours. There are still a few sentences (about 14k) marked as "unknown". For the record, reasons include:
1. Lack of creation log
2. Lack of one of the link pair
3. Too much lag between the creation log and the link log. The maximum tolerance is currently set to 310 seconds, but there are still 12 sentences which log shows more lag, including #3050070 (894 seconds) and #6192304 (more than two hours (I suspect a clock skew)).
4. Sentence #5963352 has a very strange log I still don’t understand.

Note that there are cases where it’s nearly impossible to distinguish between:
    (a) a sentence normally added as translation
    (b) a sentence created as original and then quickly manually linked to another sentence of older ID

An example of case (b) is [#1722449](https://dev.tatoeba.org/sentences/show/1722449), quickly manually linked to [#1722448](https://dev.tatoeba.org/sentences/show/1722448) after being created.

In case (b), I think it’s fair enough to consider the sentence as a translation. However, the script also needs to check the order in which the link pair is created (A-B B-A vs. B-A A-B) to avoid other kinds of false positives. So, depending on which way the link was manually added, the sentence will be considered as original or translation.

trang:
I'm gonna go ahead and merge this. It's more than good enough for a first iteration :)


> Sentence #5963352 has a very strange log I still don’t understand.

This sentence was added via the mass import feature. CK was the one who did the import, that's why the logs mentions him as the creator of the sentence.

In the file he imported, I assume there was a line:
```
Ласкаво просимо назад!	Welcome back.
```

The English sentence already existed so only the link was created.

In this case, the Ukrainian sentence is most likely a translation rather than an original sentence. It is owned by deniko, so deniko is the one who sent a file to CK to import a bunch of sentences. CK will very likely refuse to mass import sentences that are not created by native speakers, so this must have been part of a list of English sentences translated into Ukrainian by deniko.

Issue #1556, opened 2018-06-03
Log changes of license


trang:
We'll need to know when the license of a sentence was changed, from what to what.

Issue #1554, opened 2018-06-03
Updating the license of an existing sentence


trang:
On the sentence details page, we will need a new section in the sidebar where the owner of the sentence can edit the license. Admins should be able to edit the license as well.

<img width="737" alt="license-existing-sentence" src="https://user-images.githubusercontent.com/221850/40888849-97b4f5a0-675d-11e8-942a-7996452e762c.png">

It is however not as simple as that because we cannot update from and to any license.

1) The licenses of the translations will depend on the license of the original sentence. If `Sentence A` is under CC-BY, and someone translated it into `Sentence B`, then `Sentence B` cannot be CC0 unless `Sentence A` is changed to CC0.
2) A sentence cannot be moved to a more restrictive license, only to a less restrictive one. If `Sentence A` is initially under CC-BY, it cannot be changed to CC-BY-SA, but it can be changed to CC0.

We will need to think this through a lot before we allow users to change the license of their sentences. Probably this feature should be accessible to all contributors but limited to only some specific users at first.

jiru:
> The licenses of the translations will depend on the license of the original sentence. If Sentence A is under CC-BY, and someone translated it into Sentence B, then Sentence B cannot be CC0 unless Sentence A is changed to CC0.

Does the same apply for audio recordings too? If Sentence A is under CC-BY, and someone recorded it into Recording A, then Recording A cannot be CC0 unless Sentence A is changed to CC0. If this is true, it means we may have a problem for recordings that are currently licensed under Public domain.

> A sentence cannot be moved to a more restrictive license, only to a less restrictive one.

May I ask why?

trang:
> Does the same apply for audio recordings too? If Sentence A is under CC-BY, and someone recorded it into Recording A, then Recording A cannot be CC0 unless Sentence A is changed to CC0. If this is true, it means we may have a problem for recordings that are currently licensed under Public domain.

Yes, the same applies to audio. This is why Common Voice cannot just take our sentences and record them under CC0.

Do we actually have any recording licensed under Public domain?

>> A sentence cannot be moved to a more restrictive license, only to a less restrictive one.
>
> May I ask why?

Well assuming you publish something as public domain today. Then I start reusing your work in my project. One year later you decide to change it to CC-BY-NC. What happens to my project? What if published my work as public domain? It would be conflicting.

On the other other hand, if today you publish something under CC-BY-NC and one year later you decide to change it to public domain, it doesn't impact projects that have been already reusing your work.

[...]

PR #1650, opened 2018-08-16
Allow to change license and log license change

jiru:
This pull request is meant to solve #1554 and #1556. It brings the following:

1. On the right side of sentence pages, admins can now see a "change license" button.
2. When the license of a sentence is changed, a new "license changed" log entry is created. I chose a terrible color because I’m not good at it; suggestions/commits welcome.
3. This pull request also brings specific validation rules regarding license change (see the `Sentence::canSwitchLicense()` method), but those are ignored for admins and corpus maintainers when they use the "change license" button (to allow them to fix potential mess). I plan to use these validation rules for regular users, and for anyone (admin or regular users) using the future "switch all my sentences to CC0" function.
4. On the top of that, currently only admins can see and use the "change license" button.

Installation instructions:

1. Run `docs/database/updates/2018-08-14.sql`
2. Add new ACLs:
`cake acl create aco Sentences edit_license`
`cake acl deny  Group.2 Sentences/edit_license all`
`cake acl deny  Group.3 Sentences/edit_license all`
`cake acl deny  Group.4 Sentences/edit_license all`


trang:
In theory none of the admins will be changing the license yet so I'd say it's not a huge deal to display these logs. But ideally yes, we should hide them otherwise if users start to notice logs saying the license of a sentence changed to CC0, they will be a bit confused.

PR #1654, opened 2018-08-25
Changed flag for Kashmiri

RyckRichards:


trang:
For the record this change was requested by sabretou and in agreement with haziqmir.

Issue #1660, opened 2018-09-02
Log initial license


trang:
We're logging changes of license (#1556) but we also need to log the license when the sentence is added. Otherwise we won't know what was the initial license anymore.

jiru:
Note that with the current implementation, it's not possible to switch to the same license, and there are only two licenses, so it's possible to infer the initial license from the log changes and the current license.

Issue #1659, opened 2018-09-01
Some sentences are correct but shown in red


trang:
## Problem

As mentioneed in https://github.com/Tatoeba/tatoeba2/pull/1650#issuecomment-417822863
> We have many sentences that are marked as "unapproved" only because of possible copyright issues. Having them displayed in red with a warning sign may give the wrong idea to regular users, who'd think there's something incorrect in the sentence, but there isn't.
>
> The reason why they're marked as "unapproved" is because unapproved sentences are excluded from the CSV exports. This is a workaround rather than an actual solution. 

Some of these sentences can be found by looking at sentences that have a tag containing "copyright":
https://tatoeba.org/eng/tags/view_all/copyright

## Suggested solution

1. Allow admins to change the license from CC-BY to `null`. Not sure what label to put for license `null` though. No license? Unknown license? Tatoeba only?
2. Exclude sentences that are not CC-BY from the CSV exports.
3. Prevent users from translating these sentences, at least until we figure out how to handle the license of translations.
4. Display the license info to everyone, not just to admins (we can then close #1555).

jiru:
I think it’s a good solution, because people looking for the license of the sentence should be warned about licensing issues, and because it’s simple.

> Allow admins to change the license from CC-BY to null. Not sure what label to put for license null though. No license? Unknown license? Tatoeba only?

How about "Licensing issue"? Another question: any particular reason for using `null` instead of a regular value?

trang:
> How about "Licensing issue"?

That works too.

> Another question: any particular reason for using null instead of a regular value?

No particular reason. I guess it just feels more accurate to use `null` because to me regular values to should refer to actual licenses, so I naturally fallback to `null` whenever to express "stuff that's not defined".

sacredceltic:
I can't see why users should be prevented from translating these sentences. The license issue is with the sentence itself, not its potential translations, which could well be original and genuine...

jiru:
@sacredceltic A translation is a derivative work. As such, if the license of the original work doesn’t allow publishing a translation under CC-BY compatible terms, one cannot legally add a translation of it on Tatoeba.

For example, let’s say a given sentence is part of the lyrics of a song. It’s not legal to publish any unauthorized translation of the lyrics, because the song author owns the exclusive right to make derivatives from his work. So not only the sentence, as part of lyrics, cannot be published on Tatoeba because the corpus is licensed under CC-BY, but also its translations are infringing the copyright of the original lyrics.

[...]

PR #1666, opened 2018-09-05
Log initial license

trang:
This PR addresses issue #1660.

I'm not sure if we should met it yet because we will start creating a lot of logs of type license even though we're not officially handling other licenses than CC-BY yet.

We may want to wait till we make it somewhat official, or till we implement #1659.


jiru:
> I'm not sure if we should met it yet because we will start creating a lot of logs of type license even though we're not officially handling other licenses than CC-BY yet.

 Aren’t the logs hidden for non-admins anyway?

Issue #1160, opened 2016-05-22
Link to "Tools" page


trang:
When we redesigned the homepage, we dropped the link to the "Tools" page: https://github.com/Tatoeba/tatoeba2/issues/947#issuecomment-192709166.

The page still exists though (https://tatoeba.org/eng/tools/index) and we're not really planning to delete these features, since they are still of use for some people. Someone did point out to me that the Tools "disappeared". It would therefore be nice to find a place where we could add back this "Tools" link.

I currently have no idea where.

halfdan:
How about in the bottom menu "Developers"? It's not necessarily developer content - but the tools are rather raw in their UI.

trang:
Closing this.

We will never be able to invest efforts into these tools. Even if we did, we should implement them independently from Tatoeba.

The Tools page shall remain hidden for now, and one day shall be completely removed.

PR #1733, opened 2019-01-16
CakePHP 3.6

trang:
I think we're finally ready to merge the cakephp-3 branch. This pull request is just for documentation as I will merge it myself.

There are a few features that we're not migrating yet. These features will have to wait until we make sure that the website is stable enough:
- Mass import of sentences
- Import of audio
- Japanese indices
- Transcriptions

There are some features that probably won't be migrated ever: basically everything that is in the "Tools" section, which is a section that has been hidden for a couple of years now (#1160).

This is what I've roughly tested:
https://github.com/trang/tatoeba2/wiki/Tatoeba-tests-for-migration-to-CakePHP-3

This is the Wall thread where Tatoeba members reported issues they encoutered while testing:
https://tatoeba.org/eng/wall/show_message/30964#message_30964


jiru:
It's amazing to see this branch merged already! Congratulations for all the hard work Trang, and good luck with what's remaining.

trang:
Thanks @jiru :)

Issue #1789, opened 2019-02-19
New flag for Uyghur language


BaksiLi:
As posted by Leolaaziano on the wall:

> Recently, I have noticed that the language flag for Uyghur language is ![this](https://tatoeba.org/img/flags/uig.png?1549009988 ), which seemed truly terrifying to me. This symbol was the proposed flag for 'East Turkestan', which was a terrorist proto-state that not a country on Earth had ever recognised. It is NOT the flag of Xinjiang Province where most of the Uyghur people lived, nor a globally recognisable flag for the nation/language. This would have the same implication from using the flag of ISIS for the Arabic language.
> 
> I am afraid that using this symbol would hurt the feeling of many Chinese people, as well as Uyghur people who always wish peace on their homeland.
> 
> Therefore, I strongly advice you to change the flag, and if possible, to flags like <img src="https://upload.wikimedia.org/wikipedia/commons/8/84/Uyghurche.png" width="20" height="30">. It has more natural meanings and, more importantly, not biased.

https://tatoeba.org/eng/wall/show_message/31358#message_31358

I have made a new one like this:

![uig](https://user-images.githubusercontent.com/26314680/53018196-a363e280-3449-11e9-82b0-8e34a75c7029.png)

the font used is _Thabit_, which is free.

RyckRichards:
@cueyayotl what do you think about this? Can we go ahead and merge it?

soliloquist-tatoeba:
I don't think you would find any Uyghurs agreeing the change of the current flag, but I accept that some Chinese people may find it a threat to their 'national unity'. While Muslims are the easiest target in the current zeitgeist, they might want to change the Tibetan flag used on Tatoeba, too, if possible. It's also banned in China according to Wikipedia. 

https://en.wikipedia.org/wiki/Flag_of_Tibet

Btw, the Uyghur flag was adopted in 1933. Linking it with the ISIS flag is pure propaganda.

Anyway, If the Uyghur flag is going to be changed, I suggest using 'UIG' (the ISO code for Uyghur) on a blue background. It's more universal and not all Uyghurs use the Arabic script.

https://prnt.sc/mnc7ea

Also, if the Uyghur flag is replaced and the Kabyle (MAK) flag stays, that will be just one big double standard.

LiSongcheng:
Mr @soliloquist-tatoeba says:
[...]

PR #1791, opened 2019-02-19
Update Uyghur flag

BaksiLi:
Reasons given https://tatoeba.org/eng/wall/show_message/31358#message_31358.

#1789


RyckRichards:
@cueyayotl what do you think about this? Can we go ahead and merge it?

BaksiLi:
> @cueyayotl what do you think about this? Can we go ahead and merge it?

Hi mate, any replies so far? Best regards.

PR #1806, opened 2019-03-03
Allow flags to be SVG

linkmauve:
I embedded every PNG flag into a SVG file, so that they can be loaded with a normalised URL. I also replaced the French flag with a proper SVG, to demo how it will work once every flag has been converted.

There is also a cleanup of trailing whitespaces, as well as the removal of an unneeded executable bit on a file.


Ppjet6:
LGTM.

sacredceltic:
Nothing is clearer...

Le dim. 3 mars 2019 à 19:43, Maxime Buquet <notifications@github.com> a
écrit :

Ppjet6:
"Looks Good To Me".

sacredceltic:
Ben voyons...et en Mandarin, ça donne quoi ?

Le dim. 3 mars 2019 à 20:42, Maxime Buquet <notifications@github.com> a
écrit :

linkmauve:
Pull requests are traditionally done in English, not in Mandarin nor French (and by the way language names are written without a capital letter in French), but this is quite off-topic for this pull request.

sacredceltic:
And English is normally not shortened in a geeky snob way to be understood
worldwide 😉

Ppjet6:
Seriously, can we stop the useless comments already. Thanks

sacredceltic:
So according to you, to be understood by a majority of people is useless,
if i understand you well ? With this view, i think you should have nothing
to do with a site whose aim is translation...

Le dim. 3 mars 2019 à 22:30, Maxime Buquet <notifications@github.com> a
écrit :

jiru:
@linkmauve Thanks for the PR. I think it’s not ready for merge yet though.

1. The PNG files are still referenced in other parts of the code, like LanguagesHelper.php. Look for the string "flags/".
2. Most of our existing flags include a shadow (right and bottom border) that won’t show with your proposed French SVG flag.
3. The colours or your French SVG flag are too strong. We use light colours for all our flags.
4. The "Remove trailing whitespace in all source files" commit is likely to make other PRs and ongoing local work conflict and require a lot a manual merge. I think it would be better to do this when we’re sure nobody else is working on something. Let’s agree on a date and time so that it won’t make a mess.

I’d also like to discuss about whether or not it’s good idea to switch to SVG.

SVG is technically superior to PNG, but I doubt that we’re going to ever replace all the (currently 336) flags with an SVG equivalent, because it’s just too much work. SVG look better compared to PNG is when scaling the image up. As a result, if we are to scale up the flags for any reason, we’re going to have a mix of flags that scale-up nicely and badly. Some flags will "look nice" next to some others that will "look bad", which is a kind of discrimination between languages.

I also have concerns that it can be difficult to make an SVG image "look nice" at 30×20 pixels when there are lots of details. It’s like trying to use a vector font to render complex Chinese characters at very small size. A bitmap font can look nicer because the down-scaling is handmade.

Last and not least: what do the people responsible of adding new flags (@cueyayotl and @RyckRichards) think? Will they add new flags as SVG from now on?

linkmauve:
On Mon, Mar 04, 2019 at 05:05:39AM -0800, gillux wrote:
> @linkmauve Thanks for the PR. I think it’s not ready for merge yet though.

Sure, I didn’t expect it to be, that’s more of an RFC. :)

> 
> 1. The PNG files are still referenced in other parts of the code, like LanguagesHelper.php. Look for the string "flags/".

I just pushed a fix for that.

> 2. Most of our existing flags include a shadow (right and bottom border) that won’t show with your proposed French SVG flag.
> 3. The colours or your French SVG flag are too strong. We use light colours for all our flags.

For both of these points, I was expecting to apply a filter on the image
when rendering it, to avoid hardcoding the transform.  We can do it
either in the SVG or in the CSS, depending on how we want to proceed.

> 4. The "Remove trailing whitespace in all source files" commit is likely to make other PRs and ongoing local work conflict and require a lot a manual merge. I think it would be better to do this when we’re sure nobody else is working on something. Let’s agree on a date and time so that it won’t make a mess.

I removed this commit, we can do it later.

> 
> I’d also like to discuss about whether or not it’s good idea to switch to SVG.
> 
> SVG is technically superior to PNG, but I doubt that we’re going to ever replace all the (currently 336) flags with an SVG equivalent, because it’s just too much work. SVG look better compared to PNG is when scaling the image up. As a result, if we are to scale up the flags for any reason, we’re going to have a mix of flags that scale-up nicely and badly. Some flags will "look nice" next to some others that will "look bad", which is a kind of discrimination between languages.

The thing that prompted me to make this PR is that the website already
gets scaled for various reasons, be it because the user has a high
definition screen (where logical pixels don’t match with physical
pixels), or even when the user configured their browser to render it
bigger.

I’d expect flags to be trivial to import from Wikimedia Commons, they
should already have most of them.  For the few that have been made up
for Tatoeba we can probably recreate them in SVG quite easily.

> 
> I also have concerns that it can be difficult to make an SVG image "look nice" at 30×20 pixels when there are lots of details. It’s like trying to use a vector font to render complex Chinese characters at very small size. A bitmap font can look nicer because the down-scaling is handmade.

We can decide whether to simplify complex shapes (there are automated
tools for that), but I’d suggest to first determine whether this is
useful or not, after doing a full import from Commons.

> 
> Last and not least: what do the people responsible of adding new flags (@cueyayotl and @RyckRichards) think? Will they add new flags as SVG from now on?

I’m interested too. :)

-- 
Emmanuel Gil Peyrot

trang:
I have already mentioned the idea of using SVG to @RyckRichards a long time ago, and more recently to @sabretou (who joined the language team some months ago to help improving the language icons and has actually done some good work that has not been pushed yet).

They have nothing against the idea itself, but there is a technical barrier. Someone will have to assist and train them on how to migrate our self-made icons to SVG and to create new SVG icons. I would not expect them to figure this out completely on their own.

The fact that some icons may look bad next to some other icons should not stop us from pushing forward the migration to SVG, I think. Even now we have language icons that are are looking less appealing next others because we did not have the design skills or the time to make them better. It's okay if some icons get a beauty upgrade sooner than others. We cannot always introduce improvements to all icons at once, it's more of an ongoing work.

I haven't tested this PR yet but in the principles I'm okay with it. The impacts are not too huge.
* For @sabretou it doesn't introduce any immediate changes. He can still create icons in PNG until he gets more confortable with SVG.
* For @RyckRichards it complexifies a bit the process of creating PR's for new languages. But nothing insurmountable.

linkmauve:
I can provide this technical assistance if you want, probably not this week as I’m going to Brussels, but we can schedule a time for it.  I’m also almost always in the chat room, in European timezone.

jiru:
> We can decide whether to simplify complex shapes (there are automated
tools for that), but I’d suggest to first determine whether this is
useful or not, after doing a full import from Commons.

Yes, it needs some experimentation. Just to make the problem clearer [this is an example of what I’m talking about](https://graphicdesign.stackexchange.com/questions/48991/svg-looking-bad-in-small-resolution). I think it’s not only about the complexity but also the patterns, like I’m pretty sure the repeated vertical lines in the Ancient Greek flag will look terrible.

To make matters worse, the downscaling algorithm is implementation-dependant, so small flags might look different depending on the browser or system. I’m a little bit concerned about the increasing complexity of just displaying images.

trang:
I wrote to @sabretou and @RyckRichards about this PR and told them they're free to share their concerns.

@RyckRichards raised a reasonable concern about speed: is it going to make Tatoeba slower? I don't think it is something to worry about, but is it? I don't think I'm expert enough to confirm this.

He was not worried about the new things he has to learn to do, but we would have to see how it goes in practice.

jiru:
It’s certainly slower to render SVG, but I don’t think the performance decrease is gonna be noticeable.

I’m in favour of merging this PR as soon as possible even though only the French flag is converted to SVG, so that we’re not blocking the language requests for months. After all, it’s just about *allowing* the use of SVG.

The last blocking thing I’m seeing is the lightening of colours of the French flag. @linkmauve I think you can add a new CSS class with lightening rules in `elements.css`, and add that class in the French SVG.

@linkmauve Also could you commit a script to convert PNG flags into SVG the way you did for existing flags? You can put it in the `tools` directory.

trang:
I think it's possible to apply CSS on an SVG only if the SVG is inline. Here the SVG is referenced in an `` so defining the effects in the CSS then adding the class in the SVG won't work.

For this PR I'd suggest reverting the French flag back to the embedded PNG. 

We'll later update all the images at once to remove the lightening and shadow effects from the image itself and define the effects in CSS instead. We already had plans to do that on the PNG images, @sabretou already created all the flags in their "original color".

jiru:
> We already had plans to do that on the PNG images, @sabretou already created all the flags in their "original color".

Good to know!

trang:
@RyckRichards is now familiar with creating SVG's out of PNG's so I'm going to merge this and revert the French flag back to fake SVG. This way we won't block further the pending language requests.

@linkmauve If you have a script for the conversion of PNG to SVG, you can either make a new pull request later on, or just copy-paste it somewhere and I'll add it in the `/tools`.

Issue #1634, opened 2018-08-05
New flags for Ancient Greek and Latin


Guybrush88:
As posted by Alexmarcelo on the wall:

> 
> Dear community,
> 
> Ricardo and I have been discussing the development of new flags for both Latin and Ancient Greek. The current Latin flag is just not as good as the others on Tatoeba, and the current Ancient Greek flag has little or nothing to do with Ancient Greece at all.
> 
> We'd like to know which flag you like best.
> 
> Below is a brief description of each candidate, followed by a preview.
> 
> Latin
> (1) Current flag.
> (2) Contains the SPQR (Senatus Populusque Romanus) inscription, the official name of the Roman Empire.
> (3) Contains a laurel wreath, a symbol of victory for the ancient Romans.
> (4) Contains the Colosseum, a symbol of Ancient Rome.
> 
> Check the Latin flags here: https://ibb.co/hzdaJz
> 
> Greek
> (1) Current flag.
> (2) Contains the front view of the Parthenon, as it once looked.
> (3) Contains the front view of the Parthenon, as it looks today.
> 
> Check the Greek flags here: https://ibb.co/bxWvJz
> 
> The colors are not final.
> 
> You can also propose new suggestions if you like.
> 
> Thanks!

https://tatoeba.org/ita/wall/show_message/29541#message_29541

RyckRichards:
As far as nobody else has given any ideas or was opposite to the change, I'm taking it and I'm going to use # 2 flags as attached in this message:

![gre](https://user-images.githubusercontent.com/7658430/43933615-064ab17e-9c21-11e8-8026-0c146a87081e.png)
![lat](https://user-images.githubusercontent.com/7658430/43933616-0672775e-9c21-11e8-944d-a0e5532a3209.png)

trang:
@RyckRichards If you could rework a little bit the Latin flag to make the letters easier to read, that'd be great. I know there's very little space, but I'm sure you'll find a way :)

As cueyayotl reported:
> An important note, make sure that the inscription (SPQR) of the Latin flag is legible. I'm not sure if it was the contrast of colors or just the size of the letters, but I could not make them out as clearly as in the flag we already have for Latin.

Just for comparison, this is the current icon:
![lat](https://user-images.githubusercontent.com/221850/43991571-2fea2474-9d70-11e8-81b9-03368ab7c472.png)

RyckRichards:
Perhaps we'd have to use this flag instead

[...]

PR #1810, opened 2019-03-06
Changed the latin flag since the actual one is kinda cropped

RyckRichards:
In #1634, we've changed the latin flag but according to some users, it's kinda cropped. Hope this one works now.

@jiru - could you please check if there's a way to optimize this flag?


trang:
@RyckRichards This will need to wait for a decision on @linkmauve's PR (#1806). 

Also it would be a good idea to make one pull request not just for Latin but for all the language icons that @sabretou updated. Unless there is a reason for Latin to be updated separately, I'd suggest to close this pull request and make a new one with all the other languages (but not right away, only after we're done with #1806).

RyckRichards:
Agreed

-- 



Ricardo Vernaut Junior

PR #1825, opened 2019-03-14
Remove ng-cloak from <body>

jiru:
We currently put `ng-cloak` on the `<body>` tag in order to hide unstyled elements during page load. As a result, pages appear as blank while they are loading. This gives a feeling of general slowness of the website.

This PR is an attempt to make pages load faster by:
* removing `ng-cloak` from `<body>`
* dispatching it on specific parts of the pages that really needs to be hidden during load
* providing the hidden parts with basic CSS positioning styles like min-height so that the content is not jumping around when parts hidden with `ng-cloak` get displayed

On most of the pages that have been converted to AngularJS (like /register), everything after the title is hidden during load. On the main page (/), the welcome message is displayed during loading. On pages that are not converted to AngularJS yet (like /sentences/show/<n>), the content is displayed as soon as the page is downloaded.

Another indirect benefit is that CakePHP errors are not hidden by ng-cloak any more.

### DIFF
--- a/src/View/Helper/PrivateMessagesHelper.php
+++ b/src/View/Helper/PrivateMessagesHelper.php
@@ -117,7 +117,7 @@ class PrivateMessagesHelper extends AppHelper
             ]);
             ?>
             </div>
-            <div layout="row" layout-align="end center" layout-padding>
+            <div ng-cloak layout="row" layout-align="end center" layout-padding>
                 <md-button type="submit" name="submitType" value="saveDraft" class="md-raised">
                     <?php echo __('Save as draft'); ?>
                 </md-button>
--- a/src/View/Helper/PrivateMessagesHelper.php
+++ b/src/View/Helper/PrivateMessagesHelper.php
@@ -79,14 +79,12 @@ class PrivateMessagesHelper extends AppHelper
 
         <div class="body">
             <div class="pmFields">
-            <div ng-hide="true">
             <?php
             if (!$isReply) {
-                echo $this->Form->input('messageId', array('value' => $pm->id));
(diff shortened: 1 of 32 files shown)


src/View/Helper/PrivateMessagesHelper.php
```diff
@@ -79,14 +79,12 @@ public function displayForm($pm, $recipients) {
...
         <div class="body">
             <div class="pmFields">
-            <div ng-hide="true">
             <?php
             if (!$isReply) {
-                echo $this->Form->input('messageId', array('value' => $pm->id));
+                echo $this->Form->hidden('messageId', array('value' => $pm->id));
             }
-            echo $this->Form->input('submitType', array('value' => ''));
+            echo $this->Form->hidden('submitType', array('value' => ''));
```
trang: Having the inputs as hidden lead to an error when saving a draft:

> Tampered field 'submitType' in POST data (expected value '' but found 'saveDraft') Cake\Controller\Exception\AuthSecurityException

trang:
Besides the error with saving draft private messages, it looks good to me. I'm moving this to the next milestone to test a bit more.

One note on the user experience: instead of having the feeling that the website is slow, it actually gives me the feeling that the website is broken for a short moment, because I've been used to seeing everything appear at once. I suspect as a result that some users will also feel as if something is wrong at first, but will probably get used to it after some time.

src/View/Helper/PrivateMessagesHelper.php
jiru: I solved the problem.

Issue #1823, opened 2019-03-14
Preview affected sentences on the CC0 switching page


jiru:
The current CC0 switch page is working but it doesn’t allow to preview which sentences will be selected. This doesn’t make the process very trustful. My idea is to:

1. Display a paginated list of affected sentences.
2. Add a button to refresh the list.
3. Change the switching algorithm so that it filters out sentences that are not from that list, on the top of the current selection.

We may just put the previewed sentences into an unlisted private list and update the list as the sentences are being switched.

PR #1848, opened 2019-03-29
Preview affected sentences on the CC0 switching page

jiru:
This PR is an implementation of #1823.

Both the creation of the list and the license switch are performed by background jobs, so make sure to run `bin/cake Queue.Queue runworker` while testing. Note that the list can be manually edited between the two steps, as a primitive way of filtering out sentences.

Caveats:
* The list of affected sentences is displayed as such outside of the switch_my_sentences page.
* The list id is stored as a hidden user setting, which is not what it is meant for.

### DIFF
--- a/src/Controller/LicensingController.php
+++ b/src/Controller/LicensingController.php
@@ -85,22 +85,20 @@ class LicensingController extends AppController {
 
         $licensing = new Licensing();
         $isRefreshing = $licensing->is_refreshing($currentUserId);
+        $isSwitching = $licensing->is_switching($currentUserId);
         if ($this->request->is('post')) {
-            if ($currentJob) {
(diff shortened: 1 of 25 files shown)


jiru:
@trang Do you think we can merge this before the next update?

trang:
@jiru Yes, this can be merged. Sorry for not reviewing it earlier! I only noticed the merge request last week and then didn't have time to test it till now.

There is just a minor issue that I noticed: there's a closing `</div>` missing somewhere. When the message "The license switch of your sentences is in progress..." is displayed, the footer content is misplaced. I can fix that myself after merging though.

jiru:
Well spot! I fixed the closing div.

PR #1849, opened 2019-03-31
Updated uig lang flag

BaksiLi:
This is a new submission after a few discussion here, https://github.com/Tatoeba/tatoeba2/issues/1789


BaksiLi:
@RyckRichards mate,
What should I do to pass the check? It seems that I have changed only the .svg file.

jiru:
@BaksiLi The check fails because the SVG file is too big.

I recently introduced file size checks after I realized that many of the (old) PNG files were very poorly compressed (see 4b8a2c89556c3e18ba5e6a411730c56db2fa4e4e and 4b8a2c89556c3e18ba5e6a411730c56db2fa4e4e). After compressing the PNG files, I set the limit to 2KB because all the PNG files were under 2KB.

After we switched to SVG, we raised the limit to 3KB because it takes a little bit more space to include a PNG image inside SVG. However the 3KB limit is pretty arbitrary since we don’t have much experience with pure SVG files yet.

Your file is 32KB, which may be reasonable giving it’s not compressed. However, would you mind checking if there is anything you can do to reduce the size, like reducing the details of the curves?

Note that the failing check doesn’t prevent your pull request from being merged. We can merge it and adjust the size limit later. :smile:

BaksiLi:
@jiru I have made another svg with only 2KB and committed. Can you please check it again? Cheers

trang:
@BaksiLi Your PR is technically okay. The main blocker is that we have to decide whether or not we should change the flag to what you are suggesting. And for that, I think we'll first need to hear from the Uighur speakers. Until then, your PR is most likely going to remain on standby.

BaksiLi:
@trang No worries. We should wait until most of us are respected. Best wishes.

mozillakab:
I would suggest a mixed flag: 1/2 Algerian and 1/2 Kabyle

mix

trang:
@BaksiLi Our decision as of now is to keep the flag as it is. See https://github.com/Tatoeba/tatoeba2/issues/1789#issuecomment-634268557. Thank you nonetheless for your suggestion.

Issue #1851, opened 2019-04-01
Dhivehi (div)


sabretou:
* Language name: Dhivehi
* Language info: https://en.wikipedia.org/wiki/Maldivian_language
* ISO 639-3 code: div
* Public List: https://tatoeba.org/eng/sentences_lists/show/8830
* Icon: https://i.imgur.com/XKA1rVR.png
* Requested by: https://tatoeba.org/eng/user/profile/evenxzero

[for tatoeba.org]
CALL add_new_language('div', 8830);

jiru:
Note: this language uses the [Thaana script](https://en.wikipedia.org/wiki/Thaana). We need to add it to the Sphinx charset table.

RyckRichards:
Is there a reason for this language not being implemented? :/

trang:
@RyckRichards It requires to update the sphinx config. See comment: https://github.com/Tatoeba/tatoeba2/pull/1855#issuecomment-480477645

Although it's not a blocking issue to merge the request itself. But without it, the sentences in this language would not be searchable.

PR #1855, opened 2019-04-03
Added language Dhivehi

RyckRichards:
This PR solves #1851


jiru:
Thanks @RyckRichards! You also need to put the language name in `src/Lib/LanguagesLib.php`.

RyckRichards:
Thanks, @jiru . I've fixed it.

jiru:
Thank you, that’s perfect! One last thing.

As as pointed out in the issue, we also need to add the Thaana script to the search engine charset table. It means we need to tell the search engine what characters words consist of. Each character is identified by a [code point](https://en.wikipedia.org/wiki/Code_point) that looks like U+XXXX. For example, [character `ç` has a code point of U+00E7](https://www.fileformat.info/info/unicode/char/00e7/index.htm) (or U+E7 for short).

In the case of Thaana, you can easily find its code point range on the [Unicode block](https://en.wikipedia.org/wiki/Unicode_block) page. That table shows Thaana characters are in the block U+0780 and U+07BF.

So you can add `'U+0780..U+07BF'` somewhere in the [$charsetTable array of src/Shell/SphinxConfShell.php](https://github.com/Tatoeba/tatoeba2/blob/dev/src/Shell/SphinxConfShell.php#L49-L178) (if the words are separated by spaces, like in English) or [$scriptsWithoutWordBoundaries](https://github.com/Tatoeba/tatoeba2/blob/dev/src/Shell/SphinxConfShell.php#L180-L206) (if the words are glued together, like in Chinese).

Now if you want to do things even better, you can try not to add the whole range, but only the characters that are not punctuation. If you click on the Thaana link of the Unicode block page, it show you the [details of the Thaana block](https://en.wikipedia.org/wiki/Thaana_(Unicode_block)). As you can see, the cells in gray are not used. Now if you click on the [Official Unicode Consortium code chart link](https://www.unicode.org/charts/PDF/U0780.pdf), you can see even more details about the characters, like which ones are punctuation.

It looks like there is no punctuation, so we can just use the whole range `'U+0780..U+07B1'`, but if, say, character U+079C was a punctuation mark, you can exclude it by using `'U+0780..U+079B', 'U+079D..U+07B1'` instead.

For more information about the search engine charset_table, you can look at [Manticore’s documentation](https://docs.manticoresearch.com/latest/html/conf_options_reference/index_configuration_options.html#charset-table).

Issue #1609, opened 2018-07-17
Allow more export options on the download page


jiru:
Currently, we export data to CSV that are created daily using cron jobs and served as static files. This method prevents us from offering custom dumps, like options to filter the exported data. As a result, people are forced to download large dumps presumably containing a lot of unnecessary data, or they just don’t have access to the data they want because we don’t export it.

We can overcome this limitation using the following method:
* Generate exports on demand by creating CSV files in the background using [this plugin](https://github.com/dereuromark/cakephp-queue/tree/2.x).
* Make these exports showing up on the download page when they are done.

Since the CSV files would be still served statically, downloading can still be stopped and resumed for as long as we keep the file served. If we ever want to have better control over the download, we can serve the files using [x-accel](https://www.nginx.com/resources/wiki/start/topics/examples/x-accel/).

If we manage to implement this, it will ease or affect the implementation of #1788, #1687, #1557, #1515, #1358, #1283, #1168, #927, #768, #351, #283.

trang:
Another related issue: #1788

trang:
I will start reporting in this issue cases of people in need of a more customized subset of the corpus.

The latest question we received on this topic was if it's possible to download only sentences from native speakers in a certain language. The person who contacted us had some programming skills and ended up writing a Python script for this purpose. They have agreed that I publish the script. Might be useful for other people who wish to do the same.

https://gist.github.com/trang/f1a3ff4f6d00eab7171125cd857b8148

Issue #927, opened 2015-11-16
Increase the limit of sentences downloaded from a list


RyckRichards:
By [PaulP](https://tatoeba.org/eng/user/profile/PaulP) - https://tatoeba.org/eng/wall/show_message/24895#message_24895

> The download feature has been disabled for this list because it contains X sentences. Only lists containing 100 or fewer sentences can be downloaded. If you can edit the list, you may want to split it into multiple lists.
> 
> Is there an easy way to do this, other than deleting the sentences one by one?

RyckRichards:
+1
It would help a lot - even more students.

ckjpn:
If the main reason this is limited to 100 sentences is to lower the load, then perhaps you could allow downloading the whole list if it's just the sentences without linked translations, but only limit the download if it requires linking to another language.

Also, perhaps create the option to allow people to download the complete list as numbers only without translations.  I have found there are times I wish I could do this rather than wait for the weekly exports to get this information.  Maybe this wouldn't put too much load on the system.

ckjpn:
Another possibility is to allow downloading the complete list incrementally, sort of like you get it if you browse the pages.

Download the first 100.
Download the second 100.  (201-300)
Download the third 100. (301-400)
etc.

RyckRichards:
Paul says that he use the lists he create to study. After creating a list, he exports them to his phone which he can easily study and as far that it's only possible to download lists which have 100 entries, he have to chop them into many other lists which demand too much time and effort.

trang:
How many sentences does Paul have in his lists?

RyckRichards:
3-5,000 of sentences .

jiru:
For the record, I think it’s possible increase the limit tremendously without significantly increasing the server load, by using the following approach:
* In the controller method:
 * Increase the PHP script timeout
 * Get sentences using successive paged queries (to avoid filling up RAM)
 * Output the CSV already, instead of passing data to the view (necessary because of paged results)

PR #1887, opened 2019-05-10
List export

jiru:
This PR is a first step towards solving #1609 and #927. It creates a page `/exports/index` that allows to export any public or personal list into a CSV file asynchronously, using queued jobs. Once an export is queued, the list of exports updates itself every five seconds. A download button will eventually appear once the export is ready.

See the modifications to `app.php.template` for instructions to configure the export path and Nginx. You also need to run `cake migrations migrate` to create the new `exports` table.

Since this PR only aims to make the export feature available for beta testing, it dosn’t not replace the existing "download this list" functionality on the list pages yet.


jiru:
I’ve set it up on https://dev.tatoeba.org/.

@trang I also made the necessary configuration on tatoeba.org, so that you won’t have anything special to do when this goes in production.

trang:
Thank you @jiru!

Seems like this X-Accel-redirect thing is not a feature in Apache so I wasn't really able to test it locally. But I tested on the dev website and it works pretty well.

Adding the sentence ID (as suggested by Guybrush) would be nice to have in this first iteration. Besides of the use case he described, I think it would be useful for developers. They would be able to more easily process it to get additional information (for instance if they wanted to know which sentences in the list are owned by a native speaker).

It's not a blocker though. There's of course plenty of improvements we could do but that can be done later. This PR can already be merged as is for this weekend, I think.

[Note]
Related Wall thread: https://tatoeba.org/eng/wall/show_message/31855#message_31855

PR #1899, opened 2019-05-31
Fix localization bug for audio icon tooltip

AndiPersti:
The functions `__()` and `format()` are called in the wrong order. Thus the text of the tooltip for the audio icon when an audio file is available ("Play audio recorded by ...") cannot be translated.


jiru:
Thank you for your contribution, AndiPersti! :smiley:

Issue #1892, opened 2019-05-21
Simplify ACL


jiru:
ACL are hard to maintain and cause us regular problems. The most annoying thing is that ACLs are stored in database, which means we need to keep the equivalent fixtures updated.

Consider this [alternative](https://github.com/scherersoftware/cake-auth-actions).

PR #1903, opened 2019-06-13
Simplify authorization system

jiru:
This PR solves #1892.

* `User.group_id` is replaced by an enum `User.role`, which possible values are defined as constants in `src/Model/Entity/User.php`.
* Accesses are now defined in `config/auth_actions.php` instead of ACL database tables.
* Updating tatoeba.org should be transparent for logged-in users.
* dev.tatoeba.org is already migrated.

Installation instructions (local):
* `composer update`
* `cake migrations migrate`
* `cake asset_compress build -f`
* `rm tmp/cache/models/myapp_cake_model_default_users`


jiru:
@trang Note that the table `languages` still has fields like `group_1`, `group_2` etc. In `src/Model/Table/LanguagesTable.php`, the methods `decrementCountForGroup($langCode, $groupId)` and `incrementCountForGroup($langCode, $groupId)` expect a group id as parameter, but it looks like they are not used anywhere. How are these stats supposed to be maintained?

trang:
Right, this is one of the things that got left behind during the CakePHP migration. Since then the numbers have only been updated whenever a new language is added, because I would run the [update_language_stats.sql](https://github.com/Tatoeba/tatoeba2/blob/dev/docs/database/scripts/update_languages_stats.sql) script.

The fields `group_1`, `group_2` etc are supposed to be updated when a user adds a language in their profile, or when a user is being moved to another group. They basically correspond to the numbers on this page: https://tatoeba.org/eng/stats/native_speakers.

* `group_1` is for admins.
* `group_2` is for corpus maintainers.
* etc...

jiru:
Alright I updated the SQL script.

trang:
These scripts also need to be updated I think:
- https://github.com/Tatoeba/tatoeba2/blob/dev/docs/database/tables/users.sql
- https://github.com/Tatoeba/tatoeba2/blob/dev/docs/database/import/users.sql
- https://github.com/Tatoeba/tatoeba2/blob/dev/docs/database/scripts/weekly_exports.sql

And these can be deleted, I guess:
- https://github.com/Tatoeba/tatoeba2/blob/dev/docs/database/tables/groups.sql
- https://github.com/Tatoeba/tatoeba2/blob/dev/docs/database/import/groups.sql

Other than that, everything seem fine so far in my tests. I'll need a few more days to check so I can't merge for this weekend, but for the next one it should be fine.

jiru:
Thanks for the heads-up. I havn’t deleted/updated the database initialization scripts because for now Imouto is applying the new PHP migrations scripts on the top of them. For example, the migration that comes with this PR expects a `groups` table to delete. In my opinion, with the new PHP migration scripts, it’s easier to keep db init scripts unchanged and just stack new migrations on the top of them, instead of changing the db init scripts every time we update the schema. It’s less maintenance burden.

If I understand correctly, we should also convert docs/database/tables/*.sql into a [schema](https://book.cakephp.org/3.0/en/orm/schema-system.html). As for docs/database/import/users.sql, it should be converted into a [seed](https://book.cakephp.org/3.0/en/migrations.html#seed-seeding-your-database). This way we can get rid of all these SQL scripts, and a single `cake migrations migrate` creates or update your database to the latest schema.

PR #1927, opened 2019-07-24
Remove 100 sentences limitation of list downloads

jiru:
This pull request removes the 100 sentences limitation of list downloads by replacing the download feature by a functionally equivalent, asynchronous implementation. Any list, no matter how big it is, can be downloaded without overloading the server.

When clicking on the download button, the download dialog box takes more time to show up, but this delay may be reduced by optimizing the implementation later on. For very long list that take seconds to export, the spinner could also be turned into a progress circle showing the download preparation progress.


trang:
On my side I see no blocker for merging this. Local tests went fine.

I only noted that the coding style for `webroot/js/sentences_lists/download.ctrl.js` is inconsistent with the rest of the angular code, but I'm not too picky about that. It can be refactored after merging. FYI the styleguide I've been following is: https://github.com/johnpapa/angular-styleguide/blob/master/a1/README.md

Other than that, should I wait until you've optimized the download of big lists with translations (cf. [Wall post](https://tatoeba.org/eng/wall/show_message/32249#message_32249))?

jiru:
Yes, please wait. I discovered a bug that needs to be fixed. :expressionless:

jiru:
List export with translations are now a bit faster (3-4 minutes to export list 907 with Japanese translations). It’s probably okay to merge it like that.

That said, I am not really satisfied with the current implementation. I coded this as a prototype for new export functionalities, like language pairs exports, but I realize it has serious limitations. The seek pagination method doesn’t work well when joining too much tables. Outputting translations too seriously hits performance. I might have to think of an entirely different way to go.

Note that since only one background operation can be performed at a time over the whole website, it won’t impact the server load much, but users may have to wait for an unexpectedly long period of time before the download starts.

trang:
Unfortunately, it seems the queued exports can still impact server load :(

Cf. #1939

Issue #1594, opened 2018-07-02
Same tags appearing in different pages


Guybrush88:
When browsing tags, some of them are duplicated in different pages. For example, the tag "by Ovid" is appearing both here https://tatoeba.org/ita/tags/view_all/by/page:3 and here https://tatoeba.org/ita/tags/view_all/by/page:4

There are some other examples when browsing all the tags containing "by"

trang:
Cannot reproduce. I believe this has been fixed with https://github.com/Tatoeba/tatoeba2/commit/d749e3b8f3a63c8af0319f033bcef78cbfcfc289.

AndiPersti:
I can reproduce this bug locally.

Here is my `tags` table:
````
MariaDB [tatoeba]> select id, name, nbrOfSentences from tags order by nbrOfSentences;
+----+-------+----------------+
| id | name  | nbrOfSentences |
+----+-------+----------------+
| 13 | tag13 |              1 |
|  2 | tag2  |              1 |
|  4 | tag4  |              1 |
|  6 | tag6  |              1 |
| 12 | tag12 |              2 |
| 10 | tag10 |              2 |
|  1 | tag1  |              2 |
|  5 | tag5  |              2 |
|  3 | tag3  |              2 |
|  8 | tag8  |              3 |
|  7 | tag7  |              3 |
|  9 | tag9  |              4 |
| 11 | tag11 |              4 |
+----+-------+----------------+
13 rows in set (0.00 sec)
````
Now let's introduce pagination (pages contain 3 items):
````
MariaDB [tatoeba]> select id, name, nbrOfSentences from tags order by nbrOfSentences limit 3;
+----+------+----------------+
| id | name | nbrOfSentences |
+----+------+----------------+
|  2 | tag2 |              1 |
|  4 | tag4 |              1 |
|  6 | tag6 |              1 |
+----+------+----------------+
3 rows in set (0.00 sec)
````
This looks fine so far but notice that the first tag is `tag2` and not `tag13` as when selecting all rows. Next page:
````
MariaDB [tatoeba]> select id, name, nbrOfSentences from tags order by nbrOfSentences limit 3 offset 3;
+----+-------+----------------+
| id | name  | nbrOfSentences |
+----+-------+----------------+
| 13 | tag13 |              1 |
|  1 | tag1  |              2 |
|  3 | tag3  |              2 |
+----+-------+----------------+
3 rows in set (0.00 sec)
````
[...]

PR #1940, opened 2019-08-16
Pagination only works reliable with distinct values

AndiPersti:
This pull request addresses issue #1594.

Pagination depends on a stable result set because it queries the
database several times in order to split the result in pages. But
`view_all()` in the `Tags` controller uses `nbrOfSentences` as the
ordering which isn't stable because several tags could have the same
number of sentences linked to them and MySQL doesn't guarantee the
order when there is a tie.

Using `id` instead ensures that there will always be a tiebreaker
because its values are distinct.

### DIFF
--- a/src/Controller/TagsController.php
+++ b/src/Controller/TagsController.php
@@ -104,7 +104,7 @@ class TagsController extends AppController
         $this->paginate = [
             'limit' => 50,
             'fields' => ['name', 'id', 'nbrOfSentences'],
-            'order' => ['id' => 'DESC'],
+            'order' => ['nbrOfSentences' => 'DESC', 'id' => 'ASC'],
             'conditions' => $conditions
         ];
 
--- a/src/Controller/TagsController.php
+++ b/src/Controller/TagsController.php
@@ -104,7 +104,7 @@ class TagsController extends AppController
         $this->paginate = [
             'limit' => 50,
             'fields' => ['name', 'id', 'nbrOfSentences'],
-            'order' => ['nbrOfSentences' => 'DESC'],
+            'order' => ['id' => 'DESC'],
             'conditions' => $conditions
         ];
 


trang:
> If the intent is to break ties and not change the order of tags completely, wouldn't it be better to sort by both fields?

I was about to comment on that :)

There is also another issue. On the [tags page](https://tatoeba.org/eng/tags/view_all), it is possible to order [by count](https://tatoeba.org/eng/tags/view_all?sort=nbrOfSentences&direction=asc) and [by name](https://tatoeba.org/eng/tags/view_all?sort=name&direction=asc). When people choose to sort the tags, the pagination will be again unreliable.

Note that there was a similar issue with the "List of members" and I had solved it by adding the User.id to the list of fields to order in `beforeFind()` of the User model (https://github.com/Tatoeba/tatoeba2/commit/d749e3b8f3a63c8af0319f033bcef78cbfcfc289#diff-02808643f83778450f7ccbe3067eeb9cR128).

AndiPersti:
> There is also another issue. On the [tags page](https://tatoeba.org/eng/tags/view_all), it is possible to order [by count](https://tatoeba.org/eng/tags/view_all?sort=nbrOfSentences&direction=asc) and [by name](https://tatoeba.org/eng/tags/view_all?sort=name&direction=asc). When people choose to sort the tags, the pagination will be again unreliable.

No, whenever the user clicks on one of these words, the sorting will get added (at the front) to the default order by clause, i.e. first the list is sorted by count/name and then ties are sorted by id.

Here are screenshots from using the [DebugKit](https://book.cakephp.org/debugkit/3.x/en/index.html) tool. First going to `tags/view_all` without any parameters:
tags

Now after clicking on "count":
tags2

AndiPersti:
> If the intent is to break ties and not change the order of tags completely, wouldn't it be better to sort by both fields? Something like `['nbrOfSentences' => 'DESC', 'id' => 'ASC']` (I haven't tested whether the query builder handles this correctly.)

I've just tested that and it would work.
(When you click on "name" you would get all 3 order by clauses:
tags3
)

trang:
Works fine, indeed. Thanks for the fix!

Issue #1353, opened 2016-11-01
Going to last page is very slow when there are lots of pages


trang:
http://dev.tatoeba.org/eng/sentences/show_all_in/eng/none/none/indifferent/page:2
```
SELECT `Sentence`.`id`, `Sentence`.`text`, `Sentence`.`lang`, `Sentence`.`user_id`, 
`Sentence`.`hasaudio`, `Sentence`.`correctness`, `Sentence`.`script`, `User`.`id`, 
`User`.`username`, `User`.`group_id`, `User`.`level` FROM `sentences` AS `Sentence` 
LEFT JOIN `users` AS `User` ON (`Sentence`.`user_id` = `User`.`id`) 
WHERE `Sentence`.`lang` = 'eng' ORDER BY `Sentence`.`id` desc LIMIT 10, 10

// took 2 ms
```

http://dev.tatoeba.org/eng/sentences/show_all_in/eng/none/none/indifferent/page:50998
```
SELECT `Sentence`.`id`, `Sentence`.`text`, `Sentence`.`lang`, `Sentence`.`user_id`, 
`Sentence`.`hasaudio`, `Sentence`.`correctness`, `Sentence`.`script`, `User`.`id`, 
`User`.`username`, `User`.`group_id`, `User`.`level` FROM `sentences` AS `Sentence` 
LEFT JOIN `users` AS `User` ON (`Sentence`.`user_id` = `User`.`id`) 
WHERE `Sentence`.`lang` = 'eng' ORDER BY `Sentence`.`id` desc LIMIT 509970, 10

// took 3833 ms
```

The higher the `LIMIT` offset is, the worse it gets.

Earlier today the server was very slow. The load average jumped to 12-13. It seemed to be due to more people (or maybe just one person) browsing the last pages of "Browse by language".

zachleigh:
@trang Ill have a look at this one. This might be helpful for us: http://stackoverflow.com/questions/4481388/why-does-mysql-higher-limit-offset-slow-the-query-down

ckjpn:
> Earlier today the server was very slow. The load average jumped to 12-13. It seemed to be due to more people (or maybe just one person) browsing the last pages of "Browse by language".

If jumping to the last page of "Browse by language" is something members find useful, I wonder if it would speed things up if you offered the same "sort by" options as you have at the top of lists.

"Sort by: date added | sentence id"

[...]

PR #1945, opened 2019-08-25
Limit max pages on large sets of results

trang:
This PR addresses #1353.

Changes were made according to [my last comment there](https://github.com/Tatoeba/tatoeba2/issues/1353#issuecomment-476564743):

- `/sentences/show_all_in` is limited to 1000 sentences. Number of page depends on user settings (100 pages if 10 sentences per page).
- `/sentence_comments/index` is limited to 1000 comments. Title changed to "Latest comments".
- `/contributions/index` was removed.
- `/contributions/of_user` is limited to the last 1000 log entries. Title changed to "Latest contributions from {username}".


jiru:
Nice.

I suggest you replace every occurrence of 1000 by a self-explanatory constant. See https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad
-- gillux

Issue #1472, opened 2017-04-20
Big numbers in stats are difficult to read


trang:
Reported by mailohilohi on the Wall:

> Would it be feasible to add commas to the stats on the main page? It's hard to figure out exactly how large they are without manually counting the number of digits (e.g., 58167293 instead of 58,167,293).

https://tatoeba.org/eng/wall/show_message/28332#message_28332

Note: this needs to be localized. In French for instance, it should be a space instead of a comma.

Issue #1884, opened 2019-05-06
Large numbers sometimes have commas, and sometimes they don't.


ckjpn:
This is a low priority thing, I think, but this should probably be standardized throughout the website.

Here are 2 examples.  There are others.

<img width="596" alt="Screen Shot 2019-05-06 at 10 20 07" src="https://user-images.githubusercontent.com/7934273/57203174-bfa70280-6fe8-11e9-9677-e6dcd418788e.png">

<img width="691" alt="Screen Shot 2019-05-06 at 10 19 44" src="https://user-images.githubusercontent.com/7934273/57203175-bfa70280-6fe8-11e9-9e88-6e00a091fd44.png">

jiru:
Closing because it’s similar to #1472.

PR #1979, opened 2019-10-07
Add localized number formatting to language stats

AndiPersti:
This PR addresses issue #1472.


AndiPersti:
Currently there's only one commit in this PR which addresses #1472.

But as mentioned in #1884, there are many other pages with unformatted large numbers. I would like to add formatting to these pages too.

Do you prefer many single file commits added to this PR or one big commit which includes all the changes (basically adding `$this->Number->format()` at the appropiate places)?

jiru:
Thanks Andreas! If all your commits are doing the same thing, better merge them into a big one yeah.

About the commit you just pushed, I think it misses the inclusion of the Number helper in whatever View or Controller is using the sentences_by_language block.

AndiPersti:
> If all your commits are doing the same thing, better merge them into a big one yeah.

Ok, will do so in the next few days.

> About the commit you just pushed, I think it misses the inclusion of the Number helper in whatever View or Controller is using the sentences_by_language block.

According to the CakePHP docs, [loading helpers in a Controller is deprecated since version 3.0](https://book.cakephp.org/3.0/en/controllers.html#configuring-helpers-to-load). And ["you don’t have to explicitly load Helpers that come from CakePHP or your application"](https://book.cakephp.org/3.0/en/views/helpers.html#configuring-helpers). 

I you prefer explicitly loading the Number helper I guess the best place would be in the AppController because I will use it in many places:
https://github.com/Tatoeba/tatoeba2/blob/089e3ab5fe6e6c712aae21b67bedf99efc5efd5b/src/Controller/AppController.php#L71-L83

jiru:
I see, thanks for clarifying! I didn’t know that Helpers can be lazily loaded upon use. As far as I know, most of the existing code assumes that Helpers need to be explicitly loaded. Would be a good thing to clean this up one day. :roll_eyes:

AndiPersti:
I think I've found all places where number formatting is useful.

trang:
Thanks a lot :)

PR #1981, opened 2019-10-10
Change http://tatoeba.org to https://tatoeba.org

rwmpelstilzchen:
As far as I’m aware, it should not break anything.


trang:
It should be fine, yes :) Thank you for this first PR and welcome to the project!

rwmpelstilzchen:
Thanks for this and for Tatoeba in general ^_^

PR #1996, opened 2019-10-27
Fix <title> tag to improve bookmarks

### LINKED ISSUE #809: Change the order of the information in the TITLE tag of pages.
Instead of:
English example sentence: Is eating liver good for you? - Tatoeba
Have this:
Is eating liver good for you? - English example sentence - Tatoeba
Or this:
Is eating liver good for you? - English example sentence
Or this:
Is eating liver good for you? (English example sentence)
- This would put the important information at the begining.
- This would mean the in browsers like Chrome that crop page titles that you can easily see a unique name on each tab if you have multiple tabs open.
- This would make nice bookmark titles for people who bookmark a page.
- This would have sentences displays more beautifully in Google search results.
- The "Tatoeba" part at the end probably doesn't really help with Google searches, so you could remove a little clutter by taking it out of the title.

Currently Google Results look like this.

screen shot 2015-10-11 at 10 03 15

*\* Related Idea **

Perhaps it might also be a time to reconsider what's in the description META tag.  Perhaps it could be made more useful.  For example, if it doesn't slow things down too much, you could have a line like the following instead, based on what the direct translations are.

Translations in: Japanese, French, Russian

Eien18:
Fixes #809, should be merged after #1995 (sorry for messing up with the branches).

### DIFF
--- a/src/Template/Element/foot.ctp
+++ b/src/Template/Element/foot.ctp
@@ -188,7 +188,7 @@
                 __('More information'),
                 array(
                     'controller' => 'pages',
-                    'action' => 'downloads'
+                    'action' => 'terms_of_use#section-6'
                 ),
                 array(
                     'class' => 'more-info'
--- a/src/Template/Sentences/show.ctp
+++ b/src/Template/Sentences/show.ctp
@@ -36,7 +36,7 @@ if (isset($sentence)) {
     $sentenceCorrectness = $sentence->correctness;
 
     $languageName = $this->Languages->codeToNameToFormat($sentenceLang);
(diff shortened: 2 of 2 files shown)


trang:
I think you should be able to fix the branching with:

```
git rebase --onto  9b803c68b9dd75f387f47325e51967eb97d012f0 dev Eien18-patch-1
```

Cf. https://stackoverflow.com/questions/50746504/rebasing-duplicating-feature-branch-to-an-older-release

Then if the rebase worked fine, you can do a `git push --force`.

Welcome by the way :)

PR #1997, opened 2019-10-27
Fix "does not own any sentences"

### LINKED ISSUE #1931: "contributor does not have any sentence" -> "contributor does not own any sentences"
I logged in as the "contributor" user on Imouto. When I clicked on "Sentences", I saw the message "contributor does not have any sentence". The better way to put it would be "contributor does not have any sentences", or, better yet, "contributor does not own any sentences".

jilljenn:
Fixes #1931.

I used the following command:
`find . -type f -exec sed -i 's/have any sentence/own any sentences/g' {} +`

In particular, the version of the repo in the vagrant seems far from dev, do you plan to refresh it soon?


trang:
Welcome to the project, @jilljenn :)

There are some extraneous modifications in your pull request. The file `of_user.ctp` is actually the only file that needs to be modified here. All the files in the `src/Locale` folder are automatically generated so there is no need to modify them.

I plan to add some explanations on how localization works in the [README.md](https://github.com/Tatoeba/tatoeba2/tree/dev/src/Locale) in the Local folder.

Could you let me know if the information below is clear enough to you and would have helped for this pull request? Thanks!

## How localization works

1. Strings that translatable are put into specific functions in the source code (mainly `__()`, but there are [a few others](https://book.cakephp.org/3.0/en/core-libraries/internationalization-and-localization.html#using-translation-functions)).
2. Whenever there are new strings or strings that have been modified, we run the script [generate_pot.sh](https://github.com/Tatoeba/tatoeba2/blob/dev/tools/generate_pot.sh). It extracts all the strings from the `*.php` and `*.ctp` files into `*.pot` files.
3. We upload these `*.pot` files on a platform called [Transifex](https://en.wiki.tatoeba.org/articles/show/interface-translation), where people can translate the strings.
4. We use the script [update-translations.py](https://github.com/Tatoeba/tatoeba2/blob/dev/docs/update-translations.py) to download the updated `.po` files from Transifex.

trang:
> In particular, the version of the repo in the vagrant seems far from dev, do you plan to refresh it soon?

I'd like to say "yes", but I can't say for sure when we will...

AndiPersti:
If you want to keep the code in the VM updated I suggest adding a remote called 'upstream' and regularly pull from it and push the changes to your fork (called 'origin' if you followed the [instructions](https://github.com/Tatoeba/imouto#editing-the-source-code)):

    # Do the following once
    $ git remote add upstream https://github.com/Tatoeba/tatoeba2.git 
    # Do the following regularly. 
    # There should be no merge conflicts if you keep your dev branch clean, 
    # i.e. do all your work in dedicated branches
    $ git pull upstream dev
    $ git push origin dev

There seems to be a way to do the [updating with the GitHub UI](https://stackoverflow.com/q/20984802) but I prefer using the command line.

jilljenn:
Hi, I can kind of understand that there is only one file that needs to be modified.
However, in this case you should not commit the generated files, should you?
In the case you rename a translatable sentence, won't you have to create new entries in Transifex?

trang:
> However, in this case you should not commit the generated files, should you?

For the `*.po` files: 
- They are actually not 100% generated. The content of the files still comes from human translations.
- We commit these files because it is the simplest way to ensure that anyone who clones our repo will have all the UI translations available right off the bat.
- It is also useful for us to track these files in our git repo as a safety measure. If one day Transidex shuts down or if one day someone decides to sabotage our translations, we can easily restore them.

For the `*.pot` files:
- The reason why they are tracked is because Transifex has a feature to automatically update the translation strings. For that, it requires to have a URL that points to the appropriate file and by committing the files, we can have such URLs (for instance [default.pot](https://raw.githubusercontent.com/Tatoeba/tatoeba2/dev/src/Locale/default.pot)).
- However, you reminded me that I never got this feature to work properly, so at the moment, committing `*.pot` files is indeed a bit pointless.

> In the case you rename a translatable sentence, won't you have to create new entries in Transifex?

Yes, if we rename a translatable sentence, Transifex will indeed see it as the deletion of a sentence and the addition of a new sentence.

Transifex has translation memory though, so if we're only fixing a typo, it can detect that there was a similar string that was translated before and it can suggest a translation.

trang:
@jilljenn Please let me know if you plan to solve the conflict in this PR. If not I will solve it and merge. We don't need to keep this PR open forever :)

trang:
Solved in 214ea8747bf243261c8437c788fc342f2cf35cd5.

PR #1998, opened 2019-10-28
Replaced svg flags

RyckRichards:
This PR replaces .svg flags


trang:
@RyckRichards There's a problem with `mrj.svg`. It contains `alt="`, which should be removed.

The rest looks good otherwise.

RyckRichards:
@trang I'm really sorry. I've fixed that.

trang:
No worries :)

soliloquist-tatoeba:
https://tatoeba.org/eng/sentences/show/7005381#comment-1138035

I don't know if it's related to this update, but the icon of the Kamba language (kam) began to appear on Georgian (kat) sentences, too.

https://tatoeba.org/eng/Sentences/show_all_in/kat/

https://tatoeba.org/eng/Sentences/show_all_in/kam/


The Georgian flag is this.

https://upload.wikimedia.org/wikipedia/commons/thumb/0/0f/Flag_of_Georgia.svg/1200px-Flag_of_Georgia.svg.png

RyckRichards:
Thanks, @soliloquist-tatoeba 
I'm going to fix  this in a separate threat.

soliloquist-tatoeba:
Thanks, Ricardo!

There are also similar issues with Cornish (cor) and Zaza (zza).

https://tatoeba.org/eng/wall/show_message/33414#message_33414

Yorwba:
After I noticed the complaints, I decided to look at all the flags side-by-side to see if this PR introduced any more errors. I was a bit surprised that there had been many more substantial changes than just adjusting the saturation of the colors.
Based on the changes made, I've reverse-engineered the following policy for determining the flag for a language:
- Find the smallest geographic region that representative of the language and use its flag. This explains the flags for minority languages in Russia, which use the flags of republics within the Russian Federation.
- If said region doesn't have a flag, use the flag of the next higher region, usually the country. This explains the flags of most Chinese and Indian languages.
- If there are multiple languages with the same flag, add a language code to distinguish them.
- As an exception, if one language is much more prominent than the others, it doesn't need a language code
    - This explains Mandarin (official language) vs. other Chinese languages (not official)
    - This also explains Tigrinya (working language in Eritrea) vs. Tigre (recognized as national language but not a working language)
- As another exception, if there's an ongoing dispute about the flag, it is not changed. This explains the flag for Uyghur and most of the Berber languages.

I noticed some flags that don't exactly follow the rules above:
-  `dsb` and `hsb`: For Lower and Upper Sorbian the flags were previously based on the coat of arms of Lower and Upper Lusatia. They're now based on the flags of Brandenburg and Saxony, respectively.
- `roh`: The flag was previously based on the coat of arms of Grisons, now it's the flag of Switzerland.
- `bar`: Bavaria has two co-official flags. The blue-white checkered one is much more iconic, but now it was replaced by the striped one, which is visually similar to that of the Altai Republic.
- `alt`: The flag of the Altai Republic has a language code despite no other language using the same flag.
- `lug`: The flag of Uganda has a language code despite no other language using the same flag.
- `pus`: The flag of Afghanistan has a language code despite no other language using the same flag.
- `fij`: The flag of Fiji has a language code despite being official while Fiji Hindi isn't.
- The flags of `lij`, `nst`, `mal`, `mar` didn't change although they maybe should.
- `rif`: This is the only flag of a Berber language that was changed.

trang:
@Yorwba FYI, our policies for language icons are explained in this article in the wiki:
https://en.wiki.tatoeba.org/articles/show/language-icons

There isn't really any strict rule regarding which flag we will use for a language. We won't necessarily use the flag of a country/region either. Most of the time, we simply follow the suggestion that was made to us when the language was requested.

For the ISO code being inconsistently appended, I can imagine these cases:
1. The flag alone is not explicit enough.
2. The language is part of a macrolanguage and the flag can be reused for the other languages in that macrolanguage. The ISO code is added in anticipation, so that when other languages of the macrolanguage are added, there is no need to update the icon of the first language.

Anyway I'm tagging @sabretou since he is the main person responsible for the design of the language icons and was the one who decided on the changes here. Your feedback will certainly be useful for the next batch :)

jiru:
May I ask what you have been doing to the flags? Maybe a little explanation in the commit messages wouldn't hurt.

sabretou:
@Yorwba Hello! Thank you for your analysis of the flag changes.

As Trang stated, we didn't follow a strict system for which flag will be used. When picking or designing flags, I ended up prioritising which high-quality images were available while retaining an acceptable level of accuracy of relevance.

This is the chief reason for the changes to the flags of dsb and hsb.

`roh`: Using the flag of Grisons would definitely be a good idea for this.
`bar`: As both Bavarian flags have equal status, I don't think I put much thought into which one to use. We will have to wait and see if Bavarian users have issues with this.
`alt`: ISO 639-3 divided Altai into Northern and Southern Altai. At present, we only have Southern Altai on Tatoeba, but if we need to add Northern Altai, we will have to use the same template.
`lug`: Uganda has many indigenous languages that may share the flag of Uganda when added. The code on this language is to account for the possibility of other languages being added.
`pus`: This is to account for the possibility of other Afghan languages being added, such as Dari.
`fij`: This has been changed, I think, because I'm seeing codes for both languages now. However, I'm not seeing evidence that Fijian and Fiji Hindi have different official statuses.
`lij`: Agreed, this should definitely be changed.
`nst`: Agreed with this one too.
`mal`: I think I considered changing this one, but it may be best to wait for a native speaker's opinion, as the current flag was requested by a native speaker.
`mar`: I am a native speaker of this myself, and I think this flag is representative of Marathi cultural identity. However, I am amenable to changes if other native speakers feel differently.
`rif`: This was before we chose the current Berber flags. We'll have to consult a native speaker for this one.

trang:
To add a bit more context: originally, language icons had color saturation and grey borders on the right and bottom side. After some time, newer language icons were not following this pattern anymore. Generally speaking, since language icons were designed by different people over time, they were graphically inconsistent.

Sabretou worked on making them overall more consistent. For one thing, instead of applying color saturation and grey borders to all icons, it made more sense to have them all in original colors and without grey borders. Graphical effects could be applied from the CSS instead, if needed. This was not the only change, but was the main driver.

In the end, sabretou simply reviewed every icon and updated those that he could, to move closer to a consistent set of icons. Then Ricardo made this pull request from sabretou's updated icons.

There was indeed not much explanation around this pull request. I hope this clarifies things.

Yorwba:
@sabretou Makes sense that high-quality images would be preferred.

For `dsb` and `hsb`, maybe one of the higher-quality images in the Wikimedia Commons categories for [Lower Lusatia](https://commons.wikimedia.org/wiki/Category:Coats_of_arms_of_Lower_Lusatia_land) and [Upper Lusatia](https://commons.wikimedia.org/wiki/Category:Coats_of_arms_of_Upper_Lusatia_land) would be appropriate.

Regarding `fij`, I think I misread something. [The 2013 constitution](https://www.laws.gov.fj/ResourceFile/Get/?fileName=2013%20Constitution%20of%20Fiji%20(English).pdf) does treat Fiji and Fjji Hindi equally.

jiru:
@Trang Thanks for clarifying!

jeanm:
Looks like `enm`'s flag lost its language code, which confused me earlier since `lij` uses the same flag.

May I suggest adding the language code back to `enm`? Or, if you're ok with having one flag with a code and one without, then could we make `lij` the language that doesn't have the code in the flag, since (unlike Middle English) it's not extinct?

trang:
@jeanm The change was not intentional. We will add back the version with the code for `enm`.

Issue #1885, opened 2019-05-06
Shown translations disappear when contributing a sentence within another field in the same page


Guybrush88:
Since before the latest CakePHP update, if you open a field to translate a sentence that already has some translations (both direct and indirect), do nothing, translate another sentence within the same page (usually using the advanced search), and then submit the translation, the first field you opened (on another sentence) automatically closes, and every existing sentence doesn't show up anymore until you reload the page.

Steps to reproduce this issue:
1. use the advanced search to search for anything
2. click on the button to translate any sentence with existing translations (both direct and indirect) and do nothing on it
3. leave the form open
4. translate another sentence within the same page
5. submit the translation

Expected behavior: the existing translations of the first sentence you try to translate should still appear even if you translate another sentence from the same page and you leave for whatever reason the first form open.

Guybrush88:
@jiru, I don't think this is a regression, I quite remember this issue was there also before the latest CakePHP update

PR #1999, opened 2019-10-29
Only close active translation box when more than one is open

AndiPersti:
This PR addresses #1885.

`addTranslations` refers to all elements on the page having that class attribute. Thus hiding it will hide all these elements. For hiding just the currently active element, `translationFor` should be used.


trang:
As always, thanks for the fix!

Issue #1221, opened 2016-06-23
New tooltips


trang:
Currently we use the `title` attribute to display tooltips. Example of a tooltip:

![screenshot_14](https://cloud.githubusercontent.com/assets/221850/16317069/b189d828-3989-11e6-9375-fc4810416209.png)

We want to replace these tooltips with the Angular Material tooltips:
https://material.angularjs.org/latest/demo/tooltip

Parent issue: #1180

trang:
Closing because obsolete. We use `md-tooltip` in many places now already.

PR #2000, opened 2019-10-30
Wall redesign

trang:
This is a redesign of the Wall. The main goal was to make the look and feel of the Wall more consistent with the comments on sentences.

I'm opening a pull request in case someone wants to review the changes. I will merge this Sunday if there is no major issue.

Related Wall thread: https://tatoeba.org/eng/wall/show_message/33383#message_33383


src/Template/Element/wall/message.ctp
```diff
@@ -0,0 +1,162 @@
...
+        '#' => "message_" .$messageId
+    ]
+];
+
+$children = $message->children;
+$cssClass = isset($isRoot) ? 'wall-thread' : 'reply';
+$canReply = false;
+?>
+
+<md-card class="comment <?= $cssClass ?> <?= $messageHidden ? 'inappropriate' : '' ?>">
```
AndiPersti: Adding an id attribute here would allow linking to a specific message which was possible with the old design (see also the comments on the wall by [AlanF_US](https://tatoeba.org/wall/show_message/33384#message_33384) and [CK](https://tatoeba.org/wall/show_message/33385#message_33385)):

    <md-card id="message_<?= $messageId ?>" class=...

trang: I was about to do that :)

src/Template/Element/wall/message.ctp
```diff
@@ -0,0 +1,162 @@
...
+<md-card id="message_<?= $messageId ?>" class="comment <?= $cssClass ?> <?= $messageHidden ? 'inappropriate' : '' ?>">
+    <md-card-header>
+        <md-card-avatar>
+            <?= $this->Members->image($username, $avatar, array('class' => 'md-user-avatar')); ?>
+        </md-card-avatar>
+        <md-card-header-text>
+            <span class="md-title">
+                <a href="<?= $userProfileUrl ?>"><?= $username ?></a>
+            </span>
+            <span class="md-subhead ellipsis" title="<?= $fullDateLabel ?>">
```
AndiPersti: A few days ago I've come across #1221 and since you use Material tooltips already on line 106 for all the menu buttons I think you should also use them here.

Similarly for https://github.com/Tatoeba/tatoeba2/blob/923b1a726e9db2831c7da8b167307a3d1e60e16d/src/Template/Element/sentence_comments/edit_form.ctp#L50 and https://github.com/Tatoeba/tatoeba2/blob/923b1a726e9db2831c7da8b167307a3d1e60e16d/src/Template/Element/wall/edit_form.ctp#L45

PR #2001, opened 2019-11-01
Update clipboard.js to 2.0.4

AndiPersti:
The update is necessary in order to enable iPhone/iPad users to use
the copy button.
Related wall thread: https://tatoeba.org/wall/show_message/33268

While here I've also added an id attribute to sentences which aren't
editable. This simplifies the javascript code for finding the correct sentence text.


jiru:
Thanks. I just would like to point out that it breaks the "blended furigana" feature on Japanese sentences. When a Japanese sentence has its transcription (furigana) validated by a human, it gets "blended" with the sentence. You can see this in [sentence #4738018](https://tatoeba.org/fra/sentences/show/4738018) for example (you have to be logged in). The "blending" happens client-side in transcriptions.js. In that case, clicking the copy button will now copy the inner html instead. That’s why the code used to retrieve the sentence text using `.data('text')`. (See the [code history](https://github.com/Tatoeba/tatoeba2/commits/44eca4310a3be1e711425e2e36696a6e258f1633/app/webroot/js/sentences.copy.js).) I admit that that implementation was crap though.

The rationale behind that blending thing is that we want to avoid cluttering the display with the same sentence displayed twice (this can be seen when clicking the かな sentence button, which also allows to edit the furigana).

AndiPersti:
Thanks for the review. I've just reverted my changes and just updated clipboard.js.

PR #2004, opened 2019-11-07
Add missing scripts to charset_table

Yorwba:
This PR adds or improves support for a number of scripts. See issue #1970.

# Gothic

The characters used to write Gothic (got). The existing Gothic
sentences seem to use spaces between words, so using charset_table is
likely appropriate.
Unicode Block: https://www.unicode.org/charts/PDF/U10330.pdf

# Old Turkic

These characters are used to write Old Turkish (otk). The existing Old
Turkish sentences separate words with a colon, so using charset_table is
likely appropriate. Writing direction is right-to-left.
Unicode Block: https://www.unicode.org/charts/PDF/U10C00.pdf

# Warang Citi

Warang Citi is used by some to write Ho (hoc). According to Wikipedia,
Latin punctuation including spaces is used, so using charset_table is
likely appropriate.
The script has a range of uppercase characters, which I mapped to the
lowercase ones.
Unicode Block: https://www.unicode.org/charts/PDF/U118A0.pdf

# Mongolian

Mongolian script is used to write Mongolian (mon) and Manchu (mnc).
The existing Mongolian sentences all seem to use Cyrillic, though.
Spaces are separated with words, so using charset_table is likely appropriate.
There are a few punctuation marks, which I have excluded.
Traditional writing direction is vertical, but supporting that would
really mess with the layout.
Unicode Block: https://www.unicode.org/charts/PDF/U1800.pdf

# Malayalam

Some non-punctuation characters used to write Malayalam (mal) were missing from
the end of the Unicode range. Probably a typo.
Unicode Block: https://www.unicode.org/charts/PDF/U0D00.pdf

# Spacing Modifier Letters

Of the Spacing Modifier Letters already present in the Tatoeba corpus,
  - ʻ (U+2BB MODIFIER LETTER TURNED COMMA) is used to write the Hawaiian
  Okina, so I added it.
  - ʼ (U+2BC MODIFIER LETTER APOSTROPHE) is used as a stand-in for a
  "regular" apostrophe, so I left it out.
  - ʿ (U+2BF MODIFIER LETTER LEFT HALF RING) is to transcribe the Arabic
  Ayin, so I added it.
  - ˀ (U+2C0 MODIFIER LETTER GLOTTAL STOP) is used to write the glottal
  stop in Cayuga, so I added it.
  - ˈ (U+2C8 MODIFIER LETTER VERTICAL LINE) and
  - ˌ (U+2CC MODIFIER LETTER LOW VERTICAL LINE) are used in IPA
  transcriptions to mark primary and secondary stress, which I think
  means they're better left out.
  - ː (U+2D0 MODIFIER LETTER TRIANGULAR COLON) is used to mark vowel
  length in Ngeq, so I added it.
Unicode Block: https://www.unicode.org/charts/PDF/U02B0.pdf

# Cyrillic

There are five Cyrillic Unicode blocks:
- Cyrillic (some combining characters in this range were missing): https://www.unicode.org/charts/PDF/U0400.pdf
- Cyrillic Supplement: https://www.unicode.org/charts/PDF/U0500.pdf
- Cyrillic Extended-A: https://www.unicode.org/charts/PDF/U2DE0.pdf
- Cyrillic Extended-B: https://www.unicode.org/charts/PDF/UA640.pdf
- Cyrillic Extended-C: https://www.unicode.org/charts/PDF/U1C80.pdf

# Greek

The Greek Extended character set contains accented characters used to
write Ancient Greek (grc) with polytonic orthography.

Unicode Block: https://www.unicode.org/charts/PDF/U1F00.pdf

There were also some misalignments to fix in the Greek and Coptic block.

# Latin

Most Unicode blocks containing Latin characters were already covered,
but with some missing characters here and there.

There are also two new Unicode blocks:
- IPA Extensions: https://www.unicode.org/charts/PDF/U0250.pdf
- Phonetic Extensions: https://www.unicode.org/charts/PDF/U1D00.pdf

# Cherokee

Lowercase Cherokee characters were added to Unicode in 2015.
There are two Unicode blocks now:
  - Cherokee: https://www.unicode.org/charts/PDF/U13A0.pdf
  - Cherokee Supplement: https://www.unicode.org/charts/PDF/UAB70.pdf

# Armenian

The Armenian block was previously listed as a single range. However, it
contains both uppercase and lowercase characters, as well as some
punctuation (which I have excluded in the updated definition).

Unicode block: https://www.unicode.org/charts/PDF/U0530.pdf

# Miscellaneous

I had to replace `U+0` by `U+` to shave off some bytes.  This prevents
Manticore from choking on overlong lines in the configuration file.
([The buffer size is limited to 8192 bytes.](https://github.com/manticoresoftware/manticoresearch/blob/8eb6d8efb1748ef134a0c01bc5db939250a5bad3/src/sphinxutils.cpp#L1198))

Breaking `charset_table` in the generated configuration over multiple lines
when it gets too long would be more elegant, but the quick and dirty solution
is good enough for now.


trang:
This looks good to me, but like the last PR I haven't checked everything in detail. I'll leave this PR un-merged until Sunday in case someone has an objection or an issue to report.

Issue #1873, opened 2019-04-25
We used to be able to click a sentence in the logs on the side of a sentence's page, and we could jump to that sentence on the page.


ckjpn:
We used to be able to click a sentence in the logs on the side of a sentence's page, and we could jump to that sentence on the page.

I didn't miss that function until today, so maybe it's not that important to fix.

AndiPersti:
I'm not sure if I understand your problem. If I click on a sentence's id (red circled in the screenshot) I get to that sentence's page. But I guess that's not what you want, is it?
![screenshot](https://user-images.githubusercontent.com/6566321/67703013-07648080-f9b3-11e9-9cbc-d6a89e884258.png)

ckjpn:
We used to be able to click, not the link there but the whole area there, and it would jump up to the text of that sentence on the same page, so we could easily know which one of the above sentences the number referred to.

trang:
Related code: https://github.com/Tatoeba/tatoeba2/blob/dev/webroot/js/sentences.logs.js

AndiPersti:
Ok, I understand. 

Was it still possible to click on the number and get to a new page?

trang:
Yes, it was still possible to click on the number and go to a new page.

Issue #1653, opened 2018-08-25
Inconsistent logs


trang:
## Problem

The logs do not always reflect the actions of the user.

This issue combines #520, #699 and #1352.

## Solution

Suggested in https://github.com/Tatoeba/tatoeba2/issues/520#issuecomment-403515073:

> A simple way to prevent the creation of sentences without log, is to use wrap the sentence creation queries between BEGIN/COMMIT SQL statements. Here is [how to do it with CakePHP](https://book.cakephp.org/2.0/fr/models/transactions.html).

PR #2005, opened 2019-11-09
Reactivate highlighting feature for log entries

AndiPersti:
This PR addresses #1873 

We need to load the necessary JS files when at least one link contribution is shown in the logs.

I've also removed an obsolete version of the jQuery.scrollTo library.


src/Template/Element/logs/log_entry_annexe.ctp
```diff
@@ -49,10 +49,12 @@ $langDir = LanguagesLib::getLanguageDirection($langCode);
             } elseif ($type == 'license') {
                 echo $sentenceLink.' ➜ '.$this->Html->tag('span', $sentenceText, array('class' => 'license'));
             } else { // link
+                $this->Html->script(['jquery.scrollTo.min.js', 'sentences.logs.js'],
+                                    ['block' => 'scriptBottom']);
```
trang: One minor thing: if the scripts are included from here, then they will be loaded in `/users/show` where they are not needed. They are only needed in `/sentences/show` so it is better to include them from there.

AndiPersti: Yes, you're right.

Another thing I haven't considered in my current implementation is, that it will load the scripts if a sentence currently doesn't have any translation but the logs show linking and corresponding unlinking actions, i.e. it had a translation in the past which was deleted later.

I don't know if there are any cases like that in the production database but I'm thinking about avoiding loading the scripts in that case.

trang: There are many cases of sentences that don't have any translations and indeed, the script would be unnecessary for those too.

AndiPersti: I was more thinking of a case like this:
example_logs

I've just pushed my new implementation.

src/Template/Sentences/show.ctp
```diff
@@ -150,14 +148,25 @@ echo $this->element('/sentences/navigation', [
...
+                }
+                if ($contribution->type == "link" and $contribution->action == "delete") {
+                    unset($active_links[$contribution->translation_id]);
+                }
             }
             echo '</md-list>';
+            if (!empty($active_links)) {
+                $this->Html->script(['jquery.scrollTo.min.js', 'sentences.logs.js'],
+                                    ['block' => 'scriptBottom']);
+            }
```
trang: A more simple way would be to just check if `$sentence->translations` is empty. Analyzing the logs works just fine too, but it's a bit more lines of code :)

Now that doesn't take into account that logs are not 100% reliable (#1653). And if you would want to be completely accurate about when to load the javascript, you could keep your code above and additionally check if the translations in the `$active_links` actually exist in `$sentence->translations`.

But if you ask me, we don't have to go into that level of details. Such cases would be rare and we don't have to over-complexify the code for a low impact optimization.

AndiPersti: Yeah, after sleeping over it, I realized that I was overthinking the problem. :-)

But while testing for a sentence without a log (by manually inserting it into the sentences table using the mysql REPL) I've noticed that the condition on line 152 doesn't work: https://github.com/Tatoeba/tatoeba2/blob/a171662ded228c0e5be38a56b6504ea009f68e38/src/Template/Sentences/show.ctp#L152

I've expected to get the message "There is no log for this sentence" from the else branch. But since `$contributions` is a CakePHP ResultSet it will never be empty because an empty ResultSet is an object containing an empty list. Thus I think a better way for testing whether `$contributions` is empty would be to check its count.

I've updated the PR (hopefully the last time :-) )

src/Template/Sentences/show.ctp
```diff
@@ -149,7 +147,11 @@ echo $this->element('/sentences/navigation', [
             echo $this->Sentences->originText($sentence);
         }
 
-        if (!empty($contributions)) {
+        if ($contributions->count() > 0) {
+            if (!empty($sentence->translations)) {
```
trang: You're probably going to hate me but I realized while testing that `$sentence->translations` will (should?) never be empty. It's structured into two arrays.
```
[
	(int) 0 => [], // direct translations
	(int) 1 => [] // indirect translations
]
```
So we need to test that `$sentence->translations[0]` is empty.

AndiPersti: No problem, I can tolerate a lot of criticism (especially if it's justified) :-)

You probably won't believe me but I originally wrote the test with `$sentence->translations[0]` but just before committing I thought for some reason it's not necessary to be that specific and then forgot to also test for sentences without any translations where that makes a difference.

PR #2009, opened 2019-11-11
Fix typos which broke audio filter for searches

AndiPersti:
5f5c4fd1bcf5edfa2f614de1005fe673b34b31b4 introduced some typos which broke the filter for audio.


trang:
Thank you very much :)

PR #2011, opened 2019-11-13
Language

riverxside:
Fix language bug

### DIFF
--- a/src/Controller/UsersLanguagesController.php
+++ b/src/Controller/UsersLanguagesController.php
@@ -24,6 +24,7 @@
  * @license  Affero General Public License
  * @link     https://tatoeba.org
  */
+
 namespace App\Controller;
 
 use App\Controller\AppController;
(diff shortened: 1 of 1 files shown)


trang:
Welcome and thank you for this PR, @riverxside!

Before merging this, I would suggest to reword "This language is assigned" to "This language has already been added to your profile".

@alanfgh, could you confirm if the new wording is okay?

alanfgh:
@trang , yes, the new wording is okay.

jiru:
Thank you very much for your contribution, riverxside.

If you can, it would be great if you could also add a unit test that checks what happens when a adding a language already in your profile. The test should goes in `tests/TestCase/Controller/UsersLanguagesControllerTest.php`. Have a look at the existing tests in that file. Then you can run these tests using `phpunit tests/TestCase/Controller/UsersLanguagesControllerTest.php`. If you’re not sure you can do that, it’s okay, we can also merge your pull request without it.

trang:
Merging this now.

@riverxside, if you ever feel like writing the tests later on, you are welcome to open another pull request :)

Issue #1886, opened 2019-05-07
Vocabulary items count is wrong


Guybrush88:
As reported by Seael on the wall:

> Something is wrong with the counting here: https://tatoeba.org/eng/vocabul...es/spa?page=32
> 
> Two examples (there are more):
> 
> 1) It says there are 8 sentences containing the word "cigüeñas" and when you click on the link, they are actually 12 sentences.
> 
> 2) The count becomes negative with the term "fascista". It says there are 8 sentences and, when you follow the link, it only shows 6.

https://tatoeba.org/ita/wall/show_message/31800#message_31800

ckjpn:
Perhaps this is related.

https://github.com/Tatoeba/tatoeba2/issues/1388

soliloquist-tatoeba:
A similar syncing problem affects lists, too. The counts of some lists on the index page are incorrect. I suspect that removing sentences from lists triggers the issue.


<img width="519" alt="tat" src="https://user-images.githubusercontent.com/47579687/57398650-35bd8c00-71d8-11e9-9fdb-468331634104.png">

ckjpn:
Perhaps this one is related, too.  (It's been closed.)

https://github.com/Tatoeba/tatoeba2/issues/1278

ckjpn:
The one soliloquist-tatoeba mentioned is already an issue reported in 2016.

https://github.com/Tatoeba/tatoeba2/issues/1277

AndiPersti:
> 2. The count becomes negative with the term "fascista". It says there are 8 sentences and, when you follow the link, it only shows 6.

The link loads the result page with the default search options which exclude sentences which are orphan or unapproved.
[...]

PR #2015, opened 2019-11-18
Fix search link for vocabulary item

AndiPersti:
This PR addresses the second part of #1886.

Include orphan and unapproved sentences in the search for a vocabulary item. This fixes a mismatch between the count on the result page and the count on the vocabulary page.


trang:
One small thing missing: when you add a vocabulary item, the search link for the newly added item in the "Vocabulary items added" needs to be changed too.

AndiPersti:
Updated with a fix for the js code.

PR #2019, opened 2019-11-22
Private messages redesign

trang:
This is a redesign of the private messages. The main goal was to make the look and feel more consistent with the sentence comments and the Wall.

I'm opening a pull request in case someone wants to review the changes. Not sure yet if I'll merge this weekend or the next one.

Related Wall thread: https://tatoeba.org/eng/wall/show_message/33552#message_33552

### DIFF
--- a/src/Controller/PrivateMessagesController.php
+++ b/src/Controller/PrivateMessagesController.php
@@ -146,71 +146,7 @@ class PrivateMessagesController extends AppController
             return $this->redirect(['action' => 'folder', 'Inbox']);
         }
 
-        $menu = $this->_getMenu($message->folder, $messageId, $message->type);
-
         $this->set('message', $message);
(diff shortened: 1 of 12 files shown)


trang:
Following the feedback regarding the "New message" button and how it was confusing, I redesigned the entire layout of the private messages section. I couldn't find any other satisfying way to address the issue with this button without redesigning the whole environment in which it is displayed.

- The "New message" menu item was converted into a primary button.
- The string "New message" was replaced by "Compose" in order to use a verb (which makes more sense for a button).

**Notes for the responsive UI**

Remaining things to do:
- Adding a burger menu icon in the main toolbar, before the "Private messages" string.
- On small screens, the sidenav should be hidden by default. 
- Clicking on the menu icon opens/closes the sidebar.
- We could consider turning the "Compose" button into a floating action button.

Issue #1698, opened 2018-10-11
List original sentences


trang:
## Problem

I have thousands of sentences and would like to know which of them are original. My only solution right now is to go on "My sentences" and check each sentence's page.

Besides just for curiosity, I need to know which of my sentences are original so I can review them and fix the license. I know I have some sentences that are copied and should therefore not be CC-BY.

## Possible solution

Add an additional filter in "My sentences" page, below "Filter by language".

Could be a checkbox with the label "Original sentences" and checking it would display only original sentences in the chosen language.

PR #2021, opened 2019-11-24
Add original sentences filter on a user's sentences page

AndiPersti:
This PR addresses #1698.

The first commit adds a checkbox on a user's sentences page to filter original sentences.

The second commit changes the behavior of the language filter so that an existing query string isn't lost when the language is changed. I personally find that more useful but I will remove it if it is not wanted.

### DIFF
--- a/webroot/js/sentences/only_original.ctrl.js
+++ b/webroot/js/sentences/only_original.ctrl.js
@@ -1,0 +1,43 @@
+/*
+    Tatoeba Project, free collaborative creation of multilingual corpuses project
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Affero General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+(function() {
+    'use strict';
+
+    angular
+        .module('app')
+        .controller('OriginalSentencesController', [
+            '$scope', '$window', OriginalSentencesController
+        ]);
+
+    function OriginalSentencesController($scope, $window) {
+        $scope.toggle = function() {
+            var oldQuery = $window.location.search;
+            if ($scope.original) {
+                    var newQuery = oldQuery.replace(/only_original=&?/, '');
+            } else {
+                if (oldQuery.indexOf('?') > -1) {
+                    var newQuery = oldQuery.replace('?', '?only_original=&');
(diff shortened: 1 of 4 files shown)


RyckRichards:
Thanks for your PRs, Andreas!!! :D

>

AndiPersti:
As you will have noticed, this PR currently doesn't pass all tests. Should I only remove the checkbox if the user doesn't exist or the whole sidebar? (see also #2022)

AndiPersti:
@RyckRichards: You're welcome.
I hope reviewing my PRs isn't too laborious. :-) (I'm still learning a lot working on them)

AndiPersti:
Well, I think hiding the checkbox for unknown users is for the moment the best choice.

webroot/js/sentences/only_original.ctrl.js
```diff
@@ -0,0 +1,43 @@
...
+            } else {
+                if (oldQuery.indexOf('?') > -1) {
+                    var newQuery = oldQuery.replace('?', '?only_original=&');
+                } else {
+                    var newQuery = oldQuery + '?only_original=';
+                }
+            }
+            $window.location.search = newQuery;
+        }
+    }
```
trang: You can use [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) to improve a bit this code.

```
var searchParams = new URLSearchParams($window.location.search);
[...]
$window.location.search = searchParams.toString();
```

Technically if I go to http://localhost:8080/eng/sentences/of_user/contributor?only_original (without the `=` at the end), I should be able to uncheck the option and be redirected to the full list of sentence. With your current code, I won't be, because you're replacing specifically `?only_original=&` so the `.replace()` will have no effect.

trang:
> The second commit changes the behavior of the language filter so that an existing query string isn't lost when the language is changed. I personally find that more useful but I will remove it if it is not wanted.

That's fine. It's the preferred behavior.

> Should I only remove the checkbox if the user doesn't exist or the whole sidebar?

Hiding the checkbox is best indeed. We can deal with the inconsistent behavior of the sidebar separately.

> I hope reviewing my PRs isn't too laborious. :-)

Not at all. I wished there were more people like you :)

webroot/js/sentences/only_original.ctrl.js
AndiPersti: > You can use URLSearchParams to improve a bit this code.

Thanks, I've updated the code.

trang:
Just a note about the title.

I was in a situation where added a bunch of original sentences in English, then translated them into French (so I had no original sentences in French). But when filtering by French and original sentences, I had the title "contributor does not have any sentence in French", which was a bit confusing because I do have sentences in French (just not original ones). 

We could adapt the title depending on the language and on the original option:
- {username}'s sentences
- {username}'s original sentences
- {username}'s sentences in {lang}
- {username}'s original sentences in {lang}
- {username} does not have any sentence
- {username} does not have any original sentence
- {username} does not have any sentence in {lang}
- {username} does not have any original sentence in {lang}

...but I feel it's a bit too much.

I merged this pull request because I don't think we need to solve this problem here, but it's still something we need to think about in general.

Issue #1219, opened 2016-06-22
Internationalize date


trang:
Requested by [Pfirsichbaeumchen](http://tatoeba.org/eng/wall/show_message/26614#message_26614).

The date is currently displayed in a format that is not very human friendly. 
For example: `2010-10-22 23:10`

We would ideally want something like: `October 22, 2010 at 23:10`. 
This string should be translatable.

zachleigh:
How are you planning on doing this? I use Carbon (https://github.com/briannesbitt/Carbon) in other projects. Comes with a few dozen languages out of the box and looks like it would be easy to add more to.

trang:
I haven't done much research on whether or not there's already a mechanism in CakePHP for this, and I haven't researched either about existing libs.

AndiPersti:
[CakePHP changed from Carbon to Chronos since 3.2.0](https://book.cakephp.org/3/en/core-libraries/time.html).

Formatting the dates is pretty easy.

Full datetime strings as used for logs, comments, wall messages, ...:
![contributions](https://user-images.githubusercontent.com/6566321/68205820-80736180-ffcb-11e9-8f13-91cad16523c4.png)

![users](https://user-images.githubusercontent.com/6566321/68205836-86694280-ffcb-11e9-9028-e48d6a7c3fee.png)

![wall](https://user-images.githubusercontent.com/6566321/68205837-86694280-ffcb-11e9-942b-28fbdc8aff9d.png)

Just dates as used in the user profiles:
![profile1](https://user-images.githubusercontent.com/6566321/68206232-87e73a80-ffcc-11e9-9d2b-533c7818012f.png)
![profile2](https://user-images.githubusercontent.com/6566321/68206233-87e73a80-ffcc-11e9-81db-3f29f533153d.png)

What about tooltips? I guess they should still show the  ISO 8601 format for combined date and time (e.g. 2012-02-11 03:49:06).

Something to consider: There probably will not be enough room for displaying two nicely formatted datetime strings on one line if the modification date is also displayed for wall messages, comments, ... This problem already exists with the current version in some translations:
![comments1](https://user-images.githubusercontent.com/6566321/68207302-ec0afe00-ffce-11e9-94e7-db41b666e590.png)
![comments2](https://user-images.githubusercontent.com/6566321/68207303-eca39480-ffce-11e9-86e6-216ec59e0bc4.png)

ckjpn:
I personally like the 2019-11-23 02:15 format better.

[...]

PR #2025, opened 2019-11-28
Internationalize dates

AndiPersti:
This PR addresses issue #1219.

This is still unfinished work because at the moment I've just modified the `ago` and `formatBirthday` functions from the Date helper so that they output locale dependent dates and then modified the display of dates for wall messages to demonstrate how the new functionality works. If you like this I will go through the rest of the code.

The implementation enables to differentiate between stand-alone dates and dates in a phrase like "edited April 11, 2014" which is necessary for languages like French or German because these languages need to add some extra words before the date in the latter case ("le" and "am" respectively).

It also shows how another (related) problem [described on the wall](https://tatoeba.org/eng/wall/show_message/33547#message_33547) could be solved: changing the string `{createdDate} edited {modifiedDate}` to `{createdDate} {edited} {modifiedDate}` allows the translator to use the declension syntax. For example, a french translator would translate the latter string to `{createdDate} {edited.masc-sing} {modifiedDate}` and `{edited}` would be translated to `{; masc-sing: modifié; fem-sing: modifiée`. The result would be a grammatically correct output.

I've also added some tests for the Date helper, but I'm not sure about the french language tests, because they are dependent on the po-file. Thus I've commented them out for the moment.

As always I'm looking forward to your comments. :-)

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -268,3 +268,4 @@ if (!function_exists('__format_decompose_list')) {
 
 Cake\I18n\I18n::setDefaultFormatter('sprintf');
 Cake\I18n\Time::setToStringFormat('yyyy-MM-dd HH:mm:ss');
+Cake\I18n\Time::$niceFormat = [\IntlDateFormatter::LONG, \IntlDateFormatter::SHORT];
--- a/src/Template/Collections/of.ctp
+++ b/src/Template/Collections/of.ctp
@@ -151,7 +151,7 @@ $this->set('title_for_layout', $this->Pages->formatTitle($title));
                 echo $this->Html->div(
                     'correctness',
                     $this->Images->correctnessIcon($correctness),
-                    array('title' => $item->modified)
+                    array('title' => $this->Time->nice($item->modified))
                 );
 
                 echo '</div>';
--- a/src/Template/Contributions/activity_timeline.ctp
+++ b/src/Template/Contributions/activity_timeline.ctp
@@ -87,8 +87,9 @@ foreach ($stats as $stat) {
             array('style' => 'width:'.$width.'%')
         );
 
+        $formattedDate = $this->Time->i18nFormat($date, [IntlDateFormatter::SHORT, IntlDateFormatter::NONE]);
         echo '<tr>';
-        echo $this->Html->tag('td', $date, array('class' => 'date'));
+        echo $this->Html->tag('td', $formattedDate, array('class' => 'date'));
         echo $this->Html->tag('td', $this->Number->format($numSentences), array('class' => 'number'));
         echo $this->Html->tag('td', $bar, array('class' => 'bar'));
         echo '</tr>';
--- a/src/Template/Element/sentence_comments/comment.ctp
+++ b/src/Template/Element/sentence_comments/comment.ctp
@@ -56,17 +56,9 @@ $sentenceUrl = $this->Url->build(array(
     'action' => 'show',
     $sentenceId
 ));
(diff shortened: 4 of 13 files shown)


jiru:
Thanks for working on that, Andreas.

I think a string like "{createdDate} {edited} {modifiedDate}" is quite hard to understand. I think it would be more translator-friendly to avoid such complex placeholder substitutions. Correct me if I'm wrong, but since you are using a "wall message" message context, that string becomes specific to wall messages, so translators can directly decline from that context. As a result, the translatable string can stay "{createdDate} edited {modifiedDate}" and we'd have another one for sentences.

About the variant in "ago" context, don't we need a specific variant in English in the first place? Shouldn't "edited 1st of April" be actually "edited on the 1st of April"?

AndiPersti:
> Correct me if I'm wrong, but since you are using a "wall message" message context, that string becomes specific to wall messages, so translators can directly decline from that context. As a result, the translatable string can stay "{createdDate} edited {modifiedDate}" and we'd have another one for sentences.

Yes, you're right. 
I've just tested using the message context alone and it is sufficient. I don't know any more why I thought the placeholder substitution was also necessary. Probably forgetting to update the po-file while testing.

> About the variant in "ago" context, don't we need a specific variant in English in the first place? Shouldn't "edited 1st of April" be actually "edited on the 1st of April"?

Well, that's what I've asked English native speakers a few days ago [on the wall](https://tatoeba.org/eng/wall/show_message/33557#message_33557).
Using "edited on &lt;full date>" instead of just "edited &lt;full date>" as we currently do, would simplify everything because then we won't have a problem distinguishing between "edited &lt;time difference>" and "edited &lt;full date>" as is necessary for at least French and German translations. We would simply return `$formattedDate` for stand-alone dates and `format(__('on {date}', 'date' => $formattedDate))` for dates in a phrase.

AndiPersti:
I think I've modified all displayed dates now.

trang:
I notice that you changed the tooltips to only show the "ago" date label. This actually loses the main purpose of the tooltip, which was to show the exact date and time. Now when a message is `2 days ago`, I would not be able to know when exactly it was posted because the tooltip would also show `2 days ago` (instead of `2019-12-02 12:00:34` for instance).

AndiPersti:
> I notice that you changed the tooltips to only show the "ago" date label. This actually loses the main purpose of the tooltip, which was to show the exact date and time. 

Sorry I've misread your [comment on the issue](https://github.com/Tatoeba/tatoeba2/issues/1219#issuecomment-557788381). 

Just to be sure:
For dates within the 30 day period (".... ago") the tooltip should show the exact datetime in a human-friendly way.
For older dates the tooltip should show the same as the text (exact datetime in a human-friendly way).

trang:
> For dates within the 30 day period (".... ago") the tooltip should show the exact datetime in a human-friendly way.
> For older dates the tooltip should show the same as the text (exact datetime in a human-friendly way).

Yes, that's correct. 

The tooltip should indeed show human-friendly date contrary to my example above.

jiru:
Thanks for solving this issue! Since dates are displayed everywhere on Tatoeba, it will certainly improve UX.

About localization, I'd like to mention that Mozilla has been working on Fluent, an advanced localization format that handles declensions as well as other fancy stuff: https://hacks.mozilla.org/2019/04/fluent-1-0-a-localization-system-for-natural-sounding-translations/

It's interesting to see that they use a similar approach for declensions. Even better, they explicitly link terms with sentences they are to be included in (using identifiers that start with a dash).

Fluent looks like still experimental and not yet supported by Transifex, but I'd love to use that one day.

Issue #1841, opened 2019-03-25
UI not updated properly when adding sentences to list within a short period time


trang:
Reported by [Aiji on the Wall](https://tatoeba.org/eng/wall/show_message/31535#message_31535): 
> - Add two sentences, A and B.
> - Add A to a list L.
> - While A is being adding (the waiting symbol is spinning), click on the "Add to a list" button => It looks like B was added to the list L (because L disappears from the dropdown), while the waiting symbol on A spins forever.
> - Actually, A is correctly added to L (checked by opening L), and B cannot be added anymore unless going to the sentence page of B.
> 
> Expected behavior:
> - Wherever I click during the addition of A to L, the dropdown menu of A, and its waiting symbol, should be the ones affected when the operation is over, not the one with the current focus.

trang:
I realized this bug was actually not solved in #2027.

It is caused by the `sentenceId` being defined as global variable:
https://github.com/Tatoeba/tatoeba2/blob/dev/webroot/js/sentences_lists.menu.js#L22

Issue #1743, opened 2019-01-20
Adding sentence to list doesn't work on pages with multiple sentences


trang:
Reported by [soliloquist on the Wall](https://tatoeba.org/eng/wall/show_message/31166#message_31166):

> I think there's a bug with the add-to-list function on pages with numerous sentences (like list pages or "translate user's sentences" pages). On such pages, sentence selection goes wrong if a previously-selected sentence is selected again. Add-to-list command is erroneously sent to the previous sentence.
> 
> It might sound unclear so I've created a few example sentences to demonstrate.
> 
> https://dev.tatoeba.org/eng/activities/translate_sentences_of/testuser
> 
> Just add those sentences into some lists. Do it in order, not randomly. After getting done with the 3rd one, select the 1st or 2nd one again and try to add it into another list. You'll reproduce the bug.
> 
> This is present on both the main and the dev site.

soliloquist-tatoeba:
Another description.

https://tatoeba.org/eng/wall/show_message/31535#message_31535

> [BUG]

- Add two sentences, A and B.
- Add A to a list L.
- While A is being adding (the waiting symbol is spinning), click on the "Add to a list" button => It looks like B was added to the list L (because L disappears from the dropdown), while the waiting symbol on A spins forever.
- Actually, A is correctly added to L (checked by opening L), and B cannot be added anymore unless going to the sentence page of B.

Expected behavior:
- Wherever I click during the addition of A to L, the dropdown menu of A, and its waiting symbol, should be the ones affected when the operation is over, not the one with the current focus.

PR #2027, opened 2019-12-02
Use correct sentence ID when adding sentences to a list

AndiPersti:
Fix for bugs #1743 and #1841.

If there are several list selection boxes opened on a page we need to use the correct sentence ID when the user clicks on the OK button.

### DIFF
--- a/webroot/js/sentences_lists.menu.js
+++ b/webroot/js/sentences_lists.menu.js
@@ -52,6 +52,7 @@ $(document).ready(function() {
         // directly after the selection in the <select>.
         $(".validateButton").off();
         $(".validateButton").click(function(){
+            sentenceId = $(this).parent().attr("id").replace("addToList", "");
             var listId = $("#listSelection"+sentenceId).val();
             var rootUrl = get_tatoeba_root_url();
 
--- a/webroot/js/sentences_lists.menu.js
+++ b/webroot/js/sentences_lists.menu.js
@@ -52,7 +52,7 @@ $(document).ready(function() {
         // directly after the selection in the <select>.
         $(".validateButton").off();
         $(".validateButton").click(function(){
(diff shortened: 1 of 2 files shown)


webroot/js/sentences_lists.menu.js
```diff
@@ -52,6 +52,7 @@ $(document).ready(function() {
         // directly after the selection in the <select>.
         $(".validateButton").off();
         $(".validateButton").click(function(){
+            sentenceId = $(this).parent().attr("id").replace("addToList", "");
```
trang: A way of getting the sentence ID that feels less of a workaround would be to set the `data-sentence-id` attribute on the `.validateButton` element instead of the `.addToList` element.

This way you can use `sentenceId = $(this).attr("data-sentence-id")` here, and you can remove it from line 45.

AndiPersti: > A way of getting the sentence ID that feels less of a workaround would be to set the `data-sentence-id` attribute on the `.validateButton` element instead of the `.addToList` element.

Done.

> This way you can use `sentenceId = $(this).attr("data-sentence-id")` here, and you can remove it from line 45.

I don't think we can change line 45 because we need the sentence ID for toggling the correct list selection box.
And I think we don't need jQuery for reading the data attribute.

trang: > I don't think we can change line 45

True.

> And I think we don't need jQuery for reading the data attribute.

Yes, I think we can go without jQuery. It could be an issue for people who use Internet Explorer with a version below 11 but that's almost no one among our visitors and that's probably no one among our members who use the list feature.

https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

PR #2028, opened 2019-12-02
If the Clipboard API is available, checks if the clipboard content is…

### LINKED ISSUE #2017: Paste link from clipboard on click on "link to another sentence"
Posted by Inego - https://tatoeba.org/eng/wall/show_message/33545#!#message_33545

It would be really convenient if pressing the "link to another sentence" button also automatically pasted the text from the clipboard to the text field, provided it satisfies the regular expression used to validate it.

Inego:
… a number or a string ending with a slash and a sentence number. If yes, this number is automatically pasted and selected in the link input field.

Closes #2017.

### DIFF
--- a/webroot/js/sentences.link.js
+++ b/webroot/js/sentences.link.js
@@ -114,6 +114,11 @@ function linkToSentence(sentenceId, langFilter) {
                 inputField.val(inputText);
                 inputField.select();
             }
+        }, exception => {
+            if (exception instanceof DOMException) {
+                if (exception.name === "NotAllowedError") return;
+            }
+            throw(exception);
         });
         
         inputField.focus();
--- a/webroot/js/sentences.link.js
+++ b/webroot/js/sentences.link.js
@@ -99,27 +99,36 @@ function linkToSentence(sentenceId, langFilter) {
         linkTo.show();
         var inputField = $("#linkToSentence" + sentenceId);
 
-        navigator.clipboard && navigator.clipboard.readText && navigator.clipboard.readText().then(function(clipText) {
-            var inputText;
-            clipText = clipText.trim();
-            if (/^\d+$/.test(clipText)) {
-                inputText = clipText;
-            } else {
-                inputText = URLToSentenceId(clipText);
-                if (clipText === inputText) {
-                    inputText = undefined;
+        if (navigator.clipboard && navigator.clipboard.readText) {
+            navigator.permissions.query({name:'clipboard-read'})
+            .then(function(permissionStatus) {
+                if (permissionStatus.state === 'prompt') {
+                    alert("In order to allow automatic pasting of sentence numbers and links from the clipboard, the corresponding permission is required.");
                 }
-            }
-            if (inputText) {
-                inputField.val(inputText);
-                inputField.select();
-            }
-        }, function(exception) {
-            if (exception instanceof DOMException) {
-                if (exception.name === "NotAllowedError") return;
-            }
-            throw(exception);
-        });
+                return navigator.clipboard.readText();
+            })
+            .then(function(clipText) {
+                var inputText;
+                clipText = clipText.trim();
+                if (/^\d+$/.test(clipText)) {
+                    inputText = clipText;
+                } else {
+                    inputText = URLToSentenceId(clipText);
+                    if (clipText === inputText) {
+                        inputText = undefined;
+                    }
+                }
+                if (inputText) {
+                    inputField.val(inputText);
+                    inputField.select();
+                }
+            }, function(exception) {
+                if (exception instanceof DOMException) {
+                    if (exception.name === "NotAllowedError") return;
+                }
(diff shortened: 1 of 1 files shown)


webroot/js/sentences.link.js
```diff
@@ -97,6 +97,30 @@ function linkToSentence(sentenceId, langFilter) {
         linkTo.hide();
     } else {
         linkTo.show();
-        $("#linkToSentence" + sentenceId).focus();
+        var inputField = $("#linkToSentence" + sentenceId);
+
+        navigator.clipboard && navigator.clipboard.readText().then(clipText => {
```
trang: Welcome to the dev side of Tatoeba and thanks for taking care of implementing your own feature request :)

I have one issue on my side: on Firefox, this triggers the error `TypeError: navigator.clipboard.readText is not a function` and the rest of the code is not executed so the input is no longer focused upon clicking on the link icon.

Inego: How unlucky!
https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/readText says:

> Firefox only supports reading the clipboard in browser extensions, using the "clipboardRead" extension permission.

That means that Firefox essentially doesn't support reading from clipboard for Tatoeba, which leaves only Chrome and Opera that will benefit from this enhancement.

Inego: The commit below fixes the input focus issue with Firefox.

jiru:
Inego, welcome and thanks for your pull request ^^

On the code part, Javascript arrow functions [do not work in IE](https://caniuse.com/arrow-functions), could you replace them with regular anonymous functions?

On the UX part, I am bit skeptical about the permissions popup about clipboard read here. The connection between linking sentences and clipboard read is very unobvious, so we’d rather provide some explanation, especially for first-time users of the linking feature. I’m thinking a about a message among the lines of "Tatoeba can read the sentence URL directly from your clipboard when you copy and paste sentence URLs. If you’d like this, please allow."

Inego:
> On the code part, Javascript arrow functions [do not work in IE](https://caniuse.com/arrow-functions), could you replace them with regular anonymous functions?

I can, but is the arrow function branch reachable in IE in the first place?

> On the UX part, I am bit skeptical about the permissions popup about clipboard read here. The connection between linking sentences and clipboard read is very unobvious, so we’d rather provide some explanation, especially for first-time users of the linking feature. I’m thinking about a message along the lines of "Tatoeba can read the sentence URL directly from your clipboard when you copy and paste sentence URLs. If you’d like this, please allow."

Good point, I also thought about this. I wonder what the appropriate place to to display that message could be.

Inego:
Ah, I see. Indeed, if there is an API to check whether the clipboard read permission was not yet asked, then we can display a warning / explanation popup before going for that clipboard text.

jiru:
How about a good old `alert()` or `confirm()` the first time you click on the link button?

Or, if there is a way to detect that the user pasted something (as opposed to typed something), then, once the user clicks OK to perform the linking, display a popup like "It looks like you pasted something here, how about we prefill the field with the contents of the clipboard for you in the future?"

trang:
> I can, but is the arrow function branch reachable in IE in the first place?

The function branch is reachable in IE. When loading a sentence's page, I'm getting this error (tested on IE 11):
```
SCRIPT1002: Syntax error
sentences-block-for-members.js (625,110)
```
Clicking on the link button actually won't even show the text input.


> I wonder what the appropriate place to display that message could be.

An `alert()` popup, like @jiru suggested, would be good enough for a start. I'm not too fond of having a popup on top of the popup to accept/deny the permission to paste from the clipboard, but we can think of better ways later, when the link feature is migrated to the new sentence design.


> It seems the permissions API is even more limited. There's no info whether Opera supports it.

If a browser doesn't provide information regarding the copy-paste permissions, then you can just skip displaying the alert message.

jiru:
About checking whether the permission was not yet asked, I don't think there is an API for that. Maybe you could store that information as a user setting. See commit 2942742 for an example of how to add a new (hidden) user setting.

Inego:
I've added an alert before reading from clipboard, if the permission state is "Prompt".
This works well in Chrome.
Firefox, IE11 and Edge simply ignore the whole clipboard thing.
However, Opera confuses me. It keeps showing the alert all the time, but never asks for permission and so the permission state stays "Ask" forever). If I go to site settings and enable / disable clipboard, everything works. Maybe Opera doesn't ask for Clipboard access because of HTTP in local imouto installation. I don't know how to enable HTTPS inside vagrant to check whether that's really the case.

trang:
The problem with Opera is actually due to the alert popup, it's not related to HTTPS. I temporarily deployed this branch on the dev website where HTTPS is set up, and the problem remains. When I tried removing the alert, the prompt to allow/deny clipboard access did appear properly.

I will replace the alert by a message displayed below the sentence menu.

Issue #2006, opened 2019-11-10
CSRF token mismatch editing transcription after linking


Yorwba:
1. Enable "Always show transcriptions and alternative scripts" in settings.
2. Find a sentence that is indirectly linked to another sentence with an editable description.
![](https://user-images.githubusercontent.com/6555947/68544406-2a376180-03c3-11ea-93e4-fd2126695f1c.png)
3. Link the two sentences.
![](https://user-images.githubusercontent.com/6555947/68544417-43401280-03c3-11ea-9cd3-cae8a5ccb758.png)
4. Edit the transcription.
![](https://user-images.githubusercontent.com/6555947/68544427-5f43b400-03c3-11ea-9650-e30963c09eec.png)
5. Submit.
![](https://user-images.githubusercontent.com/6555947/68544574-dd548a80-03c4-11ea-92a0-b07246b19028.png)

The error message does not appear when the page is reloaded after linking. I don't know how CSRF protection works in detail, but I'm guessing that linking the sentences should update the CSRF token to avoid sending an outdated one with the edited transcription.

PR #2029, opened 2019-12-02
Always get the CSRF token from cookie in JS

Yorwba:
Debugging story on the CSRF token mismatch in #2006:
It was caused by the transcription missing the `data-csrf-token` attribute, which led to the subsequent request missing the `X-CSRF-Token` header. The `data-csrf-token` attribute was missing because it was set from the `_csrfToken` parameter, which wasn't being sent for `add` and `delete` actions on links. Those two actions are apparently intentionally excluded from CSRF protection, which is why the lack of CSRF token didn't produce an error.

My first attempt to fix this was to simply send the `_csrfToken` parameter with link changes. That didn't work. CakePHP (I guess?) completely removes that parameter if CSRF protection is disabled.

Then I realized that using hidden attributes to pass the token around is actually much too complicated, since the CSRF token is stored as a cookie. Hidden attributes are only necessary for forms that are submitted without JavaScript.

For consistency, I changed all functions where the CSRF token was accessed in JavaScript to use the cookie instead. I tested all affected endpoints at least once and didn't see any CSRF mismatches, so I probably didn't break anything.

### DIFF
--- a/webroot/js/generic_functions.js
+++ b/webroot/js/generic_functions.js
@@ -32,8 +32,11 @@ function get_tatoeba_root_url() {
 }
 
 function get_csrf_token() {
-    // Probably not worth importing jquery.cookie for this.
-    return document.cookie.replace(/^.*csrfToken=([^;]*);.*$/, "$1");
+    // Adapted from https://stackoverflow.com/a/15724300
+    // Licensed under CC BY-SA 4.0
+    var value = "; " + document.cookie;
+    var parts = value.split("; csrfToken=");
+    if (parts.length == 2) return parts.pop().split(";").shift();
 }
 
 /**
--- a/webroot/js/generic_functions.js
+++ b/webroot/js/generic_functions.js
@@ -31,6 +31,11 @@ function get_tatoeba_root_url() {
     return "//" + host + "/"+ interfaceLang;
 }
 
+function get_csrf_token() {
+    // Probably not worth importing jquery.cookie for this.
+    return document.cookie.replace(/^.*csrfToken=([^;]*);.*$/, "$1");
+}
+
 /**
  * Returns string without the new lines and stuff.
  */
--- a/webroot/js/sentences.add_translation.js
+++ b/webroot/js/sentences.add_translation.js
@@ -60,7 +60,6 @@ $(document).ready(function() {
                         'selectLang': selectLang,
                         'value': sentenceText
                     };
-                    var csrfHeader = $('#translation-form [name="_csrfToken"]').val();
                     $('#translation-form input[name^="_Token"]').each(function() {
                         data[$(this).attr('name')] = $(this).val();
                     });
@@ -70,7 +69,7 @@ $(document).ready(function() {
                         url: rootUrl + '/sentences/save_translation', 
                         data: data,
                         headers: {
-                            'X-CSRF-Token': csrfHeader
+                            'X-CSRF-Token': get_csrf_token()
                         },
                         success: function(data) {
                             translating = false;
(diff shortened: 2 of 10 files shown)


webroot/js/vocabulary/add-sentences.ctrl.js
```diff
@@ -67,7 +66,7 @@
                 method: 'POST',
                 url: rootUrl + '/vocabulary/save_sentence/' + id,
                 headers: {
-                    'X-CSRF-Token': csrfHeader
+                    'X-CSRF-Token': get_csrf_token()
```
jiru: I havn’t checked that on my own, but I think this actually shouldn’t be needed at all because Angular already sets the X-CSRF-Token header from the cookie automatically, as specified in webroot/js/responsive/app.module.js (xsrfHeaderName and xsrfCookieName).

webroot/js/vocabulary/add.ctrl.js
```diff
@@ -46,7 +45,7 @@
                 method: 'POST',
                 url: rootUrl + '/vocabulary/save',
                 headers: {
-                    'X-CSRF-Token': csrfHeader
+                    'X-CSRF-Token': get_csrf_token()
```
jiru: Same here.

webroot/js/sentences.add_translation.js
```diff
@@ -70,7 +69,7 @@ $(document).ready(function() {
                         url: rootUrl + '/sentences/save_translation', 
                         data: data,
                         headers: {
-                            'X-CSRF-Token': csrfHeader
+                            'X-CSRF-Token': get_csrf_token()
```
jiru: Same here.

jiru: Whoops, no, forget about that comment.

jiru:
@Yorwba I havn’t checked if it works, but I think your approach of getting the token from the cookie is good. Once we eventually replace all the jquery code with AngularJS equivalent, it should be much more simple as AngularJS does that for us.

webroot/js/generic_functions.js
```diff
@@ -31,6 +31,11 @@ function get_tatoeba_root_url() {
     return "//" + host + "/"+ interfaceLang;
 }
 
+function get_csrf_token() {
+    // Probably not worth importing jquery.cookie for this.
+    return document.cookie.replace(/^.*csrfToken=([^;]*);.*$/, "$1");
```
trang: Just one small issue on my side: this regex won't work if the csrfToken is defined at the end of `document.cookie`. 

For instance my cookie looked like this:
`CakeCookie=%7B%22interfaceLanguage%22%3A%22eng%22%7D; csrfToken=8ccb558280c7ad3e0729dcbf0720fd78c23305a9b559877fe8ab534216f482004e9e6991e425155aec6f5f44a4617a633501b005ee354a7a26755367f52bb015`

and therefore the function returned the whole string.

Yorwba: Good that you caught this. I hadn't considered that there's no semicolon after the value if it's the last entry. Should I just include `jquery.cookie` on all pages? It's 3121 bytes, so probably won't noticeably slow things down.

trang: Eventually we should try to remove jQuery (moving everything to angularjs), so it's probably best not to include jquery.cookie.

I would say just reuse one of the solutions suggested on Stack Overflow:
https://stackoverflow.com/questions/10730362/get-cookie-by-name
The most upvoted answer should be fine.

Yorwba: Done. I wasn't sure about the legality of including a CC BY-SA 4.0 StackOverflow answer in an AGPL codebase, but CC BY-SA 4.0 is declared compatible with inclusion in GPLv3 and GPLv3 allows combining with AGPL, so this should be fine.

trang: Thanks :)

Issue #1911, opened 2019-06-27
Make seeing collaborative lists when clicking the list icon an opt-in option in members' settings.


ckjpn:
## Suggested Enhancement

Make seeing collaborative lists when clicking the list icon an opt-in option in members' settings.

The following could be the setting description, with the checkbox unchecked by default.

Show collaborative lists when clicking the list icon

## The Reason

* Eliminating this extra data would likely improve the user experience, since pages would load faster
* There are many collaborative lists that are not being used.
* Perhaps most members never collaborative lists at all.
* So, a lot of data is being sent from the server that is never being used.

## Note

If a member is browsing 100 sentences at a time, the following code is duplicated 100 times in the page.

The following code is 22,000 bytes 

```
<optgroup label="Add to a collaborative list">
<option value="992"> # !! 0 FILTER (public list) - XXX, Culture of violence, Weaponry, Gun culture, etc. (Inappropriate for Young Children)</option>
<option value="6809">¨ es→pt (em progresso)</option>
<option value="6807">¨en→es - em progresso</option>
<option value="6808">¨es→en (em progresso)</option>
<option value="6701">¨pt→es - em progresso</option>
<option value="6722">- MEMRISE &#039;  VER OUTROS IDIOMAS -- </option>
<option value="1278">!-alexmarcelo espanhol = temporário</option>
<option value="2953">!-bunbuku-jpn=possibly</option>
<option value="2954">!-bunbuku-jpn=possibly-not</option>
<option value="6678">¿¿ Adotar ??</option>
<option value="6426">.......... .........</option>
<option value="5006">&quot;とんでもない&quot;</option>
<option value="113">[@moderator] Sentences with wrong reading for kanji</option>
<option value="2">[Chinese]HSK elementary</option>
<option value="1239">[Qualified Chinese Translation for Esperanto Sentences][带有中文翻译的世界语句子]</option>
<option value="981">[rus] Вниманию ответственного за корпус.</option>
<option value="1171">[sadhen][GRE]</option>
<option value="1069">[中文][文艺]</option>
<option value="1068">[中文][新词][网络用语]</option>
<option value="3131">* Eng = Port DIRECT translations II</option>
<option value="3459">** Adopt!</option>
<option value="6649">\\ Parafrasear //</option>
<option value="512"># temo/thème/topic: astronomio/astronomie/astronomy</option>
<option value="476"># topic/temo: animals/bestoj</option>
[...]

PR #2035, opened 2019-12-09
Avoid listing lists in the "Add to list" selection box which already contain the sentence

AndiPersti:
While working on #1743 and #1841 I've noticed that there may be lists in the "Add to list" selection which already contain the sentence we want to add. Thus I've added a clause to the query which filters out these lists.

I'm not sure about how to show to the user that the sentence is already in all possible lists. Using the optiongroup feels rather hackish. Maybe disable the "Add to list" button in that case?
(This scenario is currently very unlikely with all the collaborative lists. But after implementing #1911 it will be more realistic.)

### DIFF
--- a/src/Model/Table/SentencesListsTable.php
+++ b/src/Model/Table/SentencesListsTable.php
@@ -140,11 +140,14 @@ class SentencesListsTable extends Table
     /**
      * Returns the sentences lists that the given user can add sentences to.
      *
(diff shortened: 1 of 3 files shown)


trang:
> I'm not sure about how to show to the user that the sentence is already in all possible lists. Using the optiongroup feels rather hackish.

It's good enough for the time being :) The list feature will be redesigned anyway.

Issue #2033, opened 2019-12-09
I would suggest reverting the date format to yyyy-mm-dd on the activity_timeline.


ckjpn:
yyyy-mm-dd is much more internationally understood.

Perhaps this was an unintentional change.

<img width="210" alt="Screen Shot 2019-12-09 at 9 00 52" src="https://user-images.githubusercontent.com/7934273/70398854-c4da9d00-1a62-11ea-9905-93189907a141.png">

AndiPersti:
Have you looked at the other language versions of that page? Do you want to revert the date format only for the English version or in general?

I have no problem reverting the date format. While working on that page I've tried different versions and also thought about wether to change the date format at all. But I would like to hear more opinions.

jiru:
@AndiPersti I see no reason not to localize the date here.

@ckjpn Note that the date is now localized according to the interface language, thanks to AndiPersti’s recent localization efforts. So it shouldn’t be necessary to use a "more internationally understood" format any more.

That said, as for the English interface, the date format [varies between regions](https://en.wikipedia.org/wiki/Date_format_by_country). Since we don’t provide per-region locales, it might be better to use yyyy-mm-dd for English, if it is the most commonly understood format *among English speaking countries*.

AndiPersti:
A little background information:
The date/time formatting is based on the [PHP class IntlDateFormatter](https://www.php.net/manual/en/class.intldateformatter.php) which itself uses data from the [Common Locale Data Repository (CLDR)](http://cldr.unicode.org/), a project by the Unicode Consortium. They define 4 formats: full, long, medium and short. (for how they are defined for English see https://www.unicode.org/cldr/charts/36/summary/en.html#1989). 
I've mainly chosen the short format because it uses the least amount of space.

trang:
I would suggest medium format without the year.
```
Dec 1
Dec 2
Dec 3
etc
```

The year is already indicated in the title so there's no need to repeat it.

The month is also indicated in the title but having a number all alone might not be very explicit so it's okay to repeat it.

Poulpisator:
I would suggest to use the day of the week. Currently, I don't find this page useful (personal opinion). However, looking at December 2019, for example, I saw that the 6th had much more contributions (more than 1,000 compared to the average) so I implied it was a Saturday or a Sunday. Looking at a calendar, I realized it was actually a Friday. So now, I'm curious...

Long story short, I personally think this page would provide more information if the day of the week was displayed 
```
[...]

PR #2037, opened 2019-12-14
Some I18n changes

AndiPersti:
This PR addresses #2033.

Change date format on activity timeline and add some missing localized number formatting where years, days, hours and minutes are displayed as numbers.

### DIFF
--- a/src/Template/Contributions/activity_timeline.ctp
+++ b/src/Template/Contributions/activity_timeline.ctp
@@ -73,7 +73,6 @@ foreach ($stats as $stat) {
     <?php
     echo '<table id="timeline">';
 
-    $currentDate = null;
     $totalSentences = 0;
     $numberOfDays = 0;
 
@@ -87,7 +86,7 @@ foreach ($stats as $stat) {
             array('style' => 'width:'.$width.'%')
         );
 
-        $formattedDate = $this->Time->i18nFormat($date, [IntlDateFormatter::SHORT, IntlDateFormatter::NONE]);
+        $formattedDate = $this->Date->date_for_timeline($date);
         echo '<tr>';
         echo $this->Html->tag('td', $formattedDate, array('class' => 'date'));
         echo $this->Html->tag('td', $this->Number->format($numSentences), array('class' => 'number'));
--- a/src/Template/Contributions/activity_timeline.ctp
+++ b/src/Template/Contributions/activity_timeline.ctp
@@ -41,7 +41,9 @@ $this->set('title_for_layout', $this->Pages->formatTitle(__("Activity timeline")
 $monthName = $this->Date->monthName($month);
 $selectedMonth = format(
     __('{month} {year}'),
-    array('month' => $monthName, 'year' => $year)
+    array('month' => $monthName,
+          'year' => $this->Number->format($year, array('pattern' => '####'))
+    )
 );
  
 $maxWidth = 400;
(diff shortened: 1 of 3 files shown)


jiru:
> $formatter = datefmt_create($locale, NULL, NULL, NULL, NULL, 'EEE MMM d');
> $dateObj = date_create_from_format('Y-m-d', $date);
> list($weekday, $month, $day) = explode(' ', datefmt_format($formatter, $dateObj));

I'm afraid this is going to break if the translated terms include a space character (Lojban?) There must be some less hackish way to format the date.

I wouldn't be surprised if some languages had various ways to combine these three elements depending on the day number, the day of the week etc. that makes it impossible to express it with something as simple as '{weekday}, {month} {day}'. For example, think about how English decline the day number: 1st, but then 2nd, and then 3rd, before 4th, 5th etc. So I think it would be better if you could come up with a formatting solution that rely on localization libraries as much as possible.

AndiPersti:
> I'm afraid this is going to break if the translated terms include a space character (Lojban?)

I've just noticed that the date localization for Lojban doesn't work at all because it is not in the CLDR. So I've looked at all the other UI languages and Latin (lat) and Occitan (oci) aren't supported either.

> There must be some less hackish way to format the date. [...] So I think it would be better if you could come up with a formatting solution that rely on localization libraries as much as possible.

I've looked a little big deeper and found out that CLDR actually has so-called ["skeleton" formats](http://cldr.unicode.org/translation/date-time-1/date-time-patterns#TOC-Additional-Date-Time-Formats). They start [at line 1148 of the CLDR chart](https://www.unicode.org/cldr/charts/36/summary/root.html#1148) and I think the format [at line 1162](https://www.unicode.org/cldr/charts/36/summary/root.html#1162) is what we want.
Now the bad part: Using these skeleton formats depends on the `DateTimePatternGenerator` class in the C++ library which currently isn't available from PHP's `Intl` extension (see this [bug report](https://bugs.php.net/bug.php?id=70377)). There seems to be a [third party extension](https://github.com/ksimka/intl_dtpg) but I haven't tried it yet.
(The above is mostly a summary from https://blog.ksimka.com/a-long-journey-to-formatting-a-date-without-a-year-internationally-with-php/#header)

@Trang you can remove this PR from the milestone because I don't think I will be able to resolve this in the next days or weeks.

AndiPersti:
Just a short update:

I've found out that it would be possible to work around the missing `DateTimePatternGenerator` wrapper and use the [`MessageFormatter`](https://www.php.net/manual/en/class.messageformatter.php) instead which accepts a skeleton pattern. But unfortunately this needs at least version 64 of the ICU library and the version included in Debian stretch is 57 (even upgrading to buster would only get us to 63).

So I'm afraid this PR won't be mergeable in the near future.

(I've written a small C demo using the `MMMEd` skeleton. Running it on my computer outputs:

    ICU  version: 65.1 prefix: /usr
    # gcc  -I/usr/include  -L/usr/lib -licui18n -licuuc -licudata   -licuio   -o msgfmt_test msgfmt_test.c
    # env "LD_LIBRARY_PATH=/usr/lib:${LD_LIBRARY_PATH}"  ./msgfmt_test en fr
    Tue, Dec 17
    mar. 17 déc.

Inside the dev VM it outputs:

    ICU  version: 57.1 prefix: /usr
    # gcc  -I/usr/include  -L/usr/lib/x86_64-linux-gnu -licui18n -licuuc -licudata   -licuio   -o msgfmt_test msgfmt_test.c
    # env "LD_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu:${LD_LIBRARY_PATH}"  ./msgfmt_test en fr
     ::DecTue17
     ::déc.mar.17
 )

Issue #1948, opened 2019-08-30
Comment was added with date unknown


trang:
I received an email notification for a comment that has been saved without a date. 
Note: the date and time of the comment should be around 2019-08-30 15:12:00 (UTC).

https://tatoeba.org/eng/sentences/show/8135701#comment-1121167

![Screenshot_2019-08-30 English example sentence To those who don't know him, Mennad Bakir is a sweet and charming man - Tato](https://user-images.githubusercontent.com/221850/64036657-fda9d280-cb53-11e9-9998-1d70c33d2e28.png)

AndiPersti:
Other comments in the database with an unknown date:
https://tatoeba.org/eng/sentences/show/4012427#comment-1106349
https://tatoeba.org/eng/sentences/show/8055150#comment-1138486
https://tatoeba.org/eng/sentences/show/686252#comment-1136398

The problem exists for all UI languages which don't use arabic numbers (currently [Bangla](https://dev.tatoeba.org/wall/show_message/31684#message_31684), [Marathi](https://dev.tatoeba.org/wall/show_message/31683#message_31683) and [Arabic](https://dev.tatoeba.org/wall/show_message/31681#message_31681)) and at least for adding sentence comments and wall messages.

#1881 may be related.

AndiPersti:
@Trang I'm afraid we have a bigger problem. If you use the Arabic UI while adding sentences the date also isn't registered. Here's an excerpt from the current database dump (`sentences_detailed.csv`):

    $ grep '0000-00-00' sentences_detailed.csv | tail
    8167498 ara     دخلت على زوجتي فوجدتها وحيدة.   Tariqidufqqir   0000-00-00 00:00:00     0000-00-00 00:00:00
    8167501 ara     إنه المطر!      Tariqidufqqir   0000-00-00 00:00:00     0000-00-00 00:00:00
    8167506 ara     هطل المطر.      Tariqidufqqir   0000-00-00 00:00:00     0000-00-00 00:00:00
    8167508 ara     رآيت تلك الفتاة؟ وجهها يلمع.    Tariqidufqqir   0000-00-00 00:00:00     0000-00-00 00:00:00
    8167517 ber     Uckiɣ-nn / usiɣ-nn.     Tariqidufqqir   0000-00-00 00:00:00     0000-00-00 00:00:00
[...]

PR #2052, opened 2020-01-01
Improve handling of 'datetime' database columns

AndiPersti:
In most parts we still used strings for the content of `datetime`
columns. But in combination with the `Timestamp` behavior and locales
which don't use arabic numerals (in our current UI Arabic, Bangla and
Marathi) this is broken because CakePHP internally represents
`Timestamp` dates as `FrozenTime` objects and when these get converted
to strings, the locale numerals are used. MySQL can't use these
strings and as a consequence the dates are stored as
`0000-00-00 00:00:00` in the database.

#1948 demonstrated this problem.

I've changed all problematic tables to use `FrozenTime` objects for
representing dates. There are still tables which use strings for dates
(mainly `PrivateMessages` and the tables handling contributions) but
since we currently don't use the `Timestamp` behavior there they still
work and I left them as is (it's probably a good idea to change them in
the near future).

I've also added tests and as a bonus fixed some function documentation
comments. :-)

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -269,3 +269,5 @@ if (!function_exists('__format_decompose_list')) {
 Cake\I18n\I18n::setDefaultFormatter('sprintf');
 Cake\I18n\Time::setToStringFormat('yyyy-MM-dd HH:mm:ss');
 Cake\I18n\Time::$niceFormat = [\IntlDateFormatter::LONG, \IntlDateFormatter::SHORT];
+Cake\I18n\FrozenTime::setToStringFormat('yyyy-MM-dd HH:mm:ss');
+Cake\I18n\FrozenTime::$niceFormat = [\IntlDateFormatter::LONG, \IntlDateFormatter::SHORT];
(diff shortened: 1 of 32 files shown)


AndiPersti:
I've tried to test the changes as thoroughly as possible and all the checks pass. Nevertheless I suggest to put it on dev and do some more testing.

@jiru After changing the Sentences Table some test for the SentenceDerivation shell broke. I've noticed that it changed the `modified` date of sentences to `0000-00-00 00:00:00` because of
https://github.com/Tatoeba/tatoeba2/blob/3b73a868bb1fd8d193b4d4f654e3183b504dc2a3/src/Shell/SentenceDerivationShell.php#L191-L193
I guess we should set the current date here (at least that's what I did), shouldn't we?

trang:
No one reported issues so I will merge.

Related Wall thread: https://tatoeba.org/eng/wall/show_message/33893#message_33893

For the SentenceDerivationShell, I'm not entirely sure that the `modified` field should be set to the current date but we can review that later on. We don't use this shell on a regular basis so it is not blocking to merge this pull request.

Issue #2023, opened 2019-11-24
Make user sidebar menu collapsible


AndiPersti:
The sidebar menu on a user page is pretty long. Thus the language filter element (and other similar filter elements) on some user pages (Sentences of, Audio of, Translate sentences of and Profile) is only accessible after scrolling down a bit. How about making the sidebar menu collapsible?

Poulpisator:
Yes, please! And I also suggest to make it collapsed by default. If I check a user page, may it be its profile, its sentences, etc., it is more important to me to see information relative to this page than a list of links to other pages. 

I realized that there was an option to only show orphan sentences only because I saw the sentence untranslated in Transifex... (Yes, I know it's very recent and maybe I would have realized after a while. But maybe not)

Poulpisator:
It doesn't seem intuitive that a click on the username will bring a whole new menu. Maybe "More of <user>'s contributions" or something like that would make it clearer. @alanfgh Any suggestion of a good, concise string?

I'm not sure my second point is due to the new code but I only realized it when I checked the new page. Please have a look at this image https://imgur.com/a/HX6OLYv (using Firefox)  
You can see that the page is not correctly resized on small windows (simply split my screen in two) and that all menus exceeds the limit of the Tatoeba background. That is the case of all menus of all pages. As I said, I don't know if it's linked to the new code but it impacts it, since the arrow of the new menu is not visible if you don't slide the page to the right (making my first point even more problematic).

AndiPersti:
> It doesn't seem intuitive that a click on the username will bring a whole new menu. Maybe "More of 's contributions" or something like that would make it clearer. @alanfgh Any suggestion of a good, concise string?

Just some ideas:
- Use "xxx's Pages" or "More [info] about xxx" as the text
- Expand the menu automatically when the cursor hovers over the element (like in the top bar; won't work on mobile)
- Move the arrow to the front

> You can see that the page is not correctly resized on small windows (simply split my screen in two) and that all menus exceeds the limit of the Tatoeba background. That is the case of all menus of all pages. As I said, I don't know if it's linked to the new code but it impacts it, since the arrow of the new menu is not visible if you don't slide the page to the right (making my first point even more problematic).

As you've noticed yourself this happens on all pages with a sidebar so this should be a new issue.
(I guess this will be handled by the [Responsive UI project](https://github.com/Tatoeba/tatoeba2/projects/3). See also #1180 (especially phase 3).)

[...]

PR #2056, opened 2020-01-04
Make user sidebar menu collapsible

AndiPersti:
This PR addresses #2023

### DIFF
--- a/src/Template/Element/users_menu.ctp
+++ b/src/Template/Element/users_menu.ctp
@@ -26,6 +26,7 @@
  */
 use App\Model\CurrentUser;
 
+$this->Html->script('elements/user-menu.ctrl.js', ['block' => 'scriptBottom']);
 $menu = [
     [
         'label' => __('Profile'),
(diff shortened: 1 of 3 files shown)


AndiPersti:
[deniko raised a related issue on the wall](https://tatoeba.org/deu/wall/show_message/33862#message_33862) which this PR currently doesn't address.

The problem with the reflow could be solved by either adding the `ng-cloak` directive to the language filter element or to the whole sidebar (the latter would also improve other pages which have additional elements below the user menu in addition to the language filter, e.g. the new original sentence checkbox on the user's sentences page).
But the negative effect of `ng-cloak` would be that the whole sidebar won't show up for a few seconds (similar to the delay of the search bar at the top). IMHO that would be acceptable and we probably need to work on the rather slow performance of the AngularJS code anyways.

trang:
I wouldn't consider this design definitive but I think it's still an improvement compared to the very long menu that we currently have. I'll go ahead and merge this.

**Some notes**

My problem here is the same as with the "Compose" button during the [redesign of the private messages](https://tatoeba.org/eng/wall/show_message/33564#message_33564). It feels out of place to have something that looks like a button that is just floating around.

It would be visually more in line with our design patterns if the menu was showing a few items and had a "More" button to show the rest of the items and "Less" to hide them again (similar to the sentence component). But we would then have to decide which items should be displayed by default.

For deniko's issue, I also think it's okay to add ng-cloak on the sidebar.

RyckRichards:
@AndiPersti - May I know your username on Tatoeba?

AndiPersti:
https://tatoeba.org/eng/user/profile/rumpelstilzchen

Issue #72, opened 2014-02-17
language selected for translation is not changed inside a page


loolmeh:
for the moment it is stored in the session, which mean that it will not work if you

in a page with several sentences:
translate one sentence:
choose language X rather than the previous language Y language (by default autodetect)

translate an other
language of translation is still on Y

solution :
javascript => when you validate, it check the selected language, and change all "language of translation" select in the page 

---

Original ticket: https://www.assembla.com/spaces/tatoeba2/tickets/318

trang:
Note: this is fixed in the new sentence design. Since the plan is to completely transition to the new sentence design in the next couple of months, it is not necessary to fix it in the old design.

PR #2058, opened 2020-01-04
Translating sentences in the new sentence design

trang:
This PR implements the translation feature in the new sentence design. It also solves #72.

**Sentence menu**

- Even though only the "Translate" feature is implemented, all the buttons are already displayed. Each user should see the same buttons as they would see in the old sentence design. However, only the "Translate" button is enabled for now.
- I felt it was too much use of vertical space to create one more line for the sentence menu, so I put the menu in on the same level as the sentence header (with the id and owner). The drawback is that not all the elements can be displayed so I made the menu expandable.
- Currently, the two buttons that will always be displayed when the menu is collapsed are the "Translate", "Edit" (if the user can edit) and "List" buttons. The other buttons are always hidden by default.
- This behavior isn't going to be optimal in all situations, for instance when people are in the process of adopting or rating sentences. This is for now just a beginning of implementation to get people used to the idea of having an collapsed menu. I will think of a solution when migrating more features to the new sentence design.

**Adding translations**

For the most part, adding a translation will feel the same as in the old sentence design:
- If the user has no language in their profile, a warning message appears with a button that redirects to the page to add a language. Otherwise, all the translations are hidden and only the form is displayed.
- The text input is focused when the form is displayed.
- The translation can be submitted when pressing enter from the text input.
- The language is preselected with what was selected in the last submitted translation.
- There's a "Cancel" button to close the form. The text and language will remain there if the user clicks again on the "Translate" button.
- The translation can be edited right after being submitted.

Some differences:
- To edit a new translation, instead of clicking on the sentence itself, there's an edit button.
- This button opens a form that looks almost the same as the translation form, with the only difference that the submit button instead says "Edit translation".
- This form allows to edit both the text and the language, so changing the language is no longer done by clicking on the language icon, but by clicking on this edit button.
- In case the user edits a translation and then clicks on the "Translate" button before saving the edited translation, the edited information will be erased. In the old sentence component, the information would still be there. I don't think this is a big loss since a user who chooses to add a new translation before submitting an edited translation most likely didn't want to save the changes anyway.

### DIFF
--- a/src/Template/Element/sentences/translation.ctp
+++ b/src/Template/Element/sentences/translation.ctp
@@ -13,7 +13,7 @@ $audioBaseUrl = Configure::read('Recordings.url');
     <md-icon class="chevron">chevron_right</md-icon>
 
     <div class="lang">
-        <img class="language-icon" src="/img/flags/{{translation.lang}}.svg" />
+        <img class="language-icon" src="/img/flags/{{translation.lang ? translation.lang : 'unknown'}}.svg" />
     </div>
 
     <div class="text" dir="{{translation.dir}}" flex>
--- a/src/Template/Element/sentences/translation.ctp
+++ b/src/Template/Element/sentences/translation.ctp
@@ -41,7 +41,7 @@ $audioBaseUrl = Configure::read('Recordings.url');
         </md-tooltip>
     </md-button>
 
-    <md-button class="md-icon-button audioUnavailable" target="_blank" ng-if="translation.audios && translation.audios.length === 0"
+    <md-button class="md-icon-button audioUnavailable" target="_blank" ng-if="!translation.audios || translation.audios.length === 0"
                 href="https://en.wiki.tatoeba.org/articles/show/contribute-audio">
         <md-icon>volume_off</md-icon>
         <md-tooltip md-direction="top">
--- a/src/Template/Element/sentences/translation.ctp
+++ b/src/Template/Element/sentences/translation.ctp
@@ -1,44 +1,55 @@
 <?php
-use App\Lib\LanguagesLib;
+use Cake\Core\Configure;
 
-$showExtra = '';
-if ($isExtra) {
-    $showExtra = 'ng-if="vm.isExpanded"';
-}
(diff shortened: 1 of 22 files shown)


webroot/js/directives/sentence-and-translations.dir.js
```diff
@@ -33,31 +46,158 @@
...
     }
 
-    function SentenceAndTranslationsController() {
+    function SentenceAndTranslationsController($http, $cookies) {
+        const MAX_TRANSLATIONS = 5;
+        const rootUrl = get_tatoeba_root_url();
+
         var vm = this;
+        var allDirectTranslations = [];
+        var alIndirectTranslations = [];
```
AndiPersti: Typo? Shouldn't the name be `allIndirectTranslations`?

AndiPersti:
When I open a page like https://dev.tatoeba.org/eng/Sentences/show_all_in/abk/und the Firefox Network  DevTool shows an 404 Error because it is trying to fetch the resource `https://dev.tatoeba.org/img/flags/{{translation.lang ? translation.lang : 'unknown'}}.svg` which of course doesn't exist. The same happens whenever I expand the list of translations if there are more available.

Unfortunately I haven't found out yet when and why the AngularJS expression isn't evaluated.

src/Template/Element/sentences/translation.ctp
```diff
@@ -1,44 +1,62 @@
...
+
     <div class="lang">
-        <?= $this->Languages->icon(
-                $translation->lang,
-                [
-                    'width' => 30,
-                    'height' => 20
-                ]
-            ) ?>
+        <img class="language-icon" src="/img/flags/{{translation.lang ? translation.lang : 'unknown'}}.svg" />
```
AndiPersti: This should be changed to `<img ... ng-src="/img/flags ...` to avoid the 404 errors I've mentioned. See https://docs.angularjs.org/api/ng/directive/ngSrc

trang:
Thanks a lot for the review, @AndiPersti :)

Issue #1984, opened 2019-10-11
Outdated CakePHP link in webroot's test


Guybrush88:
Currently, in webroot's test (https://github.com/Tatoeba/tatoeba2/blob/dev/webroot/test.php) there's still a link to a reference guide for the 2 series in line 5 (http://book.cakephp.org/2.0/en/development/testing.html), which is now outdated, since now the 3 series is currently used.

I guess the proper CakePHP's page to be linked there should be, instead, https://book.cakephp.org/3.0/en/development/testing.html , so that there would be a reference to more accurate information

trang:
I don't think that page is even used anymore...

jiru:
I confirm `/test.php` (the web frontend to run the test suite) has been dropped from version 3 of CakePHP. It is safe to remove that file.

PR #2063, opened 2020-01-06
Deleting test.php as per discussion in the thread for Issue 1984

danail-georgiev:
As per the discussion here: https://github.com/Tatoeba/tatoeba2/issues/1984


trang:
Thank you for this first pull request, @danail-georgiev. I hope many more will come :)

Issue #2043, opened 2019-12-24
Make the sender's name in private messages clickable to go to their profile page.


ckjpn:
I have a feeling that we used to be able to do this, and that this is something that disappeared with this new layout.

<img width="873" alt="Screen Shot 2019-12-24 at 15 20 59" src="https://user-images.githubusercontent.com/7934273/71397573-4d5f5d00-2661-11ea-9974-7eb281241d2c.png">

PR #2065, opened 2020-01-11
Make the sender's name in private messages clickable

danail-georgiev:
Add the URL for userProfileUrl variable in the message.ctp for making the Username clickable in the Private messages page (#2043 )


jiru:
Thanks for your pull request, it looks good! I'll let Trang merge it.

Issue #1718, opened 2018-11-10
An easier way to reach the user who left a comment on a sentence I posted


RyckRichards:
**The problem**

For now, if I want to send a private message - maybe to ask something that I don't want other users to read what I'm going to ask / I'm going to talk about something unrelated to the sentence - to the user who left a comment on my sentence, I have to

1. Click on his username;
2. Click on "Contact <user>"
3. Type the message
4. Proofread and hit "send"

**Suggestion**

Add a button "Message" which would redirect me to https://tatoeba.org/eng/private_messages/write/*** 

*** = user

![screenshot_1](https://user-images.githubusercontent.com/7658430/48307015-f533cd00-e52a-11e8-9b9d-f64bd349d684.png)

danail-georgiev:
Hi, 

I created a draft pull request for this one, can someone please have a look and advise on the issues I have detailed in the MR description. I will appreciate your input on this. Thank!

PR #2068, opened 2020-01-13
Add mail icon with a link for writing a new message

danail-georgiev:
(#1718) I created a draft MR because I am not quire sure how to fix a couple of problems. 

First, If someone can advise me on the styling of the icon (height, width, color, if the tags used are the appropriate ones) . 

Second, the data label disappears after the icon is rendered but I guess this is related to the aforementioned issue with the height of the icon.

### DIFF
--- a/src/Template/Element/sentence_comments/comment.ctp
+++ b/src/Template/Element/sentence_comments/comment.ctp
@@ -112,6 +112,11 @@ if ($sentenceOwnerLink) {
         <md-card-header-text>
             <span class="md-title">
                 <a href="<?= $userProfileUrl ?>"><?= $username ?></a>
+                <span>
+                    <md-button ng-cloak class="md-icon-button" href="<?= $this->Url->build(['controller' => 'private_messages', 'action' => 'write']); ?>">
+                        <md-icon>mail</md-icon>
+                    </md-button> 
+                </span>    
             </span>
             <span class="md-subhead ellipsis">
                 <?= $dateLabel ?>
--- a/src/Template/Element/sentence_comments/comment.ctp
+++ b/src/Template/Element/sentence_comments/comment.ctp
@@ -112,11 +112,6 @@ if ($sentenceOwnerLink) {
         <md-card-header-text>
             <span class="md-title">
                 <a href="<?= $userProfileUrl ?>"><?= $username ?></a>
-                <span>
-                    <md-button ng-cloak class="md-icon-button" href="<?= $this->Url->build(['controller' => 'private_messages', 'action' => 'write']); ?>">
-                        <md-icon>mail</md-icon>
-                    </md-button> 
-                </span>    
             </span>
             <span class="md-subhead ellipsis">
                 <?= $dateLabel ?>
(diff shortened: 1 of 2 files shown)


AndiPersti:
First, thanks for helping out with the code. I've also just started to participate a few months ago and experienced some difficulties at the beginning. :-)

> First, If someone can advise me on the styling of the icon (height, width, color, if the tags used are the appropriate ones) .

The Material Design tags/directives (the ones starting with `md-`) are correct but I don't think you need the outer `span` tag.
Styling of the icon should happen with CSS classes which leads us to your second question:

> Second, the data label disappears after the icon is rendered but I guess this is related to the aforementioned issue with the height of the icon.

The date label "disappears" because the styling you've applied to the button (`class="md-icon-button"`) makes the button too big and too centered and thus the date label moves too far down.
Do you know how to use the developer tools of your browser? If you select the button and disable some properties of the `md-button` class (e.g. `margin`, `min-height`, `line-height`) you'll notice that the date line will reappear.
So you need to add some extra styling for the button.

BTW: I guess you should also prefill the recipient of the private message. See for example https://github.com/Tatoeba/tatoeba2/blob/dev/src/Template/Element/users_menu.ctp on how you can pass a parameter to the controller.

danail-georgiev:
Hi Andi, 

Thanks for the feedback! To be honest I don't know exactly how to style the button using css - I checked the css files in the webroot but still I am not able to completely understand how to structure my changes. If you can be more specific how exactly I should implement the changes, it would be welcome. (I am aware of the Web Dev tools of course but I would like to understand how exactly to include the css file in the project).
Thanks for the link, I will have a look and try to fill in the recipient of the message.

trang:
Note that the screenshot on #1718 was a screenshot of how the comments used to look like and we actually don't want to have the exact same display. Therefore in this case, there's no need to worry about styling the icon. Simply add the message icon next to the other existing icons, keeping the same default color and size.

Screenshot_2020-01-15 Tatoeba Collection of sentences and translations

Regarding the CSS structure, the code that includes the CSS is in the [layout.ctp](https://github.com/Tatoeba/tatoeba2/blob/dev/src/Template/Layout/default.ctp#L40). But in summary:
- The CSS files in the [layouts](https://github.com/Tatoeba/tatoeba2/tree/dev/webroot/css/layouts) folder will be included on every page.
- Then for each page, there's one additional CSS file included. This file is in the folder named after the controller and is named after the action. So if you are on the sentence's page (controller = sentences, action = show), then [/sentences/show.css](https://github.com/Tatoeba/tatoeba2/blob/dev/webroot/css/sentences/show.css) will be included.

Supposing you would want to have a special color for your the private message icon, you could add the CSS in the `elements.css` file (because the icon can be found on several pages). Alternatively, you can also just add the style inline (`style="color: red"`).

danail-georgiev:
I made further changes, I think that now it is ok. Thanks for the help @AndiPersti and @trang :)

AndiPersti:
Maybe don't add the icon if the comment author is the current user?

Otherwise it looks fine to me. :-)

danail-georgiev:
@AndiPersti  yes, thanks for the correction, added

Issue #1678, opened 2018-09-27
Language of sentence reverts back if it's changed while editing the sentence


trang:
Steps to reproduce:
- Open the page of one of your sentences.
- Click the edit button. The sentence becomes editable.
- Click on the flag and change the language.
- Click on "OK". This saves the sentence.
- Refresh the page. The language changes back.

PR #2077, opened 2020-01-18
Editing and deleting sentences in new sentence design

### LINKED ISSUE #33: Directionality of original lang. for sentence remains after lang. is changed
If a newly created sentence is assigned a language with one directionality, then it will continue to be displayed with the same directionality even after the language is changed. For instance, say that I add an English translation "The sky is blue." to a Hebrew sentence, but I mistakenly mark my translation as Hebrew. I see my mistake immediately because the final period clings to the beginning of the sentence (the left) rather than the end (the right). Now I click on the flag to change it to English. Even though I've changed it, it will still be displayed on my current page with right-to-left directionality (with the period clinging to the beginning rather than the end). If I actually go to the URL of the sentence, it is displayed correctly, however. Thus, this is a transient annoyance rather than a permanent problem. It would still be nice to fix the problem, though.

---

Original ticket: https://www.assembla.com/spaces/tatoeba2/tickets/407

trang:
This PR implements the edit and delete features in the new sentence design. It also fixes #33.

**Editing sentences**

- Clicking on the edit button will open a form where the user can edit both the language and the text at the same time. This makes #1678 obsolete.
- The text input is automatically focused.
- If the user clicks "Cancel", any modification is lost (meaning that if they edit again the changes will no longer be there).
- The user can submit by pressing enter from the text input.
- Corpus maintainers and admins won't have the full list of languages anymore, only the languages in their profile. We'll have to see if it is really necessary for them to have the full list (I personally don't think it is).

**Deleting sentences**

The delete button is one of the buttons that are hidden by default. The user will have to click on the "more" button (`...`) to display it. Other than that, it works the same as now.

- A confirmation popup appears upon clicking on the delete button.
- When the user confirms, the sentence gets deleted and the user is redirected to the sentence's page.


webroot/js/directives/sentence-and-translations.dir.js
```diff
@@ -221,14 +221,16 @@
         }
 
         function saveSentence(sentence) {
+            var lang = sentence.lang === 'unknown' ? '' : $sentence.lang;
```
AndiPersti: Typo: `$sentence.lang;` should be `sentence.lang;` (no `$`)

ckjpn:
> Corpus maintainers and admins won't have the full list of languages anymore, ...

Part of their jobs are to make corrections for flag changes, too, so I think this would be necessary.

At this time, though, changes could be made if these members stick with the traditional sentence design, take the extra step of clicking to each sentence's page that needs this change, or set their profile languages to every language.

ckjpn:
Bugs:

* can't change the flag, except to "unknown."
* can't edit the sentence's text, unless the flag is set to "unknown" at the same time.

Poulpisator:
> Corpus maintainers and admins won't have the full list of languages anymore, ...
I agree with CK on this. I don't feel like adding all the languages I can identify to my profile. I can identify them but not speak them, so I can change a flag but don't really want to have them in my profile, even with zero stars. That's my personal opinion. 

My less personal opinion is that changing flag of a language that don't have current corpus maintainers will ask someone to add this language to its profile to be able to do it. There are more ways of identifying a language that simply speaking it so I think it's better to leave the option to CM to modify any language 1) because CM are users that can be trusted 2) except if removing the option somehow improve performance / usability.

AndiPersti:
> Bugs:

These bugs are caused by the typo I've mentioned above.

webroot/js/directives/sentence-and-translations.dir.js
trang: Thanks!

trang:
> My less personal opinion is that changing flag of a language that doesn't have a current corpus maintainer will ask someone to add this language to its profile to be able to do it.

Is that a bad thing though? It would be more transparent if each corpus maintainer should list which languages they would/could make an intervention for. There's no corpus maintainer who needs to (nor can) take care of all the languages.

> except if removing the option somehow improve performance / usability.

Well, a dropdown of several hundreds of values is never very usable. It could be replaced by the new language selector which make languages searchable but still. I think it is worth questioning ourselves if it makes sense that every corpus maintainer can change a sentence's language to absolutely any language.

Poulpisator:
Sure, that makes sense. Then please clearly announce, in due time, to corpus maintainers that the new way to change a flag is to first add a level 0 language to your profile :)

ckjpn:
>It could be replaced by the new language selector which make languages searchable ...

I'd vote for this.

>... but still. I think it is worth questioning ourselves if it makes sense that every corpus maintainer can change a sentence's language to absolutely any language.

I think we should be able to trust our corpus maintainers.
If someone leaves a comment identifying the language, why should we not let any corpus maintainer change the flag.

Also, it's not hard to check with google translate to identify a language, and then check a member's profile to see if it's the member's native language, or one of the languages he/she knows.  In cases where google translate identifies the language and it matches the only language a member has contributed in, this would be a very easy decision to make.

Poulpisator:
> The user can submit by pressing enter from the text input.

Please make it that pressing Esc cancels the modification, as it is currently.

Issue #2072, opened 2020-01-17
Username existence check in registration form can fail


jiru:
When filling the username field on the [registration page](https://tatoeba.org/eng/users/register), the existence of the username is checked as the user is typing in. This generates a lot of requests that can easily trigger nginx’s rate limitation.

To reproduce the problem, open the browser developer tool and look at the requests while typing into the username field. As you type more, requests first get delayed, and eventually end up with a 503 response. Such 503 responses result in a "Username already taken." message, even though the username is actually free.

This is an example screenshot from Firefox developer tool:
![requests](https://user-images.githubusercontent.com/5107734/72589704-ed459900-3936-11ea-9367-35fe23c7c5ef.png)

PR #2083, opened 2020-01-19
Avoid unnecessary redirections caused by Ajax calls

AndiPersti:
While looking at #2072 I've noticed that we cause unnecessary redirections in the Ajax calls for checking the availability of the username. The reason is the missing language code in the URL.

I've also modified other js files where that happened.


RyckRichards:
Thanks a lot for this PR, Andreas!!!

>

Issue #2073, opened 2020-01-17
Remove question mark in "Set your profile public?" setting


Poulpisator:
I don't think that's supposed to be a question :)

PR #2086, opened 2020-01-21
Remove question mark in "Set your profile public?" setting 

agrodet:
This PR addresses #2073 by simply removing the unnecessary question mark.


trang:
Welcome to the dev side of Tatoeba :)

Issue #2087, opened 2020-01-21
Mingrelian (xmf)


sabretou:
- Language name: Mingrelian
- Language info: https://en.wikipedia.org/wiki/Mingrelian_language
- ISO 639-3 code: xmf
- Public List: https://tatoeba.org/eng/sentences_lists/show/8696
- Icon: http://lingvisto.org/pics/mingrelia_flag_30x20.png
- Requested by: https://tatoeba.org/eng/user/profile/lovermann

[for tatoeba.org]
CALL add_new_language('xmf', 8696);

lovermann:
If anybody has any questions about the flag, I would appreciate a discussion (here or on tatoeba.org). It was very hard to find mingrelian flag, because Samegrelo (historically correct name of the region, where mingrelians lived and live, former region of [Colchis](ukrainianconsular@gmail.com).) was never a independent state or country. That's why I've found this flag in wikipedia as mingrelian flag.

Yorwba:
Regarding the flag: I thought we were switching to SVG for the flag images, so shouldn't we use [this file](https://upload.wikimedia.org/wikipedia/commons/f/fb/Flag_of_The_Principality_of_Mingrelia_%28Portolan_1560%29.svg) instead?

lovermann:
> Regarding the flag: I thought we were switching to SVG for the flag images, so shouldn't we use [this file](https://upload.wikimedia.org/wikipedia/commons/f/fb/Flag_of_The_Principality_of_Mingrelia_%28Portolan_1560%29.svg) instead?

Sure, you can. I thought I should make small icon for it. But if you do it, I would be good!

RyckRichards:
@Yorwba - I used the .svg flag in my PR - https://github.com/Tatoeba/tatoeba2/pull/2089/files :)

Yorwba:
That's an SVG with embedded PNG, which isn't much different from using a PNG directly. In particular, it doesn't have the advantage of normal SVGs that they look sharp at all possible resolutions. If we wanted to scale the flags up for high-resolution screens, for example, the embedded PNG would look blurry.

RyckRichards:
This is the one we have been using for all the flags. We called it "fake svg" but it was an easier way to implement this kind of language icon on Tatoeba.

Yorwba:
It's a good solution for the transition period, but if there's a non-fake SVG available, that one should probably be used instead.

RyckRichards:
[...]

PR #2089, opened 2020-01-21
Added language Mingrelian

RyckRichards:
This PR addresses #2087

It adds Mingrelian to Tatoeba. This has been requested by [lovermann] (https://tatoeba.org/eng/user/profile/lovermann)


trang:
The fake SVG hasn't been replaced, but I'll merge anyway. We can take care of it along with other fake SVGs another time.

Issue #859, opened 2015-10-23
Rename "collections" to "ratings" or "lists of rated sentences", since that's what they really are.


ckjpn:
In settings, we have this option:

Activate the feature to rate sentences and build your collection of sentences.

If you accurately referred to this, I think people would more likely use this feature.

I personally wouldn't want a "collection" of bad sentences or sentences I'm unsure of, but I wouldn't mind a "list" of sentences that I had rated bad.

PaulPeer:
+1

ckjpn:
I think this should be set to a high priority, especially if you want people to start using this function.

I would still really like to see at least one more option to indicate sentences that I've read, but don't want to rate -- perhaps call it "unrated" or "ignored".  Maybe someone else can come up with a better one-word name for this.

billoliver:
+1

What do you guys think about "My Reviews"?

ckjpn:
Ratings would be better than reviews.
The word review is often used to mean more than just a rating.

ckjpn:
Related Wall Post = https://tatoeba.org/eng/wall/show_message/26501#message_26501

You can see that even someone who has been using the rating system for a while can't remember that you call them "collections."

ckjpn:
I still think that "ratings" would be a better name for this.

Related Wall Post
https://tatoeba.org/eng/wall/show_message/34053#message_34053

PR #2091, opened 2020-01-22
Rename "Collections" to "Ratings"

agrodet:
This PR addresses #859 by renaming "collection" by "ratings". 

- The page name is now My ratings / Ratings / <user>'s ratings
- The URL changed from `collections/of/<user>` to `ratings/of/<user>`
- I renamed file and functionality names to avoid confusion in future maintenance. 

The functionality in itself should not have changed.


trang:
I'm having second thoughts about the wording "Ratings" (sorry that these second thoughts came only after you worked on the change).

My issue is that "ratings" could be understood as "I (don't) like this sentence" and could lead people to rate sentences based on personal taste. Perhaps some people are already doing that, but that would not be a good direction to go in my opinion.

I was thinking, instead of "ratings", to name this "proofreading" because I think this is what this feature should aim for: something that helps us improve the quality of the corpus, not something that is used for a downvote/upvote war.

Before we go ahead with the renaming, I think we should ask on the Wall.

agrodet:
Actually, I had the same kind of thinking, when I modify the settings string to "Activate the feature to rate sentences." Without the "collection" part, it sounds like "give a grade to sentences". 

"Proofreading" sounds better yes. Maybe somebody else will come up with a better term after you ask.

trang:
Wall thread: https://tatoeba.org/eng/wall/show_message/34023#message_34023

trang:
Let's go with "Reviews".

Based on the various answers on the Wall, this seems to be the best option. It would be too premature to rename this "Proofreading". We may rather create later on a proofreading feature that makes use of the reviews.

We already have the section on the sentence's page called "Reviews" so this helps make the transition smoother.

agrodet:
I have rename "Collections" and "Ratings" to "Reviews". Hopefully, I didn't miss any sentences / page. 

I have changed the "File description" on the download page to 
>Contains sentences reviewed by users. The value of the review can be -1 (sentence not OK), 0 (undecided or unsure), or 1 (sentence OK). Warning: this data is still experimental.
 
However, I wonder if "Warning: this data is still experimental." is still necessary.

jiru:
It looks good!

Could you also:
1. rename `canRate` [here](https://github.com/agrodet/tatoeba2/blob/22e4865521475996b3e00b64f7b7639e5420982a/src/Template/Element/sentences/sentence_and_translations.ctp#L29) and [here](https://github.com/agrodet/tatoeba2/blob/22e4865521475996b3e00b64f7b7639e5420982a/src/Template/Element/sentences/sentence_menu.ctp#L46)
2. rephrase `collection` in the two comment blocks [here](https://github.com/agrodet/tatoeba2/blob/22e4865521475996b3e00b64f7b7639e5420982a/src/Model/Table/UsersSentencesTable.php#L38-L67)

agrodet:
Thank you, I renamed them.

Issue #2092, opened 2020-01-22
Chagatai (chg)


soliloquist-tatoeba:
- Language name: Chagatai
- Language info: https://en.wikipedia.org/wiki/Chagatai_language
- ISO 639-3 code: chg
- Public List: https://tatoeba.org/eng/sentences_lists/show/9614
- Icon: https://i.imgur.com/PTiur5p.png
- Requested by: https://tatoeba.org/eng/user/profile/soliloquist

[for tatoeba.org]
CALL add_new_language('chg', 9614);

soliloquist-tatoeba:
@trang

Please consider applying the same tweak (dir="auto") you previously did for Ottoman Turkish. 

https://github.com/Tatoeba/tatoeba2/issues/1780#issuecomment-471606563

It perfectly fixes the issues with writing direction.

RyckRichards:
@sabretou :) 

@soliloquist-tatoeba  0 working on it right now

PR #2095, opened 2020-01-25
Added language Chagatai

RyckRichards:
This PR addresses #2092


src/Lib/LanguagesLib.php
```diff
@@ -640,6 +641,7 @@ public static function getLanguageDirection($lang) {
             "snd",
             "div",
             "otk",
+            "chg"
```
Yorwba: To allow automatic detection of the writing direction, as @soliloquist-tatoeba requested, this should be added to `$autoLangs` below, instead of `$rightToLeftLangs`.

Also, if you add a comma to the end of the line now, you won't have to change the line later when something gets added below. It doesn't make a big difference either way; I just wanted to point it out.

RyckRichards: Thanks a lot, @Yorwba! I've fixed this

https://github.com/Tatoeba/tatoeba2/pull/2095/commits/1d5e13c5a8edd3a1a093803583e997b928a0f214

Issue #1156, opened 2016-05-21
Change wording of option "Hide random sentence on homepage"


declanmalone:
Just two places where I noticed this:
1. In options, there's a checkbox beside "Hide random sentence on the homepage". It would be better written as "Show random ..." and be ticked by default.
2. In advanced search, "Is approved" is easier to understand than "Is unapproved", especially since the default is a double-negative "is not unapproved"

trang:
1. By default the random sentence is displayed, this is why the option says "Hide random sentence" (the option is to hide, not to show).

2. It would be incorrect to say that the sentences are "approved". The current system only handles the situation where certain sentences do not comply to our standards or to our rules, and therefore are unapproved. A sentence that is "not unapproved" is not necessarily approved.
I agree with you though that it is harder to understand, and if you have an idea of a better (and accurate) wording, let me know. Otherwise I let you close this issue :)

declanmalone:
On 21 May 2016 at 21:57, Trang <notifications@github.com> wrote:

>
>    1.
>
>    By default the random sentence is displayed, this is why the option
>    says "Hide random sentence" (the option is to hide, not to show).
>
>
Yes, both the the text and the default state of the button would have to be
changed in tandem.


>
>    1.
>
>    It would be incorrect to say that the sentences are "approved". The
>    current system only handles the situation where certain sentences do not
>    comply to our standards or to our rules, and therefore are unapproved. A
>    sentence that is "not unapproved" is not necessarily approved.
>    I agree with you though that it is harder to understand, and if you
>    have an idea of a better (and accurate) wording, let me know. Otherwise I
>    let you close this issue :)
>
>
I don't actually understand what "not approved" means on the site. I
haven't gotten around to reading all the parts of the documentation, so
there are still some areas that I don't quite understand.

As for the other point you raised, about posting on the wall, I will copy
the "issue" there.

Thanks & Regards,
Declan

ckjpn:
1.

I don't see any problem with changing this the way it has been suggested.

I think the default setting for "email notifications" is "on", too.  However, I'm not sure, since I've opted out of that.

If all the options are listed as positive, then it's easier to see what's turned on and what's turned off.


2.

I agree with TRANG here that this shouldn't be changed.   We don't have any "approved" status for sentences.  However, we do have a status for sentences to warn users that the sentence is problematic -- a copyright infringement, or some other problem.

declanmalone:
Thanks for the support on #1, CK.

[...]

PR #2100, opened 2020-01-27
Modify "hide" random sentence to "show" random sentence

agrodet:
This addresses #1156 and simply rephrase the sentence displayed.

### DIFF
--- a/src/Template/User/settings.ctp
+++ b/src/Template/User/settings.ctp
@@ -44,7 +44,7 @@ $this->set('title_for_layout', $this->Pages->formatTitle(__('Settings')));
                 <h2><?php echo __('Options'); ?></h2>
             </div>
         </md-toolbar>
-        
+
         <?php echo $this->Form->create($userSettings, [
             'ng-cloak' => true,
             'url' => ['controller' => 'user', 'action' => 'save_settings']
@@ -172,21 +172,21 @@ $this->set('title_for_layout', $this->Pages->formatTitle(__('Settings')));
             </md-list-item>
 
             <md-list-item>
-                <?php $hideRandomSentence = $userSettings->settings['hide_random_sentence']; ?>
+                <?php $showRandomSentence = 1 - $userSettings->settings['hide_random_sentence']; ?>
                 <md-checkbox
                     ng-false-value="0"
                     ng-true-value="1"
-                    ng-model="hideRandomSentence"
-                    ng-init="hideRandomSentence = <?= $hideRandomSentence ?>"
+                    ng-model="showRandomSentence"
+                    ng-init="showRandomSentence = <?= $showRandomSentence ?>"
                     class="md-primary">
                 </md-checkbox>
-                <p><?php echo __('Hide random sentence on the homepage') ?> </p>
+                <p><?php echo __('Show random sentence on the homepage') ?> </p>
                 <div ng-hide="true">
                 <?php
                     echo $this->Form->input(
                         'settings.hide_random_sentence',
                         array(
-                            'value' => '{{hideRandomSentence}}'
+                            'value' => '{{1 - showRandomSentence}}'
                         )
                     );
                 ?>
--- a/src/Template/User/settings.ctp
+++ b/src/Template/User/settings.ctp
@@ -172,12 +172,15 @@ $this->set('title_for_layout', $this->Pages->formatTitle(__('Settings')));
             </md-list-item>
 
             <md-list-item>
-                <?php $hideRandomSentence = $userSettings->settings['hide_random_sentence']; ?>
+                <?php
+                    // The setting is "hide" but we display "show"
+                    $hideRandomSentence = $userSettings->settings['hide_random_sentence'];
+                ?>
                 <md-checkbox
                     ng-false-value="1"
                     ng-true-value="0"
-                    ng-model="showRandomSentence"
-                    ng-init="showRandomSentence = <?= (int)$hideRandomSentence ?>"
+                    ng-model="hideRandomSentence"
+                    ng-init="hideRandomSentence = <?= (int)$hideRandomSentence ?>"
                     class="md-primary">
                 </md-checkbox>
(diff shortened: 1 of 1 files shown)


src/Template/User/settings.ctp
```diff
@@ -172,21 +172,21 @@ $this->set('title_for_layout', $this->Pages->formatTitle(__('Settings')));
...
             <md-list-item>
-                <?php $hideRandomSentence = $userSettings->settings['hide_random_sentence']; ?>
+                <?php $showRandomSentence = 1 - $userSettings->settings['hide_random_sentence']; ?>
                 <md-checkbox
                     ng-false-value="0"
                     ng-true-value="1"
-                    ng-model="hideRandomSentence"
-                    ng-init="hideRandomSentence = <?= $hideRandomSentence ?>"
+                    ng-model="showRandomSentence"
+                    ng-init="showRandomSentence = <?= $showRandomSentence ?>"
```
trang: You can instead just swap the `ng-false-value` and `ng-true-value`.

`ng-false-value="0"` means that when the checkbox is unchecked, the value for the binded variable is 0.

agrodet: I thought so but the default value of `hide_random_sentence` being `false`, I couldn't make it work for new users. The false is not interpreted as 0 so the html is `ng-init="showRandomSentence = "` and the checkbox is unchecked. I used `1 - false` to force it to `1` and have it work as expected. Am I missing something? 

Should we set `hide_random_sentence` to `'0'` instead of `false` by default?

AndiPersti: The `ng-init` line doesn't work because a PHP false value is casted to the empty string when printing/echoing it.

If we want to keep 1 as true and 0 as false on the JS side we should use these values also on the PHP side for all the boolean settings.

trang: Yea, I remember this issue from redesigning the advanced search. I did this for the sort checkbox:
```
ng-init="sortReverse = <?= $sort_reverse == 'yes' ? 1 : 0 ?>"
```

https://github.com/Tatoeba/tatoeba2/blob/dev/src/Template/Element/advanced_search_form.ctp#L323

agrodet:
I have modified how to handle the default false value to make it consistent with the code provided by TRANG. 

As AndiPersti mentioned, it might be a good idea to use 0 and 1 as default value if we continue to use them with JavaScript, but that's a different story.

src/Template/User/settings.ctp
```diff
@@ -172,21 +172,21 @@ $this->set('title_for_layout', $this->Pages->formatTitle(__('Settings')));
...
+                <?php $showRandomSentence = $userSettings->settings['hide_random_sentence']; ?>
                 <md-checkbox
-                    ng-false-value="0"
-                    ng-true-value="1"
-                    ng-model="hideRandomSentence"
-                    ng-init="hideRandomSentence = <?= $hideRandomSentence ?>"
+                    ng-false-value="1"
+                    ng-true-value="0"
+                    ng-model="showRandomSentence"
+                    ng-init="showRandomSentence = <?= $showRandomSentence == false ? 0 : $showRandomSentence ?>"
```
jiru: Casting the value into an integer is more concise and secure: `(int)$showRandomSentence`.

src/Template/User/settings.ctp
```diff
@@ -172,21 +172,21 @@ $this->set('title_for_layout', $this->Pages->formatTitle(__('Settings')));
             </md-list-item>
 
             <md-list-item>
-                <?php $hideRandomSentence = $userSettings->settings['hide_random_sentence']; ?>
+                <?php $showRandomSentence = $userSettings->settings['hide_random_sentence']; ?>
```
jiru: With your change, the name of the variable `$showRandomSentence` doesn’t make sense any more. It should be either `$hideRandomSentence = $userSettings->settings['hide_random_sentence'];` or `$showRandomSentence = !$userSettings->settings['hide_random_sentence'];`.

agrodet: That is very true! I modified the name of the variable :)

jiru: Yes, this is still confusing. How about:
```
$showRandomSentence = !$userSettings->settings['hide_random_sentence'];
ng-false-value="0"
ng-true-value="1"
ng-init="showRandomSentence = <?= (int)$showRandomSentence ?>"
`

Comment (agrodet) [reply src/Template/User/settings.ctp]: Since the value we display and the one we send are opposite, isn't it one of the following? I have a preference for the first one but Is there a clean version without the `1 -`?
```
$showRandomSentence = !$userSettings->settings['hide_random_sentence'];
ng-false-value="0"
ng-true-value="1"
ng-init="showRandomSentence = <?= (int)$showRandomSentence ?>"

'value' => '{{1 - showRandomSentence}}'
```

```
$showRandomSentence = !$userSettings->settings['hide_random_sentence'];
ng-false-value="1"
ng-true-value="0"
ng-init="showRandomSentence = <?= 1 - (int)$showRandomSentence ?>"

'value' => '{{showRandomSentence}}'
```

Comment (jiru) [reply src/Template/User/settings.ctp]: Yes, I forgot about the label showing the opposite of the value. I think it’s simpler to keep hide* variable names and just switch the tick:
```
$hideRandomSentence = $userSettings->settings['hide_random_sentence'];
ng-false-value="1"
ng-true-value="0"
ng-init="hideRandomSentence = <?= (int)$hideRandomSentence ?>"

'value' => '{{hideRandomSentence}}'
```

AndiPersti: I agree with jiru. Renaming the model name just makes the code more complicated without any real advantage IMHO. I would rather add a comment explaining the unusual `true`/`false` values and the opposite label text.

jiru:
Can you remove the full stop in the checkbox label?

jiru:
Even though the resulting modifications are simple, I appreciate that we took time to make it right. Thank you, @agrodet!

Issue #1597, opened 2018-07-04
Setting controlling number of sentences per page is ignored


alanfgh:
My settings specify that I want to see 20 sentences per page. However, when I go to a page (for example: https://tatoeba.org/eng/sentences/of_user/AlanF_US ), I see 100 sentences on it. Changing the setting (for instance, to display 10 sentences per page) has no effect.

jiru:
Yes, this page always shows 100 sentences per page regardless of that setting. I think it makes sense on that page to show more sentences than the default 10, because the list is quite compact. It is however in contradiction with the setting.

alanfgh:
@jiru wrote: 

> I think it makes sense on that page to show more sentences than the default 10, because the list is quite compact.

Screen real estate is not the only consideration. I want to choose how many sentences are displayed on any page because it's hard to keep track of my progress when there are too many. For instance, I might want to go through a contributor's list of sentences in order to correct them, but when more than 20 are displayed on a page, I lose track of where I am.

agrodet:
I share Alan's opinion.
@jiru Is it okay to take care of this or do you have any worries about the modification? I haven't checked in detail but I think it is just a matter of modifying the following line. 

https://github.com/Tatoeba/tatoeba2/blob/3f71e6d5c206a194cf7f426f9403a0f92af9d053/src/Controller/SentencesController.php#L917

jiru:
@agrodet yes this the correct place to tune the number of sentences.

I often use this page to have an overview of the sentences of a user. It this case I prefer to see a lot of sentences at once because I'm quickly going through them (as opposed to stopping on each of them to think about it).

I think my problem is less about that particular page than about having a unique user setting that controls the number of sentences per page everywhere. I have the feeling that the number of sentences per page depends on the use case, not the user. Besides, I believe that having it "hidden" in the user settings makes it underused or unbeknown to most people. There must be some way to better allow people to control this.

So yeah, @agrodet I think you can go ahead.

alanfgh:
@agrodet, thanks for fixing this. @jiru, I agree that it would be better to be able to set the number of sentences per page by use case rather than in general. However, at least @agrodet's change lets the user control the behavior, which is an improvement, and we can think about adding more fine-grained control in the future.

PR #2102, opened 2020-01-27
Apply sentences per page setting to user's sentences page

agrodet:
This PR solves #1597 by removing the limit of 100 sentences per page and using the user's setting instead.

### DIFF
--- a/tests/TestCase/Controller/SentencesControllerTest.php
+++ b/tests/TestCase/Controller/SentencesControllerTest.php
@@ -1,6 +1,7 @@
 <?php
 namespace App\Test\TestCase\Controller;
 
+use App\Model\Entity\User;
 use App\Test\TestCase\Controller\TatoebaControllerTestTrait;
 use Cake\Core\Configure;
 use Cake\ORM\TableRegistry;
@@ -277,13 +278,12 @@ class SentencesControllerTest extends IntegrationTestCase {
         $this->assertRedirect('/jpn/sentences/show/1');
     }
 
-    public function testPaginateRedirectsPageOutOfBoundsToLastPage() {
+    public function testPaginateRedirectsPageOutOfBoundsToLastPage_asGuest() {
         $user = 'kazuki';
         $userId = 7;
-        $lastPage = 2;
-
+        $defaultNbPerPage = User::$defaultSettings['sentences_per_page'];
         $newSentences = array();
-        for ($i = 1; $i <= 100; $i++) {
+        for ($i = 1; $i <= rand(1,100); $i++) {
             $newSentences[] = [
                 'lang' => 'eng',
                 'text' => "Ay ay ay $i.",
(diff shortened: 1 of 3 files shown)


AndiPersti:
Do you know how to fix the failing test?

agrodet:
Yeah :/ I realized I didn't check the tests after I opened the PR but I had to go. Anyway, I added the tests.

tests/TestCase/Controller/SentencesControllerTest.php
```diff
@@ -277,13 +278,12 @@ public function testEditAudio_asAdmin() {
...
-    public function testPaginateRedirectsPageOutOfBoundsToLastPage() {
+    public function testPaginateRedirectsPageOutOfBoundsToLastPage_asGuest() {
         $user = 'kazuki';
         $userId = 7;
-        $lastPage = 2;
-
+        $defaultNbPerPage = User::$defaultSettings['sentences_per_page'];
         $newSentences = array();
-        for ($i = 1; $i <= 100; $i++) {
+        for ($i = 1; $i <= rand(1,100); $i++) {
```
jiru: Why did you add randomness here and on the other test? I don’t think it’s a good idea because it means we don’t run the exact same test every time everywhere. I may have a failing test while yours is passing, which can create a lot of confusion.

agrodet:
I know that's borderline but let me explain. The way I wrote the test, the goal is that *any* number of sentences will be correct. For me, we do run the same test: "do the redirect works?" I agree that if the test fails, the fact that we don't know the number of sentences might be confusing. However, this test is not supposed to fail. So if it fails, something has been broken. 

I know that "unit tests are supposed to test limit cases" and I agree that I should have added another unit test, with the exact threshold for redirection, and not doing so is a lack of testing.

All in all, I'm not the code maintainer so I'll just follow what you say.

tests/TestCase/Controller/SentencesControllerTest.php
```diff
@@ -295,12 +295,44 @@ public function testPaginateRedirectsPageOutOfBoundsToLastPage() {
...
 
         $this->get("/eng/sentences/of_user/$user?page=9999999");
+        $this->assertRedirect("/eng/sentences/of_user/$user?page=$lastPage");
+    }
+
+    public function testPaginateRedirectsPageOutOfBoundsToLastPage_withUserSetting() {
+        $user = 'kazuki';
+        $userId = 7;
+        $users = TableRegistry::getTableLocator()->get('Users');
+        $NbPerPageSetting = $users->getSettings($userId)['settings']['sentences_per_page'];
```
jiru: Could you rename this variable so that it starts with a lowercase? Let’s keep the same naming conventions within a single file.

tests/TestCase/Controller/SentencesControllerTest.php
```diff
@@ -295,12 +295,44 @@ public function testPaginateRedirectsPageOutOfBoundsToLastPage() {
...
+            ];
+        }
+        $sentences = TableRegistry::getTableLocator()->get('Sentences');
+        $entities = $sentences->newEntities($newSentences);
+        $sentences->saveMany($entities);
 
+        $nbSentences = count($sentences->find()->where(['user_id' => $userId])->all());
+        $lastPage = ceil($nbSentences / $NbPerPageSetting);
+
+        $this->logInAs('kazuki');
```
jiru: How about using `$user` instead here?

tests/TestCase/Controller/SentencesControllerTest.php
```diff
@@ -295,12 +295,44 @@ public function testPaginateRedirectsPageOutOfBoundsToLastPage() {
...
+            ];
+            $newSentences[] = [
+                'lang' => 'eng',
+                'text' => "Oy oy oy $i.",
+                'user_id' => 1,
+            ];
+        }
+        $sentences = TableRegistry::getTableLocator()->get('Sentences');
+        $entities = $sentences->newEntities($newSentences);
+        $sentences->saveMany($entities);
```
jiru: How about refactoring this code into a private method like addManySentencesAsUser()?

tests/TestCase/Controller/SentencesControllerTest.php
```diff
@@ -98,6 +98,27 @@ public function ajaxAccessesProvider() {
...
+
+        return count($sentences->find()->where(['user_id' => $userId])->all());
```
jiru: It’s way more efficient to use MySQL’s count than PHP’s count:
```
$sentences->find()->where(['user_id' => $userId])->count();
```

jiru:
Looks good. Thank you, agrodet! :+1:

Issue #1991, opened 2019-10-19
Give visual indication of current sort criterion and order for list of sentences with a given tag


alanfgh:
When you choose to show all sentences with a given tag, such as "@change", the top of the list says this:

Sort by: _date created_ | _date of tag_

However, it's not possible to determine which critierion (date created or date of tag) is currently in force. Nor is it possible to determine which order (ascending or descending) is being used. 

For "Browse by list", by contrast, it's easy to tell because:
(1) the attributes of each sentence (list name, date created, number of sentences, and list updated) are displayed, and these correspond to the sorting criteria, and
(2) the lists are displayed in a relatively dense format, so it's easy to see the trend that lets you determine which criterion is in force.

It may or may not be possible to display the attributes of sentences in the tag list view, so the same approach may or may not work. But if it doesn't, there should be some way to indicate which criterion is in effect (perhaps by changing the font weight or color in which the name of the criterion is displayed at the top of the page), and also the order (perhaps by adding an arrow to the display that points either up or down).

AndiPersti:
Something like this?
![sort1](https://user-images.githubusercontent.com/6566321/72555944-ee76b780-389d-11ea-92df-460f374e5552.png)
![sort2](https://user-images.githubusercontent.com/6566321/72555954-f3d40200-389d-11ea-96f0-f8cd871eaeaa.png)

Issue #1063, opened 2016-03-12
Move the sort section to the top of the search page.


billoliver:
I think this filter is important and we can make it more user friendly. However, this can be implemented later because you guys must be very busy with the recent changes on the site.
![sort-sentences](https://cloud.githubusercontent.com/assets/17387379/13723724/0f8b628c-e84c-11e5-90e7-4bb44e4b6baa.png)

RyckRichards:
That would help members a lot since they would see a "friendly number"
telling them "what to do".

-- 
Ricardo Vernaut Jr

billoliver:
> That would help members a lot since they would see a "friendly number" telling them "what to do" and "what they can do".

Friendly number? I better remove that round shape because it isn't a feature.

RyckRichards:
Sorry, @billoliver  I was talking about another feature. I updated my comment.

fadhelafif:
It should duplicate it or just move it on top of page?

trang:
Good question. Initially my thought was that it should be moved (not duplicated), but I can see why duplicating it makes sense too. The user may want to choose the sort order before they get the results, not after.

I'm setting this back to "pending", we need to give it a little bit more thoughts.

trang:
This has been considered and wording of the sort options have were discussed in #2108.

PR #2108, opened 2020-01-27
Indicate current sort criterion for pagination

AndiPersti:
This PR addresses #1991

Add an arrow next to the current sort key in the pagination header which indicates the current sort direction.

### DIFF
--- a/src/Controller/TagsController.php
+++ b/src/Controller/TagsController.php
@@ -108,6 +108,12 @@ class TagsController extends AppController
             'conditions' => $conditions
         ];
 
+        // We use two columns for ordering the results so in order to indicate the
+        // initial sorting direction we need to set the 'sort' key manually
+        if (empty($this->request->getQuery('sort'))) {
+            $this->paginate['sort'] = 'nbrOfSentences';
+        }
+
         $allTags = $this->paginate();
         $this->set("allTags", $allTags);
         $this->set("filter", $filter);
--- a/src/Controller/UsersController.php
+++ b/src/Controller/UsersController.php
@@ -488,6 +488,12 @@ class UsersController extends AppController
             'fields' => array('id', 'username', 'since', 'image', 'role'),
         );
 
+        // We use two columns for ordering the results so in order to indicate the
+        // initial sorting direction we need to set the 'sort' key manually
+        if (empty($this->request->getQuery('sort'))) {
+            $this->paginate['sort'] = 'role';
+        }
+
         $query = $this->Users->find()->where(['Users.role IN' => User::ROLE_CONTRIBUTOR_OR_HIGHER]);
         try {
             $users = $this->paginate($query);
--- a/src/Template/Users/all.ctp
+++ b/src/Template/Users/all.ctp
@@ -99,7 +99,7 @@ $this->set('title_for_layout', $this->Pages->formatTitle(__('Members')));
             echo ' | ';
             echo $this->Paginator->sort('since', __('Member since'));
             echo ' | ';
-            echo $this->Paginator->sort('role', __('Member status'));
+            echo $this->Pagination->sortForRole();
             ?>
         </div>
 
--- a/src/View/Helper/PaginationHelper.php
+++ b/src/View/Helper/PaginationHelper.php
@@ -132,5 +132,27 @@ class PaginationHelper extends AppHelper
         </ul>
         <?php
     }
+
+    /**
+     * Swap sort link templates for ordering by user role
+     *
+     * User roles are stored as an enum in the database where the highest role
+     * (admin) is represented by the smallest value (1). Thus using the default
+     * templates for Paginator->sort() would produce the wrong arrows. In order
+     * to display the correct arrows, we need to swap the 'sortAsc' and 'sortDesc'
+     * templates before calling Paginator->sort().
+     *
+     * @return string
+     **/
+    public function sortForRole() {
+        $templates = $this->Paginator->getTemplates();
+        $this->Paginator->setTemplates([
+            'sortAsc' => $this->Paginator->getTemplates('sortDesc'),
+            'sortDesc' => $this->Paginator->getTemplates('sortAsc')
+        ]);
+        $sortLink = $this->Paginator->sort('role', __('Member status'));
+        $this->Paginator->setTemplates($templates);
+        return $sortLink;
+    }
 }
 ?>
--- a/webroot/css/layouts/elements.css
+++ b/webroot/css/layouts/elements.css
@@ -1561,6 +1561,20 @@ div.hideLink {
     margin: 20px;
 }
 
+.sortBy .desc, .sortBy .asc {
+/*    font-weight: bold; */
+}
+
+.sortBy .desc::after {
+    content: "\02B07";
+    vertical-align: middle;
+}
+
+.sortBy .asc::after {
+    content: "\02B06";
+    vertical-align: middle;
+}
+
 
 
 /*


AndiPersti:
Some remarks:

* The design (color, type of arrow, size, ...) is just a suggestion and open to debate. :-)

* On the Users page ("List of all members") the sort keys start with an uppercase letter, on all other pages the sort keys are all lowercase. What style is preferred?

* Should we also indicate the sorting on the search page?:
search
I still don't have a working solution because it's a little bit more complicated on that page without breaking backward compatibility but I think I will get there. :-)

agrodet:
Ha ha, it's funny that I thought about this thing just yesterday! Thanks for taking care of it :) 

Personally, I don't find the arrow very useful. Probably the arrow is used to express "Ascending" and "Descending" but some people won't get that. Also, it is not clear for non-programmers what the ascending order would be for dates. That being said I don't have any better idea at the moment :/ The best option may be to use a different sentence for each option instead of a common icon, but I don't know if it would render nicely. Something like
Word count (Higher first)
Date created (Older first)
Date modified (Most recent first)
Random and Relevance, I dont't know ^^ 

Maybe somebody will get a better idea.

AndiPersti:
Interesting, I thought using arrows for indicating the sort direction is rather common. For example Wikipedia uses it with their tables and probably any GUI file manager when showing the files in list mode. But I admit that the context there is a table with columns while our results are lists.

A similar approach to the one you suggested would be [using a select box](https://ux.stackexchange.com/a/23272). YouTube and I think most online shops use this approach for sorting (and filtering) their search results.

agrodet:
Going with a select box-like thing, I can imagine something like   
Sort by: Date created (Older first)   
centered in the middle, above the pagination line. when clicking on it, the select box would open with the five (for your screenshot) options listed in a way similar to what the post you linked shows.

From the user point of view, we would loose the opportunity to see at a glance all the sorting possibilities. The maximum number of clicks would be equal to what we have now: 2. Overall, I think it would be much clearer than the current version and therefore a better experience. 

I think that not seeing all the sorting possibilities at a glance is something one can get used to.

trang:
For the responsive UI, my plan was to add a "Sort by" button in the toolbar and a menu similar to the "Sort" menu in the GitHub issues.

Quick prototype in the list page of Tatoeba:

Screenshot_2020-02-01 All public lists (2) - Tatoeba(1)


Corresponding code:
```
        <md-toolbar class="md-hue-2">
            <div class="md-toolbar-tools">
                <h2 flex><?= $title ?></h2>

                <md-menu md-offset="5 50" md-position-mode="target-right target">
                    <md-button ng-click="$mdOpenMenu($event)">
                        <md-icon>sort</md-icon> Sort by
                    </md-button>
                    <md-menu-content>
                        <md-menu-item ng-repeat="item in ['list', 2, 22223, 9324, 02934, 12]">
                            <md-button>
                                <md-icon>{{ $index === 1 ? 'check' : 'blank'}}</md-icon>
                                <span style="padding-right: 12px">Option {{item}}</span>
                            </md-button>
                        </md-menu-item>
                    </md-menu-content>
                </md-menu>
            </div>
        </md-toolbar>
```

This implies creating new strings for the sort options, but it can be difficult to find a good word for some cases. As a start, we could consider keeping the current string and add "(ascending)" or "(descending)" in parenthesis next to the option.

For the list, that would look like:
- list name (ascending)
- list name (descending)
- date created (ascending)
- date created (descending)
- number of sentences (ascending)
- number of sentences (descending)
- last updated (ascending)
- last updated (descending)

Also, with this solution, the user won't know right away how a list is sorted, they would have to click on the "Sort by" button and see which option has a check mark next to it.

We could consider:
- Adding the selected value next to "Sort by". But this might take too much space if option has a very long label.
- Adding a `<subheader>` below the toolbar to specify how the items are sorted.

AndiPersti:
> For the responsive UI, my plan was to add a "Sort by" button in the toolbar and a menu similar to the "Sort" menu in the GitHub issues.

Looks good to me. :-)
Will you implement this yourself or do you need some help?
 
> For the list, that would look like:
> 
>     * list name (ascending)
> 
>     * list name (descending)
> 
>     * date created (ascending)
> 
>     * date created (descending)
> 
>     * number of sentences (ascending)
> 
>     * number of sentences (descending)
> 
>     * last updated (ascending)
> 
>     * last updated (descending)

How about 
* List/Tag/User name (A-Z)
* List/Tag/User name (Z-A)
* Newest +
* Oldest +
* Highest number of sentences/words
* Lowest number of sentences/words
* Most recently updated
* Least recently updated
* Most relevant
* Least relevant
* Highest rank *
* Lowest rank *
(* for member status on "List of all members")
(+ for creation dates or "Member since")

I don't think it would be necessary to add "first" after each option, is it? 
Would "(A-Z)"/"(Z-A)" be too Eurocentric?
 
> Also, with this solution, the user won't know right away how a list is sorted, they would have to click on the "Sort by" button and see which option has a check mark next to it.

I personally think the button itself is sufficient but adding a subheader is a good option.

agrodet:
I also think that meaningful words are better than ascendind / descending. Again, ascending date is not very clear.

> Would "(A-Z)"/"(Z-A)" be too Eurocentric?
I think so, yes :D

trang:
> Will you implement this yourself or do you need some help?

Help is always welcome :) You are free to go ahead and rework this PR with the design I suggested, if you wish to. I'm still on the sentence component so you wouldn't be interfering with any un-pushed work that I may have on this area. You are also free to leave it to me if you prefer to dedicate yourself to less frontend'ish tasks.

Either way, I would still like to consult @alanfgh for the wording.

alanfgh:
@trang, is there a change I should look at, or did you want me to comment on the list of options above?

trang:
@alanfgh There is no change to look at yet. We need to figure how to reword the sort options if the sort feature was redesigned as in [my screenshot above](https://github.com/Tatoeba/tatoeba2/pull/2108#issuecomment-581054498). We would need your opinion on the wording suggested so far.

Of course feel free to comment as well on the suggested design itself, in case it doesn't really fulfill your requirements.

I'm listing below the mapping between current string and corresponding new strings. Each sort option would be split into two strings.

https://tatoeba.org/eng/sentences_lists/index
- list name 
  - List name (A-Z)
  - List name (Z-A)
- date created
  - Newest
  - Oldest
- number of sentences
  - Highest number of sentences
  - Lowest number of sentences
- last updated
  - Most recently updated
  - Least recently updated

https://tatoeba.org/eng/sentences_lists/show/9633
- date added to list 
  - Most recently added
  - Least recently added
- date created
  - Newest sentences
  - Oldest sentences

https://tatoeba.org/eng/tags/view_all
- count
  - Highest number of sentences
  - Lowest number of sentences
- name
  - Tag name (A-Z)
  - Tag name (Z-A)

https://tatoeba.org/eng/tags/show_sentences_with_tag/2139
- date created 
  - Newest sentences first
  - Oldest sentences first
- date of tag
  - Most recently tagged
  - Least recently tagged

https://tatoeba.org/eng/users/all
- Username 
  - Username (A-Z)
  - Username (Z-A)
- Member since
  - Newest
  - Oldest
- Member status
  - Highest rank
  - Lowest rank

https://tatoeba.org/eng/sentences/of_user/TRANG
- date modified 
  - Most recently updated
  - Least recently updated
- date created
  - Newest
  - Oldest

https://tatoeba.org/eng/reviews/of/TRANG
- date modified
  - Most recently updated
  - Least recently updated
- date created
  - Newest
  - Oldest
- sentence id
  - Newest sentences
  - Oldest sentences

**Search ("Order" dropdown in sidebar)**
- Relevance
  - Most relevant
  - Least relevant
- Fewest words first
  - Fewest words
  - Most words
- Last created first
  - Newest sentences
  - Oldest sentences
- Last modified first
  - Most recently updated
  - Least recently updated
- Random
  - Random (only case where we don't need to duplicate the strings)

agrodet:
As mentioned above A-Z is not very appropriate. Since there's also a problem of space, I would suggest two things.

1. Every time you used (A-Z), you only repeated the name of the sort option, so I suggest to remove that name (name, username, etc.)
2. Find an appropriate string for the sort order. Alphabetical would be a better but still far from ideal candidate, I think. Actually, what people are the most used to is "computer order". For example, Japanese are used to Latin alphabet coming first and kanjis last. Similarly, I *guess* speakers of other languages are used to the same thing nowadays. I don't know if there is a name for this "computer order", though...

alanfgh:
Several things:

(1) You could use "alphabetical" or "reverse alphabetical" in place of "A-Z" and "Z-A". If I saw one of those, I would assume that it extended beyond A-Z and Z-A to mean what agrodet is calling "computer order". (I believe the technical term for "computer order" is "collation", but I wouldn't suggest using it.)

(2) If you changed, for example, "Newest" to "Newest first", it would be a bit more verbose, but clearer. Otherwise, it could be interpreted as meaning only the newest sentences (down to some cutoff) would be included.

(3) I never was crazy about "Relevance", which (a) has no descriptive power and (b) is inaccurate for those people (for instance, those who like long sentences) for whom it doesn't actually bring up the most relevant sentences. What about either "Closest matches first" or (sacrificing conciseness for explicitness) "Closest matches and fewest words first"? 

(4) Not sure about "Member status: highest rank" and "Member status: lowest rank". This wording might make people feel competitive. We could sidestep this by using "Member status: admin to contributor" or "Member status: contributor to admin".

Otherwise, I think the wording is fine.

trang:
Thanks for your feedback, @alanfgh.

(1) We'll then go with "Alphabetical" and "Reverse alphabetical".

(2) "Newest first" / "Oldest first" is fine. It's still pretty short compared to some of the other strings.

(3) I'm okay with "Closest matches first" instead of "Most relevant" (and I suppose "Closest matches last" instead of "Least relevant").

(4) I was also a bit uncomfortable with "highest rank" / "lowest rank" but had not better idea. "Member status: admin to contributor" seems like a good solution. I suppose we can shorten this a bit with just "Status: admin to contributor".

trang:
As discussed during our weekly call, this would be okay to merge as is, with the initial design proposed (that is, with the arrow next to each sort criterion).

The second design (with the "Sort" button and the options listed in a menu) can be implemented in a couple of months and should be done together with #1063, I think.

jiru:
I’m going to merge this PR because it’s getting old and it already solves the ticket it refers to. Feel free to open a new issue or PR for further enhancements. If not, the new design should make it better at some point.

Issue #1579, opened 2018-06-19
Some gui strings that are visible to admins are not translated/translatable


Guybrush88:
Hello, I'd like to report that the strings related to audio that are visible just to admins on any sentence page aren't translated and aren't translatable on Transifex. I'm attaching a screenshot of a sentence page to show this.
![schermata del 2018-06-19 14-24-08](https://user-images.githubusercontent.com/1702194/41597077-b1ac87ea-73cc-11e8-8671-8af69e7d5120.png)

trang:
To clarify, the strings in question are in the "Audio" block:
- yes
- no
- Owner Name

agrodet:
Wouldn't it be better to be the following?

- Yes
- No
- Owner name

trang:
Yes.

PR #2110, opened 2020-01-28
Make admin audio menu strings translatable

agrodet:
This PR solves #1579.

### DIFF
--- a/src/Template/Element/sentences/audio.ctp
+++ b/src/Template/Element/sentences/audio.ctp
@@ -63,8 +63,8 @@ if (CurrentUser::isAdmin()) {
             "label" => false,
             "type" => "radio",
             "options" => array(
-                1 => __("Yes"),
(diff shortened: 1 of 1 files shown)


jiru:
Thanks for the PR, @agrodet. :-)

Since the strings are only visible to admins, they should go in the admin domain using the macro `__d()`. See the translated strings in `src/Model/Table/AudiosTable.php` for an example.

Translatable strings marked with `__()` actually belong to the domain called "default", which all end up into the default.pot file. Strings with a specified domain end up in a different file, like admin.pot. These domains are [translated separately on Transifex](https://www.transifex.com/tatoeba/tatoeba_website/translate/#fr).

agrodet:
Aaah, that's why. Sorry for not making the connection.

jiru:
Thank you!

Issue #1732, opened 2019-01-11
Sentences can be added/removed even when "Who can add/remove sentences" is set to "No one"


alanfgh:
If I set the "Who can add/remove sentences" option for a list to "No one (list inactive)", I would expect that if I click the red "X" above a sentence, the UI will stop me from deleting it (or better yet, will not display a red "X" in the first place). But that's not true. Either the wording of the GUI elements relating to this option should be changed or the UI should prevent me from deleting sentences from this list. (The UI already prevents me from adding sentences to it by stripping the name of the list from the drop-down list above the sentence.) 

Note that this is true on the production site as well as on the dev site.

agrodet:
I'll check this.

agrodet:
@trang I'm not 100% percent sure of how sql files in the tables folder and the updates folder are supposed to interact with each other. I was expecting the usual migration stuff, but in the `CREATE TABLE` we have 
https://github.com/Tatoeba/tatoeba2/blob/22e4865521475996b3e00b64f7b7639e5420982a/docs/database/tables/sentences_lists.sql#L30

And we have two updates on this table, one is 
https://github.com/Tatoeba/tatoeba2/blob/22e4865521475996b3e00b64f7b7639e5420982a/docs/database/updates/2016-01-03.sql#L1-L5
the other is 
https://github.com/Tatoeba/tatoeba2/blob/22e4865521475996b3e00b64f7b7639e5420982a/docs/database/updates/2016-05-22.sql#L1

Aside my surprise that the CREATE TABLE code has the same column definition than the first update, looking at the second update, I was expecting the `editable_by` column definition to be `enum('creator', 'anyone', 'no_one')`. However, looking at the table in my dev environment, the column definition is `enum('creator', 'anyone')` so it's not surprising that the setting doesn't work there (although I think there are other logic issues). Is the problem in my dev environment? Maybe the table was not updated when I created the environment? Does the prod table has the correct column definition?

PS: I looked at the history of the files and the `CREATE TABLE` was modified at the same time than the first update file.

trang:
We forgot indeed to update the sentences_lists.sql. The prod database has the value `no_one`. In your dev environment it is normal that you won't have it, because Imouto will use the database/table scripts to create the database.

```
MariaDB [tatoeba]> describe sentences_lists;
+-------------------+-------------------------------------+------+-----+----------+----------------+
| Field             | Type                                | Null | Key | Default  | Extra          |
[...]

PR #2111, opened 2020-01-29
Prevent removing sentences of lists editable by no one

agrodet:
This PR addresses #1732.

When the `editable_by` parameter is modified from or to "No one", the page is reloaded so the delete icons appear or disappear accordingly. 

I took the liberty of modifying the return value of the `set_option()` method as I think `$result['SentencesList']` is not used.


jiru:
@agrodet Thanks for working on this.

About the reloading of the page to update the delete buttons, I think your code can be improved. Since it’s more of a UI problem than a Controller problem, how about calculating the value of the reload variable in Javascript instead? You could compare the returned `$result['SentencesList']` with $scope.editable.

Also, could you add unit tests to validate your modifications in `src/Model/Table/SentencesListsTable.php`? You’d have to add a new list to the fixture with the setting `editable_by => no_one`. If you don’t feel like adding unit tests, no worries, I can merge your PR and add tests after.

webroot/js/sentences_lists.set_option.js
```diff
@@ -40,9 +40,12 @@
                 rootUrl + "/sentences_lists/set_option/",
                 { "listId": listId, "option": "editable_by", "value": $scope.editable },
                 function () {$(".is-editable.loader-container").hide();}
-            ); 
-            
+            ).then(function success(response) {
+                if(response.reload)
+                    location.reload();
+            });
+
```
jiru: There is already a callback function up there that does `$(".is-editable.loader-container").hide();`. Could you group the callback code into a single function?

agrodet:
Thank you for your feedback and advice, as usual! I finally had some time to have a look at it. Note that I'm not very familiar with AngularJS (I think that's obvious :D). I used 
```ng-change='editableChanged("{{editable}}")'``` 
to get the old value of the setting and compare it to the new one. Is that the correct way to do it? I couldn't find any direct `$scope.editable.old()`-like way to get the old value.

Please wait some more time for the tests (sorry!)

File: src/Model/Table/SentencesListsTable.php [hunk 2]
```diff
@@ -82,7 +82,7 @@ private function _getPermissions($list, $currentUserId) {
             'canView' => $visibility != 'private' || $belongsToUser,
             'canEdit' => $belongsToUser,
             'canAddSentences' => $belongsToUser && $editableBy !== 'no_one',
-            'canRemoveSentences' => $belongsToUser || $editableBy == 'anyone',
+            'canRemoveSentences' => $editableBy == 'anyone' || $belongsToUser && $editableBy == 'creator',
```
jiru: The new value for `canRemoveSentences` is definitely correct now. Now I realize that the value for `canAddSentences` is incorrect. So the "Add a sentence to this list" form on the list page doesn’t show up for other users when the list is editable by anyone. Feel free to submit a new PR that fixes that.

jiru:
Thank you for the update, agrodet. I am also learning AngularJS, so I cannot tell you whether that’s the correct way to do it or not. I think we both read the same Stackoverflow page. :upside_down_face: 

Thank you also for taking the time to add unit tests so that we are now really sure that nobody can remove sentences from lists editable by no one. :+1:

tests/TestCase/Model/Table/SentencesListsTableTest.php
```diff
@@ -198,7 +198,33 @@ function testAddSentenceToList_failsBecauseUserNotAllowed() {
...
+
+        $this->assertFalse($result);
+        $this->assertEquals($after, $before);
+    }
+
+    function testAddSentenceToList_otherUserFailsBecauseEditableByNoOne() {
+        $before = $this->SentencesList->SentencesSentencesLists->find()
+            ->where(['sentence_id' => 12, 'sentences_list_id' => 6])
+            ->count();
+        $result = $this->SentencesList->addSentenceToList(12, 6, 4);
```
jiru: Could you make the tests a little bit more easy to understand by using variables instead of numbers? For example:
```php
$userId = 4;
$listId = 6;
$sentenceId = 12;
$this->SentencesList->addSentenceToList($sentenceId, $listId, $userId);
```

This also ensures that you’re not mixing up numbers, forcing the use of same numbers in the `where()` clause and in the `addSentenceToList()` call.

webroot/js/sentences_lists.set_option.js
```diff
@@ -33,14 +33,13 @@
...
             $.post(
                 rootUrl + "/sentences_lists/set_option/",
-                { "listId": listId, "option": "editable_by", "value": newSetting },
-                function () {
+                { "listId": listId, "option": "editable_by", "value": $scope.editable },
+                function (response) {
                     $(".is-editable.loader-container").hide();
-                    if (newSetting === "no_one" || oldSetting === "no_one")
+                    if (response.editable_by === "no_one" || oldSetting === "no_one")
                         location.reload();
```
AndiPersti: Wouldn't it better to avoid the reloading and just change the visibility of the remove icon based on the value of `editable`?

I'm not an AngularJS expert either, but one way to share a model outside of the controller's element subtree is using the `$rootScope` (i.e. a global object). See https://github.com/AndiPersti/tatoeba2/commit/047ce40e89c47847270d150ca5a9e7f76c93bd71

Alternatively we could add a controller to the sentences list and inject a [service](https://docs.angularjs.org/guide/services) in both the `OptionsCtrl` and that new controller.

trang: The X button is actually not included in the HTML when the list is not editable at first, so reloading would be necessary if changing from "editable by no one" to "editable by someone".

-------

Assuming we include the HTML every time and just hide it based on the situation, we have the following possibilities:

(1) Since this script is still a hybrid of jQuery and AngularJS, to change the visibility of the X button you could simply do this:
```
if (response.editable_by === "no_one") {
    $(".removeFromList").hide();
} else {
    $(".removeFromList").show();
}
```

(2) If you want to go for the Angular way, using the $rootScope would be a solution (people usually don't recommend using $rootScope but I think in our case it would not be a big deal).

The other way would be to set `ng-controller='optionsCtrl'` on the whole page rather than just the options block. But in that case the name `optionsCtrl` is not very accurate anymore. It should rather be "SentencesListsShowCtrl" and the file could be moved to `/webroot/js/sentences_lists/show.ctrl.js`.

In both cases you'll have to remember to add `ng-if="editable"` on the X button.

-------

That being said, I would suggest to keep the reloading for now. We can later on rewrite this part of the code to be 100% angular (which will allow us to remove the reloading of the page).

jiru: I, too, think that reloading the page is acceptable for now.

tests/TestCase/Model/Table/SentencesListsTableTest.php
agrodet: Sure. I'll try to find some time today. What about the other tests? Do you want them rewritten or should we let them as they are?

src/Model/Table/SentencesListsTable.php
agrodet: I'll do that later, yes.

tests/TestCase/Model/Table/SentencesListsTableTest.php
```diff
@@ -347,12 +353,18 @@ function testRemoveSentenceFromList_failsBecauseUnknownListId() {
...
+        $sentenceId = 20;
+        $listId = 6;
+        $userId = 4;
+        $result = $this->SentencesList->removeSentenceFromList($sentenceId, $listId, $userId);
         $this->assertFalse($result);
     }
 
     function testRemoveSentenceFromList_ownerFailsBecauseEditableByNoOne() {
-        $result = $this->SentencesList->removeSentenceFromList(20, 6, 7);
+        $sentenceId = 12;
```
jiru: Shouldn’t sentence ID equal 20 here?

agrodet: Sorry...

Issue #2024, opened 2019-11-26
Internal error on page listing sentences in Kirundi, Yiddish and other languages


trang:
Reported by [driini on the Wall](https://tatoeba.org/eng/wall/show_message/33586#!#message_33586):

> Wenn ich auf der Seite
> https://tatoeba.org/deu/stats/sentences_by_language
> auf kirundi klicke, kommt eine Fehlermeldung:
> "an internal error has occurred"
> (https://tatoeba.org/deu/sentences/show_all_in/run/none/none/indifferent)

AndiPersti:
Yiddish also results in an internal error. All other languages are fine.

Do you have any logs?

trang:
For [Kirundi](https://tatoeba.org/deu/sentences/show_all_in/run/none/none/indifferent):

> 2019-11-27 11:05:27 Error: Fatal Error (1): Allowed memory size of 134217728 bytes exhausted (tried to allocate 2097160 bytes) in [/var/www-prod-a079f71bef8b6b93a2d7137906db99b76a15a113/vendor/cakephp/cakephp/src/Database/Statement/BufferedStatement.php, line 218]
>
> Trace:
> Cake\Error\BaseErrorHandler::handleFatalError() - CORE/src/Error/BaseErrorHandler.php, line 229
> Cake\Error\BaseErrorHandler::Cake\Error\{closure}() - CORE/src/Error/BaseErrorHandler.php, line 109
> [main] - [internal], line ??

For [Yiddish](https://tatoeba.org/eng/sentences/show_all_in/yid/none/none/indifferent):

> 2019-11-27 11:06:26 Error: Fatal Error (1): Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in [/var/www-prod-a079f71bef8b6b93a2d7137906db99b76a15a113/vendor/cakephp/cakephp/src/Database/Type/IntegerType.php, line 119]
> 
> Trace:
> Cake\Error\BaseErrorHandler::handleFatalError() - CORE/src/Error/BaseErrorHandler.php, line 229
> Cake\Error\BaseErrorHandler::Cake\Error\{closure}() - CORE/src/Error/BaseErrorHandler.php, line 109
> [main] - [internal], line ??

I'm also noticing that on the dev website, there is no error when *not* authenticated:
https://dev.tatoeba.org/eng/sentences/show_all_in/run/none/none/indifferent

alanfgh:
[...]

PR #2115, opened 2020-01-31
Optimize paginateLatest

AndiPersti:
For calculating the lowest id we only need the associations used in the 'select' and 'order' part of the query.

Since CakePHP eagerly loads all the associations for a table, using the original 'contain' part of the query results in unnecessary database queries.

For an unregistered user, this new version reduces the number of queries sent to the database from 13 to 9 for the 'Browse by language' page. For a registered user the number of queries is reduced from 26 to 16 with the new design and from 56 to 46 with the old design.

### DIFF
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -327,24 +327,55 @@ class AppController extends Controller
         return $this->redirect($url);
     }
 
+    /**
+     * Calculate minimal associations for a query
+     *
+     * Helper function for paginateLatest which filters the associations to
+     * load for the given query. Only the associations mentioned in the 'select'
+     * and 'order' part are necessary for calculating the lowest id we need.
+     *
+     * @param array $conditions The conditions for the query
+     * @param array $order      The ordering for the query
+     * @param array $contain    The original contain part of the query
+     *
+     * @return array
+     **/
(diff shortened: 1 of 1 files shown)


AndiPersti:
`paginateLatest` is currently used in 3 controllers: `ContributionsController`, `SentencesController` and `SentenceCommentsController`. In all 3 controllers we don't use any associations in the conditions or the orderings so simply removing the `contain` part of the query at [line 340](https://github.com/Tatoeba/tatoeba2/blob/42c74ed8cf9735e172ad1dd97b0d07b74500e440/src/Controller/AppController.php#L340) would be sufficient.

But I've chosen to use a more general solution which calculates the minimal `contain` part that is necessary.

This PR may also fix #2024 but I don't have a large enough database to test it locally.

jiru:
Thank you for this PR, @AndiPersti :-)

I confirm your patch optimizes things a lot, not only in terms of number of queries but query complexity too. When testing on dev on show_all_in, I saw is a huuuge request joining an enormous list of translations ids. It’s likely that this is what causing bug #2024, if the displayed sentences happen to have a lot of translations.

> I've chosen to use a more general solution which calculates the minimal contain part that is necessary.

It’s okay.

PR #2118, opened 2020-02-02
Change alternate URLs to absolute URLs

### LINKED ISSUE #507: Have proper localized versions of the site returned by SERP
Currently, when someone searches tatoeba on Google, the English version always comes first, even if you use a Google country-specific TLD (such as google.fr), and have your browser preferred language set to the country’s main language. To make the matter worse, all the results from tatoeba.org are grouped into a mix of localized and English pages.

[Having a proper sitemap may solve the problem](https://stackoverflow.com/questions/11518239/google-organic-results-localization).

AndiPersti:
This may fix #507 

As explained in that issue, using absolute alternate URLs may improve the ranking for our localized versions when we search for 'tatoeba' on a country-specific Google version (google.fr, google.es, ...).

### DIFF
--- a/src/Template/Element/seo_international_targeting.ctp
+++ b/src/Template/Element/seo_international_targeting.ctp
@@ -29,19 +29,15 @@ use App\Lib\LanguagesLib;
 use Cake\Core\Configure;
 
 $uiLanguages = Configure::read('UI.languages');
-$currentLang = Configure::read('Config.language');
-$currentUrl = $this->request->here();
+$currentLang = $this->request->getParam('lang');
 
-$pos = strpos($currentUrl, $currentLang);
-if ($pos !== false) {
-    foreach ($uiLanguages as $langs) {
-        $alternateURL = substr_replace(
-            $currentUrl,
-            $langs[0],
-            $pos,
-            strlen($currentLang)
-        );
-        $hreflang = LanguagesLib::languageTag($langs[0], $langs[1]);
+if ($currentLang) {
+    $pathWithoutLang = substr($this->request->getPath(), 4);
+    $host = $this->request->host();
+    $scheme = $this->request->scheme();
+    foreach ($uiLanguages as $lang) {
+        $alternateURL = $scheme . '://' . $host . '/' . $lang[0] . $pathWithoutLang;
+        $hreflang = LanguagesLib::languageTag($lang[0], $lang[1]);
         ?>
         <link rel="alternate"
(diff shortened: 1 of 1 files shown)


src/Template/Element/seo_international_targeting.ctp
```diff
@@ -29,19 +29,15 @@ use App\Lib\LanguagesLib;
...
-    foreach ($uiLanguages as $langs) {
-        $alternateURL = substr_replace(
-            $currentUrl,
-            $langs[0],
-            $pos,
-            strlen($currentLang)
-        );
-        $hreflang = LanguagesLib::languageTag($langs[0], $langs[1]);
+if ($currentLang) {
+    $pathWithoutLang = substr($this->request->getPath(), 4);
```
jiru: Your code works, but I’m a little bit concerned about the assumption that the language part of the URL consists of the first four characters. It mostly does, but it’s more of a convention of ours than something restricted by code. Actually we had no such convention in the past. For example we used to use https://tatoeba.org/pt_BR/ for Portuguese, and this now redirects to /por. If you look at `config/routes.php`, you’ll see that the `lang` parameter is currently constructed with a regexp fed by values from the array `Configure::read('UI.languages')`. The values used in that array are the first and the last, for example we have configured:
```
['por', 'BR', 'Português (Brasil)', ['pt_BR']],
```

So the constructed regexp looks like `/(eng|spa|fra|por|pt_BR|...)/` (as opposed to `/(...)/` like your code assumes). In practice when opening /pt_BR, the code in `seo_international_targeting.ctp` will never be reached because it always redirects before, but it’s just because we configured it to do.

In other words, Tatoeba does support language codes that are not 3 chars long and I’d like we keep it like that.

src/Template/Element/seo_international_targeting.ctp
```diff
@@ -29,19 +29,15 @@ use App\Lib\LanguagesLib;
 use Cake\Core\Configure;
 
 $uiLanguages = Configure::read('UI.languages');
-$currentLang = Configure::read('Config.language');
-$currentUrl = $this->request->here();
+$currentLang = $this->request->getParam('lang');
```
jiru: Good point, it makes more sense to get the lang from getParam() than Config.language here.

src/Template/Element/seo_international_targeting.ctp
AndiPersti: Agreed. It wasn't a good idea to rely on a magic number.

Issue #2121, opened 2020-02-03
Internal error when opening the page for adopting sentences


Guybrush88:
Currently, each time I want to open the page that contains the orphan sentences from the top menu (i.e. https://tatoeba.org/ita/activities/adopt_sentences/ita, according to the interface language), I get an internal error message, so I can't open the page

alanfgh:
I can reproduce this.

jiru:
It looks like the error only happens with the option "(native)" is turned on.

jiru:
In addition, it looks that it’s not possible to unadopt with the setting enabled.

AndiPersti:
I'm sorry for the inconvenience. I've just pushed a fix.

Guybrush88:
@AndiPersti no problem, and thanks for the very quick fix :D

ckjpn:
## This can also be done with the advanced search.

This mimics the current way, without hiding the translations.
https://tatoeba.org/eng/sentences/search?orphans=yes&from=eng&sort=created

You can adjust the sorting and other additional search functions, too, if you wanted to.

Sort: words
https://tatoeba.org/eng/sentences/search?orphans=yes&from=eng&sort=words

Sort: random
https://tatoeba.org/eng/sentences/search?orphans=yes&from=eng&sort=random

Other options would be ...

- limit showing only ones with certain translations
- don't show any translations

jiru:
I’ve pushed Andreas’ fix to tatoeba.org, so you guys may now use the native indicator again.

PR #2122, opened 2020-02-03
Make "native" feature more specific

AndiPersti:
This PR addresses #2121 

The implementation for the "native" feature was too general and tried to modify queries which have nothing to do with that feature.

I've added a flag to mark the queries which should add the `is_native` column to the result.


jiru:
Thank you for the prompt and elegant fix!

Issue #2064, opened 2020-01-10
Furigana is no longer hidden with "Always show transcriptions and alternative scripts" unchecked.


ckjpn:
## Problem

Furigana is no longer hidden with "Always show transcriptions and alternative scripts" unchecked.

Until recently, I wasn't seeing the furigana, but it's now showing on Japanese sentences.  I'm pretty sure it's related to the above setting option now working, but I could be wrong.

ckjpn:
I think the problem is that if someone has edited the automatically-generated furigana, then it gets shown regardless of a member's settings.

The recent Japanese contributions by this member have manually-created furigana.

https://tatoeba.org/eng/transcriptions/of/small_snow

Yorwba:
Having this setting disabled currently has the effect that automatically-generated transcriptions are hidden by default, but manually-confirmed ones are shown. I think that is intended to hide information that is potentially incorrect or misleading, but to show it when someone has confirmed that is correct. (Which is why this setting is disabled by default.)

It sounds like you never want to see any transcriptions at all. I think the current behavior of the setting is desirable, so your request might better be implemented as an additional option "Never show transcriptions and alternative scripts".

FWIW, furigana are supposed to not interfere with copy-pasting. If I select a Japanese sentence with furigana and copy it (or if I use the experimental "copy button" option), the furigana are removed when I paste it somewhere else.

agrodet:
Yorwba already explained it all. The opposite of "always" is not never, that's all. If you want to add another setting, you can open the discussion.

ckjpn:
It's OK to close this if you want to.

However, if someone wants to make this into a user option, see the CSS on this page.

https://userstyles.org/styles/132460/tatoeba-org-hide-all-furigana-on-japanese

I'd forgotten I had written this since it wasn't needed in most cases after the "Always show transcriptions and alternative scripts" option was added.

I didn't have it installed on my newer computer, but it still works and accomplishes what the title of this issue says.

jiru:
@CK Why do you want to hide the furigana? What’s your use case?

Yorwba:
Earlier I wrote

> FWIW, furigana are supposed to not interfere with copy-pasting. If I select a Japanese sentence with furigana and copy it (or if I use the experimental "copy button" option), the furigana are removed when I paste it somewhere else.

I was assuming that CK was simply unaware that the furigana getting highlighted does not necessarily mean they will be present after pasting.

[...]

PR #2123, opened 2020-02-03
Stop furigana from interfering with copying (fixes #1077) (for real this time)

### LINKED ISSUE #1077: The furigana interferes with copying the main Japanese sentence.
https://tatoeba.org/sentences/show/5006900

screen shot 2016-03-25 at 19 02 41

Yorwba:
#1077 was marked as fixed by 70ef314, but unfortunately that change was later reverted in 0fbcf53 because it broke the copy button functionality.

This PR:

1. Restores the Chrome copy-paste hack as it existed in 0fbcf53
2. Adds a workaround for ignoring copy events generated by the hidden textarea ClipboardJS uses in Firefox.
3. Adds some CSS to make ruby text non-selectable, so that it is obvious to users that it won't be copied.

The Chrome copy-paste hack actually also fixes the problem that copying in Firefox and pasting in LibreOffice Writer loses the kanji, which I mentioned in [this issue comment](https://github.com/Tatoeba/tatoeba2/issues/2064#issuecomment-581401790)

Tested in Firefox and Chromium on LInux, as well as Firefox for Android, Firefox Preview and Chrome for Android.

### DIFF
--- a/webroot/css/layouts/default.css
+++ b/webroot/css/layouts/default.css
@@ -181,6 +181,14 @@ svg {
 rt {
     font-size: 50%;
     color: #A1A1A1;
+    /* Prevent selecting ruby text, for which copying is disabled.
+     * From https://stackoverflow.com/a/4407335 */
+    -webkit-touch-callout: none;
+    -webkit-user-select: none;
+     -khtml-user-select: none;
+       -moz-user-select: none;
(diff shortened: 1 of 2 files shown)


jiru:
Thanks for working on this, @Yorwba!

PR #2137, opened 2020-02-10
Favorites and adoption in new sentence design

trang:
This PR implements the favorite and adopt features in the new sentence design.

*Favorites**

This feature keeps pretty much the same behavior as in the old design:
- When the sentence is not favorited, an outlined "heart" icon is displayed with a tooltip "Add to favorites".
- When the sentence is favorited, a filled "heart" icon is displayed with a tooltip "Remove from favorites".

The differences:
- I did not add any progress animation. I think it's not really necessary here.
- The favorite button is not displayed by default so it takes more clicks to favorite sentences. This shouldn't have too much impact. The favorite feature is not heavily used.

**Adoption**

- The adopt/unadopt button is displayed at the end of the sentence menu whenever adoption or unadaoption is possible. It is hidden by default and is displayed after clicking on the expand (...) button.
- When the sentence can be unadopted (it belongs to the current user), a filled "person" icon is displayed with the tooltip "Click to unadopt".
- When the sentence can be adopted, an outlined "person" icon is displayed with the tooltip "Click to adopt". There are two cases for this:
	- The sentence has no owner.
	- The sentence is owned by an inactive member and the current user is advanced contributor or higher.
- The sentence header is updated upon adoption/unadoption to display the correct ownership.
- On the "Adopt sentences" page, the sentence menu is by default expanded in order to avoid extra clicks to adopt sentences. This hides the sentence ID but the sentence ID is not a useful information in the context of adopting sentences.

### DIFF
--- a/src/Controller/SentencesController.php
+++ b/src/Controller/SentencesController.php
@@ -129,8 +129,6 @@ class SentencesController extends AppController
           'edit_sentence'
         ]);
 
-        $this->loadComponent('RequestHandler');
-
         return parent::beforeFilter($event);
     }
 
@@ -411,6 +409,7 @@ class SentencesController extends AppController
         ]);
 
         if ($acceptsJson) {
+            $this->loadComponent('RequestHandler');
             $this->set('user', $sentence->user);
             $this->set('_serialize', ['user']);
             $this->RequestHandler->renderAs($this, 'json');
--- a/src/Controller/SentencesController.php
+++ b/src/Controller/SentencesController.php
@@ -346,10 +346,11 @@ class SentencesController extends AppController
      *
      * @return void
      */
-    public function edit_sentence($type = 'html')
+    public function edit_sentence()
     {
+        $acceptsJson = $this->request->accepts('application/json');
         $sentence = $this->Sentences->editSentence($this->request->data);
-        if ($type == 'json') {
+        if ($acceptsJson) {
             $sentence->dir = LanguagesLib::getLanguageDirection($sentence->lang);
             $this->set('result', $sentence);
             $this->viewBuilder()->setLayout('json');
@@ -373,13 +374,13 @@ class SentencesController extends AppController
      *
      * @return void
      */
-    public function adopt($id, $type = 'html')
+    public function adopt($id)
     {
         $userId = $this->Auth->user('id');
 
         $this->Sentences->setOwner($id, $userId, CurrentUser::get('role'));
 
-        $this->renderAdopt($id, $type);
+        $this->renderAdopt($id);
     }
 
     /**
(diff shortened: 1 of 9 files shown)


src/Model/CurrentUser.php
```diff
@@ -382,4 +382,15 @@ public static function hasAcceptedNewTermsOfUse()
...
         return !self::isMember() || self::getSetting('new_terms_of_use');
     }
+    
+    public static function canAdoptOrUnadoptSentenceOfUser($user)
+    {
+        if (!$user || !$user->id || $user->id === self::get('id')) {
+            return self::isMember();
+        } else {
+            $userAccountDeactivated = isset($user['role']) ?
+                in_array($user['role'], [User::ROLE_SPAMMER, User::ROLE_INACTIVE]) : false;
```
jiru: `$user['role']` works, but better use `$user->role` instead since you are using `$user->id` just before.

src/Controller/SentencesController.php
```diff
@@ -373,13 +373,13 @@ public function edit_sentence($type = 'html')
      *
      * @return void
      */
-    public function adopt($id)
+    public function adopt($id, $type = 'html')
```
jiru: There are better ways transmit the information "I want json" or "I want html". Although I never played with that, it’s all [explained in the docs](https://book.cakephp.org/3/en/views/json-and-xml-views.html). Basically we have the HTML header `Accept` for that already, or we can use CakePHP’s Routing File Extensions and append `.json` to the requested URL.

From what I’ve seen, the old design uses `Accept: */*` while the new design uses `Accept: application/json, text/plain, */*`, so you could already use that to decide what to return.

jiru:
It’s nice to see that you are making progress on the new design. :smile: 
Looking forward to get this merged.

trang:
Thanks for the review, gillux :) I have applied your improvements.

src/Template/Sentences/adopt.ctp
```diff
@@ -37,4 +37,10 @@
  * @link     https://tatoeba.org
  */
 
-$this->Menu->adoptButton($sentenceId, $owner);
+if ($acceptsJson) {
+    $result = $this->Sentences->getSentenceData($sentence);
+    echo json_encode($result);
```
jiru: If possible, I think it would be better to avoid passing `$acceptJson` to the view and using a helper to format json data here. It’s not really the helper’s responsibility to select which data to return in the response. You could use the `_serialize` key from the controller like I did in `ExportsController.php`.

trang: Thank you for the pointers. I refactored the code for adoption. There will be other parts to refactor but I will handle that when I get to remove obsolete code from the old sentence design.

There was some issue with loading the `RequestHandler` component in the `SentencesController::beforeFilter`.

```
1) App\Test\TestCase\Controller\SentencesControllerTest::testEditSentence_doesntWorkForUnknownSentence

Failed asserting that response has header 'Location'.
```

Was not sure how to solve it so I just loaded the component only in the renderAdopt() function.

jiru:
Looks good! :+1:

Issue #2131, opened 2020-02-09
Sentences added with spaces at the start


trang:
Cf. logs of the following sentences:
https://tatoeba.org/eng/sentences/show/8539416
https://tatoeba.org/eng/sentences/show/8539417
https://tatoeba.org/eng/sentences/show/8539418

Those spaces should be automatically removed before saving.

AndiPersti:
The sentences are rather recent and not available in the weekly dump but I guess these all contain some sort of unicode white space characters. At least I can reproduce this on dev: https://dev.tatoeba.org/eng/sentences/show/7841128
The reason is that we use `trim()` for cleaning the sentence text, but this function just removes the code points 9, 10, 11, 13, 32 of all possible [Unicode whitespace characters](https://en.wikipedia.org/wiki/Whitespace_character#Unicode).

I've looked at the current export and noticed that there are a few similar sentences included:

Space at the beginning:

    $ cut -f 1,3 sentences.csv | grep '  [[:space:]]'
    375562   قدمي ما زالت تؤلمني.
    375751   تتكلم ماريكو الإنجليزية جيداً.
    377093   كما يسبب استمرار هذا النزاع تكاليف باهظة ومصاعب سياسية جمة..
    2250282  لقد حصل على العلامة الكاملة.
    8185348  I'm so grateful to work here every day. It's so nice.
    8185349  I'm so grateful to work here every day.

With the 4 Arabic sentences, the punctuation looks strange, e.g. 377093:

    $ grep '     377093  ' contributions.csv
    saeb    2010-04-08 12:28:44     insert  sentence        377093  ara     \N      . كما يسبب استمرار هذا النزاع تكاليف باهظة ومصاعب سياسية جمة.
    saeb    2010-04-08 12:28:44     insert  link    377093  \N      377092
    saeb    2010-04-08 12:28:44     insert  link    377092  \N      377093
    $ grep '^377093      ' sentences.csv
    377093  ara      كما يسبب استمرار هذا النزاع تكاليف باهظة ومصاعب سياسية جمة..

The sentence was contributed with a period at the beginning and at the end, but is stored now with two periods at the end. Maybe related to #2075?

Both English sentences have a `LINE SEPARATOR` (U+2028 / e2 80 a8) at the beginning:

    $ grep '^8185348     ' sentences.csv | hexdump -C
    00000000  38 31 38 35 33 34 38 09  65 6e 67 09 e2 80 a8 49  |8185348.eng....I|
[...]

PR #2141, opened 2020-02-12
Improve text cleaning

AndiPersti:
This PR addresses #2131 and #2138 

In order to improve the text cleaning I've written a new version of `trim()` which removes all 25 Unicode whitespaces by default.


jiru:
Thank you for your contribution Andreas, especially the tests.

If I were to implement this, I would have used regexps and Unicode classes, for example:
```php
$frontTrimmed = preg_replace('/^[\p{Zs}\p{Cc}]/u', '', $sentence);
$endTrimmed = preg_replace('/[\p{Zs}\p{Cc}]$/u', '', $sentence);
$middleReplaced = preg_replace('/[\p{Zs}\p{Cc}]{2,}/u', ' ', $sentence);
```
`\p{Cc}` represents [all control characters](https://www.compart.com/en/unicode/category/Cc) and `\p{Zs}` [all spaces](https://www.compart.com/en/unicode/category/Zs).

What do you think?

AndiPersti:
> If I were to implement this, I would have used regexps and Unicode classes, for example:

Actually I've thought about using regexps but I can't remember now why I decided not to use them. But they are definitely a better solution.

I've updated the PR with the following changes:
- Used `\p{Z}` because we also want to remove `LINE SEPARATOR` (category `Zl`) and `PARAGRAPH SEPARATOR` (category `Zp`).
- Used `+` quantifier to catch more than one problematic character and added a test for this case.
- Modified the regex for reducing consecutive whitespace characters to one as suggested.
- Removed `unicode_trim()` and the tests for it

jiru:
Looks good, thank you! :+1:

PR #2146, opened 2020-02-15
FixHashesCommand: Added option for reading ids from stdin

AndiPersti:
In addition this new version also checks whether any of the given
columns changed after creating a new entity from the old values. If this
is the case, all columns are updated and not only the hash column.

As a result it is now possible to fix specific entities in cases where
any of the source columns is updated with new contstraints (e.g.
normalizing Unicode text).


AndiPersti:
I've implemented the reading from stdin with the standard `file()` function but this has the disadvantage that writing tests for it is impossible AFAIK because the Cake test suite expects to use `Cake\Console\ConsoleIo` for input/output in command/shells. But there is only one usable method for getting input (`ask()`) which will prompt the user. Unfortunately it is not possible to disable the output of the prompt completely. It will always display at least `>` which is rather annoying if you want to use the command in a pipe:

    $ cake fix_hashes -i Sentences < ids.txt
    >
    > 
    > 
    ....
    123 rows checked, ...

Instead of a test here is an example session from my local dev environment:

	$ echo -e "1\n3\n87\n999" > ids
	$ cat ids
	1
	3
	87
	999
    # Dry run over the whole database
    $ cake fix_hashes -n Sentences
	73 rows checked, 3 rows changed:
	id 84 - hash changed from "3ea23d" to "614uh7"
	id 87 - hash changed from "23ds5k" to "15cmflk"
	id 91 - hash changed from "61hrs2w" to "27v89dg"
	$ echo "select id, concat('#', text, '#'), hash from sentences where id in (84, 87, 91);" | mysql tatoeba
	id      concat('#', text, '#')  hash
	84      #  test   #     3ea23d\0\0\0\0\0\0\0\0\0\0
	87      #  La ĉielarka flago estas la simbolo de respekto de seksa diverseco.#  23ds5k\0\0\0\0\0\0\0\0\0\0
	91      #  请问您贵姓 # 61hrs2w\0\0\0\0\0\0\0\0\0
    # Run it with some ids (even wrong ones)
	$ cake fix_hashes -i Sentences < ids
	999 ignored: Record not found
	4 rows checked, 1 rows changed:
	id 87 - hash changed from "23ds5k" to "15cmflk"
	$ echo "select id, concat('#', text, '#'), hash from sentences where id in (84, 87, 91);" | mysql tatoeba
	id      concat('#', text, '#')  hash
	84      #  test   #     3ea23d\0\0\0\0\0\0\0\0\0\0
	87      #La ĉielarka flago estas la simbolo de respekto de seksa diverseco.#    15cmflk\0\0\0\0\0\0\0\0\0
	91      #  请问您贵姓 # 61hrs2w\0\0\0\0\0\0\0\0\0
    # Fix all problematic sentences
	$ echo -e "84\n87\n91" | cake fix_hashes -i Sentences
	3 rows checked, 2 rows changed:
	id 84 - hash changed from "3ea23d" to "614uh7"
	id 91 - hash changed from "61hrs2w" to "27v89dg"
	$ echo "select id, concat('#', text, '#'), hash from sentences where id in (84, 87, 91);" | mysql tatoeba
	id      concat('#', text, '#')  hash
	84      #test#  614uh7\0\0\0\0\0\0\0\0\0\0
	87      #La ĉielarka flago estas la simbolo de respekto de seksa diverseco.#    15cmflk\0\0\0\0\0\0\0\0\0
	91      #请问您贵姓#    27v89dg\0\0\0\0\0\0\0\0\0
    # Database is fixed now
	$ cake fix_hashes Sentences
	73 rows checked, no problems found.

jiru:
I’ve looked into how you can make your code testable. It seems your use case is too specific to be able to use CakePHP’s ConsoleIo. So instead, you could add a method that allows to change php://stdin to something else:

```php
class FixHashesCommand extends Command
{
    private $input = 'php://stdin';

    public function setInput($input) {
        $this->input = $input;
    }

    public function execute(...) {
        $ids = file($this->input, ...);
        ...
    }
}
```

And your test would look like:
```php
class FixHashesCommandTest extends TestCase
{
    public function testInput() {
        $io = new \Cake\Console\ConsoleIo();
        $args = ['Sentences'];
        $cmd = new \App\Command\FixHashesCommand();
        $cmd->setInput(TESTS.'Fixture'.DS.'sentence_ids.txt');
        $exitCode = $cmd->run($args, $io);
    }
```

And then commit along with your test input file `tests/Fixture/sentence_ids.txt`.

AndiPersti:
> So instead, you could add a method that allows to change php://stdin to something else:

@jiru Good idea :+1: This pushed me into the right direction :smile: 

We don't need that method. I've just changed the `-i` option to accept a filename and use a temporary file for testing.

jiru:
Yay, that looks really nice, Andreas!

About using "stdin" as a keyword for stdin, I’d like to suggest using "-" instead. It’s a convention among many CLI tools to use "-" as a synonym for stdin and stdout.

It looks like the temporary files created by `testExecute_withInputOption()` never get deleted. Try `ls /tmp/php*` after running that test. While it’s not that much of a problem on your local development machine, I’m afraid it may clutter the continuous integration server.

You could just remove them yourself after `$this->exec()`, but this only works when the tested code doesn’t error. That’s why I suggest removing the file from the `tearDown()` method of your test file. See `ExportsControllerTest.php` for some inspiration.

AndiPersti:
> About using "stdin" as a keyword for stdin, I’d like to suggest using "-" instead. It’s a convention among many CLI tools to use "-" as a synonym for stdin and stdout.

I wanted to use `-` instead of `stdin` but then I get a warning because there is a bug in the `ConsoleOptionParser` (reading out-of-bounds):

    $ echo "" | cake fix_hashes -i - Sentences
    Notice Error: Uninitialized string offset: 1 in [/home/vagrant/Tatoeba/vendor/cakephp/cakephp/src/Console/ConsoleOptionParser.php, line 1076]
    
    Notice Error: Uninitialized string offset: 1 in [/home/vagrant/Tatoeba/vendor/cakephp/cakephp/src/Console/ConsoleOptionParser.php, line 1077]
    
    0 rows checked, no problems found.

See https://github.com/cakephp/cakephp/blob/1e89aa0399ce4c0f9af764e3b61aa24826278feb/src/Console/ConsoleOptionParser.php#L1076-L1077
So `stdin` was a workaround.

> It looks like the temporary files created by `testExecute_withInputOption()` never get deleted. Try `ls /tmp/php*` after running that test. 

No wonder. I've just noticed that [line 58](https://github.com/Tatoeba/tatoeba2/blob/1b0ca8efeac89fd60138a3835de669527411af35/tests/TestCase/Command/FixHashesTest.php#L58) is completely broken :man_facepalming: 

> You could just remove them yourself after `$this->exec()`, but this only works when the tested code doesn’t error. That’s why I suggest removing the file from the `tearDown()` method of your test file. See `ExportsControllerTest.php` for some inspiration.

Ah, I've missed that CakePHP already provides some helpers for working with files.
I'll change that part later today.

jiru:
> I wanted to use - instead of stdin but then I get a warning

Fair enough.

If I set `tmp` unwritable, I get:

```
vagrant@stretch:~/Tatoeba$ phpunit tests/TestCase/Command/FixHashesTest.php 
PHPUnit 6.5.14 by Sebastian Bergmann and contributors.

Warning Error: mkdir(): Permission denied in [/home/vagrant/Tatoeba/vendor/cakephp/cakephp/src/Filesystem/Folder.php, line 661]

Couldn't create test directory '{self::TESTDIR}'vagrant@stretch:~/Tatoeba$ 
```

I think this is not what you want. I think it’s reasonable not to check for `$folder->errors()` at all because we get the warning from PHP already and the test will fail.

Thinking about it, die()ing here may not be a good idea. If I start the whole test suite and go make some tea while it’s running, I don’t want it to stop in the middle just because that particular test failed.

AndiPersti:
> Thinking about it, die()ing here may not be a good idea. If I start the whole test suite and go make some tea while it’s running, I don’t want it to stop in the middle just because that particular test failed.

Good point about the `die()`ing. 
I've removed the check and also use `mkdir()` for creating the directory because having just `new Folder(self::TESTDIR, true);`, i.e. creating the object just for the side effects, looks a little bit strange to me.

I've noticed that there are two other tests which `die()` if creating a directory fails. We should change them too, shouldn't we?

    vagrant@stretch:~/Tatoeba$ grep -rnF 'die(' tests/
    tests/TestCase/Controller/ExportsControllerTest.php:32:            die("Couldn't create test directory '{$this->testExportDir}'");
    tests/TestCase/Model/Table/ExportsTableTest.php:42:            die("Couldn't create test directory '{$this->testExportDir}'");

jiru:
Yes, we should definitely remove the other `die()` calls. I will take care of it.

PR #2147, opened 2020-02-16
Don't try to change clipboard data in IE (fixes #2142)

### LINKED ISSUE #2142: Copy-paste stopped working on Internet Explorer
LanguageExpert [reported](https://tatoeba.org/wall/show_message/34146) that copy-pasting broke on Internet Explorer. I am pretty sure that #2123 is causing the problem because I reverted these changes on dev.tatoeba.org and asked LanguageExpert to try again and confirmed that the problem disappeared.

Yorwba:
IE doesn't support the `clipboardData` property and there doesn't seem to be another way to modify the clipboard contents while copying. So this PR just does nothing when a copy event fires.

Tested with IE 11.

### DIFF
--- a/webroot/js/generic_functions.js
+++ b/webroot/js/generic_functions.js
@@ -156,9 +156,12 @@ $(document).on('copy', function (e) {
 
     var sel = window.getSelection();
     if (sel.rangeCount > 0) {
-        e.preventDefault();
         $('rt').css('visibility', 'hidden');
-        e.originalEvent.clipboardData.setData('text', sel.toString());
+        var clipboardData = e.originalEvent.clipboardData;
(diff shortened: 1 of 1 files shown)


jiru:
Thank you for the fix, Yorwba. Let’s hope this will be the last one. :smile:

Shouldn’t we prevent the execution of `$('rt').css('visibility', 'hidden');` and `        $('rt').css('visibility', 'visible');` too on IE?

Is the furigana actually copied in IE?

Yorwba:
IE copies the furigana, so you're right, altering the visibility doesn't do anything. I moved the check.

jiru:
Thank you! :+1: I pushed your fix on production too.

Issue #1709, opened 2018-10-23
"Add to list" icon is slightly clipped under some situations


alanfgh:
Under some situations, the "Add to list" icon is slightly clipped at the bottom. See screenshot. If I zoom in or zoom out in the browser, the clipping often disappears. I see this with Firefox on Windows. I don't see it with Chrome on Windows.

![clipped_list_icon](https://user-images.githubusercontent.com/1213027/47368274-55172080-d6af-11e8-88f2-229e19b1a6b6.png)

Issue #1665, opened 2018-09-04
Design issues


Poulpisator:
[Sorry if this has already been mentioned somewhere]

Recently, I've been using Tatoeba on a Mac (not my home, not my PC) and I realized some issue with the current design. I've checked a little bit more in details and here are some of my observations. 

First of all **Mac + firefox** (sorry I cannot tell the exact version but Mac OS X + Firefox last version up to last week), there is this ugly difference between the arrows of the select boxes of the horizontal bar on the top and the menu on the right. 
![mac_firefox](https://user-images.githubusercontent.com/30520443/45032141-ebee3600-b08b-11e8-81bc-475be2d9f999.png)
I faced this strange situation myself on a project and found that the css attributes `border` `border-radius` and `background-color` together makes the arrow different (on the top bar, if you unselect those, the arrow becomes the blue one). Don't ask me the logic of that, but that's a clue. I guess...

For a comparison, here what it looks like on my usual **Windows (8.1) + Firefox**. I never use the "new" design, so never realized before today. 
![firefox_win](https://user-images.githubusercontent.com/30520443/45032432-c44b9d80-b08c-11e8-9a8f-ea9c151ee375.png)

Next is the same page on **Mac + Chrome**
![mac_chrome](https://user-images.githubusercontent.com/30520443/45032326-89496a00-b08c-11e8-84d8-e85f363094b9.png)
Here the issue is that the texts of the select boxes on the right are more "down" that they should be.

Finally, on **Mac + Firefox** (haven't tested others), a font issue, that I found more problematic, somehow echoing [this post on the Wall ](https://tatoeba.org/fra/wall/show_message/29594#message_29594).
![mac_firefox_font](https://user-images.githubusercontent.com/30520443/45032624-3cb25e80-b08d-11e8-9d00-656dc2183f46.png)
By default, the vertical space of a sentence is not big enough to display the accent on the capital letters. Here, the É is hardly readable (I actually missed it when I found the sentence). 
![mac_firefox_font2](https://user-images.githubusercontent.com/30520443/45032665-53f14c00-b08d-11e8-9df9-a6085364d6df.png)
Changing the font shows what it should look like. The problem is the same on the first line of the text area for comments (although I don't have screenshots).  
I think I was connected when I saw this page so it is not related to the "new" design. 

This font issue does not appear with my usual Win+Firefox.

trang:
[...]

PR #2149, opened 2020-02-17
Correct position of warning icon

### LINKED ISSUE #1598: Copy button overlaps with red sentences warning sign
When a sentence is in red and you activate the copy button (and do not use the new design), the [warning sign overlaps with the copy button](https://tatoeba.org/sentences/search?query=&from=und&to=und&orphans=no&unapproved=yes&user=&tags=&list=&has_audio=&trans_filter=limit&trans_to=und&trans_link=&trans_user=&trans_orphan=&trans_unapproved=yes&trans_has_audio=&sort=words).
screenshot-2018-7-4 recherche avancee tatoeba

agrodet:
This small PR is supposed to fix #1598. The result is the one given in my comment https://github.com/Tatoeba/tatoeba2/issues/1598#issuecomment-586194945 

It also modifies the height of `.copy.column` to its container height so the icon entirely fits inside the column, like the audio icon and the link icon. It shouldn't change anything visually. It's just to have correct values.

### DIFF
--- a/webroot/css/layouts/elements.css
+++ b/webroot/css/layouts/elements.css
@@ -841,6 +841,7 @@ html[dir="rtl"] .comment-sentence .text {
 
 .sentence .copy.column {
     float: right;
+    height: 21px;
 }
 .sentence .copy-btn {
     width: 16px;
--- a/webroot/css/layouts/elements.css
+++ b/webroot/css/layouts/elements.css
@@ -1238,11 +1238,11 @@ input.sentenceId {
 }
 
 .mainSentence .correctnessNegative1 {
-    background-position: right top 5px;
+    background-position: right 5px;
 }
 
 .translations .correctnessNegative1 {
-    background-position: right top;
+    background-position: right 2px;
 }
 
 .correctness-info .ok,
--- a/webroot/css/layouts/elements.css
+++ b/webroot/css/layouts/elements.css
@@ -629,12 +629,12 @@ html[dir="rtl"] .comment-sentence .text {
     background: none;
 }
 
-.paging li.prev md-icon, 
+.paging li.prev md-icon,
 .paging li.next md-icon {
     color: #4caf50;
 }
 
-.paging li.prev a:hover md-icon, 
+.paging li.prev a:hover md-icon,
 .paging li.next a:hover md-icon {
     color: #DD8800;
 }
@@ -652,7 +652,7 @@ html[dir="rtl"] .comment-sentence .text {
  * --------------------------------------------------------------------
  */
 
- #private-message-form { 
+ #private-message-form {
     background: #fafafa;
 }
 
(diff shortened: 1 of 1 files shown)


jiru:
Thanks for this fix, @agrodet! It looks like it solves the problem.

> It also modifies the height of .copy.column to its container height so the icon entirely fits inside the column, like the audio icon and the link icon. It shouldn't change anything visually. It's just to have correct values.

It does change something visually. The warning icon is not properly aligned any more:

css_diff

jiru:
Note that you may have to run `bin/cake asset_compress build` to see the changes.

agrodet:
Hmm... There's some browser display messing with us here. I wonder what browser you used to take your screenshot.

I don't think it's due to `.copy.column` height. On my Firefox, the bottom of the warning icon is actually cropped because of these 2 pixels. Even on the prod site, where the icons overlap, I can see that the red icon is cropped. That's the reason why I removed the 2px shift. See my https://github.com/Tatoeba/tatoeba2/issues/1598#issuecomment-586194945 

> We can see that the red icon is slightly higher than the other icons. This can be arranged on the main sentence, but not on the translations, because the height of one div is too small, nearly over 16px. So if we align the red buttons, they will actually be cropped (like they are now on the main site, I think?). One solution would be to slightly increase the div height to be able to align the red icons with the two other.


In more detail, the heights of the different div don't add up correctly. Following the hierarchy: 
```
content column => 20px (it would make sense for it to be ≥ 21px)
    sentenceContent => 16.5px
```
and the warning icon that is set as a background-image, not an svg in a div, has height 16px. That's why it can't be pushed on my browser. There's no space :) So I wonder what those heights are on your browser. 

trang said 
> We don't have that many red sentences and the old sentence design is meant to disappear fairly soon (max two months, I hope). I won't bargain for 2 pixels of misalignment :P
 
That's why I opened the PR. However, the discussion might give more insight than the problem (or the solution). It might also be related to #1709

jiru:
I use Firefox 69.0.1. About the list icon, I remember it used to be cropped before, but it doesn’t look cropped any more on my browser.

jiru:
I confirm that your changes move up the warning icon using Firefox safe mode too.

agrodet:
Alright, I figured it out. The problem was not Firefox but the odd combination of Mac + Firefox. The problem is the same as the last one I explained in #1665. Positioning is not as expected so things get cropped. Hopefully the new design will arrange all that (I will test that later). See what I currently get on Tatoeba 
image

The middle one is placed correctly because there are two lines of text, and the last one is cropped (because the div is too small). Anyway, that's an odd issue, and it should disappear with the new design. I put back the correct values so things should be okay now.

jiru:
Thank you for trying so hard to reproduce the problem. I think we can go on and put the copy button out of beta now. :tada:

PR #2150, opened 2020-02-17
Move Chinese translations to src/Locale/cmn/

Yorwba:
I noticed that even after the translations have been updated, there's zero Chinese text at https://tatoeba.org/cmn/ Turns out that this is because the translations are at `src/Locale/zh` and the locale parser doesn't automatically turn `cmn` into `zh`.

The easiest fix is to move `src/Locale/zh` to `src/Locale/cmn`.


Yorwba:
This is not a good fix because it only makes the user-contributed translations work, other localized strings (e.g. date formatting) remain broken. Closing this.

Issue #2139, opened 2020-02-11
It's possible to add a duplicate in a different but canonical equivalent form


AndiPersti:
Examples on dev:
https://dev.tatoeba.org/deu/sentences/show/4367051
https://dev.tatoeba.org/deu/sentences/show/7841131

![duplicate](https://user-images.githubusercontent.com/6566321/74216240-09bdc280-4ca4-11ea-9136-d1cca80bb91f.png)

I guess we should normalize the text of a sentence, shouldn't we?

I've found [2961 sentences](https://gist.github.com/AndiPersti/b000faf011ddc1611a9eca8b4a7b18e3) which are not in NFC form. (Format of data: id [tab] text as stored in database [tab] text in NFC form)

trang:
For people like me who don't (didn't) know what is this "NFC form":
https://en.wikipedia.org/wiki/Unicode_equivalence

agrodet:
I have faced this several times in the past with Japanese but that's the first time I see it with a European language, that's interesting! :) 
PHP, as many other languages, has its normalizer: https://www.php.net/manual/en/class.normalizer.php 

Also, you can get rid of half-width katakanas by using KC instead of C. See this fiddle https://dotnetfiddle.net/SjIMnk It's in C# but the normalization is (should be) the same. I don't know if you want to get rid of half-width katakanas but I don't see any case where we would want to keep them since we're not cashiers :D (just joking...)

AndiPersti:
> PHP, as many other languages, has its normalizer: https://www.php.net/manual/en/class.normalizer.php

Yes, that's what I want to use.

> Also, you can get rid of half-width katakanas by using KC instead of C. See this fiddle https://dotnetfiddle.net/SjIMnk It's in C# but the normalization is (should be) the same. I don't know if you want to get rid of half-width katakanas but I don't see any case where we would want to keep them since we're not cashiers :D (just joking...)

I don't know anything about half-width katakanas so I'm no help here. But looking at some [normalization charts](http://www.unicode.org/charts/normalization/) this would also influence other languages/scripts so maybe use NFKC only for Japanese?

jiru:
Also note that NFKC [normalizes all kind of spaces into a regular space](https://www.unicode.org/charts/normalization/chart_Separator-Space.html).

AndiPersti:
[...]

PR #2151, opened 2020-02-18
Avoid storing duplicates not in Unicode NFC

AndiPersti:
This PR addresses #2139.

I've added a normalization step to the text cleaning process and refactored the duplicate tests.


jiru:
Looks good. Thank you for dealing with this issue, Andreas! :+1:

PR #2152, opened 2020-02-18
Base i18n locale on the language tag, not the 3-letter ISO code

Yorwba:
Second attempt after #2150. The `languageTag` function is currently mainly used to mark the language of sentences in the HTML to select appropriate fonts (well, I think that's what it's for). So it should be suitable for selecting the i18n locale as well. At the very least making this change displays dates in Chinese as well as selecting the `zh` translations when the UI language is set to `cmn`. The only other language where this changes the strings on the homepage is Malaysian (`zsm`), which is also currently broken, because its translations are stored as `ms`.

We'd run into trouble if we ever wanted to have `zh_TW` or `zh_HK` locales, but for those we'd have to rethink the current system of indicating the language by its ISO 639 three-letter code in the URL anyway.

### DIFF
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -186,7 +186,7 @@ class AppController extends Controller
             $this->Cookie->write('CakeCookie.interfaceLanguage', $lang, false, "+1 month");
         }
         Configure::write('Config.language', $lang);
-        $locale = Locale::parseLocale($lang);
+        $locale = Locale::parseLocale(LanguagesLib::languageTag($lang));
         I18N::setLocale($locale['language']);
 
         // If the Router did not parse the URL, we don't know if the URL


jiru:
Thank you for your PR, Yorwba. :smile:

> The languageTag function is currently mainly used to mark the language of sentences in the HTML to select appropriate fonts (well, I think that's what it's for).

Yes, that’s right. :-)

> So it should be suitable for selecting the i18n locale as well.

While your code works, I disagree with this. If you look closely at the code of `languageTag()`, you’ll see that it handles the script suffix part too. That function really is supposed to give a language tag as defined in [BCP47](https://www.ietf.org/rfc/bcp/bcp47.txt). I suggest you use `iso639_3_To_Iso639_1()` instead because as you said, since we do not handle per-country locales like `zh_TW` and `zh_HK`, that’s really the only thing we need to do now.

> We'd run into trouble if we ever wanted to have zh_TW or zh_HK locales, but for those we'd have to rethink the current system of indicating the language by its ISO 639 three-letter code in the URL anyway.

On a side note, I wish we could provide per-country locales. We could switch from the URL 3-letters language code into a locale name like pt-br or zh-hk. I am willing to help if you’d like to make a PR for that. There is already a system in app_local.php that redirects old URLs that we had in the past like https://tatoeba.org/pt_BR → https://tatoeba.org/por.

Yorwba:
> If you look closely at the code of languageTag(), you’ll see that it handles the script suffix part too

My thought process here was that if `languageTag` explicitly defines a script for a language, that means the language should *always* be displayed with the given script, including when it's the UI locale.

The only language which currently has a script defined in this way is `lzh`. So I tried adding a Literary Chinese locale:

LIterary Chinese

Obviously the strings aren't actually Literary Chinese; they're the same as for Mandarin:

Mandarin

However, the script selection does make a tiny difference for the character 随:
随随

So I think that using `languageTag` is slightly more correct.

Your proposal to switch to per-country locales for the URL would of course eliminate the problem of translating between the way the locale is represented in the URL and the representation that the i18n libraries expect. However, there probably was a good reason for the switch from per-country locales to three-letter codes. If we want to switch back now, it would be good to know what the reason was.

jiru:
Sorry I think I wasn’t very clear. I understand very well that we need to pass a more complete definition of the locale to `Locale::parseLocale()`, not just the language code, but in some cases also the country or the script.

My point was that `languageTag()` is meant to follow BCP47, while `Locale::parseLocale()` does not. But my bad, I was wrong, they both use the same standard, so it’s okay. I thought that the `tag` of `languageTag()` referred to HTML tags, so it could only used in that context, but actually language tags have nothing to do with HTML tags.

> However, there probably was a good reason for the switch from per-country locales to three-letter codes.

Actually I was the one who pushed that change in #629. Back in the days, the code base was much more messy and we had several lists of language codes duplicated all around. That change occured in the process of centralizing the language list into a single place. I replaced the two-letters code with the three-letters code in the URL so that the whole code can rely on a unique list, and we only need to change that one list when adding a new language. I realize now it wasn’t a good idea to drop the country part from the URL.

jiru:
Thanks again for taking care of this! :+1:

Issue #1394, opened 2017-01-15
It takes too long to add several lists 


RyckRichards:
I work so much with lists as far I'm both a language teacher and a student and them are **really** useful for me - I have about 70 lists (due my absence. I'd have about 200+ lists) - since I can ask people to translate sentences, I pick up some I'm going to study/teach and add in a list - español: nationalities, Français: musique, etc, and also I ask to **few** students of mine to translate some sentences into their native languages to check if they got what we had learned

The problem is that whenever I have to add several lists, I have to click over "All public  lists" from the sentence page I'm on - https://tatoeba.org/eng/sentences_lists/show/6842 - whenever I want to add a list or rename it. In order words, if I want to add 30 lists, I have to 

1- create a sentence from [Browse > Browse by list](https://tatoeba.org/eng/sentences_lists/index)
2-from there, type the name of the new list and press enter
![screenshot_2](https://cloud.githubusercontent.com/assets/7658430/21959356/31038440-daa3-11e6-911e-ab7849a7ce59.png)
3-choose who will be able to see/edit the list 
4-click over "All public lists"
![screenshot_1](https://cloud.githubusercontent.com/assets/7658430/21959346/e181664e-daa2-11e6-9cfc-131265219b23.png)
5 - repeat the steps 1,2,3 and 4 29 times more.

Wouldn't that be easy - and/or possible - if we have something like

![screenshot_3](https://cloud.githubusercontent.com/assets/7658430/21959398/3779943a-daa4-11e6-99e0-84d6ae7f84e6.png) 

?

Thanks so really much!

ckjpn:
@RyckRichards 

All you have to do is hold down the correct key to force the resulting list to open in a new tab.

On a Macintosh, it's the "command" key.

Once you edit the list, just close the new tab, and you'll be back on the "create list" page.

Note that if you just copy the URL, and give that to your students, you don't need to set the list to "public", I think.  The list doesn't have to be publicly listed for your students to see the list if they have the URL.

RyckRichards:
@ckjpn 

Thanks for your comment.

The point is that on this way, I'd have to that even 29 times ( 29 holding down the correct key to force the resulting list to open in a new tab) and worse: I could create a duplicate list.

![screenshot_1](https://cloud.githubusercontent.com/assets/7658430/21967777/05b95872-db6c-11e6-8c32-cb5e4ffb1f34.png)

[...]

Issue #161, opened 2014-04-01
Allow the creation of a list when "Add to list" is chosen for a sentence


alanfgh:
Often, I find that I want to add a sentence to a list that doesn't exist yet. However, I can't create a list from the place where the sentence is open. Instead, I need to go to "Browse by list" to create it. It would be nice if one of the options in the drop-down list (preferably near the top) was "[New List]". Selecting this item would give a user the option to create a new list (and specify whether it should be personal or collaborative).

PR #2154, opened 2020-02-19
Add sentences to lists in new sentence design

trang:
This PR implements the list feature in the new sentence design. This is probably the toughest feature to migrate from all the sentence features and it won't be 100% perfect in this first iteration but I hope it brings enough improvement to make people want to switch to the new design.

Related Wall thread: https://tatoeba.org/eng/wall/show_message/34213#message_34213

## Current implementation

**List form**

* The list form contains a text input that fulfills two functionalities:
	* Searching a list. After entering at least one character, all the selectable lists that have this character in their name will be displayed.
    * Creating a new list. The sentence from which the new list is created is automatically added to that list. This addresses #161 and perhaps #1394 as well.
* The list of lists is no longer displayed in a dropdown but in a what looks like a multi-selection box.
	* Each row has a checkbox, the name of the list and a `group` icon if it's a collaborative list.
* The initial list of lists contains the current user's 10 most recently updated lists (only lists to which the sentence can be added).
* On top of that, the first list will be the last selected list (regardless if it belongs to the current user or not). It will have an indicator `(last selected)`.
	* The last selected list is saved in a cookie so it will not appear if the user clears their cookies.
* If the user has no list to select from, the message `You have no lists which this sentence can be added to.` is displayed.
* The list form can be closed by:
	* Clicking again on the list button.
	* Clicking on the "Close" button at the bottom.
	* Pressing escape while the search input is on focus.
* Closing the list form will reset the search.

**Adding/removing a sentence to a new list**

* The sentence is added to the list upon checking the checkbox.
	* The progress bar is displayed while the sentence is being added to the list, even though the response time is very short. Users otherwise don't intuitively understand that checking the checkbox triggers the sentence to be saved in the list right away.
	* The progress bar is also displayed while a new list is being created and the sentence being added.
* The sentence can be removed from a list by unchecking the checkbox.
	* This only applies to lists that have been added without reloading the page. This allows the user to easily revert their action if they added a sentence to a list by mistake.
	* Refreshing the page will lose this possibility because the multi-selection box is initialized with only the lists in which the sentence can be added.

## Improvements for another iteration

1) Instead of displaying only the lists to which the sentence can be added, it will be more intuitive to display the all the available lists.
	* If the sentence already belongs to the list, then the checkbox is checked.
	* Otherwise it is left unchecked.
2) The initial list of lists could display the 10 last selected lists from the user instead of the user's last updated lists.
3) There can be duplicate list names so we would need to find a way to distinguish these lists from one another. We could show the username of the owner, plus the list ID in case the user has two lists with the same name.
4) The user might want to browse the list to see what kind of sentences are already in the list. There should be a button or something to open the list in a new tab.

### DIFF
--- a/src/Controller/SentencesListsController.php
+++ b/src/Controller/SentencesListsController.php
@@ -488,4 +488,15 @@ class SentencesListsController extends AppController
         $this->set("translationsLang", $translationsLang);
         $this->set("sentencesWithTranslation", $results);
     }
+
+    public function choices() {
+        $lists = $this->SentencesLists->getUserChoices(
+            CurrentUser::get('id'), 1, true
+        );
+
+        $this->set('lists', $lists);
+        $this->loadComponent('RequestHandler');
+        $this->set('_serialize', ['lists']);
+        $this->RequestHandler->renderAs($this, 'json');
+    }
 }
--- a/src/Controller/SentencesListsController.php
+++ b/src/Controller/SentencesListsController.php
@@ -68,7 +68,8 @@ class SentencesListsController extends AppController
         $noCsrfActions = [
             'set_option',
             'save_name',
-            'add_new_sentence_to_list'
+            'add_new_sentence_to_list',
+            'add_sentence_to_new_list'
         ];
         if (in_array($params['action'], $noCsrfActions)) {
             $this->components()->unload('Csrf');
@@ -86,7 +87,8 @@ class SentencesListsController extends AppController
         $this->Security->config('unlockedActions', [
             'set_option',
             'save_name',
-            'add_new_sentence_to_list'
+            'add_new_sentence_to_list',
+            'add_sentence_to_new_list'
         ]);
 
         return parent::beforeFilter($event);
@@ -363,6 +365,28 @@ class SentencesListsController extends AppController
         }
 
         $this->set('sentence', $result);
+    }
+
+    public function add_sentence_to_new_list() {
+        $userId = $this->Auth->user('id');
+        $sentenceId = $this->request->getData('sentenceId');
+        $list = $this->SentencesLists->createList(
+            $this->request->getData('name'), 
+            $this->Auth->user('id')
+        );
+
+        if ($this->SentencesLists->addSentenceToList($sentenceId, $list->id, $userId)) {
+            $list->hasSentence = true;
+            $this->set('result', $list);
+            $this->Cookie->write('most_recent_list', $list->id, false, '+1 month');
+        } else {
+            $this->set('result', 'error');
+        }
+        
+        $this->loadComponent('RequestHandler');
+        $this->set('result', $list);
+        $this->set('_serialize', ['result']);
+        $this->RequestHandler->renderAs($this, 'json');
     }
 
     /**
--- a/src/Controller/SentencesListsController.php
+++ b/src/Controller/SentencesListsController.php
@@ -384,7 +384,6 @@ class SentencesListsController extends AppController
         }
         
         $this->loadComponent('RequestHandler');
-        $this->set('result', $list);
         $this->set('_serialize', ['result']);
         $this->RequestHandler->renderAs($this, 'json');
     }
--- a/src/Controller/SentencesListsController.php
+++ b/src/Controller/SentencesListsController.php
@@ -369,20 +369,20 @@ class SentencesListsController extends AppController
 
     public function add_sentence_to_new_list() {
         $userId = $this->Auth->user('id');
+        $listName = $this->request->getData('name');
         $sentenceId = $this->request->getData('sentenceId');
-        $list = $this->SentencesLists->createList(
-            $this->request->getData('name'), 
-            $this->Auth->user('id')
-        );
+        $list = $this->SentencesLists->createList($listName, $userId);
 
-        if ($this->SentencesLists->addSentenceToList($sentenceId, $list->id, $userId)) {
-            $list->hasSentence = true;
-            $this->set('result', $list);
-            $this->Cookie->write('most_recent_list', $list->id, false, '+1 month');
-        } else {
-            $this->set('result', 'error');
+        $result = 'error';
+        if ($list) {
+            if ($this->SentencesLists->addSentenceToList($sentenceId, $list->id, $userId)) {
+                $list->hasSentence = true;
+                $result = $list;
+                $this->Cookie->write('most_recent_list', $list->id, false, '+1 month');
+            }
         }
-        
+
+        $this->set('result', $result);
         $this->loadComponent('RequestHandler');
         $this->set('_serialize', ['result']);
         $this->RequestHandler->renderAs($this, 'json');
--- a/src/Model/Table/SentencesListsTable.php
+++ b/src/Model/Table/SentencesListsTable.php
@@ -152,9 +152,6 @@ class SentencesListsTable extends Table
                     'editable_by' => 'no_one'
                 ]
             ])
-            ->notMatching('SentencesSentencesLists', function ($q) use ($sentenceId) {
-                return $q->where(['SentencesSentencesLists.sentence_id' => $sentenceId]);
-            })
             ->select([
                 'id',
                 'name',
@@ -168,6 +165,9 @@ class SentencesListsTable extends Table
             return $results->toList();
         } else {
             $results->order(['name']);
+            $results->notMatching('SentencesSentencesLists', function ($q) use ($sentenceId) {
+                return $q->where(['SentencesSentencesLists.sentence_id' => $sentenceId]);
+            });
             $listsOfUser = array();
             $collaborativeLists = array();
     
--- a/src/Model/Table/SentencesListsTable.php
+++ b/src/Model/Table/SentencesListsTable.php
@@ -142,8 +142,8 @@ class SentencesListsTable extends Table
      */
     public function getUserChoices($userId, $sentenceId, $forNewDesign = false)
     {
-        $results = $this->find()
-            ->where([
+        $query = $this->find();
+        $query->where([
                 'OR' => [
                     'user_id' => $userId,
                     'editable_by' => 'anyone'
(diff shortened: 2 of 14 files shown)


ckjpn:
I tested it on the dev site for a while with and it seems to work as explained.



## Comments on "Improvements for another iteration"
```
Instead of displaying only the lists to which the sentence can be added, it will be more intuitive to display the all the available lists.
If the sentence already belongs to the list, then the checkbox is checked.
Otherwise it is left unchecked.
```
This would welcome a good addition.

`There can be duplicate list names so we would need to find a way to distinguish these lists from one another. We could show the username of the owner, plus the list ID in case the user has two lists with the same name.`

I like the idea of showing the List ID in this case.  If there are 2 with the same name, then the higher-numbered list would be the last-created one.  For collaborative lists, though, perhaps the list owner's username might be better, or you could include both.

Example:

Sentences about food (2920)
Sentences about food (6340)
Sentences about food (1001) (Username)

ckjpn:
It's not obvious how to removed a sentence from a list, once it's been added

How would I delete a sentence from this list?

https://dev.tatoeba.org/eng/sentences_lists/show/8886

src/Controller/SentencesListsController.php
```diff
@@ -365,6 +367,28 @@ public function add_new_sentence_to_list()
...
+        if ($this->SentencesLists->addSentenceToList($sentenceId, $list->id, $userId)) {
+            $list->hasSentence = true;
+            $this->set('result', $list);
+            $this->Cookie->write('most_recent_list', $list->id, false, '+1 month');
+        } else {
+            $this->set('result', 'error');
+        }
+        
+        $this->loadComponent('RequestHandler');
+        $this->set('result', $list);
```
AndiPersti: I think this line should be removed.

You've already set `result` above (on line 380 and 383) and in case of an error, this line would overwrite the error response and would show the sentence as added to the list although it wasn't.

AndiPersti:
>     3\. There can be duplicate list names so we would need to find a way to distinguish these lists from one another. We could show the username of the owner, plus the list ID in case the user has two lists with the same name

I can understand that list names may be duplicate if they belong to different users.
But what's the use case for allowing a user to have multiple lists with the same name?

trang:
> But what's the use case for allowing a user to have multiple lists with the same name?

There's no actually use case that I can think of. At least no cases where someone would intentionally name two of their lists with the same name. But it could happen unintentionally, since we don't check for duplicate names.

src/Controller/SentencesListsController.php
trang: Good catch. Thanks!

trang:
@ckjpn 

> It's not obvious how to removed a sentence from a list, once it's been added

You made me realize this case is not yet covered with the new design. I will remove the new sentence design from the list page until I get to rework it.

src/Controller/SentencesListsController.php
```diff
@@ -365,6 +367,27 @@ public function add_new_sentence_to_list()
         $this->set('sentence', $result);
     }
 
+    public function add_sentence_to_new_list() {
+        $userId = $this->Auth->user('id');
+        $sentenceId = $this->request->getData('sentenceId');
+        $list = $this->SentencesLists->createList(
+            $this->request->getData('name'), 
+            $this->Auth->user('id')
+        );
```
jiru: `createList()` will return false if the list name is invalid (empty or empty after trimming). However, you are not checking for this case.

src/Model/Table/SentencesListsTable.php
```diff
@@ -155,33 +155,44 @@ public function getUserChoices($userId, $sentenceId)
...
-        foreach ($results as $result) {
-            $listId = $result['id'];
-            $listName = $result['name'];
-            $userId = $result['user_id'];
+            ->select([
+                'id',
+                'name',
+                'user_id',
+                'is_mine' => 'IF(SentencesLists.user_id = '.$userId.', TRUE, FALSE)',
+                'is_collaborative' => 'IF(SentencesLists.editable_by = "anyone", TRUE, FALSE)',
```
jiru: `if(condition, true, false)` can be shortened to `(condition)`.

You can also use the query builder like this, to be sure we are not letting any chance for SQL injections:
```php
$query = $this->find();
$query->select([
                'id',
                'name',
                'user_id',
                'is_mine' => $query->newExpr()->eq('SentencesLists.user_id', $userId),
                'is_collaborative' => $query->newExpr()->eq('SentencesLists.editable_by', 'anyone'),
            ]);
```

src/Model/Table/SentencesListsTable.php
```diff
@@ -155,33 +155,44 @@ public function getUserChoices($userId, $sentenceId)
...
-                $listName = __('unnamed list');
+        if ($forNewDesign) {
+            $results->order(['is_mine DESC', 'modified DESC']);
+            return $results->toList();
+        } else {
+            $results->order(['name']);
+            $listsOfUser = array();
+            $collaborativeLists = array();
+    
+            $currentUserId = CurrentUser::get('id');
```
jiru: I know that this is not an addition from you (existing code just got indented), but I think we should use `$userId` provided as parameter instead of calling CurrentUser here.

src/Template/Element/sentences/list_form.ctp
```diff
@@ -0,0 +1,43 @@
+<?php
+use App\Model\CurrentUser;
+use Cake\ORM\TableRegistry;
+
+$SentencesLists = TableRegistry::getTableLocator()->get('SentencesLists');
+$lists = $SentencesLists->getUserChoices(
+    CurrentUser::get('id'), $sentenceId, true
+);
+$mostRecentList = $this->request->getSession()->read('most_recent_list');
+$listsJSON = htmlspecialchars(json_encode($lists), ENT_QUOTES, 'UTF-8');
```
jiru: I believe this is temporary? It’s not a good idea, for two reasons:
1. The variable should be passed to the view instead.
2. The page is going to be cluttered with that same json data duplicated over and over in every sentence block.

We already have this problem for the autocomplete language list that is duplicated once, I think it’s a good time to find an alternative way to do this. Maybe [something like this](https://stackoverflow.com/questions/28402199/how-to-initiate-global-data-in-angularjs)?

src/Template/Element/sentences/list_form.ctp
```diff
@@ -0,0 +1,43 @@
...
+?>
+
+<form layout="column" style="border-top: 1px solid #f1f1f1; padding-top: 10px;"
```
jiru: Any reason for not putting this style (and the other ones later in that file) into `elements.css`?

webroot/js/directives/sentence-and-translations.dir.js
```diff
@@ -76,28 +76,38 @@
         };
     }
 
-    function SentenceAndTranslationsController($http, $cookies) {
+    function SentenceAndTranslationsController($rootScope, $scope, $http, $cookies, $timeout) {
```
jiru: I thought that declaring the injections like that, as opposed to e.g.
```js
.controller('SentencesListsDownloadCtrl', ['$scope', '$http', '$timeout', function ($scope, $http, $timeout)`
```
wasn’t working because the JS minifier is renaming the parameters.

webroot/js/directives/sentence-and-translations.dir.js
```diff
@@ -110,6 +120,26 @@
...
+        /////////////////////////////////////////////////////////////////////////
+
+        $scope.$on('newListCreated', function(event, data, sentenceId) {
+            var list = angular.copy(data);
+            $cookies.put('most_recent_list', list.id);
+            list.hasSentence = vm.sentence.id === parseInt(sentenceId);
+            list.is_mine = '1';
+            allLists.unshift(list);
+            resetLists();
+            moveRecentListToTop();
```
jiru: I think there is no need to call moveRecentListToTop() here because resetLists() does that already.

jiru:
Sorry it took me so much time to review your PR, Trang. Please have a look at my comments.

src/Template/Element/sentences/list_form.ctp
trang: It was temporary, yes.

I made an attempt at a solution using a `.factory()` to store the data and initializing it from `app.run()`. At the same time it deals with the 1st point I mentioned in the "Improvements for another iteration" :)

src/Template/Element/sentences/list_form.ctp
trang: No reason other than laziness. I find it easier to add the styles directly inline rather than in a CSS file when I'm in the process of designing something new. I usually don't have the final picture in mind and I go back and forth a lot between different ideas. It's more comfortable to go through this design phase when the code is in one place and doesn't need to be interrupted by a build.

With https://github.com/Tatoeba/tatoeba2/commit/645b58ec2e29311b17da1c930f29855119049347 it is a bit less of an issue.

webroot/js/directives/sentence-and-translations.dir.js
trang: Encapsulating the code inside an "Immediately Invoked Function Expression" prevents issues with minification, if I'm not mistaken.

I mean this thing:

```
(function() {

// code

})();
```

src/Template/Element/sentences/list_form.ctp
jiru: That's great!

webroot/js/directives/sentence-and-translations.dir.js
jiru: No, this is not what I’m talking about. Your code currently works because that file isn’t minified yet. However if we minify it, it *will* break. Have a look at `config/asset_compress.ini` `src/Template/Layout/default.ctp` to see how minification is handled.

trang: You're right.

I pushed a commit that minifies the `sentence-and-translations.dir.js` while I was at it.

jiru: Great! :+1:

webroot/js/directives/sentence-and-translations.dir.js
```diff
@@ -283,6 +336,102 @@
...
+            
+            vm.inProgress = true;
+            var data = { 
+                name: vm.listSearch,
+                sentenceId: vm.sentence.id
+            };
+            $http.post(rootUrl + '/sentences_lists/add_sentence_to_new_list', data).then(function(result) {
+                vm.inProgress = false;
+                vm.listSearch = '';
+                $rootScope.$broadcast('newListCreated', result.data.result, vm.sentence.id);
```
jiru: If it fails, the action `add_sentence_to_new_list` will return the value `{result: 'error'}` in `result.data` (as opposed to the list data). It looks like you are not checking for that case.

src/Template/Element/sentences/list_form.ctp
```diff
@@ -0,0 +1,35 @@
+<?php
+use App\Model\CurrentUser;
+use Cake\ORM\TableRegistry;
+?>
```
jiru: It looks like this is not needed.

PR #2155, opened 2020-02-19
 Add pinyin tone marks to uppercase characters as well (fixes #2153) 

### LINKED ISSUE #2153: Capital letters in Pinyin transcriptions don't get tone marks
I'm not sure whether capital letters and spaces in Pinyin even make sense, since there's not really a standardized orthography. However, given that we do have them, "A1" should be displayed as "Ā", just like "a1" is displayed as "ā".
This is what it looks like right now:
Screenshot

Yorwba:
I think the title is self-explanatory.

Aside: Since reporting #2153, I have learned that there *is* [a standard for pinyin orthography  that requires capitalization](http://www.moe.gov.cn/ewebeditor/uploadfile/2012/08/21/20120821100233165.pdf).

### DIFF
--- a/src/View/Helper/PinyinHelper.php
+++ b/src/View/Helper/PinyinHelper.php
@@ -192,6 +192,30 @@ class PinyinHelper extends AppHelper
     );
 
     /**
+     * Uppercase the first character in a UTF-8 string.
+     */
+    private function utf8_ucfirst($string) {
+        // Based on https://stackoverflow.com/a/2518021
+        // Licensed under CC BY-SA 4.0
+        $strlen = mb_strlen($string, 'UTF-8');
+        $firstChar = mb_substr($string, 0, 1, 'UTF-8');
+        $then = mb_substr($string, 1, $strlen - 1, 'UTF-8');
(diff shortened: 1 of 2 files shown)


jiru:
Thank you for taking care of the Pinyin, Yorwba! And yay for the tests. :v: 

Since the list of possible toned vowels is short, finite, and will never change, it feels a bit overkill to call a general-purpose case conversion function. How about just adding all the uppercase chars in the `_vowelTone2Unicode_search` and `_vowelTone2Unicode_replace` arrays?

Yorwba:
I mostly wrote it this way because manually copy-pasting the correct uppercase version of each accented character didn't seem like a fun thing to do, so I preferred doing it programmatically (although it probably took me longer to write this way...).

I agree that running the conversion function every time is overkill, so I took the intermediate arrays and hardcoded them into the script.

jiru:
Thank you! I understand why you wanted to do it programmatically. But speed does matter in this function because after your code gets on production, I will have to regenerate the Pinyin for all of the Mandarin sentences.

PR #2162, opened 2020-02-24
Replace embedded newlines with spaces in exports

AndiPersti:
Currently the text of sentence comments and wall posts is exported as it is
stored, i.e. with embedded newlines. This makes it harder to process the data
with other tools (e.g. importing it into a local database).

Replacing the embedded newlines with spaces ensures that every row in
the table is represented by a single line in the downloadable file.

### DIFF
--- a/docs/database/scripts/weekly_exports.sql
+++ b/docs/database/scripts/weekly_exports.sql
@@ -74,13 +74,13 @@ ORDER BY c.datetime ASC
 INTO OUTFILE '/var/tmp/contributions.csv';
 
 -- Wall posts
-SELECT w.id, u.username, w.parent_id, w.date, w.content
+SELECT w.id, u.username, w.parent_id, w.date, REPLACE(w.content, "\r\n", " ")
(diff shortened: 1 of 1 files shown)


AndiPersti:
I don't know how many people use the exports for sentence comments and wall posts but I personally use them e.g. for offline searching. I've written already a script for preprocessing the data which merges all split up lines so I have no problem if you reject this PR. But I still think this change will make these files easier to use for other people.

jiru:
Well, the space character has a different meaning than the newline, so you are altering the meaning of the original content.

I never did it myself, but I thought that it was easy to import a CSV into a database precisely because we export it with MySQL, which uses its own CSV format with escaped newlines and stuff. We even have an old import script in `docs/database/import/restore_dumps.sql`.

It looks like [MySQL is able to export as XML](https://dev.mysql.com/doc/refman/8.0/en/load-xml.html) too, it might be a good idea to explore for multiline content as XML can be easily used by other tools too.

AndiPersti:
I admit I haven't thought to much about this and I'm sorry for wasting your time.

Issue #1729, opened 2018-12-27
Reduce bandwidth required to download sentence exports


Kentoseth:
This is a suggestion, mostly because I can do it.

Instead of having 1 giant sentences.csv file that people need to frequently download to update, you can add git-tracking and split the sentences.csv file by language.

This will conserve bandwidth and you can also use github to handle big file downloads if necessary.

jiru:
@Kentoseth Thanks for you suggestion. Our sentences distribution system clearly needs to be improved, but what exactly is the problem you are trying to solve here? I’d like to know what is your use-case. What are the benefits of git-tracking *for you*? What are the benefits of splitting the sentences.csv file by language *for you*?

jiru:
@Kentoseth Thanks for your suggestion. Our sentences distribution system clearly needs to be improved, but what exactly is the problem you are trying to solve here? I’d like to know what is your use-case. What are the benefits of git-tracking *for you*? What are the benefits of splitting the sentences.csv file by language *for you*?

Kentoseth:
@jiru 

> but what exactly is the problem you are trying to solve here? I’d like to know what is your use-case. What are the benefits of git-tracking for you?

I don't have a specific use-case, but when I saw how big the sentences.csv tarball was, I thought I would suggest this solution so that it can conserve bandwidth. Example:

20 people download the 100MB tarball each week. This is 20GB of bandwidth a month. By using a version-controlled alternative for the large files, you can reduce bandwidth to a fraction of what is being consumed currently.

> What are the benefits of splitting the sentences.csv file by language for you?

This is also another bandwidth-conserving measure. I think most people are generally interested in translations between 2/3 languages only. Being able to download just English-French instead of having to download the entire CSV might save bandwidth too.

jiru:
Thank you for your answer, this is very helpful.

jeanm:
Having worked with Tatoeba raw data before, I also ended up downloading a massive file to only use maybe 5% of it. I guess the admins will have to see whether the bandwidth costs are high enough to warrant doing extra work.

If the costs are a problem, then instead of splitting the files yourselves or you could see if the good folks at [opus.nlpl.eu](http://opus.nlpl.eu/Tatoeba.php) would be willing to do this for you regularly. They already host some data, although there's no indication of how recent it is. You could then put a fairly prominent link to that page in your download section.

alanfgh:
[...]

PR #2163, opened 2020-02-24
Split sentences_detailed.csv into single language files

AndiPersti:
As discussed on #1729 here's one possible way to serve single language exports.

I don't think it's necessary to split both `sentences.csv` and `sentences_detailed.csv` so I've chosen the latter because it is a superset of the former.

Even if we decide to not use this implementation we should still change line 7 and line 10:
- `ROOT` is defined on line 5 so we should use it on line 7
- `var/tmp` is also used by `systemd` so we should just move the `csv` files we've created

### DIFF
--- a/docs/cron/export.sh
+++ b/docs/cron/export.sh
@@ -4,10 +4,11 @@ set -e
 PATH=$PATH:/usr/local/mysql/bin
 ROOT='/var/www-prod'
 
-mysql -u "$DB_USER" -p"$DB_PASS" "$DB" < /var/www-prod/docs/database/scripts/weekly_exports.sql
+mysql -u "$DB_USER" -p"$DB_PASS" "$DB" < $ROOT/docs/database/scripts/weekly_exports.sql
 
 DL_DIR="/var/www-downloads/exports"
-mv /var/tmp/* "$DL_DIR"
+mkdir -p "$DL_DIR/single_language"
+mv /var/tmp/*csv "$DL_DIR"
 
 cd "$DL_DIR"
 tar -cjf sentences_detailed.tar.bz2 sentences_detailed.csv
@@ -27,3 +28,8 @@ tar -cjf sentences_with_audio.tar.bz2 sentences_with_audio.csv
 tar -cjf user_languages.tar.bz2 user_languages.csv
 tar -cjf tags_detailed.tar.bz2 tags_detailed.csv
 tar -cjf sentences_CC0.tar.bz2 sentences_CC0.csv
+
+# Split sentences_detailed.csv into single language files
+cd "$DL_DIR/single_language"
+awk -F"\t" '{print >> ($2 == "\\N" ? "unknown_sentences_detailed.tsv" : $2 "_sentences_detailed.tsv")}' ../sentences_detailed.csv
+bzip2 -f *_sentences_detailed.tsv
--- a/docs/cron/export.sh
+++ b/docs/cron/export.sh
@@ -7,7 +7,6 @@ ROOT='/var/www-prod'
 mysql -u "$DB_USER" -p"$DB_PASS" "$DB" < $ROOT/docs/database/scripts/weekly_exports.sql
 
 DL_DIR="/var/www-downloads/exports"
-mkdir -p "$DL_DIR/single_language"
 mv /var/tmp/*csv "$DL_DIR"
 
 cd "$DL_DIR"
@@ -29,7 +28,26 @@ tar -cjf user_languages.tar.bz2 user_languages.csv
 tar -cjf tags_detailed.tar.bz2 tags_detailed.csv
 tar -cjf sentences_CC0.tar.bz2 sentences_CC0.csv
 
-# Split sentences_detailed.csv into single language files
-cd "$DL_DIR/single_language"
-awk -F"\t" '{print >> ($2 == "\\N" ? "unknown_sentences_detailed.tsv" : $2 "_sentences_detailed.tsv")}' ../sentences_detailed.csv
-bzip2 -f *_sentences_detailed.tsv
+# Create per-language files for the different sentences files
+TEMP_DIR='/var/tmp/per_language'
+trap "rm -rf $TEMP_DIR" EXIT
+
+echo 'select code from languages;' |
+( mysql --skip-column-names tatoeba; echo 'unknown' ) |
+xargs -L 1 -I @ mkdir -p $TEMP_DIR/@
+
+split_file () {
+    local BASENAME=_${1%csv}tsv
+    local DIR=${TEMP_DIR}/
+
+    awk -F"\t" -v basename=$BASENAME -v dir=$DIR \
+    '{print >> ($2 == "\\N" ? dir "unknown/unknown" basename : dir $2 "/" $2 basename)}' \
+    $1
+}
+
+split_file sentences_detailed.csv
+split_file sentences.csv
+split_file sentences_CC0.csv
+find $TEMP_DIR -path '*tsv' -exec bzip2 -qf '{}' +
+rm -rf $DL_DIR/per_language
+mv -f $TEMP_DIR $DL_DIR
--- a/docs/cron/export.sh
+++ b/docs/cron/export.sh
@@ -6,7 +6,6 @@ ROOT='/var/www-prod'
 
 mysql -u "$DB_USER" -p"$DB_PASS" "$DB" < $ROOT/docs/database/scripts/weekly_exports.sql
 
-DL_DIR="/var/www-downloads/exports"
 mv /var/tmp/*csv "$DL_DIR"
 
 cd "$DL_DIR"
@@ -49,5 +48,5 @@ split_file sentences_detailed.csv
 split_file sentences.csv
 split_file sentences_CC0.csv
 find $TEMP_DIR -path '*tsv' -exec bzip2 -qf '{}' +
-rm -rf $DL_DIR/per_language
+rm -rf ${DL_DIR}per_language
 mv -f $TEMP_DIR $DL_DIR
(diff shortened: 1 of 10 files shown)


jiru:
Thank you, Andi. :smile_cat: 

A few comments.

* The wording `single_language` feels not self-explicit enough. How about `per_language`?
* You should update the Downloads page to give some visibility to the new files.
* If we are to provide per-language versions of other existing CSV exports, it may be a good idea to group them into per-language directories, so we’d have something like `exports/*.csv` for the current files, and `exports/per_language/deu/*.csv`, `exports/per_language/fra/*.csv` etc. for the per-language ones. This way, since the per-language files have the same format, we’d only need to put an additional link to the `exports/per_language/` directory on the Downloads page, which will give a show of language codes one can choose from.

AndiPersti:
>The wording `single_language` feels not self-explicit enough. How about `per_language`?

Yes, `single_language` also sounded a little bit strange to me but I couldn't find a better name.

>You should update the Downloads page to give some visibility to the new files.

I've added a note about them under the `Sentences` and `Sentences (CC0)` sections but I'm not sure about the placement and the wording.
(BTW: There's text in the code which doesn't show up on the page because it misses the `echo` call. Is it still useful: https://github.com/Tatoeba/tatoeba2/blob/2f40784c100bdf5cfec850865cb7d6bbec10599b/src/Template/Pages/downloads.ctp#L199-L202 )

>If we are to provide per-language versions of other existing CSV exports, it may be a good idea to group them into per-language directories, so we’d have something like `exports/*.csv` for the current files, and `exports/per_language/deu/*.csv`, `exports/per_language/fra/*.csv` etc. for the per-language ones.

I've changed the script to split all three `sentences` files and put the per-language files into their corresponding subdirectories.

src/Template/Pages/downloads.ctp
```diff
@@ -146,6 +146,9 @@ $this->set('title_for_layout', $this->Pages->formatTitle(__('Download sentences'
         $license_str = __('License');
         $attribution_url_str = __('Attribution URL');
         $tab_str = __('tab');
+        $per_language_str = __('Per-language files are available at <a ' .
+            'href="https://downloads.tatoeba.org/exports/per_language/">' .
+            'https://downloads.tatoeba.org/exports/per_language</a>.');
```
jiru: There are two problems with this translatable string.

1. It includes a URL, however the URL shouldn’t be showed to translators because they may modify it by mistake. Instead, use:
```php
$per_language_str = format(
    __('Here is my <a href="{}">link</a>.'),
    'https://downloads.tatoeba.org/exports/per_language'
);
```

2. The [hyperlink text is not good](https://agsci.psu.edu/communications/how-to/websites/web-best-practices/appropriate-link-text). I know that other existing links on that page are like that too, but feel free to make it right.

AndiPersti: > I know that other existing links on that page are like that too, but feel free to make it right.

How about something like this:
downloads

trang: You could have a dropdown to select the language and have a download button.

Screenshot_2020-03-03 Download sentences - Tatoeba

For the files where there's no possibility to download in a specific language, you just don't have the dropdown.

jiru: The language drop down seems like a better idea. I think it’s important however to let people do "right-click → save as" or "right-click → copy link URL", so the download button should allow to do so. I think you could add a checkbox that makes the language dropdown selectable, like this:

> [✘] Only include sentences in: [Select a language]
> [Download]

This way, it makes it easier to understand that the language dropdown is a filtering option.

@AndiPersti Or, if you don’t feel like doing all that frontend stuff, make it simple:

> <h3>Sentences</h3>
>
> <b>Filename</b>
> sentences.csv
>
> <b>Download</b>
> Download this file <a href="https://downloads.tatoeba.org/exports/sentences.tar.bz2">including all languages</a> or <a href="https://downloads.tatoeba.org/exports/per_language/">only for a single language</a>.

Or

> Download this file <a href="http://example.com/">including all languages</a>.

AndiPersti:
I'm sorry for missing this problem and try to fix the error tomorrow.

AndiPersti:
Some screenshots:
downloads
downloads2

src/View/Helper/DownloadsHelper.php
```diff
@@ -0,0 +1,121 @@
...
+use App\Model\Entity\LanguageNameTrait;
+
+class DownloadsHelper extends AppHelper
+{
+    use LanguageNameTrait;
+
+    public $helpers = ['Html', 'Languages'];
+
+    const PER_LANGUAGE_DIR = '/var/www-downloads/exports/per_language';
+    const DOWNLOAD_URL = 'https://downloads.tatoeba.org/exports';
```
jiru: We have the habit of putting this kind of constants in `app_local.php`, so that it’s easier to integrate with Imouto or any other custom instance. If you go for this, please add new variables in `app_local.php.template` that can be readily used by Imouto. Within your tests, you can then add the appropriate `Configure::write()` calls in the `setUp()` method.

src/View/Helper/DownloadsHelper.php
```diff
@@ -0,0 +1,121 @@
...
+     * @param string $basename Name of the file containing all languages
+     *                         (without extension)
+     *
+     * @return array           A mapping of language code => URL for file
+     **/
+    private function availableFiles($basename) {
+        $command = 'find ' . self::PER_LANGUAGE_DIR . ' -type f ' .
+                   "-name '*$basename.tsv.bz2' -printf '%P\\n' " .
+                   '2> /dev/null';
+        $stdout = trim(shell_exec($command));
```
jiru: Executing shell commands within a PHP script is not a good practice in terms of security. If `$basename` is tainted, it allows all sorts of shell injection. Please consider rewriting this part in [pure PHP](https://book.cakephp.org/3/en/core-libraries/file-folder.html) instead.

src/View/Helper/DownloadsHelper.php
```diff
@@ -0,0 +1,121 @@
...
+        $options[0] = [
+            'language' => __('All languages'),
+            'url' => $urlForAll
+        ];
+
+        $urls = $this->availableFiles($basename);
+        if (!empty($urls)) {
+            $languages = array_intersect_key(
+                $this->Languages->onlyLanguagesArray(false) +
+                ['unknown' => __x('dropdown-list', 'Unknown language')],
```
jiru: Feel free to make `unknownLanguagesArray()` accept a parameter and use `$this->Languages->unknownLanguagesArray(false)`.

docs/cron/export.sh
```diff
@@ -4,10 +4,10 @@ set -e
 PATH=$PATH:/usr/local/mysql/bin
 ROOT='/var/www-prod'
 
-mysql -u "$DB_USER" -p"$DB_PASS" "$DB" < /var/www-prod/docs/database/scripts/weekly_exports.sql
+mysql -u "$DB_USER" -p"$DB_PASS" "$DB" < $ROOT/docs/database/scripts/weekly_exports.sql
 
 DL_DIR="/var/www-downloads/exports"
```
jiru: If you make this a variable in `app_local.php`, you should also remove this line and edit the cron job runner `docs/cron/runner.sh` so that it passes the variable to this script instead.

src/Template/Element/per_language_files.ctp
```diff
@@ -0,0 +1,51 @@
...
+                <md-select class="file-selection" ng-model="<?= $model ?>Select"
+                ng-change="<?= $model ?> = <?= $model ?>Select" flex="initial">
+                    <?php foreach (array_slice($options, 1) as $option): ?>
+                        <md-option value="<?= $option['url'] ?>">
+                            <?= $option['language'] ?>
+                        </md-option>
+                    <?php endforeach; ?>
+                </md-select>
+            </div>
+        <md-radio-group>
```
jiru: This should be a closing tag instead.

jiru:
Wow, it looks good! Thank you for taking the time to reformat the contents of download.ctp, it’s so much cleaner now. :clap: 

If you don’t feel like making `/var/www-downloads/exports/` and `https://downloads.tatoeba.org/exports` configurable, no problem, I can still merge your PR without that, but the other changes I asked are rather important.

It would be great if I could merge your PR by Wednesday 11th because it introduces some translatable string changes. If not, it’s okay because the string changes are minor anyway.

src/View/Helper/DownloadsHelper.php
AndiPersti: For my very first prototype I've constructed the array myself and `unknown` was in between the other languages after localized sorting.

Then I've looked around and found the methods in the `Languages` helper. But in addition to the splitting problem, `unknowLanguagesArray()` also puts `unknown` at the beginning which I personally don't like very much (after loading the page it would show the two options `All languages` and `Only sentences in Unknown language`).

So I've used the current version which puts `unknown` at the end.

Where do you think `unknown` should be put: at the beginning, at the end or in between?

AndiPersti:
> It would be great if I could merge your PR by Wednesday 11th because it introduces some translatable string changes. If not, it’s okay because the string changes are minor anyway.

I'll try my best but I'm not sure if I will have enough time tomorrow.

src/View/Helper/DownloadsHelper.php
jiru: The only place `unknownLanguagesArray()` is used is the language selector on `/sentences/show_all_in`. I think it’s reasonable to put it at the end because it’s less frequently used than other languages.

AndiPersti:
I've implemented all requested changes and rebased the branch onto the current dev branch because there was a merge conflict in `webroot/js/responsive/app.module.js`.

jiru:
Thank you very much, Andreas! :+1:
You’re doing such a good job, it’s awesome.

alanfgh:
Many thanks for this. As a result of this change, and probably some previous ones that got rid of nulls and other bad characters, I can now search through all the English sentences with a text editor without any problem. Very nice!

PR #2164, opened 2020-02-24
Add username option to FixHashesCommand

AndiPersti:
The script now works by default under the username "FixHashesCommand" but I've also added an option which allows changing the username.


jiru:
Looks good! :+1:

PR #2165, opened 2020-02-25
Code cleaning

AndiPersti:
There are still some residues from the old CakePHP 2 era in the codebase which aren't used any more and can be removed.


jiru:
Thank you Andreas. :smile:

PR #2167, opened 2020-02-26
Add the 'Add Vocabulary Request' submenu in the 'Contribute' menu

### LINKED ISSUE #1836: Vocabulary item feature is hard to discover
In order for someone to discover the vocabulary item feature via the user interface, they need to follow this path:

Contribute
Add sentences
Sentences added (the heading for the bottom half of page)
Sentences wanted (clickable link that uses small capitals, unlike most links on Tatoeba)

At this point, they see a page that has the heading "Vocabulary that needs sentences", as well as a listing of items in different languages. They need to choose "Filter by language" in the right-hand sidebar in order to transform it to something that they can work with.

Now they need to choose between the following clickable links under "Vocabulary items" in the right-hand sidebar:
- My vocabulary items
- Add vocabulary items
- Sentences wanted

And whenever they go back to "Sentences wanted", the UI forgets the language that was selected, so the user needs to press "Filter by language" again.

This path is nonintuitive (if I want to request vocabulary, I'm not thinking that I want to add sentences), circuitous (many clicks required to get to the desired place), and inconsistent, especially with regard to terminology ("Sentences" vs. "Vocabulary items").

My suggestions:

(1) Use "vocabulary request" throughout:

"Vocabulary requests" (instead of "Vocabulary items")
"My vocabulary requests" (instead of "My vocabulary items")
"Add vocabulary requests" (instead of "Add vocabulary items")
"Existing vocabulary requests" (instead of "Sentences wanted")

(2) Add an item ("Add vocabulary request" or "Request vocabulary") to the "Contribute" menu that goes directly to vocabulary requests. Possibly remove the vocabulary-related part of the "Add sentences" page.

(3) Store and remember the value selected for "Filter by language" when it pertains to vocabulary requests.

danail-georgiev:
This resolves #1836. Add the 'Add Vocabulary Request' submenu in the 'Contribute' menu

### DIFF
--- a/src/Controller/VocabularyController.php
+++ b/src/Controller/VocabularyController.php
@@ -118,14 +118,8 @@ class VocabularyController extends AppController
     public function add_sentences($lang = null)
     {   
     
-        $filteredLang = $this->request->getSession()->read('filter_by_language');
-        if ($filteredLang == null && $lang != null) {
-            $this->request->getSession()->write('filter_by_language', $lang );
-        } else if ($filteredLang != null && $lang == null) {
-            $lang = $this->request->getSession()->read('filter_by_language');
-        } else if ($lang != null && $filteredLang != null && $lang != $filteredLang) {
-            $this->request->getSession()->write('filter_by_language', $lang);
-        }
+        $this->request->getSession()->write('vocabulary_requests_filtered_lang', $lang );
+        
         $this->helpers[] = 'Pagination';
         $this->helpers[] = 'CommonModules';
         $this->helpers[] = 'Languages';
--- a/src/Controller/VocabularyController.php
+++ b/src/Controller/VocabularyController.php
@@ -43,7 +43,7 @@ use Cake\Datasource\Exception\RecordNotFoundException;
 class VocabularyController extends AppController
 {
     public $uses = array('UsersVocabulary', 'User', 'Sentence', 'Vocabulary');
-    public $components = array ('CommonSentence', 'Flash', 'Cookie');
+    public $components = array ('CommonSentence', 'Flash');
     public $helpers = array(
         'Vocabulary',
     );
--- a/src/Controller/VocabularyController.php
+++ b/src/Controller/VocabularyController.php
@@ -43,7 +43,7 @@ use Cake\Datasource\Exception\RecordNotFoundException;
 class VocabularyController extends AppController
 {
     public $uses = array('UsersVocabulary', 'User', 'Sentence', 'Vocabulary');
-    public $components = array ('CommonSentence', 'Flash');
+    public $components = array ('CommonSentence', 'Flash', 'Cookie');
     public $helpers = array(
         'Vocabulary',
     );
@@ -87,7 +87,6 @@ class VocabularyController extends AppController
                   'action' => 'all')
             );
         }
-        
         $this->loadModel('UsersVocabulary');
         $this->paginate = $this->UsersVocabulary->getPaginatedVocabularyOf(
             $userId,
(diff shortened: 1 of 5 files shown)


src/Controller/VocabularyController.php
```diff
@@ -117,7 +116,16 @@ public function add()
...
+    {   
+    
+        $filteredLang = $this->request->getSession()->read('filter_by_language');
+        if ($filteredLang == null && $lang != null) {
+            $this->request->getSession()->write('filter_by_language', $lang );
+        } else if ($filteredLang != null && $lang == null) {
+            $lang = $this->request->getSession()->read('filter_by_language');
+        } else if ($lang != null && $filteredLang != null && $lang != $filteredLang) {
+            $this->request->getSession()->write('filter_by_language', $lang);
+        }
```
jiru: If you set the value of `$lang` here whereas the action parameter `$lang` was null, it means the page `/vocabulary/add_sentences` will show filtered languages. This is a problem because the URL says that it’s not filtered.

For example, let’s say I visit `/vocabulary/add_sentences` and I have filter_by_language=por stored in my session. It will show Portuguese vocabulary items to me. That’s good, but now let’s say I want to share this page to a friend. So I copy and paste the URL `/vocabulary/add_sentences` and send it to my friend. My friend opens the page, but the Portuguese vocabulary items won’t show up because my friend doesn’t have filter_by_language=por stored in session.

To solve this problem, you need to modify the link for "Sentences wanted" in `src/Template/Element/vocabulary/menu.ctp`, to make it point to `/vocabulary/add_sentences/por` if I have that value in session.

jiru:
Thanks for your pull request Danail! :smiley: 

I hope my comment about the action parameter will make sense to you. We already have a similar "language remembering" mechanism for the "Adopt sentences" menu item, I think you can get inspiration from it.

I also would like to suggest using a more specific name than "filter_by_language" for the session variable name, because we have a lot of "Filter by language" drop downs all around the website, whereas your variable is only for the one in the "Sentences wanted" page.

src/View/Helper/CommonModulesHelper.php
```diff
@@ -88,7 +88,14 @@ public function createFilterByLangMod($maxNumberOfParams = 1)
...
             }
-
+            
+            if ($this->request->params['controller'] == 'Vocabulary' && $this->request->params['action'] == 'add_sentences') {
+                $filteredLang = $this->request->getSession()->read('vocabulary_requests_filtered_lang');
+                if ($filteredLang != null && $lang == 'und') {
+                    $lang = $filteredLang;
+                }
+            }
+
```
jiru: I think this code is not needed any more because the preselected value of the dropdown can be now directly set from the URL.

src/Template/Element/top_menu.ctp
```diff
@@ -32,6 +32,12 @@ $session = $this->request->getSession();
 $currentLanguage = $session->read('browse_sentences_in_lang');
 $showTranslationsInto = $session->read('show_translations_into_lang');
 
+$filteredLanguage = $session->read('vocabulary_requests_filtered_lang');
+
+if ($filteredLanguage == null) {
+    $filteredLanguage = 'und';
+}
```
jiru: This if-block is not needed because the page contents will be the same wether the URL ends with `/und` or with nothing. It’s okay to leave `$filteredLanguage` as null.

src/Template/Element/vocabulary/menu.ctp
```diff
@@ -1,6 +1,12 @@
 <?php
 use App\Model\CurrentUser;
 
+$filteredLanguage = $this->request->getSession()->read('vocabulary_requests_filtered_lang');
+
+if ($filteredLanguage == null) {
+    $filteredLanguage = 'und';
+}
```
jiru: Same here.

src/Template/Sentences/add.ctp
```diff
@@ -30,10 +30,15 @@ $this->set('title_for_layout', $this->Pages->formatTitle(__('Add sentences')));
 
 $this->Sentences->javascriptForAJAXSentencesGroup();
 $this->Html->script(JS_PATH . 'sentences.contribute.js', ['block' => 'scriptBottom']);
+$filteredLanguage = $this->request->getSession()->read('vocabulary_requests_filtered_lang');
 
+if ($filteredLanguage == null) {
+    $filteredLanguage = 'und';
+}
```
jiru: And same here too.

src/Template/Element/top_menu.ctp
```diff
@@ -104,7 +110,12 @@ $menuElements = array(
             __('Show activity timeline') => array(
                 "controller" => "contributions",
                 "action" => "activity_timeline"
-            )
+            ),
+            __('Add vocabulary request') => array(
+            "controller" => "vocabulary",
+            "action" => "add_sentences",
+                $filteredLanguage
+             )
```
jiru: Could you indent these lines correctly please?

jiru:
Good job with the update, Danail! :+1:

Please have a look at my comments.

We are almost there. I am looking forward to merge your pull request.

Issue #2080, opened 2020-01-19
Standardize the way sentences as displayed in the "new look."


ckjpn:
Wouldn't it be better to have the sentences displayed the same way in the various pages on the site? 
## This is one pattern.
<img width="592" alt="Screen Shot 2020-01-19 at 18 25 03" src="https://user-images.githubusercontent.com/7934273/72678510-faef5000-3ae9-11ea-99ac-cc9cbe0edd91.png">
<img width="580" alt="Screen Shot 2020-01-19 at 18 24 39" src="https://user-images.githubusercontent.com/7934273/72678511-faef5000-3ae9-11ea-9fa6-a31c00eb09d4.png">
<img width="581" alt="Screen Shot 2020-01-19 at 18 24 06" src="https://user-images.githubusercontent.com/7934273/72678512-fb87e680-3ae9-11ea-8adc-420f0c005904.png">

## Comments are almost the same, but not quite.
<img width="590" alt="Screen Shot 2020-01-19 at 18 23 26" src="https://user-images.githubusercontent.com/7934273/72678513-fb87e680-3ae9-11ea-9e84-75ebb9fb5dd0.png">

## These are different.
Add a sentence.
<img width="704" alt="Screen Shot 2020-01-19 at 18 28 23" src="https://user-images.githubusercontent.com/7934273/72678527-10fd1080-3aea-11ea-86f5-effd24b22623.png">

Each sentence's page.
<img width="585" alt="Screen Shot 2020-01-19 at 18 23 40" src="https://user-images.githubusercontent.com/7934273/72678528-10fd1080-3aea-11ea-8b29-43f8f73b13bb.png">

trang:
The sentence in the comment is meant to be just a preview of the sentence and this won't change. The focus should be on the comment and not on the sentence.

The sentence on the "Add sentence" page is still using the old design because we would have to migrate the functionality from jQuery to AngularJS to be able to use the new design.

Same thing for the sentence on the homepage or the lists. The code would need to be rewritten with AngularJS.

As for the sentence page, it will remain with the old design until all the features are ported to the new design because it ensures that people who have the new design enabled can still access all the sentence features if they need to.

Issue #1933, opened 2019-07-28
Wrong "Show sentence's details" button behavior in Firefox


fjay69:
When right click on sharp image (field of <path> tag) I can't open link in new tab. So I need to click a space inside the green circle to hit a link.
This what's happening when I click on sharp:
https://i87.fastpic.ru/big/2019/0728/90/80a3bd9555b9c3a866b240ddb4324990.png
And this - when I click somewhere inside the green circle, but outside the sharp:
https://i87.fastpic.ru/big/2019/0728/fc/f4c7db6a81557108305cc789bde826fc.png
This is what I see in Developer's Tools:
https://i86.fastpic.ru/big/2019/0728/f3/_2cca2c7614e6c63b7ccb9426ce1b40f3.png

Windows 7, Firefox.

trang:
Note: this issue will become obsolete when we finally finish the new sentence design. But there's still a while so until then, if anyone has a fix for this, go ahead.

fjay69:
This is my workaround:
```
var ams = document.getElementsByClassName("navigationIcon");
for(i=0; i<ams.length; i++){	  
  var newd = document.createElement("div");
  newd.style.cssText = "position: relative; width: 25px; height: 25px; top: -25px";  
  ams[i].appendChild(newd);  
}
```

Issue #1874, opened 2019-04-26
I wonder if it would be easy to add a "horizontal rule" <hr /> before the first indirect translation. 


ckjpn:
I wonder if it would be easy to add a "horizontal rule" <hr /> before the first indirect translation. That would likely make it clearer where the "indirect translations" start.

If so, that might help new members more easily realize that some of the sentences are "translations of translations."

If it's not hard to do, then this might be worth the effort, since the new layout is taking a long time.

This was on the Wall today.

https://tatoeba.org/eng/wall/show_message/31764#message_31764

ckjpn:
I don't have things set up yet to do a pull request, but if I go into the code and figure out where the <hr /> needs to be inserted, would someone be willing to insert this into what's being used on tatoeba.org.

jiru:
Your coding help is certainly welcome, but I think the real problem is not about finding where in the code to put the *hr* tag. It is more about how to make a design that is less confusing (the *hr* tag being one of the possible solutions). Writing code is easy, but writing code that works well is much harder.

If I just put an *hr* tag in there, here is how it looks:

![Screenshot_2019-06-05 Turkish example sentence Tom'u bize yardım ettirmek istiyorum - Tatoeba](https://user-images.githubusercontent.com/5107734/58966374-c93ca980-87e4-11e9-903f-9710d3866a3c.png)

Now I feel like the "translations of translations" part became a whole separate section independent from the main sentence itself.

I am not very comfortable when it comes to designing interfaces. If you have better ideas, and if you can express them using images instead or HTML tags, I am all ears.

trang:
This is an issue I already tried to solve in the new design. If we decide that it's worth investing time into changing the old sentence's design instead of investing that time to into finishing the new sentence design, then I would suggest to use the same pattern as in the new design. This would help confirming that the new design works, so it wouldn't be a waste of time to rework the old sentence design.

The pattern would be:
- Add a separator below the main sentence (light grey).
- Add a label "Translations" above the translations.
- Add a label "Translations of translations" above the indirect translations.

The labels is what help people realize that grey sentences are indirect translations. I used "Translations of translations" as a label because I felt that for completely new users, "indirect translations" may not be completely clear. At least to me, it feels a bit more technical.

We could also add a separator below the group of direct translations, but I think the label already serves as some sort of separator.

[...]

PR #2168, opened 2020-02-27
Use new sentence design everywhere when option is enabled

trang:
This PR addresses #2080. Merging this PR will also merge the PR for the lists (#2154).

Related Wall thread: https://tatoeba.org/eng/wall/show_message/34267#message_34267

I initially wanted to keep the sentence's page with the old design so that people can fallback to it in case they need to do something they cannot do with the new design but I changed my mind. I instead added a tooltip to mention on features that are not yet implemented. If someone still badly needs a feature that is not yet migrated, they will have to stay with the old design.

The only feature where I cannot really add a tooltip is the transcriptions but it will be the next feature to be migrated and hopefully will be available within one or two weeks.

Since I'd like to have more people starting to actively use this new design, I added an info message about it and how to activate it.
* Users can simply close this info message and do nothing more. The info message won't appear again unless they clear their cookies.
* This info message is only displayed to users who have not enabled the new design.
* The label of the option in the Settings has been reworded to let people know that the missing features will be soon implemented.

Additionally, the `/sentences/add` and `/sentences_lists/show` pages have been reworked, but these changes apply only when the new sentence design is enabled.
* The new version of the "Add sentences" page looks almost the same as its old version. The main difference is that a message `No sentences added yet` is displayed at first, when the list of sentences added is empty.
* The list page has gone through a bit more changes.
  * The form to add new sentences to the list is hidden by default. It is displayed by clicking on the `+` icon in the toolbar.
  * This form is the same form as the one on the `/sentences/add` and therefore includes the language dropdown (which addresses #1528), as well as the the license dropdown.
  * One extra feature was added: the possibility to undo after removing a sentence from the list.

A bunch of issues should be solved or become obsolete for people who use the new design.
* https://github.com/Tatoeba/tatoeba2/issues/761
* https://github.com/Tatoeba/tatoeba2/issues/1394
* https://github.com/Tatoeba/tatoeba2/issues/1462
* https://github.com/Tatoeba/tatoeba2/issues/1528
* https://github.com/Tatoeba/tatoeba2/issues/1614
* https://github.com/Tatoeba/tatoeba2/issues/1678
* https://github.com/Tatoeba/tatoeba2/issues/1709
* https://github.com/Tatoeba/tatoeba2/issues/1874
* https://github.com/Tatoeba/tatoeba2/issues/1911
* https://github.com/Tatoeba/tatoeba2/issues/1933


trang:
[Note to self]
When creating new sentences from the list page, if the sentence is a duplicate of a sentence that is already in the list, the error message displayed is not correct. It should say instead: `The sentence you tried to create is already in your list.` and there should be no sentence displayed below that error message.

jiru:
About the message telling about the new design being rolled out, I think you shouldn’t put it inside the random sentence header, because members who activated the option to hide the random sentence block won’t see it. How about using the good old blue banner instead?

jiru:
Or you can force showing the random sentence block for members who hide it, only for the time it includes the message.

src/Model/Entity/Sentence.php
```diff
@@ -86,4 +95,24 @@ protected function _getOldFormat()
 
         return $result;
     }
+
+    protected function _getLangName()
+    {
+        return LanguagesLib::getLanguageName($this->lang);
+    }
```
jiru: I have created `src/Model/Entity/LanguageNameTrait.php` for that. See 56497bfb748fb88eed7308595fb64eba27790d1f.

src/Model/CurrentUser.php
```diff
@@ -383,9 +383,9 @@ public static function hasAcceptedNewTermsOfUse()
         return !self::isMember() || self::getSetting('new_terms_of_use');
     }
     
-    public static function canAdoptOrUnadoptSentenceOfUser($user)
+    public static function canAdoptOrUnadoptSentenceOfUser($userId)
     {
-        if (!$user || !$user->id || $user->id === self::get('id')) {
+        if (!$userId || $userId === self::get('id')) {
```
jiru: You removed the variable `$user`, but it is still used later in that method.

src/Model/Table/SentencesTable.php
```diff
@@ -352,11 +352,13 @@ public function findFilteredTranslations($query, $options) {
                     if ($translation->indirect_translations) {
                         foreach ($translation->indirect_translations as $indirectTranslation) {
                             if (!in_array($indirectTranslation->id, $indirectIds)) {
+                                $indirectTranslation = $this->newEntity($indirectTranslation->toArray());
```
jiru: It looks like this line doesn’t do anything but duplicating the object. Is that really needed?

src/Model/Table/SentencesTable.php
```diff
@@ -352,11 +352,13 @@ public function findFilteredTranslations($query, $options) {
...
                                 $indirectTranslations[] = $indirectTranslation;
                                 $indirectIds[] = $indirectTranslation->id;
                             }
                         }
                         unset($translation->indirect_translations);
+                        $translation = $this->newEntity($translation->toArray());
```
jiru: Same here.

src/Model/Table/SentencesTable.php
```diff
@@ -377,12 +379,56 @@ public function findFilteredTranslations($query, $options) {
                 $indirectTranslations = Hash::sort($indirectTranslations, '{n}.lang', 'asc');
                 
                 $result['translations'] = [$directTranslations, $indirectTranslations];
+                $result['extraTranslationsCount'] = $this->getextraTranslationsCount($result);
+                $result['expandLabel'] = $this->getExpandLabel($result['extraTranslationsCount']);
+                if (CurrentUser::isMember()) {
+                    $result['menu'] = $this->getMenu($result);
```
jiru: I think you shouldn’t call that `menu` because we are in the context of the model, not the view. How about `perms`?

src/Model/Table/SentencesTable.php
```diff
@@ -377,12 +379,56 @@ public function findFilteredTranslations($query, $options) {
...
+        return $directTranslationCount + $indirectTranslationCount - $maxDisplayed;
+    }
+
+    private function getExpandLabel($extraTranslationsCount)
+    {
+        return format(__n(
+            'Show 1 more translation',
+            'Show {number} more translations',
+            $extraTranslationsCount,
+            true
```
jiru: This `true` parameter passed to gettext functions is a relic of a previous version of CakePHP that was used to make the function echo instead of return. You can safely remove it.

src/Template/Element/random_sentence_header.ctp
```diff
@@ -25,40 +25,47 @@
  * @link     https://tatoeba.org
  */
 
-$this->Html->script(JS_PATH . 'sentences.show_another.js', array('block' => 'scriptBottom'));
```
jiru: You should remove the file `webroot/js/sentences.show_another.js` too, since it’s not used anywhere else.

jiru:
I found a bug:

1. Enable the new design.
2. Go to `/sentences/add`.
3. Click on the `Add sentence` button.
4. The loader shows forever.

I was also able to somehow trigger this bug by clicking on the `Add sentence` button while a sentence was in the process of  being added.

src/Template/Element/sentences_lists/sentence_in_list.ctp
```diff
@@ -0,0 +1,43 @@
...
+            <md-icon>remove_circle_outline
+                <md-tooltip><?= __('Remove from list') ?></md-tooltip>
+            </md-icon>
+        </md-button>
+    </div>
+
+    <div class="remove-from-list" ng-if="vm.isRemoved[<?= $sentenceId ?>]">
+        <md-button class="md-icon-button" ng-click="vm.undoRemoval(<?= $sentenceId ?>)">
+            <md-icon>undo
+                <md-tooltip><?= __('Undo') ?></md-tooltip>
```
jiru: About the tooltip "Undo", maybe it’s just me, but I find it not clear enough, because it lacks context.

undo

I asked myself "undo what"?
I suggest you either put the button next to the text "Sentence removed from list", or make the tooltip say "Undo removal" or something clearer.

src/Template/SentencesLists/show_angular.ctp
```diff
@@ -0,0 +1,228 @@
...
+    ?>
+    
+    <md-progress-linear ng-if="vm.inProgress"></md-progress-linear>
+
+    <md-content ng-cloak>
+    <?php
+    if ($permissions['canAddSentences'] && count($sentencesInList) == 0) {
+        ?>
+        <div class="no-sentence-info" ng-if="!vm.showForm && vm.sentences.length === 0">
+            <p><?= __('There are no sentences yet.') ?></p>
```
jiru: I think "This list is empty." would be a clearer message.

src/Template/SentencesLists/show_angular.ctp
```diff
@@ -0,0 +1,228 @@
...
+            </div>
+            <div class="hint">
+                <?php
+                echo format(
+                    __(
+                        'You can also add existing sentences to this list, from other pages, by clicking on'.
+                        'the list icon {addToListButton} in the menu of the sentence.', true
+                    ),
+                    ['addToListButton' => '<md-icon>list</md-icon>']
+                );
```
jiru: Good job for putting useful information in this empty state. :tada:

jiru:
After adding a sentence from the list page, when I click on the list icon of that newly-added sentence, it shows as not belonging to the list. Only refreshing the page can fix this issue. This might be seen as a minor issue, but I think it can be confusing for people who are trying to understand the lists functionality by exploring and comparing.

src/Template/Element/sentences/list_form.ctp
```diff
@@ -2,7 +2,7 @@
     <div layout="row" layout-margin>
         <md-input-container flex>
             <label><?= __('Search list or enter new list name') ?></label>
-            <input id="list-form-<?= $sentenceId ?>" ng-model="vm.listSearch"
+            <input id="list-form-{{vm.sentence.id}}" ng-model="vm.listSearch"
```
jiru: Shouldn’t this attribute be `ng-attr-id="..."`?

src/Template/Element/sentences/sentence_and_translations.ctp
```diff
@@ -1,4 +1,24 @@
...
+ *     'sentence' => $sentence,
+ *     'translations' => $translations
+ * ]);
+ * 
+ * (2) Example for displaying a sentence with Angular variables:
+ * 
+ * $this->element('sentences/sentence_and_translation', [
+ *     'sentenceData' => 'sentence',
+ *     'directTranslationsData' => 'sentence.translations[0]',
+ *     'indirectTranslationsData0' => 'sentence.translations[1]'
```
jiru: indirectTranslationsData0 → indirectTranslationsData

src/Template/Element/sentences/sentence_and_translations.ctp
```diff
@@ -11,31 +31,31 @@ if (!isset($menuExpanded)) {
...
+}
+
 $langs = $this->Languages->profileLanguagesArray(false, false);
 
-$userLanguagesJSON = htmlspecialchars(json_encode($langs), ENT_QUOTES, 'UTF-8');
-$sentenceJSON = $this->Sentences->sentenceForAngular($sentence);
-$directTranslationsJSON = $this->Sentences->translationsForAngular($directTranslations);
-$indirectTranslationsJSON = $this->Sentences->translationsForAngular($indirectTranslations);
+if (!isset($userLanguagesData)) {
+    $userLanguagesData = htmlspecialchars(json_encode($langs), ENT_QUOTES, 'UTF-8');
```
jiru: By the way, you can use [`h($var)`](https://book.cakephp.org/3/en/core-libraries/global-constants-and-functions.html#h) instead of `htmlspecialchars($var, ENT_QUOTES, 'UTF-8')`.

src/Template/Element/sentences/sentence_and_translations.ctp
```diff
@@ -46,10 +66,14 @@ $sentenceUrl = $this->Url->build([
     'action' => 'show'
 ]);
 ?>
-<div ng-cloak
+<div ng-cloak flex
      sentence-and-translations
-     ng-init="vm.init(<?= $userLanguagesJSON ?>, <?= $sentenceJSON ?>, <?= $directTranslationsJSON ?>, <?= $indirectTranslationsJSON ?>)"
+     ng-init="vm.init(<?= $userLanguagesData ?>, <?= $sentenceData ?>, <?= $directTranslationsData ?>, <?= $indirectTranslationsData ?>)"
      class="sentence-and-translations md-whiteframe-1dp">
+    <div ng-if="vm.sentence.duplicate" layout="row" layout-padding style="border-bottom: 1px solid #f1f1f1">
```
jiru: Maybe put this style in the CSS, otherwise it will be duplicated for each sentence block.

src/Template/Element/sentences/sentence_form.ctp
```diff
@@ -1,19 +1,16 @@
-<?php 
-$hasAudio = count($sentence->audios) > 0;
-?>
 <form layout="column" layout-margin style="padding-top: 10px" ng-if="vm.visibility.sentence_form">
-<?php if ($hasAudio) { ?>
 
+<div ng-if="vm.sentence.audios && vm.sentence.audios.lengh > 0">
```
jiru: I think you made a typo: lengh → lenght

src/Template/Element/sentences/sentence_form.ctp
```diff
@@ -1,19 +1,16 @@
...
     <div layout="row" layout-align="end center">
         <md-button class="md-raised" ng-click="vm.cancelEdit()">
             <?= __('Close') ?>
         </md-button>
     </div>
+</div>
 
-<?php } else { ?>
-
+<div layout="column" ng-if="!(vm.sentence.audios && vm.sentence.audios.lengh > 0)">
```
jiru: Same typo here.

src/Template/Element/sentences/sentence_form.ctp
jiru: I mean length :laughing:

src/Template/Element/sentences/sentence_form.ctp
```diff
@@ -33,7 +30,7 @@ $hasAudio = count($sentence->audios) > 0;
 
     <md-input-container>
         <label><?= __('Sentence') ?></label>
-        <textarea id="sentence-form-<?= $sentence->id ?>" ng-model="vm.sentence.text" 
+        <textarea id="sentence-form-{{vm.sentence.id}}" ng-model="vm.sentence.text"
```
jiru: id → ng-attr-id

webroot/js/directives/sentence-and-translations.dir.js
```diff
@@ -146,14 +125,45 @@
...
+            }
+            vm.inProgress = true;
+            $http.get(url).then(function(result) {
+                var sentence = result.data.random;
+                var directTranslations = sentence.translations[0];
+                var indirectTranslations = sentence.translations[1];
+                init(vm.userLanguages, sentence, directTranslations, indirectTranslations);
+                initMenu(false, sentence.menu);
+                vm.inProgress = false;
+            });
```
jiru: You are not checking for errors here. I think you can at least reset `vm.inProgress` in case of errors, like that:

```js
$http.get(url).then(function(result) {
    //...
}).finally(function() {
    vm.inProgress = false;
});

**Comment by jiru** [comment]
From a page with several sentences, like `/sentences/show_all_in/fra/none`, when I click on the "..." button of a sentence, it expands the menu for all the sentences of the page. I think this is not the intended beheavior.

**Comment by trang** [comment]
As always, thanks for the review @jiru! I've addressed the various issues you mentioned.

> I have to say it feels strange from a UI perspective to see all these sentence menus opening and closing simultaneously.

Yes, it was definitely strange. I went for another solution which is to have a button in the toolbar (where applicable) to expand/collapse all menus.

When toggling the menu, it saves the expanded state in a cookie so that people can maintain the expanded or collapsed state when going to other pages. 

I hope this will be enough for users to transition smoothly from the old menu to the new menu.

> About the message telling about the new design being rolled out, I think you shouldn’t put it inside the random sentence header, because members who activated the option to hide the random sentence block won’t see it.

I think it's okay if those who have deactivated the random sentence are not aware of the new design because they are not really impacted at this stage. Later on there will be a second announcement stating that the new design is rolled out on every page, so they will be aware eventually.

File: src/Lib/LanguagesLib.php [hunk 19]
```diff
@@ -676,4 +676,14 @@ public static function languageExists($code)
...
+
+    public static function getLanguageName($code) {
+        $languages = self::languagesInTatoeba();
+        if (isset($languages[$code])) {
+            return $languages[$code];
+        } else {
+            return __('unknown');
+        }
+        
+    }
```
jiru: You can remove this now. :koala:

jiru:
@trang, do you agree that this PR is more or less ready for merge? I will read it again but I think it’s okay. I’d like to merge it by Wednesday and update the POT files so that translators have time to translate the many new strings.

Note that you havn’t answered my [earlier comment about object duplication](#pullrequestreview-369497749). I don’t really understand why this is needed.

src/Model/Table/SentencesTable.php
trang: Ah yes, sorry, I forgot to answer this.

It is needed because of the virtual fields (`lang_name`,  `dir`, `is_favorite`, `is_owned_by_current_user`). These virtual fields won't be set in the `contain`'ed entities... So recreating the entity was a way to have them set. If you have another solution I'm all ears.

trang:
> do you agree that this PR is more or less ready for merge?

Yes, I agree it is more or less ready to merge. I will include the implementation of the copy feature in the new design since the build is breaking because of it at the moment.

But in any case, you can merge and I (or you as well) can make whatever remaining fixes directly in `dev`. It would be good indeed if translators could get the chance to translate ahead of time.

src/Model/Table/SentencesTable.php
jiru: Okay I found the problem. The entity is not a sentence, it’s a translation, so the virtual fields have to be defined in src/Model/Entity/Translation.php.

webroot/js/directives/sentence-and-translations.dir.js
```diff
@@ -146,16 +147,53 @@
...
+            var url = rootUrl + '/sentences/random';
+            if (lang) {
+                url += '/' + lang;
+            }
+            vm.inProgress = true;
+            $http.get(url).then(function(result) {
+                if (!result.data || !result.data.random) {
+                    // TODO Show error
+                    return;
+                }
```
jiru: This won’t work when the returned HTTP code is 4xx or 5xx. You have to do it like this:
```js
$http.get(url).then(
    function(response) {
        // Server did handle the request normally
        // We got whatever data has been set by the action method
    },
    function(response) {
        // server couldn’t handle the request (not logged in,
        // unauthorized, maintenance ongoing, ddos ongoing...)
    }
).finally(function() {
    // that gets additionally executed in the end no matter what
});
```
By the way, if you don’t provide the second parameter to `.then()`, a warning is printed in the browser console.

src/Model/Table/SentencesTable.php
trang: Thanks! I created the Translation entity and made it extend Sentence. It works fine from there.

webroot/js/directives/sentence-and-translations.dir.js
trang: I would keep this for some refactoring session. This is an issue for every $http request. We should set up some generic error handling for status 4xx and 5xx.

PR #2171, opened 2020-02-29
Make it easier to add new languages

### LINKED ISSUE #1916: It's hard to add languages you speak/know
I was trying to add some more languages I know on my profile. I visited [my own profile](https://tatoeba.org/fra/user/profile/Ricardo14) to do so and clicked on "Add language".  I added Bavarian to the list but to add other languages, I had to click on "save" and do the progress all over again. It may make that new members don't add all the languages they know and the progres is kinda tricky since it's not obvious I have to repeat the process all over again. On many websites we can add all the languages we speak at once - including adding your proficiency.

jiru:
This PR fixes #1916 by allowing to add languages directly from the profile page. Edition and deletion of languages stay unchanged.

In addition, it brings back the "no languages added" message when the list is empty.


src/Template/User/profile.ctp
```diff
@@ -37,6 +37,10 @@ use App\Model\CurrentUser;
  * @link     https://tatoeba.org
  */
 
+$this->Html->script('user/language.ctrl.js', ['block' => 'scriptBottom']);
+$this->Html->script('/js/directives/sentence-and-translations.dir.js', array('block' => 'scriptBottom')); // TODO: this is just to get the languageIcon directive
```
trang: I think we could consider just moving the `languageIcon` directive to `app.module.js`. As we migrate more pages to AngularJS, this directive will be used in many more places so I think it would make sense to have it in there.

If not, you can extract the directive into its own file (`language-icon.dir.js`), then inject it in both `profile.ctp` and `sentence_and_translations.ctp`.

Or, after merging this into `dev`, you can include it in the `[sentence-component.js]` of the [asset_compress.ini](https://github.com/Tatoeba/tatoeba2/blob/dev/config/asset_compress.ini#L55]) instead of injecting it in `sentence_and_translations.ctp`.

src/Template/User/profile.ctp
```diff
@@ -302,106 +306,180 @@ $this->set('title_for_layout', h($this->Pages->formatTitle($title)));
...
+$userLanguages = htmlspecialchars(json_encode($userLanguages), ENT_QUOTES, 'UTF-8');
+?>
+    <div class="section md-whiteframe-1dp"
+         ng-cloak
+         ng-controller="LanguageController as vm"
+         ng-init="vm.init(<?= $userLanguages ?>)">
+        <h2><?= __('Languages'); ?></h2>
 
-            echo '<p>';
+        <p ng-if="vm.langs.length == 0">
```
trang: In the Javascript world, the recommendation is to use `===` by default, rather than `==`. At least it seems that developers have had enough headaches from using `==` that there's plenty of articles explaining why `===` is almost always the correct way to test equality.

In this particular case here I don't think there will be unwanted side effects, but it's just to make sure you're aware, since I've seen you have been using double equality elsewhere.

trang:
On the visual side, I would recommend the following small changes:
- Use the `md-toolbar` for the "Languages" section and include the "Add language" button in the toolbar. We have this pattern on the homepage.
- Move the language form to the top of the section, so that it is displayed below the toolbar when clicking on "Add language".
- The "Add language" title could be removed from the form. The user already clicked on "Add language" so they already know the form is about adding a language.
- Remove the `md-whiteframe` from the form and use instead a `border-bottom` to separate the form from the list of languages. If needed, also add a very light grey background color to the form (`#fafafa`) so that it is more clearly separated from the list of languages. In general we want to avoid stacking whiteframes inside each other.

If you're up for a more challenging redesign, you could try to use a [panel](https://material.angularjs.org/latest/demo/panel) to open the form in a dialog box.
- Clicking on "Add language" opens the dialog box.
- Once the user reaches the last step and clicks on "Add this language", it closes the dialog box and updates accordingly the list of languages on the profile.

The dialog box can then be reused on various places to avoid disrupting the workflow of the contributor whenever they want to contribute in a language that is not yet in their profile. For instance, for the case where a contributor does not have any language in their profile yet and wants to add new sentences:
  - Currently they will have quite unnecessary redirections. 
      - When they click on "Add language", it redirects to the "Add language" page.
      - Then once they save, they are redirected to the profile. 
      - Then they have to find their way back to the "Add sentences" page...
  - A more friendly workflow would be:
     - When click on "Add language" it opens the dialog box.
     - The user selects the appropriate language and level, then saves.
     - The dialog box is closed and the info message about how they must add languages to their profile is replaced by the sentence form.
     - The language dropdown in the form contains the language that was just added by the user.

jiru:
> Use the md-toolbar for the "Languages" section and include the "Add language" button in the toolbar. We have this pattern on the homepage.
> Move the language form to the top of the section, so that it is displayed below the toolbar when clicking on "Add language".

I moved the button down because I saw how hard it was for users to find the button during my UX tests. While the button looks "nicer" on the top, I don’t think this is where people are looking when they want to add a new language. I’d like to leave it down and do another UX test.

> The "Add language" title could be removed from the form. The user already clicked on "Add language" so they already know the form is about adding a language.

I think it’s better to keep it because it makes it clearer what’s going on. If the user leaves the page and then comes back, the title makes it easier to understand where the user was at.

> Remove the md-whiteframe from the form and use instead a border-bottom to separate the form from the list of languages. If needed, also add a very light grey background color to the form (#fafafa) so that it is more clearly separated from the list of languages. In general we want to avoid stacking whiteframes inside each other.

Makes sense. I’ll make the form title a bit smaller then.

> If you're up for a more challenging redesign, you could try to use a panel to open the form in a dialog box.

I see the point of using a dialog, but I won’t do it this time because I want to work on the tutorial after registration.

jiru:
@trang I’m gonna merge this now so that we have something better on production already, but it doesn’t mean I’m ignoring your comments. I’m open for more feedback.

trang:
> @trang I’m gonna merge this now so that we have something better on production already, but it doesn’t mean I’m ignoring your comments. I’m open for more feedback.

@jiru No worries. You do what you can, when you can :) You may anyway get better feedback than I can give by having this on production and being used in real conditions by the rest of the users.

Issue #2039, opened 2019-12-16
UTF-8 characters in the CSS sometimes get interpreted as another encoding instead


Yorwba:
[Reported by Pandaa on the wall](https://tatoeba.org/eng/wall/show_message/33747#message_33747)

![](https://i.imgur.com/iRtGRrl.jpg)

The character before "Advanced Search" should be `▸ `. The `â–¸ ` visible in the screenshot is the result of decoding the UTF-8 bytes as any of several other encodings like Windows code page 1250.
I'm guessing that the browser is incorrectly using that encoding because it has not been declared explicitly by the point the `layout.css` file is being parsed.

[W3.org says the following about declaring the character encoding of CSS](https://www.w3.org/International/questions/qa-css-charset):
> You should always use UTF-8 as the character encoding of your style sheets and your HTML pages, and declare that encoding in your HTML. If you do that, there is no need to declare the encoding of your style sheet.
> Other approaches are only needed if your style sheet contains non-ASCII characters and, for some reason, you can't rely on the encoding of the HTML and the associated style sheet to be the same. In this case you should use `@charset` or HTTP headers to declare the encoding. (If your HTML and CSS files use the same encoding, the latest versions of major browsers will apply the encoding of the HTML file to the CSS stylesheet.)

There is no `@charset` declaration in `layout.css`, but the CSS uses the same encoding as the HTML. Under the assumption that someone who knows how to open the developer tools also keeps their browser up to date, I guess that makes it likely the HTML encoding was not declared properly.

There is a `<meta charset="utf-8">` tag in the HTML, but it comes after a long list of inline styles. [W3.org says that the declaration should be within the first 1024 bytes](https://www.w3.org/International/questions/qa-html-encoding-declarations):
> Always declare the encoding of your document using a meta element with a charset attribute, or using the http-equiv and content attributes (called a pragma directive). The declaration should fit completely within the first 1024 bytes at the start of the file, so it's best to put it immediately after the opening head tag.

Maybe that is the root cause.

AndiPersti:
> There is a `<meta charset="utf-8">` tag in the HTML, but it comes after a long list of inline styles. [W3.org says that the declaration should be within the first 1024 bytes](https://www.w3.org/International/questions/qa-html-encoding-declarations):

But the meta tag comes directly after the opening head tag:

    $ curl -s https://tatoeba.org/eng | head -n 5
    <!DOCTYPE html>
    <html lang="en">
    <head>
[...]

PR #2174, opened 2020-03-02
Add charset declaration at the beginning of layout.css

AndiPersti:
This PR addresses #2039 

`layout.css` contains Unicode characters so we need to declare the encoding of that file to ensure that all clients will interpret these characters correctly.


jiru:
Danke! :smile_cat:

PR #2175, opened 2020-03-03
Correct time of exports given on the download page

### LINKED ISSUE #1920: On the Downloads page "9:00 GMT" needs to be "6:00"
## The Problem

The wrong time is listed on this page: https://tatoeba.org/eng/downloads
"The files provided here are updated every Saturday at 9 a.m. (GMT). "

## Correction Needed

6:00 a.m. (GMT) or possibly 06:00 UTC

## Notes 

Perhaps you might want to change GMT to UTC, too.

The Difference Between GMT and UTC
https://www.timeanddate.com/time/gmt-utc-time.html

See the timestamps on files here.
https://downloads.tatoeba.org/exports/

I usually start my downloads at 6:25 UTC since it takes time for the files to be created.
You may want to compensate for this in what's written on the page by just saying 6:30 or something like that.

## Request

Please don't change the time the files are created.  This is a very convenient time for me  to download and process files.   I wouldn't mind an hour or two earlier, but please don't make this later.

agrodet:
This small PR modifies the message giving the time of the exported files available on the download page, to fix #1920.

### DIFF
--- a/src/Template/Pages/downloads.ctp
+++ b/src/Template/Pages/downloads.ctp
@@ -57,8 +57,8 @@ $this->set('title_for_layout', $this->Pages->formatTitle(__('Download sentences'
         <p>
             <?php
             echo __(
-                'The files provided here are updated every <strong>Saturday at 9 a.m.</strong> '.
-                '(GMT).'
+                'The files provided here are updated every <strong>Saturday at 6:30 a.m.</strong> '.
+                '(UTC).'
             );
             ?>
         </p>


jiru:
Why not change the message the way you [suggested](https://github.com/Tatoeba/tatoeba2/issues/1920#issuecomment-586860464)?

> I'm not sure there is added value to tell the user about when the exports are "triggered". The interesting information is at what time we can download the files. Taking a long enough insurance, shall we go with the following?
>
> `The files provided here are updated every Saturday at 6:30 a.m. (UTC). `

agrodet:
I'm not sure what you mean. Isn't the string in the commit the same as the one you gave?

jiru:
Yes, you are right, I got confused.

PR #2176, opened 2020-03-03
Remove obsolete 'time' option for cache

AndiPersti:
CakePHP 3 doesn't use the `time` option for caching view elements anymore.


jiru:
Thank you! :+1:

Issue #2144, opened 2020-02-15
Activity timeline page is hard to understand


jiru:
[UX test 2](https://en.wiki.tatoeba.org/articles/show/ux-test-2) and [3](https://en.wiki.tatoeba.org/articles/show/ux-test-3) showed that the [Activity timeline](https://tatoeba.org/contributions/activity_timeline) page is hard to understand. Let’s make it more self-explicit.

I am not familiar with this page actually. I would like to ask what’s the purpose of this page to members who use it (as far as I know @Guybrush88 and @ckjpn).

As mentioned in #1068, the location of this item in the top menu is also questionable. Maybe we could group it with other stats pages?

ckjpn:
Reverse sorted, like it was years ago, might make it easier to understand, since today's date would be at the top, with a lower number.

Adding a title, other than just the month and year, would also likely help.

**Number of Sentences Added**, or something like that.

trang:
The purpose of this page is mostly just to fulfill the curiosity of people who like stats.

ckjpn:
Note that I don't really use it much, even though I occasionally harvest the data from there to put it on this page, so I can get a quick overview of things.

http://tatoeba.ueuo.com/timeline/

This has color-coding, somewhat like what TRANG used to have.

ckjpn:
Note that if we had the last 2 or 3 days listed on the main page, mentioned in another issue, then members could sort of get the feeling of how things were going without needing to go to this page at all.

Guybrush88:
> I would like to ask what’s the purpose of this page to members who use it (as far as I know @Guybrush88 and @ckjpn).

Personally I use it out of curiosity, to have an overall idea on how much Tatoeba is used to add new sentences over time, so from time to time I enjoy seeing how the whole database has grown (and how it is growing) day by day

agrodet:
There's also #1882. I've been thinking of looking at it for a while, but I think there are other issues I should be solving first so I didn't suggest anything. Anyway, since the occasion has come, here's my ideal "Activity timeline page". Note that I didn't feel like building a prototype so the explanation might be lengthy. 

- Remove it from the "Contribute" menu. Place it under "STATS PER LANGUAGE".  
**Remarks:** That makes it necessary to go back to the homepage to access the page. However, the "stats per language" page has the same issue, I think. We could imagine putting them both under the "Community menu". 

- Name it "Activity timeline - Month Year"

[...]

Issue #1882, opened 2019-05-05
The activity timeline should be more flexible with deleted sentences


Guybrush88:
Currently, the activity timeline (https://tatoeba.org/ita/contributions/activity_timeline/2019/05) keeps track of every single sentence that it's added. However, sentences that are eventually deleted are still kept in the count, and, in my opinion, they should be removed from it, since this could improve the accuracy of the overall contributions' stats.

If this isn't feasible, maybe we could have a distinction between the sentences that are contributed and the sentences that are deleted, so anyone who is interested in stats can see a clearer distinction between these two types of contributions

jiru:
I think that removing a sentence is also a contribution and should be counted as such.

> If this isn't feasible, maybe we could have a distinction between the sentences that are contributed and the sentences that are deleted, so anyone who is interested in stats can see a clearer distinction between these two types of contributions

That seems like a good idea to me, but I think we need to think more about the problem you are trying to solve here, in order to come up with a good solution. Currently, the graph answers the question: how many contributions were done each day? Now, what other questions would you like the graph to answer?

Guybrush88:
within the graph, I'd personally like to see a clearer distinction between the sentences that have been actually contributed and the sentences that are actually deleted, since they're different types of contribution, and, if a user needs some stats about the actual growth of the whole corpus, this would be a clearer distinction for this purpose (i.e.: there could be a mass deletion of sentences and even more new sentences are added on the same day, but the graph currently doesn't show an accurate chart of sentences' contributions during that day, so I don't find this an accurate stat)

jiru:
I think it's a good idea, so I'd like to know why you want to see the number of deleted sentences per day, so that we can find a good way to display that information. Here is a non exhaustive list of questions I suggest you ask yourself. Do you want to see the actual number of deleted sentences? Or you just want to visually compare how much it represents compared to the number of added sentence on a given day? Or, since you're talking about the growth of the corpus, do you want to be able to compare the number of deleted sentences with the next/previous day?

Guybrush88:
Do you want to see the actual number of deleted sentences? Or you just want to visually compare how much it represents compared to the number of added sentence on a given day?

Maybe we can have both, if feasible. I suggest a bar with a different color and, next to it, the number of deleted sentences, so we could have: on the far left the total number of sentences that have been added, next to it on the right a visual bar, to the right of this bar another bar (with a different color) representing the deleted sentences, to its right the total number of deleted sentences

[...]

PR #2180, opened 2020-03-07
Rework the Activity timeline page

agrodet:
This PR addresses #2144. It should also solves #1882. The page looks like the following 

image

I took a screenshot of the Japanese page because the default font is bigger, thus things are displayed a little bit more differently. The original English is "Sentences added".
 
I have the following concerns:

- I wonder if we should to vertically center the text inside the div of the first line (legend). However, I cannot I cannot find satisfying CSS rules. If someone could give a hand here, that'd be nice. Although I'm not sure it would look better if the text is vertically aligned.
- I'm not sure of the English sentence "Statistics about the number of contributions over this month." For month other than the current one, the sentence looks like "Statistics about the number of contributions over February 2020" Maybe @alanfgh could tell me if that's correct English or if I should use something else.   
- Maybe CakePHP (or raw PHP) has a better way to build the stats array than what I did. 

Some technical notes:
- Until version 10.2.3, MariaDB does not support multiple triggers on the same event. I updated the part of the trigger on `last_contributions` that used to take care of filling `contributions_stats`.
- The trigger records `sentence + insert`, `sentence + delete`, `link + insert`, `link + delete`. It ignores the other contributions. For links, the number displayed is half the one recorded in the table.
- The + and - before the numbers seemed strange at first, but once the eye gets used to it, I think that gives a good feeling. The colors are similar to the ones used through Tatoeba. 
- I modified some `height` attributes because of the difference of heights between the default fonts of different languages (i.e. Japanese).
- I set a min-width attribute to be able to distinguish small numbers (like the last line of the screenshot).

### DIFF
--- a/src/Command/FillContributionsStatsCommand.php
+++ b/src/Command/FillContributionsStatsCommand.php
@@ -43,12 +43,13 @@ class FillContributionsStatsCommand extends Command
         try {
             $firstDay = new DateTime($from);
             $lastDay = new DateTime($to);
+            $lastDay->modify('+1 day');
         } catch (Exception $e) {
             $io->error($e->getMessage());
             $this->abort();
         }
 
-        for ($day = $firstDay; $day <= $lastDay; $day->modify('+1 day')) {
+        for ($day = new DateTime($from); $day < $lastDay; $day->modify('+1 day')) {
             $date = $day->format('Y-m-d');
             $stats[$date] = [
                 ['sentence', 'insert', 0],
@@ -63,7 +64,7 @@ class FillContributionsStatsCommand extends Command
         // Fetch
         $contributions = $this->Contributions->find()
                               ->where(["type !=" => "license", "action !=" => "update",
-                                       "datetime >=" => $from, "datetime <=" => $to])
+                                       "datetime >=" => $firstDay, "datetime <=" => $lastDay])
                               ->order(["datetime" => "ASC"]);
         $contributions->disableBufferedResults();
         foreach ($contributions as $contribution) {
@@ -88,7 +89,7 @@ class FillContributionsStatsCommand extends Command
 
         // Truncate table and Fill
         $contributionsStats = TableRegistry::getTableLocator()->get('ContributionsStats');
-        $contributionsStats->deleteAll(['date >=' => $from, 'date <=' => $to]);
+        $contributionsStats->deleteAll(['date >=' => $from, 'date <=' => $lastDay->format('Y-m-d')]);
         foreach ($stats as $date => $dailyStat) {
             foreach ($dailyStat as $unitRecord) {
                 if ($unitRecord[2] != 0) {
--- a/src/Command/FillContributionsStatsCommand.php
+++ b/src/Command/FillContributionsStatsCommand.php
@@ -1,0 +1,79 @@
+<?php
+namespace App\Command;
+
+use Cake\Console\Arguments;
+use Cake\Console\Command;
+use Cake\Console\ConsoleIo;
+use Cake\ORM\TableRegistry;
+use \Datetime;
+
+class FillContributionsStatsCommand extends Command
+{
+    public function initialize()
+    {
+        parent::initialize();
+        $this->loadModel('Contributions');
+        $this->loadModel('ContributionsStats');
+    }
+
+    public function execute(Arguments $args, ConsoleIo $io)
+    {
+        // Initialize
+        $firstDay = new DateTime('2007-09-30');
+        $today = new DateTime('now');
+
+        for ($day = $firstDay; $day <= $today; $day->modify('+1 day')) {
+            $date = $day->format('Y-m-d');
+            $stats[$date] = [
+                ['sentence', 'insert', 0],
+                ['sentence', 'delete', 0],
+                ['link', 'insert', 0],
+                ['link', 'delete', 0]
+            ];
+
+            $inserted[$date] = [];
+        }
+
+        // Fetch
+        $contributions = $this->Contributions->find()
+                              ->where(["type !=" => "license", "action !=" => "update"])
+                              ->order(["datetime" => "ASC"]);
+        $contributions->disableBufferedResults();
+        foreach ($contributions as $contribution) {
+            if(!is_null($contribution->datetime)) {
+                $date = $contribution->datetime->format('Y-m-d');
+                if ($contribution->type === 'sentence' && $contribution->action === 'insert') {
+                    $stats[$date][0][2] += 1;
+                    array_push($inserted[$date], $contribution->sentence_id);
+                } elseif ($contribution->type === 'sentence' && $contribution->action === 'delete') {
+                    if (in_array($contribution->sentence_id, $inserted[$date])) {
+                        $stats[$date][0][2] -= 1;
+                    } else {
+                        $stats[$date][1][2] += 1;
+                    }
+                } elseif ($contribution->type === 'link' && $contribution->action === 'insert') {
+                        $stats[$date][2][2] += 1;
+                } elseif ($contribution->type === 'link' && $contribution->action === 'delete') {
+                        $stats[$date][3][2] += 1;
+                }
+            }
+        }
+
+        // Truncate table and Fill
+        $contributionsStats = TableRegistry::getTableLocator()->get('ContributionsStats');
+        $contributionsStats->deleteAll([]);
+        foreach ($stats as $date => $dailyStat) {
+            foreach ($dailyStat as $unitRecord) {
+                if ($unitRecord[2] != 0) {
+                    $record = $contributionsStats->newEntity([
+                        'date' => $date,
+                        'type' => $unitRecord[0],
+                        'action' => $unitRecord[1],
+                        'sentences' => $unitRecord[2]
+                    ]);
+                    $contributionsStats->save($record);
+                }
+            }
+        }
+    }
+}
--- a/src/Command/FillContributionsStatsCommand.php
+++ b/src/Command/FillContributionsStatsCommand.php
@@ -4,8 +4,10 @@ namespace App\Command;
 use Cake\Console\Arguments;
 use Cake\Console\Command;
 use Cake\Console\ConsoleIo;
+use Cake\Console\ConsoleOptionParser;
 use Cake\ORM\TableRegistry;
 use \Datetime;
+use \Exception;
 
 class FillContributionsStatsCommand extends Command
 {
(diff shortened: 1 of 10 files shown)


trang:
You also need to update this script:
https://github.com/Tatoeba/tatoeba2/blob/dev/docs/database/scripts/init_contributions_stats.sql

agrodet:
I can see two ways. The first one is to simply modify it to 
```
INSERT INTO contributions_stats(`date`, `lang`, `sentences`, `action`, `type`) 
 SELECT date_format(`datetime`, "%Y-%m-%d") as `day`, NULL, COUNT(*), `action`, `type`
  FROM contributions
  WHERE type != 'license' AND action != 'update'
  GROUP BY `day`, `action`, `type`;
```
But the "deletes of today are removed from the inserts" behavior I introduced in the trigger will not be reflected.

The other way is to do it "properly". I think that's a heavier process (but not impossible) and probably a Cake command would be preferable to a SQL procedure.

What do you think?

jiru:
@agrodet 
That looks tricky indeed. Could you temporarily remove the "deletes of today are removed from the inserts" behaviour from the trigger, and modify `init_contributions_stats.sql` like you said, so that we can merge your PR already? Then, try to add the enhanced behaviour in another PR.

As for filling `contributions_stats` while keeping track of sentences that are removed on the same day, I think it would be better to write it as a Cake command because then you can write unit tests for it. By the way, I tried to run the SELECT part of the query you mentioned on dev.tatoeba.org, it took 51 seconds, so it’s okay like that, but I suspect anything more complex in SQL would likely take ages because the contributions table is very big.

That’s why I think you should go through the table in a Cake command instead. I suggest to do it this way: just `select()` the whole contributions table with [buffered results turned off](https://book.cakephp.org/3/en/orm/retrieving-data-and-resultsets.html#working-with-result-sets). This allows you to iterate over the results as if it were a stream, so it won’t hog memory. Note that in this case you won’t be able to perform any other SQL query while iterating, so keep in mind you can only start writing to `contributions_stats` *after* you finished iterating over `contributions`. That shouldn’t be a big problem though.

@ckjpn or @alanfgh
Could you please confirm [these two strings](https://github.com/agrodet/tatoeba2/blob/01c825d6a5fbaac9791b3e4fa9b3dc81ac90cb77/src/Template/Contributions/activity_timeline.ctp#L69-L72) are correct English?

agrodet:
I prefer not postpone. I'm not good at doing postponed things ^^

I added a command to fill the `contributions_table`. This command includes the "deletes of today are removed from the inserts" behavior. It completely truncates the table before refilling it (so please backup the table just in case ^^).
I also added a condition in the template to not display a stat whose count is 0.

I don't know what should be done with the `init_contributions_stats.sql` file so I let it there.

I may add the "unlinking of today are removed from the linking" one day, but the behavior is more complicated so not for this PR.

jiru:
Wow, that’s great, thank you. :smile:

It looks all good and nice, but I am afraid of what will happen to the `contributions_stats` table if a member makes a contribution while we are running your script. We could put the website under maintenance while running the script, but I would like to avoid that because the stats page is not so important.

Could your script deactivate the SQL trigger and put it back after it’s done? The contributions made while the script is running won’t be counted, but at least nothing strange will happen.

agrodet:
Oh! I didn't think about that one. Let me check and come back to you. ^^

src/Template/Contributions/activity_timeline.ctp
```diff
@@ -43,81 +43,131 @@ $selectedMonth = format(
...
     <div class="section md-whiteframe-1dp">
-    <h2><?php echo $selectedMonth; ?></h2>
+    <h2><?php echo __('Activity timeline') . ' - ' . $selectedMonth; ?></h2>
+    <p>
+    <?php
+        if ($month == date('m') && $year == date('Y')) {
+            echo __('Statistics about the number of contributions over this month.');
+        } else {
+            echo format(__('Statistics about the number of contributions over {$selectedMonth}.'),
+                        array('$selectedMonth' => $selectedMonth));
```
jiru: It’s not really good for translators to have this `$selectedMonth` inserted like that out of their control. Could you get rid of `$selectedMonth` entirely and update these three translatable strings so that they include `{month}` and `{year}`?

jiru:
Or you could run your update script in two passes. First pass: update the whole contributions tables; second pass: only update today’s contributions. The second pass should be quick enough so that the trigger won’t interfere.

You could make your script accept a date as parameter to limit updating only for that date. That could be useful to "repair" stuff otherwise.

agrodet:
I have added two options to specify two dates between which rewrite the contributions stats. Instead of truncating the whole table, the command will only delete the records corresponding the dates given. Therefore, the id order will not necessarily correspond to the date order anymore.

However, there is still a problem for today's contributions. We can't stop the trigger because it's in charge of `last_contributions`. The command will run fast for today but there's no guarantee that some record won't interfere. It could happen that we end with two `contributions_stats` records for the same action & type. One solution could be to lock the table for update the time the script is running. Another solution could be to simply let one day pass and correct the problematic results by hand the next day.

src/Command/FillContributionsStatsCommand.php
```diff
@@ -36,7 +62,8 @@ public function execute(Arguments $args, ConsoleIo $io)
 
         // Fetch
         $contributions = $this->Contributions->find()
-                              ->where(["type !=" => "license", "action !=" => "update"])
+                              ->where(["type !=" => "license", "action !=" => "update",
+                                       "datetime >=" => $from, "datetime <=" => $to])
```
jiru: The parameter `$to` won’t work like you thought.

Mysql stores the datetime as `YYYY-MM-DD HH:MM:SS`. Let’s say `$to = '2020-01-01'`. If there is a record with `datetime = 2020-01-01 01:23:45`, then it won’t match because the string comparison `'2020-01-01 01:23:45' <= '2020-01-01'` is false. Actually, it can never be equal because you are comparing a date with a datetime.

If I was to write a test for this, it could be:
```php
    function testExecute_paramsGiven_ContributionsAreRewrittenToTo() {
        $this->exec("fill_contributions_stats -f 2016-06-19 -t 2016-06-19");

        $firstContributionStats = $this->ContributionsStats->find()->order(['id' => 'DESC'])->first();
        $this->assertEquals('2016-06-19', $firstContributionStats->date);
    }
```

src/Command/FillContributionsStatsCommand.php
```diff
@@ -61,7 +88,7 @@ public function execute(Arguments $args, ConsoleIo $io)
 
         // Truncate table and Fill
         $contributionsStats = TableRegistry::getTableLocator()->get('ContributionsStats');
-        $contributionsStats->deleteAll([]);
+        $contributionsStats->deleteAll(['date >=' => $from, 'date <=' => $to]);
```
jiru: Same problem here.

src/Template/Contributions/activity_timeline.ctp
```diff
@@ -43,81 +43,131 @@ $selectedMonth = format(
...
         )
-    ); 
+    );
     ?>
 </div>
 
 <div id="main_content">
     <div class="section md-whiteframe-1dp">
-    <h2><?php echo $selectedMonth; ?></h2>
+    <h2><?php echo __('Activity timeline') . ' - ' . $selectedMonth; ?></h2>
```
jiru: Could you also fix this string the same way? We can then get rid of this `$selectedMonth` variable entirely.

jiru:
> We can't stop the trigger because it's in charge of `last_contributions`.

How about we temporarily replace that trigger with something that only updates `last_contributions` but doesn’t touch `contributions_stats`? The migration could go like this:
1. Replace the trigger so that it only updates `last_contributions`.
2. Run your script run on the whole `contributions` table at some point during the day.
3. Wait until 23:59.
4. Replace the trigger with your new trigger code.
5. Wait until midnight.
6. Run your script only for yesterday.

The stats won’t be updated on that day between step 1 and 6 but that’s reasonable. I can run the migration on dev.tatoeba.org to have an estimation on how long step 2 takes so that I can avoid waiting for too much time on step 3.

src/Command/FillContributionsStatsCommand.php
agrodet: MySQL knows how to deal with comparison between a datetime and a string. However, I did make a silly mistake because I forgot that the string will be interpreted as the datetime with 00:00:00. I just need to replace that with $lastDay + 1 day, and it should be okay.

I will update a fix later.

agrodet:
I've pushed some modifications. I removed `$selectedMonth` as you asked. And I replaced the hyphen of the page title by an em dash. I think that gives a better look. I used the HTML code inside code, I don't know if you prefer to directly display the character.

About the trigger, I think the process you describe is good. What about simply starting the trigger with the part that take care of `contributions_stats` commented, and for step 4, simply remove the comments and restart the trigger?

src/Template/Contributions/activity_timeline.ctp
```diff
@@ -64,7 +60,10 @@ if (!empty($stats)) {
 
 <div id="main_content">
     <div class="section md-whiteframe-1dp">
-    <h2><?php echo __('Activity timeline') . ' - ' . $selectedMonth; ?></h2>
+    <h2><?php
+        echo __('Activity timeline') . ' &#8212; ' .
+            format(__('{month} {year}'), array('month' => $monthName, 'year' => $year));
```
jiru: Sorry, I think I wasn’t clear. The whole string should be translatable so that translators can totally and easily rearrange the whole title. Your code forces translators to use the dash character and space and to use the two strings in that order. However, some languages don’t use the dash character or even don’t use the space character. And you can never be sure about the word order. So it should be like this:
```php
echo format(__('Activity timeline — {month} {year}'), array('month' => $monthName, 'year' => $year));
```

jiru: About the dash character, as far as I know [we use hyphen everywhere else in titles](https://github.com/Tatoeba/tatoeba2/blob/d39ecb565a894a661c137302bedbc2eb527e16a0/src/View/Helper/PagesHelper.php#L28), so for consistency it’s probably better to stick with the hyphen for now.

jiru:
> I've pushed some modifications.

Thanks for the update! (and the new tests!) :smile: Please have a look at my new comments.

> What about simply starting the trigger with the part that take care of `contributions_stats` commented?

Sure, I can do that during the migration.

jiru:
@agrodet I updated the activity page of dev.tatoeba.org with your code if you want to see how it look with real data. It only took a few minutes to run your script. :+1:

jiru:
Some of the numbers span on multiple lines on this page:
https://dev.tatoeba.org/eng/contributions/activity_timeline/2018/09
image

agrodet:
Thank you! Good for the scripts :) 
Now that I see it, of course that's a very simple mistake, the minimum width can accommodate only two digits (if you turn the UI to Japanese, some numbers will even be displayed on three lines! 😱). I'll think oof an improvement and try to push it soon.

PR #2182, opened 2020-03-08
Decrement number of sentences in list when a sentence is deleted

### LINKED ISSUE #1277: List Counts Don't Match -- On the List page and on the List of Lists
## Problem

The number of sentences is different on these 2 pages.

https://tatoeba.org/eng/sentences_lists/of_user/CK/page:1/sort:modified/direction:desc

241356

screen shot 2016-08-04 at 10 58 18 am

https://tatoeba.org/eng/sentences_lists/show/4000

241335

screen shot 2016-08-04 at 10 57 56 am
## Note

I noticed this when I imported 552 audio files today.

The count is 21 sentences off.

agrodet:
This PR modifies the `deleteSentence` method of the SentencesTable to actually decrement the value of `sentences_lists.numberOfSentences` by 1, for all lists containing the deleted sentence.  
It also provides a command to correct the values in existing lists. It uses a SELECT COUNT(*) GROUP BY on the `sentences_sentences_lists` table to correct the number of sentences present in the list. It sets this number to 0 for lists that do not appear in `sentences_sentences_lists`. 

These together should fix #1277.

As a bonus, it also fixes one test that used to check admin capabilities while using corpus maintainer id.


src/Command/CorrectNumberOfSentencesCommand.php
```diff
@@ -0,0 +1,43 @@
...
+        // The lists that do not appear have numberOfSentences set to 0
+        $subquery = $this->SentencesSentencesLists->find()->select('sentences_list_id');
+        $listsWithNoSentenceQuery = $this->SentencesLists->find('all', array(
+            'conditions' => array('id NOT IN' => $subquery)
+        ));
+
+        foreach($listsWithNoSentenceQuery as $sentencesList){
+            $sentencesList->numberOfSentences = 0;
+            $this->SentencesLists->save($sentencesList);
+        }
```
jiru: Your code is okay, but if you can, please [use `$query->update()->set(['numberOfSentences' => 0])->where(...)`](https://book.cakephp.org/3/fr/orm/query-builder.html#mettre-a-jour-les-donnees) instead. It is more efficient.

tests/TestCase/Command/CorrectNumberOfSentencesTest.php
```diff
@@ -0,0 +1,43 @@
...
+        parent::tearDown();
+    }
+
+    function testExecute_CorrectNumberIsWritten() {
+        $this->SentencesLists->updateAll(['numberOfSentences' => 99], []);
+
+        $this->exec("correct_number_of_sentences");
+
+        $sentencesLists = $this->SentencesLists->find('all')->toList();
+        $newListsAndCounts = array_combine(array_column($sentencesLists, 'id'), array_column($sentencesLists, 'numberOfSentences'));
```
jiru: Nice! Here is how to do that the CakePHP way:
```php
use Cake\Utility\Hash;
$newListsAndCounts = Hash::combine($sentencesLists, '{n}.id', '{n}.numberOfSentences');
```

src/Model/Table/SentencesTable.php
```diff
@@ -1259,7 +1259,16 @@ public function deleteSentence($id)
...
+        $listsContainingSentence = $this->loadInto($sentence, ['SentencesLists'])->sentences_lists;
+        $isDeleted = $this->delete($sentence);
+        if ($isDeleted) {
+            foreach ($listsContainingSentence as $list){
+                $list->numberOfSentences--;
+                $this->SentencesLists->save($list);
+            }
+        }
+
+        return $isDeleted;
```
jiru: This code works, but it can be improved. First, you can move it to the `afterDelete()` method, which will only be executed if the deletion succeeds, so you don’t have to carry around `$isDeleted`. That would be, let’s say, improvement A.

As you can see, `afterDelete()` is already quite busy doing a lot of things which are not of SentencesTable’s responsibility. Updating the number of sentences in the lists is also not something SentencesTable should have to care about. So we can improve this even more by moving your code into the SentencesLists class and have it executed with some kind of event/triggering system. That would be improvement B, and it’s super good because it reduces [coupling](https://en.wikipedia.org/wiki/Coupling_(computer_programming)) between SentencesTable and SentencesLists. For an example of how to do that, look at `SentencesTable::afterSave()` and `src/Event/ContributionListener.php` (and the unit tests).

Consider improvement B as a bonus. I will still merge your PR if you only do improvement A.

tests/TestCase/Command/CorrectNumberOfSentencesTest.php
```diff
@@ -0,0 +1,43 @@
...
+        $this->assertEquals($newListsAndCounts[1], 3);
+        $this->assertEquals($newListsAndCounts[2], 0);
+        $this->assertEquals($newListsAndCounts[3], 1);
+        $this->assertEquals($newListsAndCounts[4], 2);
+        $this->assertEquals($newListsAndCounts[5], 0);
+        $this->assertEquals($newListsAndCounts[6], 1);
```
jiru: I think you can do
```php
$expected = [
    1 => 3,
    2 => 0,
    ...
];
$this->assertEquals($newListsAndCounts, $expected);
```

jiru:
Thanks a lot for your PR, especially the tests. :+1:

On a side note, could you avoid committing all these changes that consist of removing spaces at the end of the lines? It makes your commits harder to read. I believe your text editor is doing that automatically. There are two ways to go about this:
1. Fix your editor.
2. Use `git add --patch` to only stage relevant code changes into your commit. Similarly, you can do `git checkout --patch` to only discard irrelevant code changes from your work tree.

src/Model/Table/SentencesTable.php
agrodet: Now, I remember why I wrote it like this. If done after deletion, the sentence is already gone so there are no associations between this sentence and the corresponding sentencesLists, hence rows to decrement cannot be fetched.   
If we check the `sentencesList.ids` before deletion, and then pass them to decrement them, might as well directly do it. That's what I thought.

Now that I think again, maybe the easiest trigger is not on `sentences` but on `sentences_sentences_lists`? Because we would have the `list_id` available directly. But I don't know if a similar behavior is possible or how to do it. By that I mean that `sentence` deletion cascades to `sentences_sentences_list` deletion, and that one could cascade to `sentences_list` decrement. 

Any advice? Is it correct to have an afterDelete on the intermediate table?

jiru: I see. Yes, putting your code in the link table model sounds like a good idea, if it works. You’d have to create a method named `afterDelete()` in SentencesSentencesListsTable using the same signature as in SentencesTable. The deleted link should be provided as a second parameter of `afterDelete()`.

agrodet: I had to do a [little research](https://book.cakephp.org/3/en/orm/deleting-data.html#cascading-deletes) about that. It seems that we need to set the `cascadeCallbacks` option of the association to `true` to be able to propagate the call to `afterDelete()` (and other callbacks) of the junction table. The documentation warns us 

> Setting cascadeCallbacks to true, results in considerably slower deletes when compared to bulk deletes. The cascadeCallbacks option should only be enabled when your application has important work handled by event listeners.

but I think it's okay because we don't delete sentences so much and there aren't many associations to delete.

jiru:
Thanks for solving the problem, agrodet! 

After testing your code a last time, I realized that the `modified` is updated by your script, but I will just fix this by myself after merging.

Issue #2183, opened 2020-03-08
Jewish Babylonian Aramaic (tmr)


RyckRichards:
* Language name: Jewish Babylonian Aramaic
* Language info: https://en.wikipedia.org/wiki/Jewish_Babylonian_Aramaic
* ISO 639-3 code: tmr
* Public List: https://tatoeba.org/eng/sentences_lists/show/9706
* Icon: https://imgur.com/a/YAzMdxi 
* Requested by: https://tatoeba.org/eng/user/profile/shekitten

PR #2184, opened 2020-03-08
Added Jewish Babylonian Aramaic

RyckRichards:
This PR solves #2183


trang:
@RyckRichards, the language icon's size is too big. Perhaps you have not resized the PNG to be 30x20 pixels before you turned it into SVG?

RyckRichards:
@trang I'm really sorry. I'm going to fix this.

RyckRichards:
@trang Fixed! Hope it's good enough :)

trang:
No problem @RyckRichards. It is fine now. As you can see, the status is green.

When you see that your commit has a red cross next to it, you can click on the cross, then on the "Details" link. Then it should tell you some information about what's wrong.

jiru:
It looks like the flag [consisted of some written text](https://imgur.com/n653OTw), but now it has been downsized to 30x20, it’s completely unreadable. I don’t think this is a good flag to represent the language.

RyckRichards:
@jiru - I've asked about a new flag suggestion

conlangbecca:
Maybe this would work better? https://imgur.com/a/YAzMdxi

jiru:
Thank you @conlangbecca! I can’t say about the meaning of the text, but it certainly looks better yes. :smiley_cat:
@RyckRichards I think we can use this new flag.

RyckRichards:
Alright!

RyckRichards:
@jiru done

Issue #2185, opened 2020-03-08
Copy button stopped working


Yorwba:
[Reported by AlanF_US on the wall](https://tatoeba.org/eng/wall/show_message/34364#!#message_34364).

PR #2187, opened 2020-03-09
Fix Copy Button

AndiPersti:
This PR addresses #2185


jiru:
Thank you so much for the quick fix, Andreas! :+1:

PR #2190, opened 2020-03-10
Remove old format for Entities

AndiPersti:
There are still a few places in the code which rely on the old CakePHP 2 Entity format.

I don't think converting to that format is necessary and thus the old format should be removed.


jiru:
Thank you very much for the cleanup, Andreas! :sun_with_face:

PR #2197, opened 2020-03-13
Lighten SQL queries and json

jiru:
This PR aims at removing unnecessary fields that are used in SQL queries and serialized as json for the initialization of Angular components.

On the Sentences table, we already had `$this->contain()` and `$this->fields()` to set the fields and containment of queries to display sentence blocks. These methods now take a parameter to return an appropriate value depending on the context: sentence block with or without translations, or sentence page. We can easily add more contexts in the future.

In addition, the new find parameter `hideFields` allows to hide certain fields that we need in the SQL queries but not in json.

I also removed `CurrentUser::hasFavorited()` because that information was already retrieved using contain.

Since this PR is mingling with core stuff, please read and test it as much as you can.

### DIFF
--- a/src/Controller/ActivitiesController.php
+++ b/src/Controller/ActivitiesController.php
@@ -66,7 +66,7 @@ class ActivitiesController extends AppController
             'limit' => CurrentUser::getSetting('sentences_per_page'),
             'conditions' => $conditions,
             'fields' => $this->Sentences->fields(),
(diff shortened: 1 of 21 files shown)


AndiPersti:
Looks good to me.

jiru:
Thank you for your review, Andreas. :smiley:

PR #2200, opened 2020-03-14
Add "Send a PM" icon to wall posts

### LINKED ISSUE #2113: Add the new mail icon to wall posts
Would be nice to add the mail icon added by #2068 to wall posts as well.

agrodet:
This PM closes #2113. I simply copied what was done for the comment menu


jiru:
Looks good! :+1:

On a side note, I don’t really like the tooltip "Send message" because it’s too vague, I think something like  "Contact {username}" that we have on the profile page is better. But at least your code is consistent with the sentence comment button, so you can leave it like that for now.

agrodet:
I added a commit to solve #2201

jiru:
Thanks! :shallow_pan_of_food:

Issue #1558, opened 2018-06-03
Update Terms of Use


trang:

Guybrush88:
maybe also ticket #126 could be kept into consideration when updating the terms of use, imho

PR #2206, opened 2020-03-15
Announcement for coding event

trang:
This PR includes an announcement about the idea of a Tatoeba coding event.

@alanfgh @ckjpn Could you quickly review the text:

> **Tatoeba coding event**
>
> In order to get more developers involved in Tatoeba and make it fun at the same time, we got the idea of organizing a coding event. If this sounds interesting, please fill up [our survey](https://forms.gle/wyLqhcyLZxkiqn1WA).
>
> Until then, if you already wish to get involved, please read our [guide for contributing as a developer](https://github.com/Tatoeba/tatoeba2/wiki/Contributing-as-a-developer) or just [contact us](https://tatoeba.org/eng/contact). We are an open source project and we welcome everyone!

@jiru Note that the announcement will only be displayed if `Announcement.enabled` is set to `true` in the `app_local.php` config file. So this is safe to merge anytime.


alanfgh:
Here's my suggested wording:

> In order to get more developers involved in Tatoeba and have some fun at the same time, we are organizing a coding event. If this sounds interesting, please fill out our survey.
> 
> Until then, if you wish to get involved, please read our guide for contributing as a developer or just contact us. We are an open source project and we welcome everyone!

trang:
Thanks @alanfgh! The changes sound good so I'll go with that. It's a bit more assertive than intended (we are actually not sure yet if we will organize this event, it depends on the results of the survey) but it's not a bad thing.

alanfgh:
I see. You could say "... we are thinking of organizing a coding event" if you decide to be less assertive. :)

ckjpn:
What alanfgh says sounds OK.

jiru:
I had to solve some bugs with the announcement code:

1. When `Announcement.enabled` is set to false, the "please accept new ToS" message isn’t displayed.
2. All the users who registered after the new ToS came into effect (20/01/2019, see #1558) still have the setting `new_terms_of_use` initialized to false.

Because of these two problems combined, users currently aren’t seeing the new ToS prompt even though they still have to accept them, just because we aren’t displaying any other announcement now. In other words, if we enable the announcement of this PR, users who registered after the 20/01/2019 will suddenly be prompted to accept the new ToS, even though they accepted them already when they registered. You can see that problem if you register now on dev.tatoeba.org, it asks you to accept the new ToS because dev.tatoeba.org has an announcement about "everything you submit here will be lost".

So I fixed problem 1 and 2, and I also added a migration to fix the user setting new_terms_of_use of users who registered after 20/01/2019.

This also means that once my fix get into production, users who registered before 20/01/2019 but didn’t get to accept the new ToS yet will suddenly be prompted, so we may have reports from users about that.

Issue #2212, opened 2020-03-18
Okinawan (ryu)


RyckRichards:
* Language name: Okinawan
* Language info: https://en.wikipedia.org/wiki/Okinawan_language
* ISO 639-3 code: RYU
* Public List: https://tatoeba.org/eng/sentences_lists/show/9735
* Icon: https://en.wikipedia.org/wiki/File:Flag_of_Okinawa_Prefecture.svg
* Requested by: https://tatoeba.org/eng/user/profile/Balamax

PR #2219, opened 2020-03-18
Added language Okinawan (ryu)

RyckRichards:
This PR is addressed to #2212 

Language info: https://en.wikipedia.org/wiki/Okinawan_language
Requested by: https://tatoeba.org/eng/user/profile/Balamax


jiru:
@RyckRichards Thank you!

I replaced (e21d88de1e0ad06d53bf8d2cbd165e71db37a22d) your SVG with the [original image from Wikimedia](https://upload.wikimedia.org/wikipedia/commons/2/2f/Flag_of_Okinawa_Prefecture.svg) because it’s smaller and it looks about the same.

Next time, if the original flag image is an SVG image, please check if it’s not better to use it as is in terms of file size and quality.

Issue #2211, opened 2020-03-18
Can't copy text from Safari on iPhone


RyckRichards:
Reported by cueyayotl via PM (He allowed me to post it here):

> Also, when using Safari on my iPhone, I noticed that I can't copy text from this site. I can highlight words, and then click on 'copy' but it doesn't put the words on the clipboard, so I am able to paste them elsewhere. On the other hand, I CAN copy text when I click on 'Edit this sentence' (which is only accessible because of my 'Corpus Maintainer' status here)... I assume normal users are unable to copy text at all. However, this isn't an issue with Google Chrome on my PC at all.

Yorwba:
This is probably related to the copy-paste hack to remove furigana from the selection. I've been the last to touch that code (#2123, #2147), so I'd like to fix this problem, but unfortunately I don't have any Apple devices to test this. [CanIUse lists support for clipboard data access on iOS Safari as "unknown"](https://www.caniuse.com/#feat=mdn-api_window_copy_event_clipboarddata), so maybe it's related to that.

@cueyayotl Does the copy button at the right-hand side of each sentence work? It would at least allow you to copy entire sentences.

cueyayotl:
Ah, I didn't realize there was a copy button at the right-hand side of each sentence. But yes, the button does work; I'm just not able to copy individual words right now.

trang:
@cueyayotl Is it a problem on the new sentence design?

jiru:
@cueyayotl Could you try if you can copy text from Safari on https://dev.tatoeba.org/ now?

jiru:
I assume the problem is solved. Feel free to re-open this issue if it isn’t.

PR #2220, opened 2020-03-18
More fixes for the furigana copy-paste hack

Yorwba:
I think that #2211 was most likely caused by `setData` being unsupported in Safari on iOS ([according to MDN](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/setData), it's the only modern browser that doesn't support this API). This PR checks whether the function is present before calling it and that hopefully should allow people using iOS to copy things again.

While testing in Firefox, I noticed a different problem, which is that when copying from an `input` element (I was using the search bar), `toString()` on the selection sometimes returns an empty string even though the selection is not empty. I'm not sure what causes this. In the debugger I could see that the selection markers on the `input` element were set, so it might be a browser bug that it returned the wrong selection content. Anyways, since it doesn't seem to affect other elements and we have no ruby tags in `input` elements, ignoring those should be an acceptable workaround.

### DIFF
--- a/webroot/js/generic_functions.js
+++ b/webroot/js/generic_functions.js
@@ -156,7 +156,8 @@ $(document).on('copy', function (e) {
 
     var sel = window.getSelection();
     var clipboardData = e.originalEvent.clipboardData; // not available in IE
-    if (sel.rangeCount > 0 && clipboardData) {
+    var setData = clipboardData && clipboardData.setData; // not available in iOS Safari
+    if (sel.rangeCount > 0 && setData) {
         $('rt').css('visibility', 'hidden');
         clipboardData.setData('text', sel.toString());
         $('rt').css('visibility', 'visible');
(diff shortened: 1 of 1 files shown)


Yorwba:
Update on the second part: it's not a browser bug, but documented behavior. [`getSelection` doesn't work for `<input>` and `<textarea>` elements in Firefox](https://developer.mozilla.org/en-US/docs/Web/API/Window/getSelection#Related_objects). I added a comment to the code linking to the documentation.

Now I just don't understand why I couldn't reliably trigger it. It never works on a page with just a single `<input>` element, but during testing on my local Tatoeba instance, I could get the selection more often than not. Maybe some of the extra markup Angular adds masks the problem somehow.

jiru:
Thank you Yorwba. I’m merging your PR for the sake of Safari users (and because I trust you), but this code is getting really ugly now. Maybe I’m just not used to see cross-browser front-end code under the hood, but I really wish we could find a cleaner, less error-prone method to prevent copy-pasting furigana.

Yorwba:
I don't usually write front-end code either, so it's totally possible that there's an elegant solution and I simply missed it. But we're kind of fighting the browser's default behavior of copying exactly what the user sees. I wouldn't be surprised if `setData` is intentionally disabled on iOS to protect the user against malicious modification of the clipboard content, but I didn't find any official Apple documentation that explains it.

Issue #1087, opened 2016-04-01
Blended furigana hides search results highlighting


jiru:
See [this example](https://tatoeba.org/jpn/sentences/search?query=%22%E5%BD%93%E7%A4%BE%22&from=jpn&to=und). While searching for 当社, sentences with furigana on a separate line have 当社 highlighted, while the ones with furigana line and sentence blended do not.

ckjpn:
Note that it's not only the furigana sentences ....

https://github.com/Tatoeba/tatoeba2/issues/1272#issuecomment-235474226

Issue #952, opened 2015-12-11
Change the colors of the "editable transcription" icon. (Reverse the colors.)


ckjpn:
Reversing the colors would likely make this icon
easier to read and would also make this icon
blend in better with the other icons.

![screen shot 2015-12-11 at 11 45 01](https://cloud.githubusercontent.com/assets/7934273/11734701/135c323e-9ffd-11e5-9709-ee7eb43ab232.png)

halfdan:
The icon needs some improvement - I tried inverting the colours, but this makes the icon barely visible. The two characters are so tiny that you can barely make out that there's an icon.

ckjpn:
Maybe just including one character would help.

か for kana
or
ふ for furigana

PR #2224, opened 2020-03-20
Transcriptions in the new sentence design

trang:
This PR implements the transcriptions in the new sentence design. 

For the major part, it works the same as in the old design.

- Transcriptions are displayed in grey below the sentence (or translation), except furigana which works a bit differently and is displayed on the sentence itself.
- If the transcription is generated by a script that is prone to mistakes, a warning icon is displayed in front of the transcription.
- When a transcription has a warning icon, it is not displayed by default. But enabling the option "Always show transcriptions and alternate script" in the Settings will make these transcriptions displayed all the time.
- The permissions on editing transcriptions remain the same:
  - Every contributors can edit transcriptions of their own sentences.
  - Advanced contributors can edit unreviewed transcriptions of other people's sentences.
  - Advanced contributors can re-edit transcriptions as long as the sentence owner didn't edit the transcriptions themselves.
  - Corpus maintainers can edit any transcription.

**Things that changed**

- There were various custom icons used to identify the type of transcription. I felt it was a bit too complicated and used a more generic icon (the [subdirectory_arrow_right](https://material.io/resources/icons/?icon=subdirectory_arrow_right&style=baseline) material icon). This makes #952 obsolete.
- The hidden unreviewed transcriptions are displayed when expanding the sentence menu rather than by clicking some transcription button in the sentence menu.
  - The transcription icon would have been part of the hidden icons in the menu otherwise. Instead of having to expand the menu then click this transcription icon, I figured we could just display the hidden transcription directly after clicking the "Expand menu" icon.
- On the search page, when the main sentence has furigana, the furigana will always be displayed on a cloned sentence below the actual sentence. This is a temporary solution for #1087.
- When a transcription could not be saved, an error message is displayed in red below the text input.
- The warning tooltips are displayed when hovering over the transcription rather than when hovering over the warning icon.
  - There was a bit of redundancy between the info tooltip (on the transcription) and the warning tooltip (on the warning icon), so I just factorized it into the info tooltip.

**Todo for another time**

- Find a long term solution for #1087, to make it possible to have search highlight and furigana at the same time.
- Non-authenticated users currently still have no way to see unreviewed transcriptions. They used to be able to reveal these transcriptions by clicking the transcription icon in the sentence menu. I'm still thinking of what is the best way to integrate this in the new design.


jiru:
I’m very happy about that! :tada: 

I havn't reviewed the code yet so I'll just comment on the visible part first.

> Transcriptions are displayed in grey below the sentence (or translation), except furigana which works a bit differently and is displayed on the sentence itself.

It looks like your implementation is a little bit different. Consider a Japanese sentence with reviewed furigana. I have the option "Always show transcriptions" turned on. I go on the sentence page. In the old design, the furigana appears on the top of the sentence. In the new design, it appears below.

Another problem: when the furigana is on the top, I can’t see figure out how to edit the transcription.

> There were various custom icons used to identify the type of transcription. I felt it was a bit too complicated and used a more generic icon (the subdirectory_arrow_right material icon). This makes #952 obsolete.

I think it’s okay. I realized in UX test 3 that some of these icons needed to be changed anyway.

> Non-authenticated users currently still have no way to see unreviewed transcriptions

That would probably annoy learners of Japanese who don’t have account.

> shot

The transcriptions button on the right are easily seen as belonging to a single group. It is not very intuitive which button is acting on what. Probably gray background on hover makes it even more non intuitive. I guess it’s a similar feeling to what agrodet mentioned about eye-matching the (i) icon with the sentence text.

I have no idea right now on how to improve that but I’ll think about it.

> shot2

I think I would look more consistent if the transcription text (if reviewed) is aligned with the sentence text.

trang:
> It looks like your implementation is a little bit different. Consider a Japanese sentence with reviewed furigana. I have the option "Always show transcriptions" turned on. I go on the sentence page. In the old design, the furigana appears on the top of the sentence. In the new design, it appears below.

This should be fixed.

> The transcriptions button on the right are easily seen as belonging to a single group.

The best solution I have for this is to combine the editing of the transcription with the editing of the sentence. I had already considered doing it since the beginning but it complexifies the "Edit" feature so I was a bit reluctant to go for it. But I also really don't like having these extra buttons next to the transcription text.

One consequence is that it's no longer possible to edit the transcriptions of translations. I think users will be okay with that. They cannot edit the translations text so it shouldn't feel too annoying that they can no longer edit the transcriptions of translations. They probably rarely did it anyway.

One other consequence is that it takes more steps to mark a transcription as reviewed. It requires to first click on the "Edit" button then click on the check icon next to the transcription in the sentence form. But I'd rather go for a longer term solution here rather than try to squeeze one more button in the sentence component view. The ideal solution, I think, would be to have page "Review transcriptions" where people can efficiently proofread transcriptions.

My implementation does not take into account yet that there can be several editable transcriptions for a same sentence. It will probably have to be implemented as part of #2189.

> Non-authenticated users currently still have no way to see unreviewed transcriptions

I have added a button for non-authenticated users. For lack of a better idea, I reused the same button as for expanding or collapsing the sentence menu.

> I think I would look more consistent if the transcription text (if reviewed) is aligned with the sentence text.

This should be fixed.

jiru:
Good job! I think you solved the buttons problem in a smart way. :smile_cat: 

I’m still commenting the visible part only.

Problem: After I click the review button of a transcription, the tooltip of that transcription says that it has been generated by a piece of software. Only after I reload the page, the tooltip correctly says "last edited...".

Problem: On a Japanese sentence, I click on the edit button, I modify both the sentence text and the transcription in a way that makes the transcription invalid. I click save, the sentence is saved, but the transcription isn’t saved and the edition box stays open, showing the autogenerated transcription. Instead, it should show me an error and prefill the transcription field with what I last entered.

By the way, do we want to allow editing both sentence and transcription? I can imagine a scenario when the autogenerated transcription is known to be wrong and the user wants to both edit the sentence and the transcription to avoid fixing every single place the autogenerated transcription would be wrong after editing the sentence. Your current implementation allows that, which I think is great.

> I have added a button for non-authenticated users. For lack of a better idea, I reused the same button as for expanding or collapsing the sentence menu.

I think it’s good enough. :+1: On a side note, I think this change may make the "always show transcriptions" setting less useful, because people could now quickly show all transcriptions on a page by clicking on "expand all menus for all sentences". Since this "expanded state" is remembered across pages, it could almost replace the "always show transcriptions" setting. I am not sure we want users to use the "expand" feature in this way though.

jiru:
The position of the tooltips feels a off on [this sentence](https://dev.tatoeba.org/fra/sentences/show/5386889). The tooltip over the simplified characters is right on the middle of the text, however on the pinyin it’s off on the right. Since the pinyin spans on two lines, the tooltip is positioned on the center of the card, even though there is quite some blank on the right.

trang:
I fixed the two problems you mentioned.

> By the way, do we want to allow editing both sentence and transcription?

Yes, we definitely want to allow it. Editing the transcription is a way to mark it as reviewed.

- If I edit the sentence but not the transcription → the transcription is auto-generated to match the new sentence text and it will not be marked as reviewed.
- If I edit the sentence and the transcription → the transcription is saved and marked as reviewed.

This way of marking the transcription as reviewed is not necessarily ideal. If I make significant changes on the sentence, I may want to first get the auto-generated transcription and then apply the necessary fixes on top of it. But that can be an improvement for another time.

> On a side note, I think this change may make the "always show transcriptions" setting less useful, because people could now quickly show all transcriptions on a page by clicking on "expand all menus for all sentences". Since this "expanded state" is remembered across pages, it could almost replace the "always show transcriptions" setting. I am not sure we want users to use the "expand" feature in this way though.

It's still a bit too early to tell but I think I wouldn't be against using the "expand" feature in this way.

> The position of the tooltips feels a bit off on [this sentence](https://dev.tatoeba.org/fra/sentences/show/5386889).

I changed the position of the tooltip to be above instead of below the transcription.

src/Controller/TranscriptionsController.php
```diff
@@ -127,10 +136,20 @@ public function save($sentenceId, $script) {
             );
         }
 
-        $this->setViewVars(array_filter(array($saved)), $sentenceId);
-        $this->render('view');
         if (!$saved) {
-            return $this->response->withStatus(400, 'Bad transcription');
+            return $this->response->withStatus(400, __('The transcription could not be saved.'));
```
jiru: :smile: Well, it is certainly not common practice to use the status line as an error message displayed to users, but nothing’s really preventing you from doing it. This message is typically not translated and only read by developers (API users). I assume that this is temporary and that you are aware that the Transcription model is already providing detailed error messages.

jiru:
Could you make PinyinHelper and TranscriptionsHelper use your new PinyinTrait and FuriganaTrait so that we avoid keeping duplicate code?

src/Model/Entity/Transcription.php
```diff
@@ -140,4 +154,80 @@ private function getSourceScript($sourceLang) {
         }
         return false;
     }
+
+    protected function _getHtml() {
+        if (!$this->sentence) {
+            return;
+        }
+
+        $lang = $this->sentence['lang'];
```
jiru: Just in case, you should also check for `$this->sentence['lang']`. And it’s clearer to use `$this->sentence->lang`. So something like:
```php
if (!$this->sentence || !$this->sentence->lang) {
    return;
}
```

jiru: And also `$this->script` and `$this->text`.

src/Model/Entity/Transcription.php
```diff
@@ -140,4 +154,80 @@ private function getSourceScript($sourceLang) {
...
+            $text = $this->rubify($this->text);
+        } elseif ($lang == 'cmn' && $this->script == 'Latn') {
+            $text = $this->numeric2diacritic($text);
+        }
+
+        return $text;
+    }
+
+    protected function _getEditingFormat() {
+        if (!$this->sentence) {
```
jiru: Same here.

src/Model/Entity/Transcription.php
```diff
@@ -140,4 +154,80 @@ private function getSourceScript($sourceLang) {
...
+            $text = htmlentities($this->text);
+            if ($this->script == 'Hrkt') {
+                $text = $this->bracketify($this->text);
+            }
+        }
+
+        return $text;
+    }
+
+    protected function _getInfoMessage() {
```
jiru: Same here, better check if we have `$this->needsReview` and `$this->type`.

src/Model/Entity/Transcription.php
```diff
@@ -140,4 +154,80 @@ private function getSourceScript($sourceLang) {
...
+            } else {
+                $message = __(
+                    'This transcription may contain errors because '.
+                    'it has been generated by a piece of software that '.
+                    'is not perfect.'
+                );
+            }
+        } else {
+            if (isset($this->user->username)) {
+                if ($this->type == 'altscript' && $this->sentence->lang == 'jpn') {
```
jiru: Check if `isset($this->sentence->lang)` before using it.

src/Model/Entity/Transcription.php
```diff
@@ -140,4 +154,80 @@ private function getSourceScript($sourceLang) {
...
+        if ($this->script == 'Hrkt') {
+            $text = $this->rubify($this->text);
+        } elseif ($lang == 'cmn' && $this->script == 'Latn') {
+            $text = $this->numeric2diacritic($text);
+        }
+
+        return $text;
+    }
+
+    protected function _getEditingFormat() {
```
jiru: Consider renaming "editing format" to "[markup](https://en.wikipedia.org/wiki/Markup_language)".

src/Template/Element/sentences/transcription_button.ctp
```diff
@@ -0,0 +1,13 @@
+<div class="menu-wrapper" ng-if="vm.hasHiddenTranscriptions">
+    <div class="menu" layout="row" layout-align="space-between center">
+        <md-button ng-if="!vm.isMenuExpanded" class="md-icon-button" ng-click="vm.toggleMenu()">
+            <md-icon>unfold_more</md-icon>
+            <md-tooltip><?= __('Show all transcriptions') ?></md-tooltip>
```
jiru: This message is used both in the context of multiple transcriptions (like Chinese) as well as single transcription (like Japanese). It would be best to have a pluralized message, but in the mean time, it may be better not to put so much emphasis on the plural by just saying "Show transcriptions" instead.

jiru: Or maybe "Show unreviewed transcriptions", to match the opposite message for unfolding.

src/Template/Element/sentences/transcriptions.ctp
```diff
@@ -0,0 +1,28 @@
...
+use App\Model\CurrentUser;
+
+if (!isset($sentenceVar)) {
+    $sentenceVar = 'vm.sentence';
+}
+
+if (CurrentUser::getSetting('show_transcriptions')) {
+    $ngIf = 'ng-if="!'.$sentenceVar.'.furigana || '.$sentenceVar.'.highlightedText"';
+} else {
+    $ngIf = 'ng-if="(vm.isMenuExpanded || !transcription.needsReview) && (!'.$sentenceVar.'.furigana || '.$sentenceVar.'.highlightedText)"';
```
jiru: Theses strings are hard to read. In particular, it’s hard to tell which dot is used to concatenate and which dot is part of the string. I suggest you rewrite it like this:
```html
<?php
$ngIf = "!$sentenceVar.furigana || $sentenceVar.highlightedText";
?>
<div ng-if="<?= $ngIf ?>">
```

jiru:
Thanks for the quick update, Trang. Please have a look at my new comments. :point_up_2: 

I think your implementation is feature-complete, except that the detailed error messages are not displayed when an invalid transcription is submitted. I think these detailed error messages are quite necessary because a number of autogenerated transcriptions are invalid already (in particular Japanese sentences that include numbers). It means an error can be produced just by clicking on the "Mark as reviewed" button.

trang:
The comments have been addressed 🙂

src/Controller/TranscriptionsController.php
```diff
@@ -84,18 +84,13 @@ public function reset($sentenceId, $script) {
...
-            $this->loadComponent('RequestHandler');
-            $this->set('_serialize', ['result']);
-            $this->RequestHandler->renderAs($this, 'json');
+            return $this->returnTranscriptionData($saved, $sentenceId, $script);
         } else {
+            if (!$saved) {
+                return $this->response->withStatus(400, 'Bad transcription');
+            }
             $this->setViewVars(array_filter(array($saved)), $sentenceId, $sentence);
             $this->render('view');
```
jiru: Isn’t this going to prevent `$this->setViewVars()` and` $this->render()` from being executed if the transcription cannot be reset?

webroot/js/directives/sentence-and-translations.dir.js
```diff
@@ -548,19 +548,19 @@
...
 
+            vm.inProgress = true;
             $http.post(url, data).then(function(result) {
-                sentence.transcriptions.push(result.data.result);
+                transcription = result.data.result;
                 hide('sentence_form');
             }, function(error) {
-                transcription.error = error.statusText;
-                sentence.transcriptions.push(transcription);
+                transcription.error = error.data[0];
```
jiru: You should check the existence of `error.data[0]` first. This function can be executed in other contexts where no error is return, such as http 500, 401 etc.

jiru: You are only displaying the first error, however there might be more than one error in the array. [See how the legacy code displays transcription errors](https://github.com/Tatoeba/tatoeba2/blob/dev/src/Template/Transcriptions/view.ctp#L21-L30).

jiru:
Thank you, Trang! :+1:
I left a few other comments. We are almost there! :relaxed:

src/Controller/TranscriptionsController.php
trang: Yes, but it's how things are done in the current code:
https://github.com/Tatoeba/tatoeba2/blob/dev/src/Controller/TranscriptionsController.php#L87

I'm not entirely sure of the impacts of moving the `if` block at the end so I kept it like that.

jiru: I see. Never mind my comment.

jiru:
Thanks for the update, Trang. I think the error handling part may be improved a bit but I’ll merge it now so that UI translators can start working already. We can continue to work on it in the dev branch.

Issue #2233, opened 2020-03-29
Old French (fro)


RyckRichards:
* Language name: Old French
* Language info: https://en.wikipedia.org/wiki/Old_French
* ISO 639-3 code: FRO
* Public List: https://tatoeba.org/eng/sentences_lists/show/9800
* Icon: https://upload.wikimedia.org/wikipedia/commons/1/19/Flag_of_France_%28XII-XIII%29.svg
* Requested by: https://tatoeba.org/eng/user/profile/Julien_PDC

PR #2234, opened 2020-03-29
Added language Old French

RyckRichards:
This PR solves #2233 

Language: Old French


jiru:
I think the we can improve the flag.

First, there is an [SVG version](https://upload.wikimedia.org/wikipedia/commons/1/19/Flag_of_France_%28XII-XIII%29.svg) that is more lightweight (I highgly suspect the flag on touslesdrapeaux.xyz just copied it from there).

Second, it is originally squared-shaped, but in your 30x20 version, it is horizontally squizzed so the proportions are not respected. However I’m not sure if it’s okay to crop a few lines instead.

For comparison, for the [flag of middle French](https://commons.wikimedia.org/wiki/File:Flag_of_France_(XIV-XVI).svg) (also on [touslesdrapeaux.xyz](https://touslesdrapeaux.xyz/france_renaissance_XIV_XVI.png)). On Tatoeba, it looks like we [simplified it](https://tatoeba.org/img/flags/frm.svg) so it looks more readable at small size.

RyckRichards:
@jiru Thanks a lot. I've fixed this

jiru:
@RyckRichards I don’t see any difference in the new flag. What did change exactly?

jiru:
I tried to rework the flag by removing the first and last line so that it fits the 30x20 proportions.

@RyckRichards Could you ask the language requester (Julien) if this would be okay to represent Old French?

My proposed flag:
fro small

The flag with proportion problems (compare with the [original file](https://upload.wikimedia.org/wikipedia/commons/1/19/Flag_of_France_%28XII-XIII%29.svg)):
ricardo flag

My proposed flag’s high resolution for reference:
fro

RyckRichards:
Alright, I'm going to ask him.

RyckRichards:
@jiru, he agreed on use this flag

RyckRichards:
@jiru I made all changes needed. Is everything OK now or should I change anything else on my side? Thank you :)

jiru:
It’s okay, thank you!

PR #2240, opened 2020-04-01
Correctly align reply block of a wall message

### LINKED ISSUE #2196: a reply to a comment is not nested at the right level until browser is refreshed
When I reply to a comment (say, on the Wall), my reply is displayed at the same level as the comment instead of one level below it. 

well_put_1

Only when I refresh the browser is it displayed at the correct level:

well_put_after_refresh

agrodet:
This PR fixes #2196


jiru:
Thank you agrodet :slightly_smiling_face:

PR #2241, opened 2020-04-02
Better display of some checkboxes

### LINKED ISSUE #2134: Make clicks on label change checkbox value
Let's take the advanced search form, for example. I think that clicking on "Reverse order" should modify the checkbox value. "Owned by a self-identified native" has a similar behavior, and probably other `md-checkbox`, I haven't check.

See this piece of code
https://github.com/Tatoeba/tatoeba2/blob/f070d7d4b574f8609b8332d70204741243f77aa6/src/Template/Element/advanced_search_form.ctp#L318-L334

agrodet:
This PR modifies three checkboxes by removing the `<label>` text and putting the corresponding text directly inside the `<md-checkbox>` tag. The three checkboxes involved are the "Remember me" on the login page, and "Owned by a native" and "Reverse order" of the advanced search.

If I didn't miss any other checkbox that would need a modification, this PR would close #2134

### DIFF
--- a/src/Template/Element/advanced_search_form.ctp
+++ b/src/Template/Element/advanced_search_form.ctp
@@ -112,7 +112,7 @@ echo $this->Form->create('AdvancedSearch', [
                     <?= __('Unapproved sentences are likely to be incorrect.') ?>
                 </div>
             </div>
-            
+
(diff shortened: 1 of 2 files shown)


jiru:
Thank you :smile:

PR #2242, opened 2020-04-02
Allow admins to set a sentence's license to a special value

AndiPersti:
This PR addresses #1659 

The first commit refactors the LicenseHelper.

The rest implement the first three points of the suggested solution. (The last one was already implemented).

### DIFF
--- a/src/Controller/SentencesController.php
+++ b/src/Controller/SentencesController.php
@@ -30,6 +30,7 @@ use App\Controller\AppController;
 use App\Model\CurrentUser;
 use App\Lib\LanguagesLib;
 use App\Lib\SphinxClient;
+use App\Lib\Licenses;
 use Cake\Core\Configure;
 use Cake\Database\Expression\QueryExpression;
 use Cake\Event\Event;
@@ -1130,15 +1131,16 @@ class SentencesController extends AppController
         } else {
             $errors = $sentence->getError('license');
             $savedSentence = $this->Sentences->save($sentence);
+            $licenseName = Licenses::allLicenses()[$newLicense]['name'] ?? $newLicense;
             if ($savedSentence) {
                 $this->Flash->set(format(
-                    __('The license of the sentence has been changed to “{newLicense}”.'),
-                    compact('newLicense')
+                    __('The license of the sentence has been changed to “{licenseName}”.'),
+                    compact('licenseName')
                 ));
             } elseif (!empty($errors)) {
                 $message = format(
-                    __('Unable to change the license to “{newLicense}” because:'),
-                    compact('newLicense')
+                    __('Unable to change the license to “{licenseName}” because:'),
+                    compact('licenseName')
                 );
                 $params = compact('errors');
                 $this->Flash->set($message, compact('params'));
--- a/src/Controller/SentencesController.php
+++ b/src/Controller/SentencesController.php
@@ -1131,7 +1131,7 @@ class SentencesController extends AppController
         } else {
             $errors = $sentence->getError('license');
             $savedSentence = $this->Sentences->save($sentence);
-            $licenseName = Licenses::allLicenses()[$newLicense]['name'] ?? $newLicense;
+            $licenseName = Licenses::getSentenceLicenses()[$newLicense]['name'] ?? $newLicense;
             if ($savedSentence) {
                 $this->Flash->set(format(
                     __('The license of the sentence has been changed to “{newLicense}”.'),
--- a/src/Controller/SentencesController.php
+++ b/src/Controller/SentencesController.php
@@ -1124,15 +1124,14 @@ class SentencesController extends AppController
         }
 
         $sentence = $this->Sentences->get($sentenceId);
-        $sentence = $this->Sentences->patchEntity($sentence, ['license' => $newLicense]);
 
         if (!CurrentUser::canEditLicenseOfSentence($sentence)) {
             $this->Flash->set(__('You are not allowed to change the license of this sentence.'));
         } else {
+            $sentence = $this->Sentences->patchEntity($sentence, ['license' => $newLicense]);
             $errors = $sentence->getError('license');
-            $savedSentence = $this->Sentences->save($sentence);
             $licenseName = Licenses::getSentenceLicenses()[$newLicense]['name'] ?? $newLicense;
-            if ($savedSentence) {
+            if (empty($errors) && $this->Sentences->save($sentence)) {
                 $this->Flash->set(format(
                     __('The license of the sentence has been changed to “{newLicense}”.'),
                     ['newLicense' => $licenseName]
--- a/src/Controller/SentencesController.php
+++ b/src/Controller/SentencesController.php
@@ -1134,13 +1134,13 @@ class SentencesController extends AppController
             $licenseName = Licenses::allLicenses()[$newLicense]['name'] ?? $newLicense;
             if ($savedSentence) {
                 $this->Flash->set(format(
-                    __('The license of the sentence has been changed to “{licenseName}”.'),
-                    compact('licenseName')
+                    __('The license of the sentence has been changed to “{newLicense}”.'),
+                    ['newLicense' => $licenseName]
                 ));
             } elseif (!empty($errors)) {
                 $message = format(
-                    __('Unable to change the license to “{licenseName}” because:'),
-                    compact('licenseName')
+                    __('Unable to change the license to “{newLicense}” because:'),
+                    ['newLicense' => $licenseName]
                 );
                 $params = compact('errors');
                 $this->Flash->set($message, compact('params'));
--- a/src/View/Helper/LicenseHelper.php
+++ b/src/View/Helper/LicenseHelper.php
@@ -66,11 +66,17 @@ class LicenseHelper extends AppHelper
     /**
      * Get the options for a selection control
      *
+     * @param boolean $admin  Whether to get the options for regular users
+     *                        or for an admin
+     *
      * @return array
      **/
-    public function getLicenseOptions() {
+    public function getLicenseOptions($admin = false) {
         foreach ($this->validLicenses as $license) {
-            $keyToName[$license] = $this->getLicenseName($license, false);
+            if (!isset($this->licenses[$license]['admin_only']) ||
+                ($this->licenses[$license]['admin_only'] && $admin)) {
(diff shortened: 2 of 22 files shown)


src/Controller/SentencesController.php
```diff
@@ -1130,15 +1131,16 @@ public function edit_license()
...
-                    compact('newLicense')
+                    __('The license of the sentence has been changed to “{licenseName}”.'),
+                    compact('licenseName')
                 ));
             } elseif (!empty($errors)) {
                 $message = format(
-                    __('Unable to change the license to “{newLicense}” because:'),
-                    compact('newLicense')
+                    __('Unable to change the license to “{licenseName}” because:'),
+                    compact('licenseName')
```
jiru, line 1143: Could you change this so that the translatable strings are not modified? Only the placeholder name is changed, so UI translators will have to add a new translation for no reason.

src/View/Helper/LicenseHelper.php
```diff
@@ -20,75 +20,76 @@
...
+    protected $licenses = [];
+
+    /* The licenses used in the current context (sentences or audio files) */
+    protected $validLicenses = [];
 
-    public function __construct(View $view, $config = array()) {
-        assert(isset($config['availableLicences']));
-        parent::__construct($view, $config);
-        $this->licenses = array(
-            '' => array('name' => __('No license for offsite use')),
```
jiru, line 37: The audio license string `''` and its name "No license for offsite use" need to be kept because they are currently used for audio recordings which have been contributed without an explicit license. When this is the case, the audio page of a user looks like that for example: https://tatoeba.org/eng/audio/of/dcampbell

A bit of history. In the past, there were no "Audio" block on the sentence page, and no `audios` table. Just a `has_audio` column in the `sentences` table and links to mp3 files on the speaker icon. So a lot of audio have been initially added at a time where we didn’t really think about licensing. Only later I changed it the way it is now, in order give audio contributors more recognition for their work. Still now, it looks like some audio contributors are not choosing a license. I believe it’s because they need to choose a license by themselves from that page, and a lot of them don’t or don’t know or don’t care about the license.

jiru:
Thanks for your contribution, Andreas. It’s always a pleasure to read your code. :sparkles:

src/Controller/SentencesController.php
AndiPersti: Sorry, I've forgot about the translators.

src/View/Helper/LicenseHelper.php
AndiPersti: I've written that version under the premise that we come up with a good name for the empty license string for both sentences and audio files as [discussed in the issue](https://github.com/Tatoeba/tatoeba2/issues/1659#issuecomment-606251828). But it looks like it is probably better to don't mix them.

Since we use completely different licenses for sentences and audio files, my new version will use two different arrays for them.

> I believe it’s because they need to choose a license by themselves from that page, and a lot of them don’t or don’t know or don’t care about the license.

I've noticed that with the current version from the dev branch I can't choose the option "No license for offsite use" in the selection because the FormHelper adds the `required` attribute to the `select` element which doesn't allow to choose an option with the empty string value. Can you confirm this? Is that on purpose?

jiru: No it’s not on purpose. I confirm there is a problem with license selection box.

jiru:
Thanks for the update.

Even though the option "Licensing issue" is hidden to regular users, they can still switch their license to "Licensing issue" by sending a custom request. While it’s not a big deal, it would be better for the UI to match the actual model validation rules. Could you edit `SentencesTable::canSwitchLicense()` is such a way that only admins can switch to "Licensing issue"?

src/View/Helper/LicenseHelper.php
AndiPersti: I've found out why the `required` attribute is added. The FormHelper uses the table validation rules (when in entity context) in order to decide whether to add it or not:
https://github.com/Tatoeba/tatoeba2/blob/24fb3bb9206e04e6899a41a247b51d155cfb3de7/src/Model/Table/UsersTable.php#L139-L141
After adding `->allowEmptyString('audio_license')` there's no `required` attribute any more.

So should all users be able to choose `No license for offsite use` or just admins?

AndiPersti:
> Could you edit `SentencesTable::canSwitchLicense()` is such a way that only admins can switch to "Licensing issue"?

Instead of editing `canSwitchLicense` I thought that using a validation rule is the better way.

At first I've tried to filter the valid licenses based on the current user and use that in the `inList` validation rule. But I've found out that whenever a validation set for a field allows an empty value for that field, the custom rules aren't called any more.
So I've modified the `allowEmptyString` rule to use a callback function instead of allowing empty strings in all cases.

Test included :-)

src/View/Helper/LicenseHelper.php
jiru: All users can choose `No license for offsite use` for their audio contribution, that’s the current behaviour.

jiru:
Thank you, Andreas, and sorry for the delay. I tested your PR and I found some other problems.

1. After an admin switched somebody’s sentence to "Licensing issue", then, if the user is allowed to switch license, she can just switch to a CC license by herself, effectively removing the "licensing issue".

2. The sentence logs show an empty entry as for "Licensed switched to 'licensing issue'":
image

AndiPersti:
Thanks for the review.

> After an admin switched somebody’s sentence to "Licensing issue", then, if the user is allowed to switch license, she can just switch to a CC license by herself, effectively removing the "licensing issue".

I don't think we should display the form at all in that case, should we? So I've added a new condition to `CurrentUser::canEditLicenseOfSentence`.
I have also refactored `SentencesController::edit_license` a little bit. After getting errors from validating the patched entity, saving it will always fail. So we should only try to save an error-free entity.

> The sentence logs show an empty entry as for "Licensed switched to 'licensing issue'":

Fixed.

jiru:
Thanks for the quick update, Andreas. I think it’s more or less ready for merge. The only thing I’m concerned about is that it’s still possible to add translations to a sentence with licensing issues, at least the following ways:
1. Locally editing the HTML or forging an HTTP request.
2. Adding an original sentence A, and then adding a duplicate of the sentence with licensing issues L as a translation of A. A gets linked to L. In that case, A is an added as original sentence so it’s not supposed to have anything to do with L. They just "happen" to be translations of each other. But this can be abused to circumvent the rule.
3. Same as 2 but by editing a sentence into a copy of L and letting the deduplication bot merge it.

That said, maybe we can ignore this if we consider that the gain of circumventing that rule is close to zero and that admins or corpus maintainers can deal with such abuse manually.

AndiPersti:
> Locally editing the HTML or forging an HTTP request.

Preventing this case is rather simple by adding an application rule which disallows adding translations to sentences with licensing issues. See 6364972.

> Adding an original sentence A, and then adding a duplicate of the sentence with licensing issues L as a translation of A. A gets linked to L. In that case, A is an added as original sentence so it’s not supposed to have anything to do with L. They just "happen" to be translations of each other. But this can be abused to circumvent the rule.

So we should not only prevent translations of sentences with licensing issues but also linking to them?
 
> Same as 2 but by editing a sentence into a copy of L and letting the deduplication bot merge it.

Two possible solutions for that case:
1. It shouldn't be possible to edit a sentence to a duplicate at all. (We currently don't check for duplicates after editing but it looks like there was [such a check in the past](https://github.com/Tatoeba/tatoeba2/pull/1311). Do you know why this changed?)
2. Horus shouldn't be allowed to link to sentences with licensing issues (i.e. fall back to case 2)

trang:
> So we should not only prevent translations of sentences with licensing issues but also linking to them?

I think linking can be allowed. We will just have to investigate the cases manually.

- Let's say you create a sentence in German as an original sentence.
- I copy-paste a sentence in French from somewhere.
- An admin considers that my sentence has licensing issues and sets the license as such.
- Someone realizes that my French sentence is a translation of your German sentence and links them.
- Two possible outcomes:
  - My copy-pasted sentence can transition to CC BY since it can be considered as a translation of an original sentence within Tatoeba.
  - Or we realize your sentence was actually not a legit original sentence and we change the license to `Licensing issue`.

> It shouldn't be possible to edit a sentence to a duplicate at all. (We currently don't check for duplicates after editing but it looks like there was such a check in the past. Do you know why this changed?)

This actually didn't change. It was implemented like that (cf. https://github.com/Tatoeba/tatoeba2/pull/1327#issuecomment-252000774).

It's because of this use case:
- I translate a sentence and make a typo.
- I correct the typo but this action makes my translation become a duplicate.
- If I'm not able to edit, then I would have to ask an admin to delete my sentence and ask an advanced contributor to link the sentence without typo. Which is not very practical.
- Therefore the solution we have right now is: let users edit their typos and let Horus do the merge.

jiru:
> My copy-pasted sentence can transition to CC BY since it can be considered as a translation of an original sentence within Tatoeba.

Is that legal? If the content is the same, isn’t it more about the way the sentence was added that makes it legal or illegal? If the sentence was actually translated from German, it means you put some real effort into creating a sentence based on a CC content, so you are good. However if you just copied it from copyrighted material, you’re infringing the law. I’m no lawyer, but shouldn’t the sentence be re-created as a translation?

trang:
> If the sentence was actually translated from German, it means you put some real effort into creating a sentence based on a CC content, so you are good. However if you just copied it from copyrighted material, you’re infringing the law.

If you copied a sentence and nothing else, yes, that's infringement. 

But what if you copied a sentence because it turns out to be the perfect translation for an existing CC BY sentence? You wouldn't randomly take any copyrighted sentence to add it as a translation. You still have to process the content of the sentence and then evaluate that it is the perfect translation of another existing sentence that you saw before.

In any case, it doesn't make too much sense to make hypothesis on single sentences. This is something we have to look at in the bigger picture.
- If 100 sentences with licensing issues have been linked to original sentences in a short period of time, we can suspect that the original sentences were not so original.
- If 100 sentences with licensing issues have been linked to original sentences sporadically, over the course of a couple of years, and the sentences are overall quite ordinary, we can rule out that these sentences were the result of actual translations.

Until we can make a judgement, we would of course just keep the licenses as they are. Which means we can have cases of CC BY sentences being linked to sentences with licensing issues. I'm not saying we should automatically convert a sentence with licensing issue to a CC BY sentence whenever it's being linked as a translation of a CC BY sentence.

Issue #1515, opened 2017-11-08
Transcripts are not available for download.


Lemmih:
I'm interested in the user-edited transcripts for Chinese sentences but they are not available for download.

Ideally they would be available in the format of: Id [tab] Senence id [tab] Script [tab] Text.

Making this data available would only require a change to 'scripts/weekly_exports.sql' and 'cron/export.sh'. May I go ahead and create a PR for this?

trang:
I'm not sure we actually have a lot of user-edited content there... For Japanese we surely have a bit, for Chinese I'm afraid we won't have anything. But that's just a guess.

Feel free to make a PR for it :)

Lemmih:
Not many transcriptions yet but they'll come. I've shown lots of sentences to people and they're quite good at spotting when the automatic transcription falls short.

I'll get cracking on that PR.

PR #2249, opened 2020-04-04
Export transcriptions

jiru:
This PR solves #1515 by adding a new section to the Downloads page allowing to download transcriptions and alternative scripts.

@alanfgh or @ckjpn Could you please check if the English is correct in the Transcriptions section of the [new version of the download page](https://dev.tatoeba.org/eng/downloads)?


ckjpn:
> Contains all transcriptions and alternative scripts. If the username is empty, it means it has been autogenerated by our tools. Otherwise, it means it has been reviewed by this user. The script name is defined according to the ISO 15924 standard.


It looks OK.

BTW, you might want to look into using the Yahoo API for Japanese furigana.
It seems more accurate, at least in a few cases.

See this comment.
https://tatoeba.org/eng/sentences/show/8611523#comment-1166322

ckjpn:
2 points

1. I wonder what the field that I have labeled as "type" on this page means, and also wonder if it needs to be in the exported data.

https://all.imfast.io/transcriptions.html

2. Perhaps instead of "username is empty", using \N as in the sentences_detailed.csv might be a good idea.  I'm not sure.

jiru:
The column you labeled as type is the script name. As explained in the file description, the script name is defined according to the ISO 15924 standard. Languages like Uzbek or Mandarin have more than one.

alanfgh:
Currently, the text says this:

> Contains all transcriptions and alternative scripts. If the username is empty, it means it has been autogenerated by our tools. Otherwise, it means it has been reviewed by this user. The script name is defined according to the ISO 15924 standard. 

Is this what is meant?

:::
Contains each transcription or alternative script, with the name of the user who either generated or reviewed it. An empty username indicates that it was autogenerated and has not been reviewed. The script name is defined according to the ISO 15924 standard.
:::

jiru:
@alanfgh Not exactly.

> the user who either generated or reviewed it

A user cannot generate a transcription; only an algorithm can. To put it another way, in languages that have transcriptions, when a sentence is created, it automatically gets an autogenerated transcription attached. From there, a user can review the transcription by marking it as proofread or by editing it.

ckjpn:
Perhaps this would be clearer.

FROM:
Otherwise, it means it has been reviewed by this user. 
TO:
Otherwise, it means it was written by this user or approved by this user.

alanfgh:
> A user cannot generate a transcription; only an algorithm can. To put it another way, in languages that have transcriptions, when a sentence is created, it automatically gets an autogenerated transcription attached. From there, a user can review the transcription by marking it as proofread or by editing it.

Okay, let me see whether I understand.

Your original text talks about transcriptions AND alternative scripts. If I understand the terminology correctly, a transcription refers to a representation of the sentence that is not meant to stand alone. An example would be furigana. A transcription is always generated automatically. However, it can be marked as proofread, or edited (which I assume will also mark it as proofread). Presumably, the name of the last user to do this is associated with the transcription.

What you are calling an alternative script refers to a version of the sentence in a language that allows multiple independent writing systems (for instance, Arabic and Latin), all of which are used in written materials for native speakers (adults). Such a version of a sentence, presumably, is always written by hand (?) and the username of the author is automatically associated with it (?).

If I understand "alternative script" correctly, I'm not enthusiastic about the term because I think of the script as referring to the writing system, not to a sentence written in it. I'd need to think some more about an alternative term. In some ways, it would be nice to come up with a term that could be used to cover both cases where the writing system is standalone (Arabic vs. Latin script) and where it isn't (furigana). But it might also be good to distinguish between them. I guess I need to know more before I can decide.

jiru:
> Your original text talks about transcriptions AND alternative scripts. If I understand the terminology correctly, a transcription refers to a representation of the sentence that is not meant to stand alone. An example would be furigana. A transcription is always generated automatically. However, it can be marked as proofread, or edited (which I assume will also mark it as proofread). Presumably, the name of the last user to do this is associated with the transcription.

That’s right.

> What you are calling an alternative script refers to a version of the sentence in a language that allows multiple independent writing systems (for instance, Arabic and Latin), all of which are used in written materials for native speakers (adults). Such a version of a sentence, presumably, is always written by hand (?) and the username of the author is automatically associated with it (?).

Your definition of alternative script is correct. However, alternative script work exactly like transcriptions: they are always autogenerated upon sentence creation, and can be further proofread or edited. For example, Mandarin Chinese can be written in either simplified or traditional characters, and the opposite script is always autogenerated. Same for Uzbek that can be written in Cyrillic or Latin.

jiru:
> NOTE: the TatoDev downloads page doesn't have the section for transcriptions and alternative scripts anymore.

Sorry. I put it back.

ckjpn:
> the user who last reviewed and possibly modified it.

Can one member change another member's transcription?
If so, is the member whose username was originally attached to the transcription informed?

jiru:
@alanfgh Thanks for all the thoughts you put into this. It makes it all easier when all the data is named with the same wording "transcriptions". I updated the [text on dev](https://dev.tatoeba.org/eng/downloads) if you wanna have a final look.

@ckjpn 
> Can one member change another member's transcription?

Regular members can only review/edit transcriptions of their own sentences. Advanced users can review/edit unreviewed transcriptions of other members. Corpus maintainers and admins can edit anything.

>  If so, is the member whose username was originally attached to the transcription informed?

No.

alanfgh:
> I updated the text on dev if you wanna have a final look.

Looks good. Thanks!

Issue #2250, opened 2020-04-04
Linefeeds in sentence text are breaking the weekly export script


ckjpn:
There may be other problems, but here is the one in sentences_detailed.csv

8649740	eng	817/5000\
Achieve easy success.	pelwan	2020-04-01 04:28:01	2020-04-01 04:28:01

My scripts I have from processing the weekly files wouldn't work because of this.

ckjpn:
The problem is that this item had a linefeed inserted into it.

https://tatoeba.org/eng/sentences/show/8649740

Maybe linefeeds should be filtered out before adding sentences to the database.

jiru:
@ckjpn I fixed the exports files if you want to download them again.

Not only the export files are not well formatted for most CSV parsing tools (except MySQL’s), but the export halted half way with the following error:
```
awk: cmd. line:1: (FILENAME=sentences_detailed.csv FNR=8243386) fatal: can't redirect to `/var/tmp/per_language/pelwan/pelwan_sentences_detailed.tsv' (No such file or directory)
```

As a result, the per_language files did not get generated.

So I manually edited this week’s sentences.csv and sentences_detailed.csv to remove the garbage sentence 8649740 is prefixed with, and I manually executed the per_language files generation, so the exports are good for this week, but the export script needs to be fixed and we want to prevent linefeeds in sentence text.

The faulty files are still available for debugging purpose: [sentences.csv.old](https://downloads.tatoeba.org/exports/sentences.csv.old) and [sentences_detailed.csv.old](https://downloads.tatoeba.org/exports/sentences_detailed.csv.old). The relevant part is at lines 8243385 and 8243386.

Note that interestingly, while sentence 8649740 broke the generation of per_languages for sentences_detailed.csv, it didn’t broke sentences.csv (but I fixed it too).

AndiPersti:
The problem is that the text cleaning I've introduced in 1b31f9c21cfc45cefdb1c93bf29a0beceec78239 has a major flaw:
https://github.com/Tatoeba/tatoeba2/blob/436ad1252a45e7d33e4f2405962bea6fae811d29/src/Model/Entity/Sentence.php#L70-L72
This only replaces a series of at least 2 whitespace or control characters with a single space.
But I don't think we can simply replace `{2,}` with `+` to replace also single occurrences of these characters because than there will be problems with eg French sentences which use different spaces in sentences (see #770).

I've checked all exported sentences for whitespace and control characters (except the normal space):
[...]

PR #2251, opened 2020-04-04
Replace all control characters in a sentence's text with space

AndiPersti:
This PR addresses #2250 

The old version only replaced a series of at least two control characters so it was still possible to add sentences which contained single control characters.


jiru:
Thanks :heart:

PR #2253, opened 2020-04-05
Correct display of activity timeline page

agrodet:
This is a following of #2180. I dynamically set `min-width` to value proportional to the number of characters in the string, which gives a big enough width (I hope). I'm not sure it's a very good way to do it. If there's a better way to do that, please let me know.

I also added a `ContributionsStatsHelper` in order to factorize the code of the page.

Hopefully, this will give us a nice activity timeline page :)

### DIFF
--- a/src/View/Helper/ContributionsStatsHelper.php
+++ b/src/View/Helper/ContributionsStatsHelper.php
@@ -51,9 +51,12 @@ class ContributionsStatsHelper extends AppHelper
     {
         if(isset($stat[$category]) && $stat[$category] > 0){
             $width = ($stat[$category] / $maxTotal) * 100;
-            $minWidth = (mb_strwidth($stat[$category]) + 1) * 13;
-            return $this->Html->div('logs_stats '.$category, '+' . $stat[$category],
-                array('style' => 'width:'.$width.'%;min-width:'.$minWidth.'px')
+            $stringWidth = mb_strwidth($stat[$category]) + 1; // + 1 for the sign
+            $sign = '+';
+            if ($category == "unlinked" || $category == "deleted")
+                $sign = '-';
+            return $this->Html->div("logs_stats ${category}", $sign . $stat[$category],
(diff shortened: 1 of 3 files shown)


AndiPersti:
> If there's a better way to do that, please let me know.

CSS3 defines `fit-content` as a possible value for `min-width` which does exactly what we want:
http://jsfiddle.net/xL20tebj/

But as with all reasonable and modern features it's not supported by IE :-(
https://caniuse.com/#feat=mdn-css_properties_min-width_fit-content

src/View/Helper/ContributionsStatsHelper.php
```diff
@@ -0,0 +1,64 @@
...
+     *
+     * @return string
+     */
+    public function statBar($stat, $category, $maxTotal)
+    {
+        if(isset($stat[$category]) && $stat[$category] > 0){
+            $width = ($stat[$category] / $maxTotal) * 100;
+            $minWidth = (mb_strwidth($stat[$category]) + 1) * 13;
+            return $this->Html->div('logs_stats '.$category, '+' . $stat[$category],
+                array('style' => 'width:'.$width.'%;min-width:'.$minWidth.'px')
```
AndiPersti, line 56: Since `fit-content` isn't supported by all browsers, another option is to move the minimum size calculation to the browser:
```php
$chars = mb_strwidth($stat[$category]);
return $this->Html->...
    array('style' => "width: ${width}%; min-width: calc(1em * ${chars});")
```
This way you can get rid of the magic constant `13`.

Example with hardcoded values for `$width` and `$chars`: http://jsfiddle.net/u7mks3qt/

agrodet:
Thanks for the information @AndiPersti, I modified the code as you suggested. I also corrected the display of the sign that had become + whatever the kind of contribution.

jiru:
Thank you guys for fixing the problem :+1:

PR #2257, opened 2020-04-06
Allow downloading lists as txt and for the Shtooka recorder

jiru:
This PR allows to download lists in different formats in order to solve two problems:

1. As shown in [UX test 4](https://en.wiki.tatoeba.org/articles/show/ux-test-4), when exporting lists, most users are not familiar our dubious about the CSV format. Now, when downloading a list, the raw text format (txt) is selected by default.

2. (reported by arh) Members who want to record sentences with the Shtooka recorder are dependent on CK to provide them with the right data to feed Shtooka with. Now, when downloading a list, a format called "list for the Shtooka recorder" can be selected so that members may create their own lists and easily record them.

The design of the download list page is still crap, it would probably need to be redesigned completely, but I didn’t want to do this in this PR.

Note that to test downloading, you have to manually run the following command so that exports are created in the background:
```sh
sudo -u www-data bin/cake queue runworker
```


ckjpn:
Let me know when this is ready to test.
It doesn't seem to be an option yet.
https://dev.tatoeba.org/eng/sentences_lists/show/8333

Briefly, if this:
Sentence id [tab] Text 
... is exported like this:
Sentence id - Text 

That is with a space before and after the hyphen, it should work.

Actually at this time, a couple of audio contributors are creating their own Shtooka Recorder ready files, but it would be easier if any member could create their own lists for the most part, even if I have to help a few.

If would be cool though, if members could create such files from search results.

For example.

* download all my own French sentences that don't yet have audio files, sorted by length or by created, etc.  (Or limited to the first 1,000 results)

* download all my own French sentences without audio that are linked to another sentence with audio.

This would perhaps save a lot of time, since members wouldn't need to create a list first.

ckjpn:
## A possible improvement
I wonder if it is possible to either make it the default or at least the default checkbox option to not export any sentence that already has an audio file.

* Making it the only option might make it less likely that 2 members record the same sentence. Of course,  if there is a long time delay between downloading the text file and sending me the audio files for uploading another member may have already recorded the sentence.

* Making it a checkbox option, with it selected would prevent the above problem, but would allow members to create a list of sentences they wanted to re-record to replace previous sentences.

## One concern

Note that if a number of audio files come in for me to upload, it will take me a lot of time to verify that the replacement audio file is by the same member who previously uploaded the audio file.  We wouldn't want to delete an audio file by someone else to replace it, I think.

Perhaps you could re-program the audio file import page to check to make sure that it was the same person who originally created to file and if not, delete the file from the folder I import into and don't move the file to the audio file directory.

jiru:
@ckjpn It’s ready to test on dev.

> If would be cool though, if members could create such files from search results.

Arh also mentioned that to me. What I’m thinking about is to allow creating (or adding to) a list from search results instead. This way, it would also allow members to create a rough list about pretty much anything, and then manually add or remove sentences to the list in order to make additional adjustments.

> I wonder if it is possible to either make it the default or at least the default checkbox option to not export any sentence that already has an audio file.

Not for now.

> Note that if a number of duplicate sentence numbered audio files come in for me to upload, it will take me a lot of time to verify that the replacement audio file is by the same member who previously uploaded the audio file. We wouldn't want to delete an audio file by someone else to replace it, I think.

> Perhaps you could re-program the audio file import page to check to make sure that it was the same person who originally created the file and if not, delete the file from the first folder I import into and don't move the file to the audio file directory.

We could compare the existing audio file with the one in the import directory, and show in the import page whether or not they are different, and make this information stand out by grouping equal files in the table (by adjusting the sort order). Would that be enough to solve the problem?

ckjpn:
Would it be possible to export that file with Windows linefeeds?
At least when I downloaded the files on my Mac, the file had Unix linefeeds, which don't work properly in the Shtooka Recorder.

I can easily change them myself, but other members perhaps can't easily do so.

jiru:
Thanks for your feedback, CK. I’ve never used Shtooka so I’m not familiar with the process. From what I’ve seen in the documentation you and arh wrote, I thought that people had to open the sentence list with a text editor, and from there, copy and paste the sentence list into Shtooka. Do you mean that the sentence list file can also be directly opened by Shtooka and that it expects Windows-style linefeeds?

ckjpn:
## Answer

I open the file in the default text editor in Windows and copy and paste from there into the Shtooka Recorder, but if I haven't converted the linefeeds using BBEdit on my Mac, the lines aren't displayed properly on Windows.

I haven't yet actually tried your exported file, since my Windows computer is dedicated only for recording and it's not in an Internet-connected area.

## How I've been creating the lists

1. I usually do a search similar to this as long as there are still sentences that meet this criteria.

Query: gillux's French sentences without audio that are linked (indirect or direct) to another sentence with audio.

https://tatoeba.org/eng/sentences/search/page:1?to=none&native=yes&has_audio=no&user=gillux&trans_has_audio=yes&sort=words&from=fra

https://tatoeba.org/eng/sentences/search/page:2?to=none&native=yes&has_audio=no&user=gillux&trans_has_audio=yes&sort=words&from=fra

etc. up to page 10 with 100 sentences each

2. I then merge all the pages together.

3. I then run a script that converts these pages to the needed format.  _I assume I'll  need to rewrite the script once the new design is forced upon us._

SENTENCE_ID - SENTENCE_TEXT

4. I then usually split the resulting file into groups of 100 to 300 sentences depending on which members has requested the sentences, so members can record and send the sentences to me in several sessions.

5. I then zip those files together and email them to the member.


Additional notes:

* Once all sentences resulting from the default search (see above) are done, I remove the "trans_has_audio=yes" requirement.

https://tatoeba.org/eng/sentences/search/page:1?to=none&native=yes&has_audio=no&user=gillux&sort=words&from=fra
(I don't do this search until I've uploaded any previous submissions, of course.)

* If the member does not have any sentences, I try to make sure that what I send them are sentences by native speakers who do not already record their own sentences.

In the past, I've had one member who asked that others don't record his sentences.

The easiest way to do this is to send sentences by one chosen member each time, since that way I don't have to filter out all the ones who are adding recordings.

* I offer to do different searches for sentences, and have done so sometimes, but most members seem to like the idea of creating audio files for sentences that are linked to other sentences with audio files. 

Here's a list I built for you to actually test using the Shtooka Recorder if you want to.

The list:
https://dev.tatoeba.org/eng/sentences_lists/show/8909

How to:
http://bit.ly/shtooka

I usually have the sentences sorted by length when I send members sentences to record.

* This way if they don't finish the list, they've recorded more sentences in the same amount of time.
* Also, I have found it's easier for me to record if I record sentences of about the same length at the same time.  As you get to longer sentences, the Shtooka Recorder allows setting the minimal sentence length to a longer time, which helps with not needing to manually hit the previous button for false starts, etc.

I wonder if you could presort the sentences by length instead of exporting them in sentence ID order.

ckjpn:
I noticed that the download option is not there today.
If it's not intentional, then it's likely a bug.

https://dev.tatoeba.org/eng/sentences_lists/show/8909

AndiPersti:
> I noticed that the download option is not there today.

The download button moved from the sidebar to an icon in the (green) header of the list.

ckjpn:
I had the green hidden by Stylish to get the traditional white background, which hid the white icons. Sorry.

I guess I'll have to rewrite this.
https://userstyles.org/styles/178319/tatoeba-org-get-rid-of-dec-2019-green-titles

Issue #2194, opened 2020-03-11
Expand/collapse translations not working correctly on the new design


jiru:
I compiled a list of related bugs:

A. Steps to reproduce:
1. Find a sentence with a total of exactly 6 translations (direct + indirect)
2. Problem: only 5 translations are displayed without expand/collapse button.

B. Steps to reproduce:
1. Find a sentence without an expand/collapse menu (5 translations or less).
2. Add several translations.
3. Problem: the new translations appear on top of the translation list while the lower translations disappear.

C. Steps to reproduce:
1. Find a sentence with an expand/collapse menu (6/7 translations or more).
2. Add one translation.
3. The expand/collapse menu says "show n more translations", but n is incorrect, it should be n+1 instead.

trang:
I fixed (A) as well as another issue I found along the way.

(B) and (C) are a bit trickier because the label has to be updated properly. The string can be singular or plural, and plural can have complicated rules in some languages.

I'm thinking of adding a parameter `currentNumberOfTranslations` when calling `/sentences/save_translation/`. This number would be incremented if the translation is saved properly and in the response we can return the new label along with the saved translation.

Not sure if someone has a better idea.

Issue #1783, opened 2019-02-15
When linking sentences in the advanced search, more languages appear, even if they're not selected in the user's profile


Guybrush88:
as reported by Aiji on the wall:

> Advanced search:
> English, display translations in French
> Limit to sentences having French indirect translations.
> 
> Only English and French are displayed.
> 
> Link one French => All translations are now displayed.
> 
> Is it intended behavior?
> (I do not think so, and that is quite a bother when doing some linking work)

second post:

> If I link one French sentence, then all translations will appear.
> The problem will not happen if you restrict the langages to be displayed in your profile. That's usually how I do it, but this time I forgot so I faced this situation (maybe I, or somebody else, already mentioned it in the past).

https://tatoeba.org/ita/wall/show_message/31332#message_31332

ckjpn:
The same thing happened before the update to the latest CakePHP, so it's not a change in the way the website works.

It never really bothered me.  However, I can see how it might be irritating on a smart phone.

ckjpn:
Actually, this is very irritating when trying to do something like that following.

1. Find all English sentences with audio that have an indirect link to a Japanese sentence.

https://tatoeba.org/eng/sentences/search?query=&from=und&to=jpn&orphans=no&unapproved=no&user=&tags=&list=4000&has_audio=&trans_filter=limit&trans_to=jpn&trans_link=indirect&trans_user=&trans_orphan=&trans_unapproved=&trans_has_audio=&sort=words

2. Link the indirectly-linked Japanese that match.

What happens is that all the other translations appear at this point, and you have to scroll past them all.


If this behavior could be changed, it would be a lot faster to get matching indirectly-linked sentences linked.

jiru:
I think this regression has been introduced by 5be6155f1ac3de8e06c124f72ac233db5a81b67b.

Issue #1332, opened 2016-10-09
Display warning when a user tries add a translation that is a duplicate


trang:
Extracting from https://github.com/Tatoeba/tatoeba2/pull/1311#issuecomment-250975600

There are 3 cases.

1) The user is copy-pasting the main sentence and is adding it as a translation. Not likely to happy, but could happen.
2) The user is adding a translation that is already a translation of the main sentence. It's possible that a user doesn't notice there is already the exact translation they want to add.
3) The user adds a translation which is an existing sentence but is not yet linked to the main sentence.

Suggestions for the warning text:
* https://github.com/Tatoeba/tatoeba2/issues/577#issuecomment-252452754 (case 3)
* https://github.com/Tatoeba/tatoeba2/issues/577#issuecomment-252456279 (case 3)

Issue #536, opened 2014-12-16
When producing a link that has already been produced, don't generate an error


alanfgh:
If you have a sentence with an indirect translation open on one page, then open it on another page where you change it to a direct translation (that is, you link it), then go back to the first page and try to link it again, you get an error message: "An error occurred while saving. Please try again or contact us to report this." I think it would be nicer to simply update the display of the sentence group without generating an error.

PR #2266, opened 2020-04-08
Sentence linking/unlinking in the new sentence design

trang:
This pull request implements linking and unlinking in the new sentence design. It also addresses the following issues: #1332, #2194, #1783, #536.

What's been implemented:
- For advanced contributors, the `>` icon in front of each translation is replaced by a scissors or link icon.
- Just like in the old design, clicking on the scissors icon will unlink the translation from the main sentence and clicking on the link icon will link the indirect translation to the main sentence.
- After linking or unlinking, the translations are refreshed and expanded.
- The link button has been removed from the sentence menu because it doesn't much added value. Contributors can already link to another sentence by adding the desired sentence as a translation.

Since the link button was removed, the translation feature has been readjusted:
- If a translation is added when all translations are displayed, all the translations are refreshed. This allows to include new indirect translations in case the translation is creating a link to an existing sentence.
  - There will be a warning icon to indicate that the sentence already exists and therefore only the link was created (#1332).
- If a translation is added while the translations are filtered in a specific language, the new translation is just added at the top (as it is today). The translations are not refreshed. This leads to some collateral issues resulting from solving #1783. I've decided to keep these issues for another time:
  1. I am viewing sentences in English with only French translations. I add a translation in Japanese that is a duplicate and it generates new indirect French translations. These indirect translations will not appear.
  2. I am viewing sentences in English with only French translations. I add a translation in Japanese. Then I decide to link or unlink one of the French translations. The Japanese translation disappears.

Wall thread for user tests: https://tatoeba.org/eng/wall/show_message/34750#message_34750


jiru:
Cool! :sunglasses: 

I quickly tested and my first impressions are:
1. The indentation is changed because the link/unlink icon is bigger. In the context of transcriptions, it looks strange because the arrow is exactly between the link icon and the flag. Here is what I mean:
    a. without linking icon (looks aligned): image
    b. with linking icon (looks out of place): image

2. The scissors icon looks like cutting the flag. I remember this comment was made to the old design too at some point, and it was decided that the scissors should be placed in front of the sentence (I don’t remember the context of that discussion).

3. About the warning to indicate that the sentence already exists and therefore only the link was created, the location of the information (a grey icon along with the "copy", "play audio" and (i) icon) doesn’t feel logical to me. Let me explain. All these buttons perform an action on the sentence, but the warning icon doesn’t (it’s not a button). All these buttons consistently stick to every single sentence, but the warning icon just appear once. It doesn’t convey an information that always sticks with the sentence, it’s just a one-time thing. I think the information would be better conveyed as a kind of notification. Maybe a text that appears on the top of the newly-added sentence.

I am sure some power users out there won’t appreciate this suggestion, but I have the feeling that these link/unlink buttons don’t need to be shown everywhere all the time like we’d need them ready to be used any time. How about bringing back the link button on the sentence menu as a way to toggle show/hide all the link/unlink buttons on the sentence an translations?

trang:
1. Alignment of transcriptions is fixed.
2. I replaced the scissors icon with the `link_off` icon. The scissors were reported to feel too big. I think with the `link_off` icon it's fine to keep it in front of the flag.
3. Showing a warning was also confusing for the case where a user is linking to another translation by copy-pasting. It gives the impression that something went wrong when on the contrary, everything went fine. I converted the warning into a tooltip that appears when hovering over the translation text. 

> How about bringing back the link button on the sentence menu as a way to toggle show/hide all the link/unlink buttons on the sentence an translations?

I think it makes sense to not have the buttons displayed all the time. I've instead chosen to display the link/unlink buttons when the sentence menu is expanded. I think it's best to avoid too many levels of toggling.

jiru:
The interface looks good to me now.

The "expand menu" button now does a lot more than expanding the sentence menu: it also shows unreviewed transcriptions and linking buttons (and it shows hidden translations (I didn’t know that)). Maybe we should rename its tooltip accordingly.

trang:
> and it shows hidden translations (I didn’t know that)

That's actually due to a shared behavior with the "Show unreviewed transcriptions" button for non-authenticated users.

In the case of non-authenticated users, the translations are expanded so that users don't get confused at not seeing anything happening if the unreviewed transcriptions happen to be part of hidden translations.

We could technically make the behaviors specific to non-authenticated users only. But I've actually found myself many times using this "Expand menu" button to expand the translations. And conversely, using it to try to collapse the translations (but nothing would happened and I would get shortly confused). So there's definitely something to rework there...

> Maybe we should rename its tooltip accordingly.

I think so too, but I'm not very inspired on what could be the tooltip... "Show basic view" / "Show extended view"?

src/Model/Table/SentencesTable.php
```diff
@@ -438,21 +439,18 @@ public function findFilteredTranslations($query, $options) {
...
-            'Show {number} more translations',
-            $extraTranslationsCount
-        ), array('number' => $extraTranslationsCount));
+        $extraTranslationsCount = $total - self::MAX_TRANSLATIONS_DISPLAYED;
+        if ($extraTranslationsCount > 0) {
+            return format(__n(
+                'Show 1 more translation',
+                'Show {number} more translations',
+                $extraTranslationsCount
+            ), array('number' => $extraTranslationsCount));
```
AndiPersti, line 450: I think it would be nice if that number is localized. 

While it is possible to use a helper (here `Cake\View\Helper\NumberHelper`) in a model, this demonstrates that `getExpandLabel` breaks the MVC philosophy where the model and the view shouldn't be coupled. Thus I don't think `getExpandLabel` should be a table method.

jiru: Related bug: #2194.

webroot/js/directives/sentence-and-translations.dir.js
```diff
@@ -614,6 +627,57 @@
...
+            if (vm.isExpanded) {
+                showAllTranslations();
+            } else {
+                showFewerTranslations();
+            }
+        }
+
+        function updateNewTranslationsInfo(translation, sentenceId, translations) {
+            allDirectTranslations = translations[0];
+            allIndirectTranslations = translations[1];
```
jiru, line 658: The `allIndirectTranslations` variable is not used.

src/Template/Element/sentences/translation.ctp
```diff
@@ -6,12 +6,26 @@ $sentenceBaseUrl = $this->Url->build([
...
+            <md-tooltip><?= __('Unlink this translation.') ?></md-tooltip>
+        </md-button>
+        <md-button class="md-icon-button" ng-if="vm.isMenuExpanded  && !translation.isDirect" ng-click="vm.saveLink('add', translation)">
+            <md-icon>link</md-icon>
+            <md-tooltip><?= __('Make into direct translation.') ?></md-tooltip>
+        </md-button>
+        <md-icon class="chevron" ng-if="!vm.isMenuExpanded">chevron_right</md-icon>
+    <?php } else { ?>
+        <md-icon class="chevron">chevron_right</md-icon>
+    <?php } ?>
```
jiru, line 27: I think it would be cleaner to have this `if (CurrentUser::isTrusted())` branch coded as an AngularJS `ng-if`. Maybe you could retrieve the value as a new canLink permission from the menu (we already have canEdit, canTranscribe…).

webroot/js/directives/sentence-and-translations.dir.js
jiru: Nevermind.

jiru:
Thanks for the update, Trang. :+1:

Having to come up with the more vague tooltip "Show more features" is a sign of bad UI design in my humble opinion. I don’t have an answer to that, so my suggestion is let’s sit down and think about it.

trang:
The answer to that is actually simple: remove features from the sentence component, don't make it do anything and everything. It's not easy to implement though as it requires to redesign the overall user experience.

My personal vision is that the sentence component should be primarily designed for exploration (meaning that the user is only seeking information). The primary buttons should just allow the user to access information. We should still provide a way for the user to transition smoothly from exploration to any specific activity. But these activities don't have to happen in the sentence component itself.

Now again, unless we take the time to redesign entirely the whole user experience for each activity, we won't be able to address the core issue.

For the shorter term, I see only these possibilities:

**(1)** We bring back the labels `Expand menu` and `Collapse menu` and we make the button do exactly just that. The consequences:

- We display the link/unlink buttons all the time.
- We rework the logic for unreviewed transcriptions. For this we have two options:
  - We introduce a "transcription" button to show unreviewed transcriptions. But this is not great because there's two levels of toggling. Plus it would be difficult to find a default Material icon that expresses the idea of transcriptions.
  - Or we hide unreviewed transcriptions by default all the time. We only show them if the user has the option to always show transcriptions.

**(2)** We give up on the toggle menu button and display the whole menu all the time above (or below) the sentence id and owner. The downside is that it makes the sentence component feel more overwhelming at first sight. But at least from there we could:

- Bring back the "link" button to show or hide the link and unlink icons.
- Have a transcription button to show or hide unreviewed transcriptions.

**(3)** We redesign the sentence menu into [an actual menu](https://material.angularjs.org/latest/demo/menu). There would be just one button that opens a menu with the items "Translate sentence", "Edit sentence", "Add sentence to list", etc. This would be a better UI for mobile display but veteran contributors will be very frustrated to have to click twice as much.

**(4)** We keep the concept of having a "basic" and "extended" mode. The basic mode only shows what we arbitrarily decide to be the most important actions and information. The extended mode shows everything.

src/Model/Table/SentencesTable.php
trang: So this would be my solution if we were to localize the number:
https://github.com/Tatoeba/tatoeba2/tree/expand-label

I don't feel very happy about this code though but I don't see another way.

jiru: There might be a way to get this done in Javascript using [angular-gettext](https://angular-gettext.rocketeer.be/), but that project is not maintained any more, and we’d have to find a way to implement NumberHelper in Javascript too. It feels like too much trouble and uncertainty for that one string. 

Maybe ecb2e3e323f20f4f099b96f838d96b93901b3b6b can be the basis of a solution. You don’t have to throw away all the the CakePHP serialization code, you can expand it. Look at `src/View/JsonView.php` in CakePHP’s code, it basically just runs `json_encode()`. So you could extend that class into, let’s say, `SentencesJsonView.php`, make it add the expand label before calling `json_encode()`, and then use `$this->RequestHandler->renderAs($this, 'sentences_json');` from the controller.

As for the views, you already have `SentencesHelper::sentenceForAngular()` that does a similar job, so you could have the expand label added from there too, and call that method from `SentencesJsonView.php`.

By the way, findFilteredTranslations has grown into something that does a lot more that filtering translations: it sets the expand label, as well as permissions and extra translations count. Maybe it’s a good opportunity to move all this out of the Sentence model too.

Issue #2265, opened 2020-04-07
Nyungar (nys)


RyckRichards:
* Language name: Nyungar
* Language info: https://en.wikipedia.org/wiki/Nyungar_language
* ISO 639-3 code: NYS
* Public List: https://tatoeba.org/eng/sentences_lists/show/9906
* Icon: https://i.postimg.cc/VNj6mkPH/nys-little.png
* Requested by: https://tatoeba.org/eng/user/profile/shekitten

PR #2268, opened 2020-04-08
Added language Nyungar (nys)

RyckRichards:
This PR addresses #2265


jiru:
Are you sure about the flag? It is just a text.

RyckRichards:
@jiru - It's just a temporary solution which both shekitten and sabretou agreed on. :) You can find more details (and how did they come up with this temporary solution) here:

»  https://tatoeba.org/eng/sentences/show/8657206#comment-1169502

Issue #2192, opened 2020-03-10
Some translators have problems to translate the UI


RyckRichards:
Some users have told me that they are not sure how to translate a specific string because

a) They don't know how it's intended to be used (as a verb? a command? a noun?)
b) They don't have a "picture" about what is the string about. Example: {By doing so you're aware that you can't revert it back} [I've just created this.] What does the string mean by "by doing so"? 

I direct them to https://en.wiki.tatoeba.org/articles/show/interface-translation - Useful information in the details - but even for me it's kinda confusing. Sometimes there's no "piece of link" (like in https://prnt.sc/regr7x ) to have a clue. We ought get something like https://prnt.sc/regrme

Perhaps we could find a way to make it clear to them what are those (every) strings about by adding a @Developer note or @Note for translators.

jiru:
My idea on how to semi-automate the provision of information to translators:

1. Generate a special locale that inserts identifiers in every string:
```po
# src/Locale/special_locale/default.po
msgid "Browse"
msgstr "XXX1234567XXX Browse"
```
3. Create a script to automate the browsing of such page using [Selenium](https://www.selenium.dev/documentation/en/). The script should keep track of which strings are displayed on which URL.
4. Make this information available to translators by automatically [adding string instructions](https://docs.transifex.com/api/examples/#adding-instructions-to-a-specific-source-string) that include all the URLs to the page.
5. (Optional improvement) Make the Selenium script remove all the identifiers, locate the position of the string inside the page, take a screenshot of the page, upload the screenshot somewhere, create a basic HTML page that displays the screenshot overlayed with a marker at the right position, and add a link to that page in the strings instructions.

jiru:
@RyckRichards Can you give us some example of actual strings that are hard to translate?

PaulPeer:
> Can you give us some example of actual strings that are hard to translate?
A typical one is #937 in default.pot:
reply
Is it a substantive, like in "Here you see the reply"?
Or a verb like "I want to reply"?
In most languages it's impossible to translate the word „reply” without knowing the context or how it is used (verb? command? noun?)

PR #2274, opened 2020-04-12
Insert developers notes to help translating ambiguous UI strings

jiru:
This PR solves #2192 by inserting 313 developers notes (that show up on Transifex) on translatable strings that are potentially difficult to translate for UI translators because of the lack of context. I focused on strings that are one-word long, or could otherwise be difficult to understand.

Not only the notes tell the context, but when a string is used a different places around, I took care of inserting a separate note on every place it is used. This might be seen as very verbose, but I believe this is especially useful for translators to see all the different contexts the string is used.

I also added msgctxt where appropriate, capitalized all tooltips, refactored a legend, and removed the `@translators` markers from the final notes in the POT files.

I believe this approach is efficient but it is quite verbose and takes a lot of space in the source code. Are you okay with that?

If you want to see how the developers notes would ultimately look like on Transifex, run `./tools/generate_pot.sh` on this branch and look for lines that start with `#. ` in `src/Locale/default.pot`.

### DIFF
--- a/src/Model/Entity/LanguageNameTrait.php
+++ b/src/Model/Entity/LanguageNameTrait.php
@@ -53,6 +53,7 @@ trait LanguageNameTrait
         if (isset($languages["$code"])) {
             return $languages["$code"];
         } else {
+            /* @translators: dropdown option for unknown language */
             return __('unknown');
         }
     }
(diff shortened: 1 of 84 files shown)


AndiPersti:
> I believe this approach is efficient but it is quite verbose and takes a lot of space in the source code. Are you okay with that?

While in sum it takes some space, these are mostly one line comments so I don't see any problems. Good work :+1:

PR #2281, opened 2020-04-16
Add Gronings and Kurdish into Tatoeba U.I.

RyckRichards:
This PR adds Gronings (Grunnegs) and Kurdish (Kurdî) on Tatoeba UI.

As far it got more about 70% of their strings translated 
(https://www.transifex.com/tatoeba/tatoeba_website/dashboard/), I agreed on adding them.

The translators are aware that many "texts" (strings) will be displayed in English yet because it's not 100% complete

Gronings - MarijnKp - https://tatoeba.org/eng/user/profile/MarijnKp (MarijnKp in Transifex)
Kurdish - Cibrani - https://tatoeba.org/eng/user/profile/Cibrani (resat7447 on Transifex)


jiru:
Thanks Ricardo!

There is no need to add `ku_KU:kur` in the tx config because the language code on Transifex is already `ku`. Only add languages to tx config when they have an underscore in Transifex.

While Kurdish has indeed reached 70%, Gronings is still only at 34% (these numbers are only referring to the completion of default.pot, excluding the other pot files).

By the way, we have a tool to print out the statistics `./tools/ui-languages-check.py` that I repaired recently.

RyckRichards:
Thanks, @jiru ! I was following [this article](https://github.com/Tatoeba/tatoeba2/tree/dev/src/Locale#how-to-add-a-new-ui-language). I've removed "ku_KU:kur" from the tx config

jiru:
My bad, that article was outdated. Thanks for telling, I updated it.

By the way, you forgot to remove a comma in the lang_map list. It makes the file invalid so please fix it. :wink:

RyckRichards:
Thanks, @jiru ! I've fixed it.

By the way, you told me that e have a tool to print out the statistics ./tools/ui-languages-check.py which you repaired recently.

How can I use that?

Issue #2231, opened 2020-03-25
Wrong warning message when searching for existing tag


AndiPersti:
[Tag 7468](https://tatoeba.org/eng/tags/show_sentences_with_tag/7468) has the rather unusual name `'-' it is used for linking two different parts of`.

After [searching for this tag](https://tatoeba.org/eng/sentences/search?query=&from=und&to=und&user=&orphans=no&unapproved=no&has_audio=&tags=%27-%27+it+is+used+for+linking+two+different+parts+of&list=&native=&trans_filter=limit&trans_to=und&trans_link=&trans_user=&trans_orphan=&trans_unapproved=&trans_has_audio=&sort=relevance&sort_reverse=) a warning message is shown (`the following criteria have been ignored ... because it's an invalid tag name`) although the correct result is displayed.

AndiPersti:
Steps to reproduce:
1) Go to some sentence
2) Add the following tag: `1234567890123456789012345678901234567890123456789 content after space gets trimmed`
3) Searching for the part before the space will produce a warning and the correct result

A tag name's length is [restricted to 50 bytes](https://github.com/Tatoeba/tatoeba2/blob/d39ecb565a894a661c137302bedbc2eb527e16a0/src/Model/Table/TagsTable.php#L86-L89). If the 50th byte happens to be a space character we will still get the correct result because [MariaDB ignores trailing spaces in VARCHAR comparisons](https://mariadb.com/kb/en/varchar/#description):

    MariaDB [tatoeba]> select 'ab ' in ('ab');
    +-----------------+
    | 'ab ' in ('ab') |
    +-----------------+
    |               1 |
    +-----------------+
    1 row in set (0.00 sec)

But since we [removed trailing spaces from the tag query parameter](https://github.com/Tatoeba/tatoeba2/blob/d39ecb565a894a661c137302bedbc2eb527e16a0/src/Controller/SentencesController.php#L642) we don't find these tags in the result set later on [when we create the list for the warning message](https://github.com/Tatoeba/tatoeba2/blob/d39ecb565a894a661c137302bedbc2eb527e16a0/src/Controller/SentencesController.php#L653-L660).

There is [another tag in the database](https://tatoeba.org/eng/tags/show_sentences_with_tag/7084) which has the same problem.

[...]

PR #2284, opened 2020-04-17
Improve add tag

AndiPersti:
This PR addresses #2231 

In order to fix this bug I've refactored `TagsTable::addTag`:

1) Avoid a redundant SQL query when we try to add an already existing tag
2) Move the trimming part to `beforeMarshal` event listener and use CakePHP's validation process
3) Make it work when no sentence id is provided
4) Add the `TagsSentence` entity to the `Tags` entity instead of throwing away the `Tags` entity. We need the name which is actually stored in the database later in the controller

### DIFF
--- a/tests/TestCase/Model/Table/TagsTableTest.php
+++ b/tests/TestCase/Model/Table/TagsTableTest.php
@@ -159,9 +159,9 @@ class TagsTableTest extends TestCase {
 
     public function testAddTag_correctDateUsingArabicLocale() {
         I18n::setLocale('ar');
-        $added = $this->Tag->addTag('arabic', 4, 1);
+        $added = $this->Tag->addTag('arabic', 4);
         $returned = $this->Tag->get($added->id);
-        $this->assertEquals($added->added_time, $returned->created);
+        $this->assertEquals($added->created, $returned->created);
     }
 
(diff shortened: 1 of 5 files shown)


jiru:
Good job! :+1:

It’s interesting to see that suggestd was being unnecessarily poked on every tag operation! Good thing that you made the code call `/add` only when a new tag is created.

tests/TestCase/Model/Table/TagsTableTest.php
```diff
@@ -159,8 +159,25 @@ public function testAdminDoesRemoveTag() {
...
+
+    public function testAddTag_addEmptyTag() {
+        $added = $this->Tag->addTag('', 1);
+        $this->assertFalse($added);
+    }
+
+    public function testAddTag_noTrailingSpaceAfterCuttingTo50Bytes() {
+        $tagName = '1234567890123456789012345678901234567890123456789 content after 9 gets cut';
+        $added = $this->Tag->addTag($tagName, 4);
+        $storedId = $this->Tag->getIdFromName(substr($tagName, 0, 49));
```
jiru, line 180: Could you rewrite this test without using substr? In general it’s better to keep the tests as simple as we can because complexity is a source of bugs and because it’s easier to understand (I didn’t get the 49 at first).

Issue #2289, opened 2020-04-19
On the "Languages of Members" page, the fluency symbols are out of place


sabretou:
I'm using Windows 10 64-bit desktop to use Tatoeba. On my computer, the fluency bars on the "Languages of Members" page are out of place. 

This bug can be replicated across:

- Firefox 64-bit 76.0b5
- Chrome 64-bit 81.0.4044.113
- Edge (Unknown version, thanks Microsoft)

The bug is present on both the main website and the development website. It also present regardless of UI language setting (I have tested English and Marathi).

This bug is **not** present when I use Tatoeba on Firefox for Android.

Screenshots:

Firefox (Main website):
![image](https://user-images.githubusercontent.com/7028559/79687772-3cb5a100-8267-11ea-9097-bebe40fe07f6.png)

Chrome (Dev website):
![image](https://user-images.githubusercontent.com/7028559/79687794-61117d80-8267-11ea-8561-dbd1d4d4c336.png)

jiru:
Just to clarify, are you talking about the blue bars that overlap a little bit over the bottom of their frames? On my machine running Linux, I can reproduce the problem with Chromium but not Firefox...

sabretou:
@jiru Yes, I am referring to the blue bars that are placed slightly lower than they should be, overlapping the frames.

AndiPersti:
The misalignment is dependent on the browser font size. I can reproduce it on Linux with both Firefox and Chromium if I set the font size to a value bigger than 15px (the height of the containing `div` is set to 15px).

I've pushed a fix which works for me with Firefox 75 and Chromium 81 (both on Linux) and the stock Android Browser 6.0.1, Firefox Preview 4.2.1 and IceCatMobile 68.4.2 (all on an Android tablet). But I suggest to also test it on dev.

jiru:
I pushed Andreas’ fix on [dev.tatoeba.org](https://dev.tatoeba.org/eng/stats/users_languages) if anybody wants to test.

sabretou:
I checked dev, and it looks perfect to me now!

jiru:
Thanks for confirming, @sabretou!

PR #2290, opened 2020-04-20
Fix alignment of language level bars on stats page

AndiPersti:
This PR addresses #2289 

I'm not a CSS expert but the suggested fix works for me on all browsers I've tested (Webkit- and Gecko-based).

Since the bars have no content (they are just empty `div`s with a background colour), we need to position them manually relative to the top left corner of the containing box.

### DIFF
--- a/webroot/css/stats/users_languages.css
+++ b/webroot/css/stats/users_languages.css
@@ -42,9 +42,6 @@
 .usersLanguagesLegend .languageLevel .level {
     background-color: #06A;
     height: 100%;
-    font-size: 12px;
-    float: left;
-    position: relative;
 }
 
 .usersLanguagesStats .languageLevel .unknownLevel,
--- a/webroot/css/stats/users_languages.css
+++ b/webroot/css/stats/users_languages.css
@@ -42,6 +42,9 @@
 .usersLanguagesLegend .languageLevel .level {
     background-color: #06A;
     height: 100%;
+    font-size: 12px;
+    float: left;
+    position: relative;
 }
 
 .usersLanguagesStats .languageLevel .unknownLevel,
--- a/src/View/Helper/LanguagesHelper.php
+++ b/src/View/Helper/LanguagesHelper.php
@@ -459,9 +459,10 @@ class LanguagesHelper extends AppHelper
             $text = '';
             $opacity = $opacity = 0.5 + 0.5 * ($level / Language::MAX_LEVEL);
             $size = ($level / Language::MAX_LEVEL) * 100;
+            $padding = (100 - $size) / 2;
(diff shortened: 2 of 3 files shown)


jiru:
Nice. :+1:

I wonder if having the bar at the center of the frame was the original intention of @trang, because it’s quite uncommon represent a "progress" bar like this. If not, you may as well remove the left padding.

trang:
> I wonder if having the bar at the center of the frame was the original intention of @trang

No, indeed, it wasn't my original intention. It was supposed to look like progress bar (where native level is 100% progress).

webroot/css/stats/users_languages.css
```diff
@@ -42,6 +42,9 @@
 .usersLanguagesLegend .languageLevel .level {
     background-color: #06A;
     height: 100%;
+    font-size: 12px;
+    float: left;
+    position: relative;
```
trang, line 47: Instead of setting the font-size, float and position, just adding `display: block` might solve everything.

The misalignment seems to be because of the `level` div being set with a display `inline-block` and for some reason, inline-block messes up the vertical alignment.

AndiPersti: > The misalignment seems to be because of the `level` div being set with a display `inline-block` and for some reason, inline-block messes up the vertical alignment.

Well, your comment motivated me to look for the reason why this happens and the real problem lies somewhere else. :grin:   

There's another CSS declaration block for the `.languageLevel .level` selector:
https://github.com/Tatoeba/tatoeba2/blob/0ac8959db06237f110834e9aa0b0a3bef1f1dd65/webroot/css/layouts/elements.css#L1707-L1719
As you can see the second inner `div` block is also set to `inline-block`. That's why setting `display:block` helps and the inner `div` behaves as it should.

I've searched through the whole codebase to find the places where we use elements with class `languageLevel` and I've found the following three places:
- https://github.com/Tatoeba/tatoeba2/blob/0ac8959db06237f110834e9aa0b0a3bef1f1dd65/webroot/js/user/language.ctrl.js#L38-L43
- https://github.com/Tatoeba/tatoeba2/blob/0ac8959db06237f110834e9aa0b0a3bef1f1dd65/src/View/Helper/MembersHelper.php#L102-L117
- https://github.com/Tatoeba/tatoeba2/blob/0ac8959db06237f110834e9aa0b0a3bef1f1dd65/src/View/Helper/LanguagesHelper.php#L468-L472

Only in the third place (i.e. the stats page) we also use the inner `level` class so I suggest to simply remove the conflicting blocks from `elements.css` (the empty rule is unnecessary and `.languageLevel.unknown` isn't used anywhere).

Issue #2156, opened 2020-02-20
Rework the Stats block


agrodet:
See the discussion started on #2144. Extracting some information

One idea: 
<img width="274" alt="image" src="https://user-images.githubusercontent.com/42261966/74811041-feb60280-5333-11ea-85e2-a45f27a36982.png">

trang said

> My only issue is that it feels a bit strange to have two stats (number of sentences + number of languages) associated to one page. It would feel less broken to have each stat followed by a link to some related page.
> 
> For the {x} sentences, I have two ideas of related page:
> 
>  the random sentence
>  or the advanced search

CK said
> Perhaps the following might be better.
> 468 sentences since 0:00 UTC
> 
> I still think it would be a good idea to add a "yesterday's" number on the main page, too. This gives both regular contributors and new contributors better idea of what's going on without needing to click to the timeline every day.
> 
> 468 sentences since 0:00 UTC
> 3,141 sentences yesterday.

@trang I don't think the random sentence would bring any benefit. The chance to find a useful sentence to you is unlikely. A search might be better, I'm not very sure... I like the idea that this block is entirely dedicated to stats. 

@CK I think "since 0:00 UTC" is kind of noisy. Most of people don't naturally refer to "0:00 UTC", some don't even know what it is, so it would make more sense from the human point of view to use "today". However, "today" can be adapted to the user, like it was suggested elsewhere.

ckjpn:
[Additional Idea]

Instead of since 0:00 UTC, perhaps you could use the same relative time code you use on this page https://tatoeba.org/eng/contributions/latest so you would get "8 minutes ago, "an hour ago", etc.

ckjpn:
I've mentioned this elsewhere, but I really wonder if the supported languages line is really necessary for this block.  It's not something that changes all that often, and members can easily find this information by clicking the "stats per language" page.

Eliminating that from the block might help minimize the text on the page and also solve what TRANG mentioned,  "My only issue is that it feels a bit strange to have two stats ...." 

However, this one line doesn't really clutter up the page, unless including it also means more text has to be added to accommodate it.

I do think, however, that this is a good stat to have on the non-logged in home page that new visitors to the website see, since it gives them an idea of the scope of the project.

jiru:
Thank you agrodet for starting a new topic.

[...]

PR #2291, opened 2020-04-21
Move activity timeline link out of Contribute menu

jiru:
This PR implements the suggested enhancement in #2156.

Block for logged-in users:
image

Block for guests:
image

Don’t forget to run `rm tmp/cache/tatoeba_stats_element_homepage_stats_*` if you want to see the result of this change.


agrodet:
Phrased like this, there will be a discrepancy between "412 contributions today" and the activity timeline page where it will be 412 added sentences. But I think we can take care of that later :)

Issue #2157, opened 2020-02-21
"Browse by language" page is inconsitent with the other "Browse by" pages


trang:
Looking that the "Browse" menu:
- The page "Browse by list" shows the list of lists. Selecting a list displays the sentences in that list.
- The page "Browse by tags" shows the list of tags. Selecting a tag displays the sentences with that tag.
- The page "Browse by language" shows directly a list of sentences in a somewhat arbitrary language (for the first time you visit the page).

If we were to be consistent, "Browse by languages" should display the list of languages. Then selecting a language displays sentences in that language.

The most simple change we can do is to use the "[Number of sentences per language](https://tatoeba.org/eng/stats/sentences_by_language)" for "Browse by language".

- The URL could be changed to `/languages/index`.
- The title could indicate the number of supported languages instead of the total number of sentences.
- The "Number of sentences with audio by language" could be removed. Those stats are already displayed in the "Browse audio" page.

Later on we can improve the page with more information about each language.

agrodet:
> The title could indicate the number of supported languages instead of the total number of sentences.

The title could indicate both. "8 498 843 sentences in 358 languages" gives more useful information and (, more important from my view :P), makes sense both as a title for the "Browse by language" and the link from the Stats Block.

If you allow me to draw on the "consistency" issue, I wonder if the [Browse audio](https://tatoeba.org/eng/audio/index) page is useful as it is. I don't mean that browsing all languages is useless, actually it can be fun to hear languages we don't understand (although English is overwhelming).
I wonder if the best landing page when a user click on "Browse audio" is simply not the same as the one you mention (after thinking longer about it, no, it's not). On the current stats page, we have both sentences in languages and audio. To me, the title in the audio column `Number of sentences with audio by language` should be replaced by `{totalAudio} sentences with audio` or `{totalAudio} sentences have audio in {totalLang} languages`. That way, the title would give useful information, while it is currently just a plain column title.  
Then, the `Browse audio for all languages` link, at the bottom of the column, should send to the current `audio/index`. However, the page would give more insight if the sentences were displayed randomly or if both `random` and `last added` orders were available (default to random is the best for this page in my opinion).  

Now, I realize that I spoke mainly of the "Browse audio" while the title of the issue is about the "Browse by language". Excuse my bad manners, but since you mentioned the stats page, I think it's a good idea to improve this page at the same time. To summarize my points: 

- The title of future `languages/index` would give more information if it was "8 498 843 sentences in 358 languages"
[...]

PR #2294, opened 2020-04-24
New "browse by language" page

jiru:
Solves #2157. Got most of the design inspiration from some other website. :roll_eyes: 

The page feels quite heavy to load, very likely because of all the flags. Suggestions about optimization welcome.


src/Template/Element/sentences/index/all.ctp
```diff
@@ -0,0 +1,32 @@
...
+  <div class="language-list">
+    <ul>
+      <?php
+        $l = h(json_encode($languages));
+      ?>
+
+      <li ng-repeat="lang in <?= $l ?>">
+        <a ng-href="<?= $baseUrl ?>/{{lang.code}}/none">
+          <img class="language-icon" width="30" height="20"
+               ng-attr-title="{{lang.name ? lang.name : lang.code}}"
```
AndiPersti, line 25: Is it really necessary to show the tooltip on the flag when the language name is displayed right next to it? (This is more a general question which also concerns other pages outside the scope of this PR.)
Mobile users won't see it anyways.

AndiPersti:
I've just created a little demo page: http://tatoeba.forono.be

http://tatoeba.forono.be/flags.html:
````html
<!DOCTYPE html>
<html>
<head>
    <title>Flags demo</title>
    <meta charset="utf-8">
</head>
<body>
    <span id="flag_eng"><svg><use href="flags.svg#eng" /></svg></span>
    <span id="flag_fra"><svg><use href="flags.svg#fra" /></svg></span>
</body>
</html>
````

http://tatoeba.forono.be/flags.svg
````svg
<?xml version="1.0"?>
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
    <symbol id="eng" width="30" height="20">
        <image  xlink:href="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAB4AAAAUCAMAAACtdX32AAACJVBMVEX////4///SHDMAJH3JABDZMUTTHzTKABbMAxzIAAoAE3MAAGEAJ4DrT
1vvYGr0/v/ULUIAFXYNNYkAB2vOECdmjcIPOY3RHjQAAFbKAQ0AAGcAAGTQFy/PFiwAAmnVMUbVLkP//v4AH3rn9v1GbK0ADHDOAADk7fbcAAD88PEIMonr+v/87/D/+/v7///mjZbHAAOlyuwAAFHXRlXp8/nX
AAD/+fzvs7n/7PAAFnz/k4nLAAYDGnfWUlzPEyoAJX7XP1L//f3bTF3hAAHQLDj59/njSFb9S0H74ujonKLuUl3mWGYWJ37/9vbyY2vaWWT/pp0OIHr///qywNshPZFGgcXt//9Gb64hV6tpn9eiq81VZKP3Ojf
1y8kfMIS0vdj+8fUAAFhib6rQMz/0KSScpcnVPktHVJjxv8X/h3u50erNDRbODiYAAm3aNzpWktH0hobi0tr/uLTKARPXO04OQZsQOo1ifrcAIIPP3e0LPJfZUGLj1t1QW5uotNPSPkP3aGRBUZi0p8E/ZKj88v
PcU1b/u7gAK4r10NTw+v6rtNLje4TtkJHrgYG8prz66u5AW6Q1cbxghbyostLUISd9v+7C1u4XSKD/n5NVYKCtzexyptrs1djSOkVui8EABXBCXqbUFhthoNlgmNPHq7pTj87nl55AZafqo6r4Ny/tzc9hisAoX
7CsxeTqpq3vJy7ZKi/soKni0dmWn8Z4suLptbvOJzA7uKARAAAB3ElEQVQoz12Sd1cTURDFb7JrWd4qLLsxy26yJCEhkSakAGlgUFSUCNIRRIp0e++9N+y994ZK/Xxs8g6HwO+vefeeeeeemYGvG+4O1cUzTFqm
YTXHrkpfk8YwvPj+6+T/0Qn8C/8Fdu4gDn7BrnSQ0yX1gb5yYDzkGSiF/6XqEuZtQdx34Ozm3HXAbgTbf/0M/0h8IOcVJWxDpnyspH7Dxi0w9GqIkLaRkMcXh3/Qlb+dY7fmnGw+RFu7SASaKJPmLx+L9ff5nG0
cu8ndfzDReuqSVxY1OJ2t35s+tdy7UAbgDMf6cHh/DZB/oulKq9OJJIUGIB6z26s4dn2sEECs2k6dZUlstuzsjAybRZIsemXTS6pjRSp7LRZpkYCVqUhWq7RIwPIkRqPRrGOtqLCazYnSSHUaoGGs4CpQuLYq68
jRAh2gOk4dk8n0+8O7FmDo2nXU6ckB3Ky7W/T0wSPdgqL1eLsaURoNXLxMx9KIhjvFT+4/JF5NAaNGpoCyrD272m/QoQrnbqG7z/Pi1XPiABl0A9HA8LO2Hp6uJE+uvQ3UhDvffHahn7a+DirzCz0uiEpHOsrfe
kLjAKK5wyN/VCHlHHgHqX0MfAt3Ymh2dGY6qPCVS45J7fVjbGAOXcZoRuW5mukAAAAASUVORK5CYII=" />
    </symbol>
    <symbol id="fra" width="30" height="20">
        <image  xlink:href="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAB4AAAAUBAMAAABohZD3AAAABGdBTUEAALGPC/xhBQAAAAFzUkdCAK7OHOkAAAAbUExURRMii////90nN/73+
NoVJt83Rubp8xYljQQUhOYQryIAAAAaSURBVBjTY2AAgvZEQUFB4RAlIGAY5ZPEBwBdayjJpQdErwAAAABJRU5ErkJggg==" />
    </symbol>
</svg>
````

**Comment by jiru** [comment]
Hey, this is great, Andreas! Thank you for the code example. I will try to implement that next week.

**Comment by jiru** [comment]
Alright I’ve made some progress on this but it’s not 100% complete yet. All the flags are now served as one big SVG sprite file. The sprite file is generated by asset_compress.

There are still some limitations however:
1. Long language names overflow on the flag image (some CSS needs to be fixed).
2. When displayed from the SVG sprites, flags look blur (compared to their `` equivalent).
3. When the SVG file is not directly served from `webroot/cache_svg/`, asset_compress returns a content type of "application/octet-stream". This prevents the image from displaying, at least on Firefox. I’ll have to fix this in the asset_compress repository.
    ```sh
    # After running `cake asset_compress build`,
    # you should get this and images will display
    $ curl -s -v http://localhost:8080/cache_svg/allflags.svg 2>&1 >/dev/null | grep Content-Type
    < Content-Type: image/svg+xml
    
    # If you run `cake asset_compress clear` and have debug enabled,
    # you should get this and images won’t display
    $ curl -s -v http://localhost:8080/cache_svg/allflags.svg 2>&1 >/dev/null | grep Content-Type
    < Content-Type: application/octet-stream
    ```
    This problem shouldn’t prevent us from merging this PR because the production server is not affected by the problem; only local development environment are.

Also, I ran `composer update` to get a newer version of asset_compress, which incidentally updated all the other dependencies to their newer versions. This shouldn’t break anything but we never know.

**Comment by AndiPersti** [comment]
Some flag images are "real" svg images (currently the flags `ryu`, `frr`, `fro`, `evn`) so they won't display correctly if you don't set the correct `viewBox`. E.g. `ryu` should look like 
ryu_correct
but it is displayed like 
ryu_wrong

The reason is that without a `viewBox` `<svg class="language-icon" width="30" height="20">` will only render a `30x20` box in the upper left corner from the full image whose full size is `1050x700`. Setting the `viewBox` in the flag file (and removing `width` and `height` will allow the browser to scale down the image to `30x20`:
```svg
<symbol id="ryu" viewBox="0 0 1050 700">
```

(I've noticed this after [my comment](https://github.com/Tatoeba/tatoeba2/pull/2294#issuecomment-619361264) above).

AndiPersti:
> 2. When displayed from the SVG sprites, flags look blur (compared to their `` equivalent).

There seems to be a difference between browsers. The flags display sharper if I view the page on Chromium.

Firefox 77/Linux | Chromium 83/Linux
----- | ------
firefox | chromium

jiru:
(It looks like my last comment didn’t make it :disappointed:)
I solved all the problems I mentioned earlier. The only thing I’m still not satisfied with is the slowness of the page when switching tabs as a logged-in user. Maybe this can be solved with a simpler design.

jiru:
I updated AngularJS Material and it solved CPU issues. Thanks for the pointer @AndiPersti! :+1: Switching tabs is still slow, but not so slow compared to a page load, so we might as well keep it like this.

There is one last strange bug: after opening the page as a logged-in user, clicking on the second tab brings the contents of the third tab. Further tab switches show the expected contents.

AndiPersti:
> There is one last strange bug: after opening the page as a logged-in user, clicking on the second tab brings the contents of the third tab. Further tab switches show the expected contents.

Can you still reproduce it?
I've started a new VM and updated all necessary parts.

jiru:
> Switching between the Top10 and the profile languages is fast enough IMHO.

I agree, but the third tab is more laggy.

> Since almost everything is static on that page, I wonder whether generating the language lists on the server instead of using AngularJS `ng-repeat` would make any difference.

I’m going to check that.

> Can you still reproduce it?

Yes. Quite strange. I’m going to install this branch on dev and try from other browsers. Please let me know if you can reproduce the bug on dev too.

jiru:
It turns out removing `ng-repeat` on static content solved the tab-switching bug. :tada:

jiru:
Alright I found the real cause of the tab-switching bug: minification of the already-minified `angular.min.js`.

But I found another bug caused by the upgrade of AngularJS Material: the language drop-downs of the search bar are quite not opening (when clicking on the white area, not the ✘). You can see that on dev at the moment.

@AndiPersti It can’t see a substantial difference between server generated templates vs. `ng-repeat`. I left a the commit that removes the `ng-repeat`s, feel try to compare.

AndiPersti:
> Alright I found the real cause of the tab-switching bug: minification of the already-minified `angular.min.js`.

Yes, this seems to fix it. I can't reproduce it neither locally nor on dev with all devices/browsers I've tested.

> But I found another bug caused by the upgrade of AngularJS Material: the language drop-downs of the search bar are quite not opening (when clicking on the white area, not the ✘). You can see that on dev at the moment.

It looks like another problem with Firefox to me. On Chromium the language drop-downs still work. [*]
But even on Firefox it works when I don't click exactly on `Any language`.

[*] The only thing I've noticed is that clicking on the dropdown causes the window scrollbar on the right side to disappear and thus the whole content "moves". In an older version that movement doesn't happen although the scrollbar also disappears.

> @AndiPersti I can’t see a substantial difference between server generated templates vs. `ng-repeat`.

Neither do I.
I've compared some performance statistics (e.g. `DOMContentLoaded`, `First ContentfulPain (FCP)`, `Time to interactive (TTI)`) but the numbers are pretty close.
So I think the main problem is the layout reflow caused by the high number of elements in the third tab (`All languages`). Here's what I get in the browser console:
```
document.querySelectorAll('*').length
2328
document.querySelector('#tab-content-6').querySelectorAll('*').length
1737
```

jiru:
@AndiPersti Thanks for your continuous support. :smiley:

> The only thing I've noticed is that clicking on the dropdown causes the window scrollbar on the right side to disappear and thus the whole content "moves".

I solved that problem.

As for the dropdown flickering on Firefox, I’m quite sure it is triggered by our `onFocus()` and `onBlur()` handlers that clear/refill the text automatically. To be honest I’ve always been unsatisfied with this beheavior and I always wanted to get rid of this, so this could be the perfect excuse. :wink: I’d like to use "any language" as a placeholder (as opposed to a selectable item) and let users type in or remove text as they see fit. UI-wise, I think the absence of text matches better the absence of filtering. Another rationale is that during my UX tests I’ve seen at least 2 new users who treated the dropdown as a "regular" one and *not taking advantage at all of the keyboard to select languages during the whole session*. @trang What do you think?

jiru:
Alright I solved the dropdown issue on Firefox by using AngularJS Material 1.1.12 instead (so this PR only upgrades from 1.1.5 to 1.1.12).

jiru:
So this PR is ready for merge but I’m postponing the milestone once more because of the risk of regressions caused by updates of both AngularJS Material and all the libs of composer.json. I’m going to merge this branch on dev.tatoeba.org as of today to let everyone test.

jiru:
Note to self: South Levantine Arabic flag displays incorrectly.

trang:
> I’d like to use "any language" as a placeholder (as opposed to a selectable item) and let users type in or remove text as they see fit. UI-wise, I think the absence of text matches better the absence of filtering. Another rationale is that during my UX tests I’ve seen at least 2 new users who treated the dropdown as a "regular" one and _not taking advantage at all of the keyboard to select languages during the whole session_. @trang What do you think?

I have no objection to that. I also think that the absence of text can intuitively be associated with "Any language". I believe the addition of "Any language" as an item of the dropdown list might have been motivated by some technical reason... Perhaps because there was issue making the "X" really clear the selection?

If additionally to using "any language" as a placeholder, you also expect keyboard input before showing suggestions, this might be annoying to some users who have been used to selecting their language profile via click only. I can tell you that I sometimes use only the mouse to select languages even though I know I could take advantage of the keyboard. But that's because the effort it takes to use the keyboard doesn't feel worth the trouble, for instance if I know that the language I want is towards the top of the list or if I'm holding something in my hand (glass of water, ice cream...). I will only use the keyboard if I don't remember exactly the name of the language and don't know where it is in the list (or if it even exists).

jiru:
If I change the interface into Finnish, some language names, such as "jamaikankreoli/patois" or "ladino/juutalaisespanja" overflow. This happens because of our current "word-wrap: break-word;" CSS rule on all `div` and `p`. There are [several options to handle line wrap](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap), we should check carefully and do this right.

src/Template/Element/sentences/index/search.ctp
```diff
@@ -0,0 +1,29 @@
...
+  <md-button class="md-raised md-primary"
+             ng-cloak
+             ng-if="vm.selectedLanguage"
+             ng-href="<?= $baseUrl ?>/{{vm.selectedLanguage.code}}/none">
+    <?= format(
+            /* @translators: button appearing after a language search
+                             in Browse by languages page */
+            __('Show all sentences in {language}'),
+            ['language' => '{{vm.selectedLanguage.name}}']
+        ) ?>
```
jiru, line 27: This code does not allow using [declensions in language names](https://en.wiki.tatoeba.org/articles/show/interface-translation#declensions) (currently used in Finnish, Kirundi, Ukrainian, Russian and Arabic). It will always use the first declension even though the translator uses {language.something} in her translation.

jiru:
I fixed the language word wrap problem.

I also fixed the language declension limitation by fetching the formatted string using an extra query. Not an ideal solution, but I think it’s a good compromise. Let me know if you think it’s too hackish or overkill.

jiru:
> I can tell you that I sometimes use only the mouse to select languages even though I know I could take advantage of the keyboard. But that's because the effort it takes to use the keyboard doesn't feel worth the trouble, for instance if I know that the language I want is towards the top of the list or if I'm holding something in my hand (glass of water, ice cream...).

@trang You are right. I agree that it makes sense to select with the mouse in the case of a profile language. But if you want to select a particular item among the 380+ languages, I still think it would be considerate to the users not to initially drag their attention on such a long list. That said, I can imagine some users will want to go through the list no matter what. I wonder if we could add a sort of `v` icon exactly like where the `x` is, to allow opening the list when the field is empty.

src/Controller/AngularTemplatesController.php
```diff
@@ -17,4 +18,15 @@ public function beforeRender(Event $event)
...
     }
-}
\ No newline at end of file
+
+    public function show_all_sentences_button_text($code)
+    {
+        $this->response->header('Content-Type: application/json');
+
+        if (!LanguagesLib::languageExists($code)) {
+            return $this->response->withStatus(404);
```
AndiPersti, line 27: `Unknown language` is found in the search field, but selecting it will result in a 404 response because `LanguagesLib::languageExists('unknown')` is `false`.

trang:
> I can imagine some users will want to go through the list no matter what.

I don't think any user would really want to browse the whole list of languages through the dropdown list. It's not very comfortable because it's a small box, and even less so when there's this new "Browse by language" which makes browsing the list of languages much nicer. When there's no character input yet, I think the dropdown list can just to have suggestions of languages from the profile, or/and recently selected languages.

> I wonder if we could add a sort of `v` icon exactly like where the `x` is, to allow opening the list when the field is empty.

A `v` would very much make sense, yes. It would also match better the current behavior, as clicking on `x` doesn't exactly clear anything... It only visually clears the input while it is focused, but if you click outside, the last selected value comes back. There doesn't seem to be a way to customize this icon in the md-autocomplete.

Anyway, we don't have to solve this now. This pull request can be merged IMO and other improvements can come later 🙂 

@AndiPersti Do you take care of merging?

AndiPersti:
> @AndiPersti Do you take care of merging?

Yes, I can take care of it.

Issue #1613, opened 2018-07-19
Better way to prevent the creation of spam users


Guybrush88:
While browsing the users list, I noticed that there are some users that were created just for spam, like this one: https://tatoeba.org/ita/user/profile/mingletrain
Maybe the code should be refined to have better and stricter rules for the creation of new users.

ckjpn:
Looking at recently added members today, I noticed the same problem.

https://tatoeba.org/eng/user/profile/smvec
https://tatoeba.org/eng/user/profile/tradingexact
https://tatoeba.org/eng/user/profile/slikbeauty
https://tatoeba.org/eng/user/profile/jawlineme

Access to this profile is PUBLIC. All the information can be seen by everyone.


Perhaps making it impossible to set profiles to "Access to this profile is PUBLIC." until a member has contributed sentences would help solve this problem.  Likely it's more of an SEO, hoping Google will count them as links, than it is an attempt to spam our members.

If Google can't find the links, then perhaps a lot of this would be eliminated.

Seeing how the formatting of the profiles are so similar, likely only one or just a few SEO companies are doing this.

ckjpn:
Somewhat related, if we could somehow make it clearer that people didn't need to create an account to browse or download the data, then we would probably not have so many accounts being created.

I know that the number look impressive when looking at the number of "members," but that is a very deceptively high number when compared to the number of contributors.

It would likely help to very clearly state on the main page that you only need to create an account to write to the website (comments, wall, sentences, etc.) and view non-public members' profiles.  This might also encourage those who may incorrectly think registration is necessary to stay and use the website longer.

jiru:
> Maybe the code should be refined to have better and stricter rules for the creation of new users.

Maybe I’m wrong, but looking at the profiles you and CK mentioned, I think they have been created by a human rather than a machine. The country in the profile even match the company the link is pointing to. Which means captcha and the like are probably useless. If we make the creation of new accounts more difficult, it will be a barrier for legitimate users too. I think we need to tackle this problem differently.

jiru:
Related: #349.

trang:
Preventing the creation of spam users seems to be a suggested solution for a problem that is not clearly defined.

Let's talk first about the negative impacts of registration of spam users, then we can figure out what to do.

Guybrush88:
[...]

Issue #1528, opened 2018-02-25
When adding new sentences to directly on a list's page, the auto-language detect doesn't work.


ckjpn:
## The Problem
When adding new sentences to directly on a list's page, the auto-language detect doesn't work, and also the "cookie" of the last contributed language that's used elsewhere on the site isn't used.

![list-additions](https://user-images.githubusercontent.com/7934273/36637804-72cc9a3c-1a27-11e8-8c88-38f722ed4507.png)

I was assuming that if I contributed a sentence directly to a list, that instead of creating a new sentence, it would find the correct sentence and add it to the list, with a comment that the sentence already existed, in the same way it happens when trying to use the "add a sentence" forms works.

Guybrush88:
I just checked, and apparently the auto-detection feature still doesn't work anywhere on the site, not just within lists

ckjpn:
Perhaps, at least temporarily, maybe the same select menu that is on the "Add new sentences" page could be added.  Or, make it so that the cookie that's set for the last sentence added is read and used by this.

<img width="345" alt="screen shot 2018-04-13 at 8 52 00" src="https://user-images.githubusercontent.com/7934273/38709829-2482691a-3ef8-11e8-8f49-40066b8cddc4.png">

trang:
For the language detection, we already have an issue (#1479). But I think it makes sense to add the language dropdown in the list page. I'm changing the title of this issue.

ckjpn:
This is related:

It would be nice if adding a sentence directly to a list would use the same default language (set in a cookie, I think) that the "add a sentence" page uses.

Here are 2 examples of what happens.

https://tatoeba.org/eng/sentences_lists/show/7966
<img width="696" alt="screen shot 2018-07-26 at 9 45 20 am" src="https://user-images.githubusercontent.com/7934273/43240496-c026c926-90d1-11e8-96a4-fa7a72a137fb.png">


https://tatoeba.org/eng/sentences_lists/show/7073
<img width="673" alt="screen shot 2018-07-26 at 9 56 37 am" src="https://user-images.githubusercontent.com/7934273/43240500-c418ea8c-90d1-11e8-9a0b-a194f237e79e.png">

ckjpn:
This problem still exists.

This sentence was already in the database, so I thought I'd be able to add it to the list by just typing in the sentence instead of going to the sentence and then adding it to the list.

https://tatoeba.org/eng/sentences/show/31725
Once more, please.

[...]

PR #2295, opened 2020-04-24
Prevent adding sentences without profile language

AndiPersti:
While the UI prevents users without any profile language to add a sentence, it was still possible to bypass this rule by sending a direct POST request to `/sentences/add_an_other_sentence`.

### DIFF
--- a/src/Controller/SentencesController.php
+++ b/src/Controller/SentencesController.php
@@ -313,7 +313,7 @@ class SentencesController extends AppController
     {
         // Users without a profile language should not be able to add sentences
         if (empty(CurrentUser::getProfileLanguages())) {
-            throw new \Cake\Http\Exception\ForbiddenException;
+            return;
         }
 
         $userId = $this->Auth->user('id');
@@ -325,8 +325,7 @@ class SentencesController extends AppController
         $sentenceLang = $this->request->getData('selectedLang');
         $sentenceText = $this->request->getData('value');
 
-        if (is_null($sentenceText) || is_null($sentenceLang)) {
-            //TODO add error handling
+        if (empty($sentenceText) || empty($sentenceLang)) {
             return;
         }
 
--- a/src/Controller/SentencesController.php
+++ b/src/Controller/SentencesController.php
@@ -311,6 +311,11 @@ class SentencesController extends AppController
      */
     public function add_an_other_sentence()
     {
+        // Users without a profile language should not be able to add sentences
+        if (empty(CurrentUser::getProfileLanguages())) {
+            throw new \Cake\Http\Exception\ForbiddenException;
+        }
+
         $userId = $this->Auth->user('id');
         $userLevel = $this->Sentences->Users->getLevelOfUser($userId);
         if ($userLevel < 0) {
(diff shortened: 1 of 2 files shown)


ckjpn:
Note that some of the spam sentences came in yesterday with languages specified.


I wonder if we should ask for email confirmation when registering new accounts, or at least have the possibility for an admin to temporarily change a setting to force this.  This would slow down the kind of attack we had yesterday.

It's more friendly, I think, to not have to verify email addresses, but it does open us up to this kind of problem.

trang:
> Note that some of the spam sentences came in yesterday with languages specified.

Do you have some specific examples? In which language were they?

> I wonder if we should ask for email confirmation when registering new accounts

This has already mentioned in #1613. Note that email confirmation won't help much against bots. Checking emails and opening a confirmation link can be automated too.

AndiPersti:
I'm aware that this fix won't help much in preventing attacks by a determined attacker (it should be really easy to adapt their process to add a profile language).

ckjpn:
> Do you have some specific examples? In which language were they?

Here is one.

https://tatoeba.org/eng/users/show/89213

trang:
> https://tatoeba.org/eng/users/show/89213

Thanks @ckjpn. 

This user indeed added Mandarin Chinese to their profile before they started adding sentences. But none of the sentences added were set to Mandarin Chinese. They were all created with a language `null`.

This would normally not be possible from the current UI. On the UI, if you only have one language in your profile, this language is pre-selected in the dropdown list and you cannot switch to another language. This may be another rule we want to implement on the backend side.

AndiPersti:
> ...They were all created with a language `null`.

https://github.com/Tatoeba/tatoeba2/blob/66e4c9386641c2da7def190f0bc7288bbc395f44/src/Controller/SentencesController.php#L325-L331
So the test on line line 328 should be `empty($sentenceLang)`, shouldn't it? And I think we should also change the test for `$sentenceText` because an empty text isn't allowed by the UI either (and wouldn't be stored any ways).

trang:
As mentioned in the weekly call, we would need to solve #1528 before we can merge this. If this PR is merged now, any sentence which language could not be auto-detected will not be added, if created from the list page with the old design.

#1528 is solved for those who use the new sentence design, but not for those who are still in the old design. It would be really worth to implement the language selection in the old design. We're very close to transitioning to the new design as the default one so let's just wait.

We'll deal with the spam by checking for patterns in the sentence text.

jiru:
If I understand correctly, #1528 is not related to this PR because the protection happens on the controller level, not the model. The SentencesLists controller is not affected by this PR. So I think it is safe to merge as-is.

AndiPersti:
Yes, this PR doesn't change the behavior of the SentencesLists controller. With this PR, sentences without a language can still be added from the old list page.
(When I've written the patch I've assumed the spammers abused the POST endpoint in the SentencesController.)

trang:
> The SentencesLists controller is not affected by this PR. So I think it is safe to merge as-is.

Ah yes, indeed. This is safe to merge.

PR #2299, opened 2020-04-28
Add cleanup section to Jenkinsfile

AndiPersti:
This should cleanup the workspace after running the tests and therefore should prevent filling up the disk on the dev server.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -28,4 +28,9 @@ pipeline {
       }
     }
   }
+  post {
+    always {
+      deleteDir()
+    }
+  }
 }


jiru:
Well, I already configured Jenkins to only keep the latest 20 branches and remove the rest. I wonder how your approach differs. In particular, will `deleteDir()` remove the branch from the list of branches displayed by Jenkins?

AndiPersti:
> In particular, will `deleteDir()` remove the branch from the list of branches displayed by Jenkins?

It looks like, it will only display branches which have subdirectories in the workspace.

AndiPersti:
> It looks like, it will only display branches which have subdirectories in the workspace.

Actually I think I was wrong. If I'm not misunderstanding the docs, the setting you've used removes everything of the old branches (workspace directory, logs, ...) while `deleteDir()` will only remove the subdirectory in `/var/lib/jenkins/workspac`.

So it mostly depends on whether we want to keep the logs of past branches. But I wonder whether we even need to keep the workspaces of the latest 20 branches. These are just "disposable directories' according to the [Jenkins glossary](https://www.jenkins.io/doc/book/glossary/#workspace).

jiru:
I see. So your PR would free up the space used by the latest 20 branches, right? In this case, it makes sense to merge it as a complementary measure to keep used disk space low.

On the other hand, I wonder if we would ever need to manually go check something in the Jenkins workspace when we are unable to reproduce a problem locally. I think I did that once.

AndiPersti:
> So your PR would free up the space used by the latest 20 branches, right?

It would only delete the working directory of the test which has finished. It wouldn't even delete the related `@script` and `@temp` directories. See https://issues.jenkins-ci.org/browse/JENKINS-41805
(I don't know why some PRs have a `@script` directory and others not)

PR #2306, opened 2020-05-05
Reviews in new sentence design

trang:
This PR implements the reviews feature in the new sentence design. Nothing much more to say on that. It works the same way as in the old sentence design.

Since this is the last feature, I updated the announcement about the new sentence design and refactored the code at the same time. The updated announcement will be visible to every logged in user, not just people who have the random sentence displayed on the homepage.

### DIFF
--- a/src/Model/Entity/Sentence.php
+++ b/src/Model/Entity/Sentence.php
@@ -34,7 +34,8 @@ class Sentence extends Entity
         'lang_tag',
         'is_favorite',
         'is_owned_by_current_user',
-        'permissions'
+        'permissions',
+        'current_user_review'
     ];
 
     protected $_hidden = [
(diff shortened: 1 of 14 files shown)


src/Model/Entity/Sentence.php
```diff
@@ -136,4 +137,9 @@ protected function _getPermissions()
             return null;
         }
     }
+    
+    protected function _getCurrentUserReview()
+    {
+        return CurrentUser::correctnessForSentence($this->id);
+    }
```
jiru, line 144: Why not use a containment of the UsersSentences model from the Sentence model to get the correctness?

jiru:
With the new design, when clicking on one of the review buttons, there is no more spinner to indicate that something’s loading. It makes the clicking experience not as smooth as it used to be.

jiru:
About the banner. My browser is configured to remove all cookies after closing, for obvious privacy reasons. Because of this, the banner is showing up every time I open my browser, regardless of me clicking on "Close" every time.

jiru:
Thanks for the update. The code looks all good. :+1:

PR #2313, opened 2020-05-08
Avoid unnecessary calls to the search engine

AndiPersti:
CakePHP's paginator sends two queries to the database: the first for getting the items of a page and the second for the total count.

But when we use the search engine we get the total count already with the first call. The `withSphinx` finder from the `Sentences` model already uses the cached result for the total count but we still queried the search engine twice in the `Sphinx` behavior.

### DIFF
--- a/src/Model/Behavior/SphinxBehavior.php
+++ b/src/Model/Behavior/SphinxBehavior.php
@@ -63,7 +63,12 @@ class SphinxBehavior extends Behavior
      */
     function beforeFind($event, $query, $options, $primary)
     {
-        if (empty($options['sphinx'])) {
+        /* CakePHP's paginator makes two calls to the database: the first for the actual
+         * query and the second for the total count. But when we use the search engine
+         * we already get the total count with the first call. The 'withSphinx' finder
+         * from the model will use the cached count so we don't need to do anything
+         * when this callback gets called the second time. */
+        if (empty($options['sphinx']) || $this->_cached_result) {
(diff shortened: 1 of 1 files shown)


jiru:
Thank you. I’d like to mention some background about this.

This code is [based on a 10+ years old code](https://github.com/nabeelio/Sphinx-CakePHP) initially written by the Sphinxsearch team.

We used to have an if-block that prevented running the extra query. But when we migrated from CakePHP 2 to CakePHP 3, that if branch got removed (in 245c463a02ebddd56f23336e21a38dfbf43fc65d), and it remains are still visible:
https://github.com/Tatoeba/tatoeba2/blob/1d2e08e3eaf65a739b6247f0635bd42000fcca00/src/Model/Behavior/SphinxBehavior.php#L74-L78

However, the SphinxBehavior’s beforeFind is not supposed to be triggered by the paginator count call, because we are using a custom counter defined there:
https://github.com/Tatoeba/tatoeba2/blob/c2fa6405430913bc8efea2e96ff32d6b86828011/src/Model/Table/SentencesTable.php#L448-L457

Instead, the call is triggered by some internal call to `\Cake\ORM\Query::cleanCopy()` before our counter is actually executed. Strange.

Note that Manticore recently published a [brand new client](https://manticoresearch.com/2020/03/27/manticore-3-4-0-online-schema-change-new-php-client-and-docker-image/) that I hope we can migrate to at some point.

Issue #2046, opened 2019-12-26
Replace CSS property "box-shadow" under language icons with "filter: drop-shadow"


sabretou:
Currently, we use the CSS rule box-shadow to create the shadows under language icons. This is good enough for most icons, which are rectangular in shape. However, this ends up misrepresenting non-rectangular icons, such as the flags of Marathi, Nepali, Awadhi, Old Saxon, and so on.

For eg. this is our current implementation:
![box-shadow example](https://i.imgur.com/AVdk3By.png)

On the other hand, if we used the "filter: drop-shadow" property, we can have the shadow adapt to the actual shape of the icon.

This is how it can look with filter: drop-shadow:
![filter: drop-shadow example](https://i.imgur.com/nq2W7HB.png)

The exact parameters I used for the above effect are:
`filter: drop-shadow(2px 2px 1px #737373);`

These can of course be adjusted as necessary, especially the colour. I only chose these parameters to throw together a quick mock-up.

sabretou:
@trang Just pinging to see if there's a hold up regarding this. I've personally been using a Stylish script for some time now to achieve this effect, but I see this hasn't been implemented yet on the site itself.

trang:
No there's no hold up (not on my side at least). There just needs to be someone to volunteer for this.

As you can see it has been marked as "good first issue" meaning that it's easy to solve and would be perfect as a first task for a new developer.

sabretou:
Ahh, I assumed good first issue was for the ones starting the issue. It left me quite perplexed, to say the least. Thank you for clarifying!

trang:
In case you are puzzled by other labels, know that you can read the description of a label if you hover over it.

You can also find the descriptions on the "Labels" page: https://github.com/Tatoeba/tatoeba2/labels

And even more information on our wiki page: https://github.com/Tatoeba/tatoeba2/wiki/Labels

🙂

speedysera:
I'm going to work on this

PR #2314, opened 2020-05-08
Replace CSS property "box-shadow" under language icons with "filter: drop-shadow"

speedysera:
This pull request addresses issue #2046 - use filter: drop-shadow

### DIFF
--- a/webroot/css/layouts/default.css
+++ b/webroot/css/layouts/default.css
@@ -723,7 +723,7 @@ rt {
  }
 
 .language-icon {
-    box-shadow: 1px 1px 3px 0px rgba(204,204,204,1);
+    filter: drop-shadow(1px 1px 3px rgba(204,204,204,1));
 }
 
 


jiru:
Hi speedysera! Welcome and thank you for your pull request! :smile:

I had a look at how the flag shadow looked with your PR. If I understand correctly, it makes the shadow fit the actual contour of the flag instead of the square image, right? That is all good, but it seems that your shadow is much bigger than what we currently use:
flags_shadow

If the current shadow is barely perceivable (I have to zoom in to actually see it), yours, in contrast, it a bit too thick in my opinion. Maybe we can find something in between?

speedysera:
Hi jiru! 

   Thanks for the feedback! This time, I've copied the old parameters, please take a look if it's acceptable.

thanks,
speedysera

AndiPersti:
Welcome speedysera!

Just a technical note: For the future I would recommend that you create a separate branch for each PR. Do you know how to do that with git?

jiru:
@speedysera I don’t know why, but it looks like giving the same parameters to `drop-shadow` and `bow-shadow` produce different results (tested on Firefox and Chromium). `drop-shadow` produces a much bigger shadow. It seems that
```css
filter: drop-shadow(1px 1px 3px black);
```
is equivalent (in terms of rendering) to:
```css
drop-shadow(1px 1px 1.33px black)
```

So I suggest that you reduce the blur-radius to 1.33px.

speedysera:
> Just a technical note: For the future I would recommend that you create a separate branch for each PR. Do you know how to do that with git?

Thanks, @AndiPersti! Did you mean something like `git flow feature` or a manual branch creation? I have a chance to try it out to address @jiru 's comment :)

AndiPersti:
It depends on your work style. I'm working on the command line and thus I create branches manually using `git checkout -b <new-branch> dev` (usually after updating the `dev` branch to the current version). But you could also use some plugins or a GUI for creating a new branch.

speedysera:
Thanks @AndiPersti ! Please refer to https://github.com/Tatoeba/tatoeba2/pull/2327 for the new Pull Request

PR #2317, opened 2020-05-11
Improve AngularJS performance a little bit

AndiPersti:
According to the [docs](https://docs.angularjs.org/guide/production) disabling debug info and comment and CSS class directives should improve the perfomance a little bit.

### DIFF
--- a/webroot/js/responsive/app.module.js
+++ b/webroot/js/responsive/app.module.js
@@ -3,9 +3,13 @@
 
     angular
         .module('app', ['ngMaterial', 'ngMessages', 'ngCookies', 'ngSanitize'])
-        .config(['$mdThemingProvider', '$mdIconProvider', '$httpProvider', '$cookiesProvider', function(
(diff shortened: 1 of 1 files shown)


jiru:
Shouldn’t AngularJS’s debug be activated when debug is enabled in app_local.php?

AndiPersti:
You can always enable the debug info from the console using `angular.reloadWithDebugInfo()` so I don't think it's necessary that it is enabled by default even in a development environment.

jiru:
I’m just afraid that we won’t see the stacktraces when something fails while loading the page.

AndiPersti:
This change has no effect on stacktraces. See [Disabling Debug Data](https://code.angularjs.org/1.6.1/docs/guide/production#disabling-debug-data) for the things which will be disabled by this change.

jiru:
Alright then! Thank you for answering my questions :slightly_smiling_face:

Issue #2296, opened 2020-04-25
Refactor search code


jiru:
The `search()` action of the sentences controller is fat and untested. We need to refactor it into a reusable and tested module. I plan to do that next week. :factory_worker:

PR #2321, opened 2020-05-12
Refactor search

jiru:
My efforts to refactor the search action of the sentences controller (#2296). This PR introduces two new classes:

1. App\Model\Search, a lightweight class that represents a set of search criteria and sort order. It can output it all as an array usable by SphinxBehavior. One may use that class from any model to run `find()` queries with search:
```php
$search = new Search();
$search->filterByLanguage('por');
$search->filterByAudio(true);
$search->filterByOwnerId(42);
$this->Sentences->find('all', ['sphinx' => $search->asSphinx()]);
```

2. App\Form\SentencesSearchForm, a class that represents everything that is specific to the sentences search form. It parses the GET parameters, apply default criteria, warns about unwanted combinations, etc. Everything that is related to search but only in the context of the form.


AndiPersti:
I like the general approach but there is a little problem: it doesn't work yet :-)

https://github.com/Tatoeba/tatoeba2/blob/0057bffa6db34fe2033a64149489e49d4915b210/src/Controller/SentencesController.php#L512-L513
`$sphinx` isn't used anymore and thus this settings aren't sent to the search engine. Here's is what the query log shows when I search with the current version of your changes:
````
[Thu May 14 15:59:24.224 2020] 0.000 sec 0.000 sec [ext2/2/ext 38 (0,1)] [eng_main_index,eng_delta_index]
                                                                  ^^^^^
                                                      offset/page = 0, limit = 1
````
i.e. I always get only one sentence.

jiru:
Thanks for taking time to test it, @AndiPersti :smile_cat: The problem should be fixed now.

AndiPersti:
> The problem should be fixed now.

Yes, looks good after a few tests :+1:

PR #2322, opened 2020-05-13
Fix licences bug in add sentence form (new design)

AndiPersti:
There is a bug in the form for adding a new sentence in the new design:

1) Create a new user or log in as a user who cannot switch the license. The default license should be `CC BY 2.0 FR`.
2) Set the new design in the settings.
3) Add a new sentence using the form on `sentences/add`.

=> The new sentence will be stored with the CC0 license.

The reason: 
https://github.com/Tatoeba/tatoeba2/blob/2dd631f83a255d4817c409f678b0ce60c64ec557/webroot/js/sentences/add.ctrl.js#L36
The first entry in `vm.licenses` isn't `CC BY 2.0 FR`.

My fix simply removes the line because 
* Users who cannot switch the license will submit an empty license and thus the license will be set to the user's default license in the model.
* Users who can switch the license will have a select control where the user's default license is preselected.

### DIFF
--- a/webroot/js/sentences/add.ctrl.js
+++ b/webroot/js/sentences/add.ctrl.js
@@ -33,7 +33,6 @@
                 vm.newSentence.lang = vm.showAutoDetect ? 'auto' : langCodes[0];
             }
             vm.licenses = licenses;
-            vm.newSentence.license = Object.keys(vm.licenses)[0];
         }
 
         function addSentence() {


AndiPersti:
The bug was introduced in ee7f5824 which was deployed on March 15.

There are currently 1975 sentences which have the wrong license:
````sql
MariaDB [tatoeba]> select count(s.id) from sentences s join users u where s.user_id = u.id and s.license = 'CC0 1.0' and u.settings not regexp '"can_switch_license":"1"' and s.created > "2020-03-14";
+-------------+
| count(s.id) |
+-------------+
|        1975 |
+-------------+
1 row in set (0.93 sec)
````

jiru:
Good catch! How are we gonna deal with the license information stored in the `contributions` table? Should we overwrite the logs, or add an additional "License changed" log entry?

trang:
Hmm... I'd personally prefer to add an additional log entry. It's more transparent. I can't find an advantage for overwriting the logs in this case.

PR #2323, opened 2020-05-13
changed tooltip description from "show less features" to "show fewer …

bakananbanjin:
Same as the description just a quick change in words

changed tooltip description from "show less features" to "show fewer features"

(I have an other problem, since I fetched the last version, see mail.)

greetings


trang:
Thank you for this pull request @bakananbanjin :)

You actually didn't need to change the *.po and *.pot files because they are updated automatically. You can read more about it in the [README of the Locale folder](https://github.com/Tatoeba/tatoeba2/blob/dev/src/Locale/README.md).

In itself it doesn't break anything, but it makes the diff a bit crowded. Could you revert the changes on these locale files so that only the changes on the *.ctp file remain?

PR #2324, opened 2020-05-14
changed tooltip description from "show less features" to "show fewer …

bakananbanjin:
…features"

second try this time it should be the correct way


jiru:
Yes, it’s alright now! :+1:
Thank you for your first contribution, bakananbanjin, and welcome here. :slightly_smiling_face:

PR #2327, opened 2020-05-16
Replace CSS property "box-shadow" under language icons with "filter: drop-shadow" - improved version

speedysera:
This pull request addresses issue Tatoeba#2046 - apply better shadows to non-rectangular icons, with considerations for both Firefox and Chromium

### DIFF
--- a/webroot/css/layouts/default.css
+++ b/webroot/css/layouts/default.css
@@ -723,7 +723,7 @@ rt {
  }
 
 .language-icon {
-    box-shadow: 1px 1px 3px 0px rgba(204,204,204,1);
+    filter: drop-shadow(1px 1px 1.33px black);
 }
 
 


webroot/css/layouts/default.css
```diff
@@ -723,7 +723,7 @@ rt {
  }
 
 .language-icon {
-    box-shadow: 1px 1px 3px 0px rgba(204,204,204,1);
+    filter: drop-shadow(1px 1px 1.33px black);
```
jiru, line 726: You changed the color of the shadow to `black`. Because of this, the shadow now looks much stronger than it used to be. Could you revert it to the original value `rgba(204,204,204,1)`?

speedysera: Thanks @jiru, please refer to https://github.com/Tatoeba/tatoeba2/pull/2334

Issue #1858, opened 2019-04-04
Contributing translations in CC0


trang:
The CC0 license is starting to get momentum and our users are starting to feel limited for not being able to create translations in CC0, as shown in this Wall thread:
https://tatoeba.org/eng/wall/show_message/31587#message_31587

We will need to:
1. Allow users to choose the license when adding a translation to a CC0 sentence.
2. Allow users to change the license from CC BY to CC0 not just on their original sentences, but also on their translations of CC0 sentences.
3. Provide a way to switch all eligible translations to CC0 (either by evolving the `switch_my_sentences` page, or creating another similar page dedicated for translations).

jeanm:
May I also suggest two small quality of life improvements to consider?

1. To ensure that translations added via the “add translation” icon on the “contribute > add sentences” interface match the license specified at the top.
2. More generally, to have a way for users to set CC0 as their default license, wherever possible. Or more radically, a way for users to specify “upgrade my license to cc0 wherever possible, even in the future” which is more or less what I was suggesting on the wall. I realise the latter is probably more annoying to implement.

I imagine these would significantly decrease the reasons why users would need to use the “switch all my sentences to cc0” command.

aucampia:
I think it is quite critical to try and encourage a move to CC0. Wikidata is only CC0 and therefore use of Tatoeba data inside Wikidata will be problematic.

The current situation is also absurd in that very simple sentences such as ["How are you doing?"](https://tatoeba.org/eng/sentences/show/38926) are CC BY 2.0 FR and there is no way to make a new sentence that is CC0 for the same.

trang:
Could you elaborate why you feel that the CC BY license makes our data useless? Are you facing any particular issues in reusing our data due to the CC BY license? Or have you seen other projects not being able to reuse our data because of CC BY?

aucampia:
@trang Wikidata has CC0 policy for data: https://www.wikidata.org/wiki/Wikidata:Licensing#Official_policy so importing data or derived data from Tatoeba to Wikidata would be a problem.

AndiPersti:
As I understand it, Wikidata is about structured data. So how would the sentences in Tatoeba be useful for Wikidata?

Yorwba:
[...]

PR #2329, opened 2020-05-16
Allow users to set default license

AndiPersti:
Add a select control to the settings page for users who can switch license.

This was requested [on the wall](https://tatoeba.org/eng/wall/show_message/34688#message_34688) and also [in a related issue](https://github.com/Tatoeba/tatoeba2/issues/1858#issuecomment-480433240).


jiru:
Thank you! :+1:

Issue #2293, opened 2020-04-22
Show sentences as tooltips when the mouse is positioned on sentence numbers in the Logs block


soliloquist-tatoeba:
I mean this block.

![Logs_2](https://user-images.githubusercontent.com/47579687/79962765-2723bf80-8491-11ea-85ca-6da3f229b2f6.png)

This feature is already present on the Wall, in comments and private messages. If the Logs block had it too, it would be easier to see who linked/unlinked what, without leaving the page.

jiru:
Did you know that you can click on the log entry to highlight the corresponding sentence?

soliloquist-tatoeba:
Yes, it's very useful, but it naturally doesn't include unlinkings.

agrodet:
I have wrote the requested improvement, extending it to users last contributions. I have tested several combinations. At first I only added the tooltip to "link" logs, since the sentence logs only display those. Besides, I argued with myself that "Since the sentence is displayed, there is no need for the tooltip."  
However, in the user contributions, it felt strange to have some "#nnn" with a tooltip and others without. Eventually, even if `sentence` type (`add` and `delete` action) logs display the text of the sentence, I think displaying the redundant tooltip is better than not at all. 

I mean that it becomes like this. 
<img width="246" alt="image" src="https://user-images.githubusercontent.com/42261966/80505451-4ea5eb00-89af-11ea-8a26-aa946cd5b918.png">

That's a very small point. If it's okay with everybody, I'll open a PR.

trang:
Having the tooltip would definitely be an improvement to the current situation but I don't think it will be the long term solution.

For the long term solution, we should explore the idea of showing the sentence's text when it's a log of type `link` (that's why I assigned this issue to myself).

I've been thinking for a while now on how to redesign the sentence's page to be responsive, which led me to think about how to redesign the logs. One of my ideas is to have several [tabs](https://material.angularjs.org/latest/demo/tabs) on the sentence's page and one of these tabs would be for the logs. This idea led me to ask myself what was need for highlighting a translation when clicking on a log entry because if the logs are displayed in a separate tab, then highlighting translations doesn't make sense anymore.

If the translation highlight mechanism was just for users to know what is the text corresponding to the ID in the logs, then why not just display that text right away? If we would just display the text, then there would be no need of having a tooltip. It would remove the need of solving #2252.

[...]

PR #2330, opened 2020-05-17
Add tooltips to #n log links

agrodet:
This PR adds a tooltip containing the target sentence text to `#n` links that appear in logs. It addresses #2293.

I factorized the code of the logs a little bit.  
I would also like to simplify the return statement of the following piece of code but I'm not sure how I should correct the corresponding tests
https://github.com/Tatoeba/tatoeba2/blob/70d6fc5200c9697d57f9f1db59096a3d78b44766/src/View/Helper/ClickableLinksHelper.php#L131-L138
Also, I'm not sure about how to add a note to translators so I tried to add it the same way I saw it somewhere else in the code.

PS: @jiru I didn't really understand your intention in your [comment about using view cells](https://github.com/Tatoeba/tatoeba2/issues/2293#issuecomment-627499720)

### DIFF
--- a/src/View/Helper/ClickableLinksHelper.php
+++ b/src/View/Helper/ClickableLinksHelper.php
@@ -190,7 +190,7 @@ class ClickableLinksHelper extends AppHelper
                 'action' => 'show',
                 $sentenceId
             ),
-            ['title' => $sentenceText]
+            ['title' => $this->_View->safeForAngular($sentenceText)]
         );
     }
 }
--- a/src/View/Helper/ClickableLinksHelper.php
+++ b/src/View/Helper/ClickableLinksHelper.php
@@ -177,6 +177,11 @@ class ClickableLinksHelper extends AppHelper
      */
     public function buildSentenceLink($sentenceId, $sentenceText = null)
     {
+        $tooltipTag = $this->Html->tag(
+            'md-tooltip',
+            $this->_View->safeForAngular($sentenceText),
+            ['ng-cloak']
+        );
         $linkText = format(
             __(
                 /* @translators: You can translate the sharp in the link to a sentence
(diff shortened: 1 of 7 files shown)


jiru:
About the tests of that part of `ClickableLinksHelper.php`, they are actually in `tests/TestCase/View/Helper/MessagesHelperTest.php`. The ClickableLinks helper code is a bit different from the use case of your PR, because it needs to build a full URL (including the prefix https://tatoeba.org/...) because the link will show up in email notifications too. As opposed to that, your code doesn’t need the full links because it’s only used within pages. These two pieces of code could probably be factorized, but it’s important to understand this difference.

About the notes to translators, the way you did it is okay. If in doubt, you can try to run `./tools/generate_pot.sh` and see if your note show up in `src/Locale/default.pot`. I think the note can be either inside or outside the `__()`, as long as it is located on the line before the string itself.

About the view cells, I never used them so I’m not 100% sure, but basically, calling a model method from a view (using `TableRegistry::getTableLocator()` and so on) is a violation of the MVC principle, so we should avoid it as much as we can. But there are some specific cases when it’s sometimes needed, such as the clickable links. View cells give a proper way to achieve this. But it’s okay if you don’t want to go that far. :wink:

jiru:
There is actually a way to get the sentence directly from the controller. You can modify `ContributionsTable::getContributionsRelatedToSentence()` so that it returns the sentence text of related sentences and translations using containments. Then, from the view, you can just get the related sentence text with `$log->sentence->text` and the related translation text with `$log->translation->text`. That would be a better design I think.

jiru:
I did what I mentioned on my previous comment. Additionally, I made it so that the message "This sentence was initially added as a translation of sentence #n" also shows the sentence text as a tooltip.

src/View/Helper/ClickableLinksHelper.php
```diff
@@ -183,16 +183,14 @@ public function buildSentenceLink($sentenceId, $sentenceText = null)
...
         return $this->Html->link(
             $linkText,
             array(
                 'controller' => 'sentences',
                 'action' => 'show',
                 $sentenceId
             ),
-            // $sentenceText could come as the empty string
-            array('title' => $sentenceText != null ? $sentenceText : $model->getSentenceTextForId($sentenceId))
+            ['title' => $sentenceText]
```
AndiPersti, line 193: The sentence text needs to be sanitized using `$this->_View->safeForAngular()`.

BTW: Shouldn't we start using `md-tooltip` for all new tooltips? (see #1221)

jiru: Good catch about the XSS! :sunglasses: 
I also updated the code to use `md-tooltip`.

Issue #1429, opened 2017-02-15
for each tag displayed with autocompletion, show number of sentences with the tag


alanfgh:
When we get tag autocompletion working again (see issue #1415), we could think about implementing a suggestion by wallebot that we display next to each possible tag the number of sentences that have that tag. (We currently do this from the "Browse all tags" page.) This would help people make sure that they're not misspelling the name of a common tag.

jiru:
Setting the 'good first issue' label because the sentence number is already there but unused. The value could be displayed by editing `webroot/js/autocompletion.js`.

Ynue:
I'd like to work on this issue!

AndiPersti:
> Setting the 'good first issue' label because the sentence number is already there but unused.

Even if we [fix the bug](https://github.com/Tatoeba/tatoeba2/pull/2332) I've introduced a few weeks ago, the numbers we get from `suggestd` aren't really useful unless we restart it after each tagging event. The number of sentences is only current at the time of initialization but will not get updated later on. So the longer `suggestd` is running more and more tags will have a wrong count.

Here's how one tag looks like on my dev machine:
````
vagrant@stretch:~/Tatoeba$ echo 'select t.name, t.nbrOfSentences, count(ts.id) from tags t join tags_sentences ts where t.id = ts.tag_id and t.id = 30;' | mysql tatoeba
name    nbrOfSentences  count(ts.id)
foo     6       6
````
After restarting `suggestd` I get the expected result back:
````
vagrant@stretch:~/Tatoeba$ sudo systemctl restart suggestd
vagrant@stretch:~/Tatoeba$ wget -q -O - 'http://127.0.0.1:8080/suggest?str=foo&sort=1&size=10'
<?xml version="1.0" encoding="utf8"?>
<items>
        <item value="6">foo</item>
</items>
````
But after simulating adding a tag, i.e. use the API call from `SuggestdListener` the value doesn't change:
````
vagrant@stretch:~/Tatoeba$ wget -q -O - 'http://127.0.0.1:8080/add?str=foo&value=1'
OK
vagrant@stretch:~/Tatoeba$ wget -q -O - 'http://127.0.0.1:8080/suggest?str=foo&sort=1&size=10'
<?xml version="1.0" encoding="utf8"?>
<items>
        <item value="6">foo</item>
</items>
````

Looking at the source code for `suggestd` that's no surprise:
[...]

PR #2331, opened 2020-05-17
Add number of sentences with tag in autcomplete

Ynue:
This pull request addresses issue #1429.

### DIFF
--- a/src/Controller/TagsController.php
+++ b/src/Controller/TagsController.php
@@ -260,7 +260,8 @@ class TagsController extends AppController
         $query = $this->Tags->find();
         $query->select(['name', 'id', 'nbrOfSentences']);
         if (!empty($search)) {
-            $query->where(['name LIKE :search ESCAPE "|"'])->bind(':search', str_replace('%', '|', $search).'%', 'string');
+            $pattern = str_replace(['\\', '%', '_'], ['\\\\', '\%', '\_'], $search).'%';
+            $query->where(['name LIKE :search'])->bind(':search', $pattern, 'string');
         }
         $allTags = $query->order(['nbrOfSentences' => 'DESC'])->limit(10)->all();
 
--- a/src/Controller/TagsController.php
+++ b/src/Controller/TagsController.php
@@ -260,7 +260,7 @@ class TagsController extends AppController
         $query = $this->Tags->find();
         $query->select(['name', 'id', 'nbrOfSentences']);
         if (!empty($search)) {
-            $query->where(['name LIKE :search ESCAPE "_"'])->bind(':search', str_replace('%', '_', $search).'%', 'string');
+            $query->where(['name LIKE :search ESCAPE "|"'])->bind(':search', str_replace('%', '|', $search).'%', 'string');
         }
         $allTags = $query->order(['nbrOfSentences' => 'DESC'])->limit(10)->all();
 
--- a/src/Controller/TagsController.php
+++ b/src/Controller/TagsController.php
@@ -259,12 +259,10 @@ class TagsController extends AppController
 
         $query = $this->Tags->find();
         $query->select(['name', 'id', 'nbrOfSentences']);
-
         if (!empty($search)) {
-            $query->where(['name LiKE' => "$search%"]);
+            $query->where(['name LIKE :search ESCAPE "_"'])->bind(':search', str_replace('%', '_', $search).'%', 'string');
         }
-        $query->limit(10);
-        $allTags = $query->all();
+        $allTags = $query->order(['nbrOfSentences' => 'DESC'])->limit(10)->all();
 
         $this->loadComponent('RequestHandler');
         $this->set('allTags', $allTags);
--- a/src/Controller/TagsController.php
+++ b/src/Controller/TagsController.php
@@ -257,22 +257,18 @@ class TagsController extends AppController
     {
         $this->helpers[] = 'Tags';
 
-        $conditions = [];
-        if (!empty($search)) {
-            $conditions = [
-                'name LIKE' => "$search%"
-            ];
-        }
-        $this->paginate = [
-            'limit' => 10,
-            'fields' => ['name', 'id', 'nbrOfSentences'],
-            'order' => ['nbrOfSentences' => 'DESC'],
-            'conditions' => $conditions
-        ];
+        $query = $this->Tags->find();
+        $query->select(['name', 'id', 'nbrOfSentences']);
 
-        $this->autoRender = false;
-        $this->response = $this->response->withType('json');
-        $allTags = $this->paginate();
-        echo json_encode($allTags);
+        if (!empty($search)) {
+            $query->where(['name LiKE' => "$search%"]);
+        }
+        $query->limit(10);
+        $allTags = $query->all();
+
+        $this->loadComponent('RequestHandler');
+        $this->set('allTags', $allTags);
+        $this->set('_serialize', ['allTags']);
+        $this->RequestHandler->renderAs($this, 'json');
     }
 }
--- a/src/Controller/TagsController.php
+++ b/src/Controller/TagsController.php
@@ -246,4 +246,33 @@ class TagsController extends AppController
             $search
         ]);
     }
+
+    /**
+     * Display list of tags for autocompletion.
+     *
+     * @param String $filter Filters the tags list with only those that contain the
+     *                       search string.
+     */
+    public function autocomplete($search)
+    {
+        $this->helpers[] = 'Tags';
+
+        $conditions = [];
+        if (!empty($search)) {
+            $conditions = [
+                'name LIKE' => "$search%"
+            ];
+        }
+        $this->paginate = [
+            'limit' => 10,
+            'fields' => ['name', 'id', 'nbrOfSentences'],
+            'order' => ['nbrOfSentences' => 'DESC'],
+            'conditions' => $conditions
+        ];
+
+        $this->autoRender = false;
+        $this->response = $this->response->withType('json');
+        $allTags = $this->paginate();
+        echo json_encode($allTags);
+    }
 }
--- a/webroot/js/autocompletion.js
+++ b/webroot/js/autocompletion.js
@@ -74,8 +74,7 @@ function suggestSelect(suggestionStr) {
 function suggestShowResults(suggestions) {
     // we remove the old one
     removeSuggestList();
-
-    if (suggestions.length == 0) {
+    if (suggestions.allTags.length == 0) {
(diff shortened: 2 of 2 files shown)


AndiPersti:
@Ynue thanks for working on that issue.

I'm not sure if your suggested solution is the one @jiru was thinking of because you've replaced the autocompletion service we use (called `suggestd`; the repo is https://github.com/Tatoeba/suggestd) with a call to the main database.

But as [I've mentioned on the issue](https://github.com/Tatoeba/tatoeba2/issues/1429#issuecomment-629779063) it turns out that `suggestd` has its flaws.

Out of curiosity I've done some timing measurements on our dev machine with "real world" data:
````
root@envy:~# time wget -q -O - 'http://127.0.0.1:8080/suggest?str=a&sort=1&size=10' >/dev/null

real    0m0.006s
user    0m0.004s
sys     0m0.000s
root@envy:~# time wget -q -O - 'http://127.0.0.1:8080/suggest?str=abcd&sort=1&size=10' >/dev/null

real    0m0.011s
user    0m0.004s
sys     0m0.004s
root@envy:~# time wget -q -O - 'http://127.0.0.1:8080/suggest?str=abcdefg&sort=1&size=10' >/dev/null

real    0m0.010s
user    0m0.004s
sys     0m0.004s
root@envy:~# time wget -q -O - 'http://127.0.0.1:8080/suggest?str=a+long+search+string&sort=1&size=10' >/dev/null

real    0m0.010s
user    0m0.004s
sys     0m0.004s
````
````
root@envy:~# echo 'set profiling = 1; select name, id, nbrOfSentences from tags where name like "a%" order by nbrOfSentences desc limit 10; show profiles; set profiling = 0;' | mysql tatoeba | tail -2
Query_ID        Duration        Query
1       0.00155089      select name, id, nbrOfSentences from tags where name like "a%" order by nbrOfSentences desc limit 10
root@envy:~# echo 'set profiling = 1; select name, id, nbrOfSentences from tags where name like "abcd%" order by nbrOfSentences desc limit 10; show profiles; set profiling = 0;' | mysql tatoeba | tail -2
Query_ID        Duration        Query
1       0.01371438      select name, id, nbrOfSentences from tags where name like "abcd%" order by nbrOfSentences desc limit 10
root@envy:~# echo 'set profiling = 1; select name, id, nbrOfSentences from tags where name like "abcdefg%" order by nbrOfSentences desc limit 10; show profiles; set profiling = 0;' | mysql tatoeba | tail -2
Query_ID        Duration        Query
1       0.01668651      select name, id, nbrOfSentences from tags where name like "abcdefg%" order by nbrOfSentences desc limit 10
root@envy:~# echo 'set profiling = 1; select name, id, nbrOfSentences from tags where name like "a long search string%" order by nbrOfSentences desc limit 10; show profiles; set profiling = 0;' | mysql tatoeba | tail -2
Query_ID        Duration        Query
1       0.01035478      select name, id, nbrOfSentences from tags where name like "a long search string%" order by nbrOfSentences desc limit 10
````
````
root@envy:~# time echo 'select name, id, nbrOfSentences from tags where name like "a%" order by nbrOfSentences desc limit 10;' | mysql tatoeba >/dev/null

real    0m0.017s
user    0m0.008s
sys     0m0.008s
root@envy:~# time echo 'select name, id, nbrOfSentences from tags where name like "abcd%" order by nbrOfSentences desc limit 10;' | mysql tatoeba >/dev/null

real    0m0.009s
user    0m0.004s
sys     0m0.000s
root@envy:~# time echo 'select name, id, nbrOfSentences from tags where name like "abcdefg%" order by nbrOfSentences desc limit 10;' | mysql tatoeba >/dev/null

real    0m0.009s
user    0m0.008s
sys     0m0.000s
root@envy:~# time echo 'select name, id, nbrOfSentences from tags where name like "a long search string%" order by nbrOfSentences desc limit 10;' | mysql tatoeba >/dev/null

real    0m0.008s
user    0m0.000s
sys     0m0.004s
````
All in all it looks to me that `suggestd` hasn't any speed advantages.

**Comment by jiru** [comment]
@AndiPersti But your measurement doesn’t include the overhead of nginx and php.

File: src/Controller/TagsController.php [hunk 1]
```diff
@@ -246,4 +246,33 @@ public function search()
...
+        $this->paginate = [
+            'limit' => 10,
+            'fields' => ['name', 'id', 'nbrOfSentences'],
+            'order' => ['nbrOfSentences' => 'DESC'],
+            'conditions' => $conditions
+        ];
+
+        $this->autoRender = false;
+        $this->response = $this->response->withType('json');
+        $allTags = $this->paginate();
```
jiru, line 275: I don’t think you need to use pagination here, because you are never going further than page 1. If all you need to do is to limit the number of results to 10, you can use a simple query, like this:
```php
$query = $this->Tags->find();
$query->select(['name', 'id', 'nbrOfSentences'])
      ->limit(10);
if (!empty($search)) {
    $query->where(['name LIKE' => "$search%"]);
}
$allTags = $query->all();
```

src/Controller/TagsController.php
```diff
@@ -246,4 +246,33 @@ public function search()
...
+        echo json_encode($allTags);
```
jiru, line 276: If you’re trying to echo something, this should happen in the View, not the Controller. :point_up_2: Otherwise, errors like "headers already sent" are likely to happen.

In addition, if all you need to do is to echo the results in json, you can [do it from the controller](https://book.cakephp.org/3/fr/views/json-and-xml-views.html), like this:
```
$this->loadComponent('RequestHandler');
$this->set('allTags', $allTags);
$this->set('_serialize', ['allTags']);
$this->RequestHandler->renderAs($this, 'json');
```

webroot/js/autocompletion.js
```diff
@@ -69,35 +69,27 @@ function suggestSelect(suggestionStr) {
...
-        }
-        suggestion = suggestions[i].firstChild.data;
+    suggestions.forEach(function(suggestion, index) {
         var li = document.createElement("li");
-        li.innerHTML = "<a id='suggestItem" + i + "' onclick='suggestSelect(this.innerHTML)' style='color:black;'>"+
-            suggestion +
-        "</a>";
+        li.innerHTML = `<a id='suggestedItem${index}' onclick='suggestSelect("${suggestion.name}")' style='color:black';>
+        ${suggestion.name} (${suggestion.nbrOfSentences})
+        </a>`;
```
jiru, line 90: Could you rewrite this part without using backticks? They are not supported by IE.

src/Controller/TagsController.php
Ynue: Thank you! I actually did not know what `paginate` did, now I know.

src/Controller/TagsController.php
Ynue: Thank you, this makes more sense, I'm new to CakePHP so I have to practise :+1:

webroot/js/autocompletion.js
Ynue: Thank you for letting me know this compatibility issue, I've fixed it.

AndiPersti:
> @AndiPersti But your measurement doesn’t include the overhead of nginx and php.

I don't think that there will be much difference because the overhead should be almost the same for both approaches but I've planned to do some more testing tomorrow.

AndiPersti:
I've imported all current tags into my local database (in addition to the ones which were already there):
````
vagrant@stretch:~/Tatoeba$ echo 'select count(id) from tags;' | mysql tatoeba
count(id)
7632
````

Then I run a [test script](https://gist.github.com/AndiPersti/676ccecb235357f7cbc8be536dff1960#file-autocompletion_test-sh) which calls each autocompletion version 10 times for several search strings and [prints out the average and the result](https://gist.github.com/AndiPersti/676ccecb235357f7cbc8be536dff1960#file-results-txt)

Summary: It looks like the database approach is really faster than `suggestd`.

**Comment by jiru** [comment]
@AndiPersti I tried your benchmarking script and my results are a bit different. While the database approach is faster, it is only faster by a slightly bit: 6 to 26ms. I’m not sure what’s the reason. Anyway, I think Ynue’s approach is good to go.

@Ynue Your implementation returns different results than the current one:
1. It is case-insensitive (I think that’s a welcomed improvement though).
2. The sort order is different. As a result, since we only return the first 10 matches, the results can be totally different if the total number of matches is large. For example, when you type "by ", the results are totally different. It looks like suggestd’s sort order is alphabetical, but I think sorting by number of sentences would be a welcomed improvement too.

File: src/Controller/TagsController.php [hunk 4]
```diff
@@ -246,4 +246,29 @@ public function search()
...
+     */
+    public function autocomplete($search)
+    {
+        $this->helpers[] = 'Tags';
+
+        $query = $this->Tags->find();
+        $query->select(['name', 'id', 'nbrOfSentences']);
+
+        if (!empty($search)) {
+            $query->where(['name LiKE' => "$search%"]);
```
jiru, line 264: This approach may allow SQL injections. For example, I can type `%VO` and get results such as `SVO`. (CakePHP does some kind of quote escaping, but let’s not solely rely on that.) Have a look at `src/Command/FixNewUsersTOSSettingCommand.php` to see an example of how to implement it using binded variables.

Ynue: Thank you for pointing that out! Binding was not enough to prevent "%" from being interpreted, so I had to work it around by replacing and escaping "|" characters. Doing so assumes "|" is not used in tags.

jiru: I see. Sorry, my comment about the percent was misleading. Using bind() alone protects from SQL injection, but the percent problem is a different thing, that is not a security threat. It’s just that using percent will have special meaning that it not what the user expects. But you are welcome to escape the percent too.

I think you are in the right direction, but the ESCAPE operator should be used to set the escape character (which is backslash by default), not the characters you actually want to escape. So if ESCAPE is backslash, you need to write '\\%' in the LIKE pattern to match a percent. If a tag name is, say, '100% correct', then if the user types '100%', we want that name to show up. In addition, we want the tag '100 words' NOT to show up. Finally, you want to escape the underscore because it’s also a metacharacter (which means "any character"), and the escape character itself.

So maybe something like that:
```php
$pattern = str_replace(['\\', '%', '_'], ['\\\\', '\%', '\_'], $search).'%';
$query->where(['name LIKE :search'])->bind(':search', $pattern, 'string');
```

Ynue: I finally understand, I'm sorry for causing you so much trouble!

jiru:
Thank you guys for taking care of this issue! :+1: I think regular contributors will really appreciate this improvement.

PR #2332, opened 2020-05-17
Tag event dispatch

AndiPersti:
Unfortunately I've introduced a bug in 3842e94591b47bd02c07a3c2f7d1c5e9b2a8d1b3 so that only new tags dispatched the event to notify `suggestd`. :facepalm: 

(My lame excuse is that the test for event dispatching was incomplete and only tested the case when a new tag was added.)


jiru:
Nice rewrite of the event test! :sunglasses: 
About the failing test, I solved the problem on Jenkins.

PR #2334, opened 2020-05-17
Replace CSS property "box-shadow" under language icons with "filter: drop-shadow" - improved version, original color

speedysera:
This pull request addresses issue #2046 - apply better shadows to non-rectangular icons, with considerations for both Firefox and Chromium, using original color

### DIFF
--- a/webroot/css/layouts/default.css
+++ b/webroot/css/layouts/default.css
@@ -723,7 +723,7 @@ rt {
  }
 
 .language-icon {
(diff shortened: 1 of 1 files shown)


jiru:
Thank you, @speedysera! Your contribution is now visible on Tatoeba, check out the [list of languages](https://tatoeba.org/eng/stats/sentences_by_language) :sunglasses:

PR #2341, opened 2020-05-23
PoC of a REST API

AndiPersti:
This is just a demo!

Inspired by a [recent wall thread](https://tatoeba.org/eng/wall/show_message/35211#message_35211) I was looking into building an API. And it turns out it isn't really that hard with CakePHP.

It mainly requires configuring some routes and then some changes in the controller (i.e. deciding what to send to the client; for this demo I've simply serialized two view variables). It would also be possible to write specific view templates.

Here's how it looks like:
````
$ curl -s http://localhost:8080/eng/sentences/show/3.json
{
    "sentence": {
        "id": 3,
        "text": "My father was absent from his office yesterday.",
        "lang": "eng",
        ...
        "translations": [
            [
                {
                    "id": 66,
                    "text": "qwery",
                    "lang": "deu",
                    ...
                }
            ],
            ...
    }
}
$ curl -s http://localhost:8080/eng/sentences/show_all_in/eng/none.json?page=3
{
    "results": [
        {
            "id": 95,
            "text": "user",
            "lang": "eng",
            ...
        },
        {
            "id": 87,
            "text": "La \u0109ielarka\u00a0flago estas la simbolo de respekto de seksa diverseco.",
            "lang": "eng",
            ...
        },
        ...
    ]
}
````
The usual way to access the pages still work:
````
$ curl -s http://localhost:8080/eng/sentences/show/3 | head -5
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="utf-8"/>    <title>
        My father was absent from his office yesterday. - English example sentence - Tatoeba    </title>
````

### DIFF
--- a/config/routes.php
+++ b/config/routes.php
@@ -62,6 +62,18 @@ Router::scope('/', function (RouteBuilder $routes) {
     }
     $interfaceLanguages = join('|', $iso3LangArray);
 
+    $routes->setExtensions(['json']);
+    $routes->resources('Sentences', [
+        'only' => ['view'],
+        'actions' => ['view' => 'show'],
+        'map' => [
+            'showAllIn' => [
+                'action' => 'showAllIn',
+                'method' => 'GET'
+            ]
+        ],
+    ]);
+
     /**
      * Here, we are connecting '/' (base path) to a controller called 'Pages',
      * its action called 'display', and we pass a param to select the view file
--- a/src/Controller/SentencesController.php
+++ b/src/Controller/SentencesController.php
@@ -94,6 +94,7 @@ class SentencesController extends AppController
         if (in_array($params['action'], $noCsrfActions)) {
             $this->components()->unload('Csrf');
         }
+        $this->loadComponent('RequestHandler');
     }
 
     /**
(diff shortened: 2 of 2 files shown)


AndiPersti:
I'm closing this because it is far away from production ready.

jiru:
You can also set the pull request as 'draft'.

jiru:
That’s great! This approach however is tying endpoints used by browsers with endpoints of the API. I don’t think that’s a good idea, because:

1. Many controllers and actions are poorly named. While it’s not such a big problem in the browser, it wouldn’t be great for an API, where URL components are much more visible and meaningful.

2. We are currently free to change controllers and actions as long as the view is updated accordingly. Since an API needs to be stable, if we base these on API endpoints, we will loose this freedom.

So it may be better to use dedicated controllers/actions for the API.

AndiPersti:
> You can also set the pull request as 'draft'.

I set it as a draft but since this first version doesn't work and I will rebase the branch and start from a new "blank sheet" anyways I thought this can be closed for now.

AndiPersti:
> That’s great! This approach however is tying endpoints used by browsers with endpoints of the API. I don’t think that’s a good idea ... So it may be better to use dedicated controllers/actions for the API.

Agreed, the failing tests demonstrated the problems.

My plan is to use [Prefix Routing](https://book.cakephp.org/3/en/development/routing.html#prefix-routing).

agrodet:
It would be unfair to all the people always listening to my grumbling that REST APIs should stay in the 2010s if I wouldn't mention the GraphQL option. Of course it's not very straightforward and would require (much) more work so I'm not sure it would be a viable option for Tatoeba. Anyway, now that I've mentioned it, I'll disappear to the shadow once again! 

And sorry for posting this on the REST API PR. I just saw it and didn't think of a better place.

PR #2344, opened 2020-05-24
Tag selection using the keyboard

AndiPersti:
After 143cb20d2adb74906bf5e63695809cad4cbafc0b we've lost the abilitiy to select one of the suggested tags using the keyboard.

This also fixes some out-of-bounds accesses, sets "autocomplete" off in the template instead of in the Javascript code and removes an unused event handler.


jiru:
Thanks! :strawberry:

Issue #1459, opened 2017-03-22
Make it possible to rate sentences OK and add to a list when browsing "my comments."


ckjpn:
## The Problem

When browsing ["my commets"!](https://tatoeba.org/eng/sentence_comments/of_user/CK/page:8) it's not possible to easily add OK ratings or add sentences to a list, without clicking to each sentence's page.  It would be nice if browsing "my comments" had the same or similar user interface as ["browse by language"](https://tatoeba.org/eng/sentences/show_all_in/eng/und/none/indifferent).

Perhaps others don't go back and check to see if sentences have been corrected after leaving comments.  If that's the case, then this enhancement would perhaps be a low priority.

## Screenshots

![screen shot 2017-03-22 at 22 52 35](https://cloud.githubusercontent.com/assets/7934273/24201643/05a0f72a-0f54-11e7-8c32-24918b94afe5.png)

trang:
Note that we have the "Outdated ratings" feature. I don't think it's worth cluttering the comments just to be able to fix a rating. A better approach would be to make it easy to change the ratings in the "Outdated ratings" page.

agrodet:
I wonder if something like the following would help people to update their outdated reviews. It could be displayed only on the "outdated reviews" page, or more generally on all category pages, although I don't see the use case where one user would modify an up-to-date review from these pages.

<img width="664" alt="image" src="https://user-images.githubusercontent.com/42261966/78977560-45192800-7b53-11ea-965c-95323cef8d16.png">

ckjpn:
That would be great.

My work around has been to go to the outdated page(s) download them all, then filter the HTML to get the sentence numbers.  After that, using the downloaded sentences, I grab every sentence in that file that matches the sentences, and proofread them again, then send the proper URL to the website to update my rating to OK or remove the sentence from my List 907.

It would be less time-consuming to do it that way agrodet suggests.  It would be easier to just change them all directly as he suggests.

agrodet:
Among other redesign on the review pages, I'm working on implementing my screenshot above. However, I'm facing a technical difficulty. @trang I've read your recent code and commits about reviews but I can't figure out why the sentence's `current_user_review` is not set to its correct value on the review page. It's always null. `$this->users_sentences` of the `_getCurrentUserReview` method is null. It works properly on other page so I guess I'm missing some elementary thing here. I'd be very happy if you could give me a hand :)

AndiPersti:
[...]

PR #2345, opened 2020-05-26
Small redesign of review page + possibility to modify review on the reviews page

### LINKED ISSUE #2254: Sentence block on Reviews page doesn't scale down correctly
My reviews, with my browser set to 90% display. 

image

This issue doesn't occur on a list page. Although the div have the same `sentence mainSentence` class attribute, the CSS for `sentence` on the reviews page forces `width:600px`. I think that's the cause.

### LINKED ISSUE #2333: Color of review icons is inconsistent
Reported by [AlanF_US on the Wall](https://tatoeba.org/eng/wall/show_message/35153#message_35153):

> When I look at the sentences I've reviewed:
> 
> https://dev.tatoeba.org/eng/reviews/of/AlanF_US
> 
> there's a box "Filter" in the upper right-hand corner. Within that box, the items "Sentences marked as 'OK'", "Sentences marked as 'unsure'", and "Sentences marked as 'not OK'" are preceded by the checkmark, question mark, and exclamation mark icons. However, the icons are in gray. It would make more sense for them to be in green, orange, and red, as they are next to the sentences.

agrodet:
This PR is dedicated to the Reviews page. It includes a slight redesign of the U.I. and adds the possibility to modify my review directly on my Reviews page. 

If I'm looking at my own reviews, the main section now looks like 
image

If I'm looking at somebody else's reviews, it is displayed as it currently is. I removed the svg icon to use the `md-icon`.  
There is one little thing about which I'd like your opinion, however. When I'm looking at someone else's reviews, the tooltip indicates the date of the review. When I'm looking at mine, it indicates what the click will do, that is "Unmark sentence".  

This PR will close #2333, close #2254, and may or may not remove the need of #1459, I'll let you decide.

It's my first time writing an Angular controller by myself, so I'm looking forward to hear your reviews! :) Also, I'm aware that the code is similar to the one in the `sentence-and-translations` directive but I felt like factorizing this part would bring more problems than benefits. Again, I'll wait for your reviews.


src/Template/Reviews/of.ctp
```diff
@@ -26,6 +26,12 @@
  */
 use App\Model\CurrentUser;
 
+
+if (CurrentUser::isMember()) {
+    $this->Html->script('/js/services/list-data.srv.js', array('block' => 'scriptBottom'));
```
AndiPersti, line 31: You don't need this service. (It loads the selectable lists for a sentence when the user wants to add a sentence to a list.)

src/Controller/ReviewsController.php
```diff
@@ -128,9 +126,11 @@ public function of($username, $correctnessLabel = null, $lang = null)
         );
         $corpus = $this->paginate('UsersSentences');
 
+        $this->set('username', $username);
         $this->set('corpus', $corpus);
         $this->set('userExists', true);
         $this->set('correctnessLabel', $correctnessLabel);
         $this->set('lang', $lang);
+        $this->set('userIsReviewer', $userId == $this->Auth->user('id'));
```
AndiPersti, line 134: We usually use `CurrentUser::get('id')` to get the current user's id.

webroot/js/reviews/of.ctrl.js
```diff
@@ -0,0 +1,93 @@
...
+        vm.iconsInProgress = {};
+        vm.sentence = null;
+        vm.correctness = null;
+        vm.setReview = setReview;
+        vm.resetReview = resetReview;
+
+        ///////////////////////////////////////////////////////////////////////////
+
+        function initSentence(sentenceData, correctness) {
+            // console.log(sentenceData);
```
AndiPersti, line 53: Probably should be removed. Also at some lines below (56, 60, 70 and 73).

trang:
> When I'm looking at someone else's reviews, the tooltip indicates the date of the review. When I'm looking at mine, it indicates what the click will do, that is "Unmark sentence".

If you just want to fix the fact that you don't have the date when you reviewed a sentence when looking at your own reviews, then you could add the date tooltip on the sentence text instead of the icon.

If you want to make the design of the page more consistent between your reviews and other people's reviews, here's one possibility:

- You display the review info as an icon on both pages, with the date tooltip.
- On your reviews page, you add a [more_horiz](https://material.io/resources/icons/?icon=more_horiz&style=baseline) or [more_vert](https://material.io/resources/icons/?icon=more_vert&style=baseline) button. Clicking on this button displays a [menu](https://material.angularjs.org/latest/demo/menu) with the options:
  - Unmark sentence
  - Mark as X
  - Mark as Y (where X and Y are the 2 other possible reviews)

Adding the "Copy sentence" actions and "Show sentence's details" into the menu is also a possibility. All actions would then be gathered under the menu button and the remaining icons (language + review status) would be just information icons. In this case you would save some horizontal space (can be helpful for mobile display) and the reviews page would look more consistent between your page and someone else's page.

Shoving all the actions behind a menu is however asking for one more click for performing these actions. As long as people don't intensively perform these actions from this page, then it's fine. Otherwise it will be annoying. Unfortunately we don't have any methods to evaluate this.

All that being said, I have no issue with the PR being merged as is. We can rework the UI/UX later on (and refactor at the same time).

agrodet:
Thanks for all the comments. I tried several things. I couldn't find a good way to add a second module because it seems that manual bootstrap doesn't work with `ng-app`. If you know how to do that, feel free to share the information :D 

Eventually, I moved the logic to a single controller, which I attached to the sentence menu. Please tell me what you think.

About the change of design to "stick" to other pages. I don't think it's a good idea to hide the primary entity of a page behind more clicks. On "My reviews" page, I should see (and be able to modify) my reviews. For horizontal space, we can get rid of the right menu. It's just a guess, but probably the right menu will be moved for small devices design. We can think of what would be the best design for each page when we redesign all of them for small devices design.

src/Template/Element/sentences/sentence_menu.ctp
```diff
@@ -1,6 +1,8 @@
-<div class="menu-wrapper" sentence-menu flex="{{vm.isMenuExpanded ? '100' : 'none'}}" 
+<div class="menu-wrapper" sentence-menu flex="{{vm.isMenuExpanded ? '100' : 'none'}}"
      ng-init="vm.initMenu(<?= (int)$expanded ?>, vm.sentence.permissions)">
-    <div class="menu" layout="row" layout-align="space-between center">
+    <div class="menu" layout="row" layout-align="space-between center"
+         ng-controller="ReviewsController as reviewCtrl"
```
AndiPersti, line 4: The reason why you have to put the `ng-controller` directive here is that the we use `ng-if` for deciding whether we show the expanded menu or not. But `ng-if` deletes and recreates an element every time the conditions change and thus you call `reviewCtrl.initSentence` again which initializes the sentence with the old data (the original one received from the server) and consequently we loose the changes the user already made before.

Using `ng-show` instead you could move the `ng-controller` back where it was before and then check in `reviewCtrl.initSentence` whether `vm.sentence` is already initialized.

AndiPersti:
> Thanks for all the comments. I tried several things. I couldn't find a good way to add a second module because it seems that manual bootstrap doesn't work with `ng-app`. If you know how to do that, feel free to share the information :D

As I understand it you don't need to bootstrap manually.

We would have a main module which depends on other modules like 
````js
angular.module('app', ['app.sentences', 'app.reviews', ... ])
````

and then "submodules" each in their own files:
````js
/* app.sentences.js */
angular.module('app.sentences',  [/* dependencies for app.sentences */])

/* app.reviews.js */
angular.module('app.reviews',  [/* dependencies for app.reviews */])
````

see https://docs.angularjs.org/guide/module#recommended-setup and https://github.com/johnpapa/angular-styleguide/blob/master/a1/README.md#modularity

**Comment by agrodet** [comment]
When I add a new module to `app`'s dependencies, I simply get an `$injector/modulerr` error. I haven't investigated in all details but my guess is that since `reviewModule` is not defined anywhere (since the script is not added anywhere), obviously the dependency cannot be resolved. In the `Module Dependencies` section of your second link, is the code of `app.dashboard` always bundled with `app` or is there a way to require it on demand? If the latter, I didn't find how to do it. If the former, why loading a function we don't need?

**Comment by AndiPersti** [comment]
> When I add a new module to `app`'s dependencies, I simply get an `$injector/modulerr` error. I haven't investigated in all details but my guess is that since `reviewModule` is not defined anywhere (since the script is not added anywhere), obviously the dependency cannot be resolved.

You would need to add the module file to [`config/asset_compress.ini` before line 37](https://github.com/Tatoeba/tatoeba2/blob/3c68b27a2ff283e7c8f00c740398a304c0c85d96/config/asset_compress.ini#L37).

> In the `Module Dependencies` section of your second link, is the code of `app.dashboard` always bundled with `app` or is there a way to require it on demand? If the latter, I didn't find how to do it. If the former, why loading a function we don't need on most pages?

Looking a little bit closer, I think having a `reviews` module as a dependency for our main `app` module doesn't make sense at the current state of the code. As you've noticed, we would load it on every page.

IMHO we need to redesign/refactor the Javascript code. We don't have a SPA (single page application) so having just one `app` module that we load on every page isn't ideal. I think it would be better to have page specific main modules that load and configure all their necessary dependencies. This would allow us to split the code into several sub-modules and inject them as needed. But I need to think a little bit more about this. :smile: 

About optional dependencies: it looks like there are [some ways to achieve this](https://stackoverflow.com/questions/18542032/optional-dependencies-in-angularjs).

**Comment by AndiPersti** [comment]
I've rebased the code and also refactored the common parts into a directive which uses a template for rendering the icons.

**Comment by trang** [comment]
Haven't found any issue on my side. There's just the alignment between the copy button and the review buttons that is a bit off, but it's not blocking and can be fixed another time (when the page is redesigned to be responsive).

I think this can be merged 🙂

File: src/Template/Element/sentences/sentence_and_translations.ctp [hunk 5]
```diff
@@ -110,7 +119,8 @@ $sentenceUrl = $this->Url->build([
             <?php
             if (CurrentUser::isMember()) {
                 echo $this->element('sentences/sentence_menu', [
-                    'expanded' => $menuExpanded
+                    'expanded' => $menuExpanded,
+                    'sentenceData' => $sentenceData
```
jiru, line 123: Isn’t this change unnecessary?

src/Template/Reviews/of.ctp
```diff
@@ -133,7 +146,9 @@ $this->set('title_for_layout', $this->Pages->formatTitle($title));
             $withAudio = false;
             foreach ($corpus as $item) {
                 $sentence = $item->sentence;
-                echo '<div>';
+                $sentenceData = $this->Sentences->sentenceForAngular($sentence);
+                $sentenceData = str_replace('{{', '\{\{', $sentenceData);
```
jiru, line 150: Same here: I don’t think we need these two lines any more.

src/Template/Element/sentences/sentence_and_translations.ctp
AndiPersti: Yes you're right. That's a leftover from the old version that I forgot to remove. Thanks.

src/Template/Reviews/of.ctp
AndiPersti: Again, good catch.

Issue #2335, opened 2020-05-17
Fix wrong license of some sentences


jiru:
Some sentences have been affected a wrong license, as described in #2322.
TODO: write a cake shell that allows to massively fix the license of a selectable subset of sentences (with the creation of a "license changed" in the contributions log).
TODO2: make the script so that we can use it to set the license to any value, such as "licensing problem", too.

AndiPersti:
All wrong licenses should be fixed.

PR #2346, opened 2020-05-26
Add new edit_licenses command

AndiPersti:
This PR addresses #2335.

The command itself should be self-explanatory.

Sample run:
````
vagrant@stretch:~/Tatoeba$ echo 'select id, license from sentences where id in (5,7);' |mysql tatoeba
id      license
5       CC BY 2.0 FR
7       CC BY 2.0 FR
vagrant@stretch:~/Tatoeba$ echo -e "5\n7" | cake edit_licenses stdin ''
2 rows proceeded:
id 5 - license changed from CC BY 2.0 FR to
id 7 - license changed from CC BY 2.0 FR to
vagrant@stretch:~/Tatoeba$ echo 'select id, license from sentences where id in (5,7);' |mysql tatoeba
id      license
5
7
````

### DIFF
--- a/src/Command/EditLicensesCommand.php
+++ b/src/Command/EditLicensesCommand.php
@@ -1,0 +1,130 @@
+<?php
+namespace App\Command;
+
+use Cake\Console\Arguments;
+use Cake\Console\ConsoleIo;
+use Cake\Console\ConsoleOptionParser;
+use Cake\Console\Command;
+use Cake\Collection\Collection;
+use Cake\Datasource\Exception\RecordNotFoundException;
+use App\Model\CurrentUser;
+use App\Lib\Licenses;
+
+class EditLicensesCommand extends Command
+{
+    protected $log;
+    const DEFAULT_USER = 'admin';
+
+    public function initialize() {
+        parent::initialize();
+        $this->loadModel('Sentences');
+        $this->loadModel('Users');
+    }
+
+    protected function buildOptionParser(ConsoleOptionParser $parser) {
+        $parser
+            ->setDescription('Change sentence licenses.')
+            ->addArgument('file', [
+                'help' => 'Name of the file that contains the sentence ids whose ' .
+                          'license should be changed. ("stdin" will read from ' .
+                          'standard input.)',
+                'required' => true,
+            ])
+            ->addArgument('license', [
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by AndiPersti** [comment]
For simpler testing I've used `admin` as the default user. Is that ok?

**Comment by jiru** [comment]
I think it would be better to make the user a mandatory option. Otherwise, I am afraid that one day somebody will run this command without thinking about the user, and mistakenly assign the changes to an actual `admin` user on Tatoeba.

**Comment by AndiPersti** [comment]
I've updated the code.
The username is now the first required argument.

File: src/Command/EditLicensesCommand.php [hunk 1]
```diff
@@ -0,0 +1,129 @@
...
+
+        $input = $args->getArgument('file');
+        if ($input === 'stdin') {
+            $input = 'php://stdin';
+        } elseif ($input && !file_exists($input)) {
+            $io->error(format('{path} does not exist!', ['path' => $input]));
+            $this->abort();
+        }
+
+        $newLicense = $args->getArgument('license') ?? '';
```
jiru, line 101: Why null coalescing here?

AndiPersti: The command interprets both no third argument and `''` (the empty string) as `licensing issue`.
If no third argument is given, `$newLicense` will be `null` and while we allow `null` in the `license` column of `sentences` we don't allow it in the `text` column of `contributions` where a license change will be logged.

Do you think it is better to also make the license mandatory, i.e. to always run the command as `cake edit_licenses admin stdin ''` instead of `cake edit_licenses admin stdin` when we want to change the license to `licensing issue`?

jiru: Thanks for clarifying. Yes, I think it would be better to also make the license mandatory. I can see how inadvertently copy-pasting or pressing enter in the middle of typing the command could cause damage. I’m going to merge the pull request now so that it will be on production this week, but free to make this last change directly on the dev branch.

AndiPersti: The license is now mandatory too.

And after thinking more about it, using the empty string for a license isn't a good choice, so I've changed the command to accept only the license names we use in the UI.

PR #2347, opened 2020-05-26
Remove suggestd from codebase

AndiPersti:
This removes any remnants from `suggestd` from the code and moves the tests to `TagsControllerTest`.

As a result the tests on `dev` should work again.


jiru:
Thanks! I plan to merge your PR 2 weeks away from now, to allow a smoother transition between front-end and back-end during the next update (front-end will keep working just after back-end is updated.) The tests on dev should work now thanks to 6b14c7a25f1edbf1a233df97c68c66bc30935a8d.

PR #2348, opened 2020-05-27
Updated README file by deleting the entries for acl and archives

ngfuwai:
The folders for acl and archives were deleted, but the README.md file still had entries for those two folders. My task is to delete those two entries from the README file.


jiru:
Thank you for cleaning up the documentation, @ngfuwai! :blush:

PR #2349, opened 2020-05-27
Update advanced_search_form.ctp

gioele-santi:
Added clear icons to text inputs

### DIFF
--- a/src/Template/Element/advanced_search_form.ctp
+++ b/src/Template/Element/advanced_search_form.ctp
@@ -35,7 +35,7 @@ echo $this->Form->create('AdvancedSearch', [
             <?php /* @translators: section title in advanced search form */ ?>
             <h3><?= __('Sentences'); ?></h3>
 
-            <md-input-container>
+            <md-input-container class="md-icon-right">
                 <?php
                 echo $this->Form->input('query', array(
                     'label' => __('Words:'),
@@ -44,6 +44,7 @@ echo $this->Form->create('AdvancedSearch', [
                     'dir' => 'auto',
                 ));
                 ?>
+                <md-icon id="clearSearch" tabindex="-1" ng-click="ctrl.clearSearch()">clear</md-icon>
                 <div class="hint"><?= __('Enter a word or a phrase') ?></div>
             </md-input-container>
 
@@ -64,13 +65,14 @@ echo $this->Form->create('AdvancedSearch', [
                 ?>
             </div>
 
-            <md-input-container>
+            <md-input-container class="md-icon-right">
                 <?php
                 echo $this->Form->input('user', array(
                     'label' => __('Owner:'),
                     'value' => $user,
                 ));
                 ?>
+                <md-icon id="clearSearch" tabindex="-1" ng-click="ctrl.clearSearch()">clear</md-icon>
                 <div class="hint"><?= __('Enter a username') ?></div>
             </md-input-container>
 
@@ -134,13 +136,14 @@ echo $this->Form->create('AdvancedSearch', [
                 ?>
             </div>
 
-            <md-input-container>
+            <md-input-container class="md-icon-right">
             <?php
             echo $this->Form->input('tags', array(
                 'label' => __('Tags:'),
                 'value' => $this->safeForAngular($tags),
             ));
             ?>
+            <md-icon id="clearSearch" tabindex="-1" ng-click="ctrl.clearSearch()">clear</md-icon>
             <div class="hint">
                 <?= __('Separate tags with commas.') ?>
             </div>
@@ -236,13 +239,14 @@ echo $this->Form->create('AdvancedSearch', [
                     ?>
                 </div>
 
-                <md-input-container>
+                <md-input-container class="md-icon-right">
                     <?php
                     echo $this->Form->input('trans_user', array(
                         'label' => __('Owner:'),
                         'value' => $trans_user,
                     ));
                     ?>
+                    <md-icon id="clearSearch" tabindex="-1" ng-click="ctrl.clearSearch()">clear</md-icon>
                     <div class="hint"><?= __('Enter a username') ?></div>
                 </md-input-container>
 
--- a/src/Template/Element/advanced_search_form.ctp
+++ b/src/Template/Element/advanced_search_form.ctp
@@ -35,16 +35,19 @@ echo $this->Form->create('AdvancedSearch', [
             <?php /* @translators: section title in advanced search form */ ?>
             <h3><?= __('Sentences'); ?></h3>
 
-            <md-input-container class="md-icon-right">
+            <md-input-container class="md-button-right">
                 <?php
                 echo $this->Form->input('query', array(
                     'label' => __('Words:'),
                     'value' => $this->safeForAngular($query),
                     'lang' => '',
                     'dir' => 'auto',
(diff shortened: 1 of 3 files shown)


jiru:
Thanks for your pull request. :+1: Can you elaborate on what this change is doing?

gioele-santi:
This is still a draft. It should not be merged yet as it still miss some parts.

I just followed this discussion: https://github.com/orgs/Tatoeba/teams/kodoeba/discussions/5?from_comment=5#discussion-5-comment-5

jiru:
I see, I didn’t read that thread. :blush: I marked your pull request as draft. I’ll let Trang guide you because she’s more into front-end. Nice to meet you by the way! :raised_hand_with_fingers_splayed:

trang:
Good, you've passed the trial of opening a pull request :)

The next steps can be part of your first challenge for when Kodoeba starts. I'll just give you some pointers.

When you click on the icon, you want 2 things to happen:

1) Reset the associated input.
2) Focus the cursor on the associated input.

For this, we can create a [directive](https://docs.angularjs.org/guide/directive) which makes it that when we click on an element, it does the two things mentioned above.

As a starting point, you can add this piece of code in the `app.module.js` (where you'll see other directives):
```
.directive('resetButton', function() {
    return function(scope, element, attrs) {
      element.bind('click', function(e) {
            console.log('reset button clicked');
        });
    };
})
```

What it does is that on every HTML element where you add `reset-button` as an attribute, if you click that element, it will output "reset button clicked" in the browser's console.

I'll let you figure out what to replace this `console.log()` with :)

src/Template/Element/advanced_search_form.ctp
```diff
@@ -44,6 +44,7 @@ echo $this->Form->create('AdvancedSearch', [
                     'dir' => 'auto',
                 ));
                 ?>
+                <md-icon id="clearSearch" tabindex="-1" ng-click="ctrl.clearSearch()">clear</md-icon>
```
trang, line 47: Assuming you've added the directive that I posted in my other comment, you can replace this with:
```
<md-icon reset-button>clear</md-icon>
```

gioele-santi:
It is clear thank you. I will proceed as soon as I get some free time from work. As of now I have a lot of paperwork to complete.

src/Template/Element/advanced_search_form.ctp
gioele-santi: I tried this code, I still have to implement the javascript for the actual reset, but I don0t understand why in search bar the `md-icon` acts as a button while if I add it in advanced search page, it is a simple icon.
To make the button work I had to put the `md-icon` inside and `md-button`. Then it works and the directive is called correctly, only problem is that then the button is too large.

AndiPersti: It does already work as a button. But your expression for `ng-click` does nothing because `ctrl.clearSearch` is not defined.

AngularJS makes quick testing a little bit harder because some features aren't everywhere available (such as for example showing a quick pop-up with `alert()`).

But if you add `<?= "{{test}}" ?>` just before the line with `md-icon` and change the value of `ng-click` to `"test = test + 1"` you should get a simple counter which increments whenever you click the icon.

gioele-santi: I was following Trang suggestion to use a directive and it actually works when called, but that does not work with the plain `md-icon` and requires the `md-button`. For now I will focus on te actual JS to reset and then check what I can do with styling.

trang: Hmm... The directive should be called even if you don't put the `md-icon` inside a `md-button`. Perhaps you were confused at the fact that on hover, the cursor isn't converted into a hand cursor?

In the search bar, the icon acts as a button because the CSS for this element contains `cursor: pointer;`
```
.search_bar #clearSearch {
    background: transparent;
    border: none;
    cursor: pointer;
    color: #CCCCCC;
    margin-left: -30px;
}
```

gioele-santi: Probably I was confused by the hovering, I checked again on another reset button in the page and the click works even in that. Probably yesterday I messed to much with code and at beginning it didn't work JS and I trid other solutions. Anyway I go on with JS and then I will check the style in CSS.

gioele-santi: I have updated and pushed. Javascript code works, I don't know it it is the most elegant solution as it heavily depends on order of html elements or if there's a better way.
For css I can edit `advanced_search.css` right?

trang: For the CSS, you should rather edit the `default.css`. I've experimented a bit and I think you could use this code:

```
.md-button-right input {
    padding-right: 40px;
}
.md-button-right .md-icon-button {
    position: absolute;
    right: -8px;
    top: -4px;
}
```

You can add it at the end of the `default.css` in the "Angular Material override" section.

With this, you can to set `class="md-button-right"` (instead of `md-icon-right`) on the `<md-input-container>`, and you can encapsulate the `<md-icon>` inside a `<md-button class="md-icon-button">`. It does have the advantage to provide an extra little animation on click, which the `<md-icon>` alone wouldn't have.

For the Javascript, how about adding an attribute to specify the id of the target input, then read the value of this attribute to reset the input which has that id?
```
<input id="some-id"/>
<md-button reset-button target="some-id"> ...
```

webroot/js/responsive/app.module.js
```diff
@@ -68,5 +68,15 @@
                     }
                 });
             };
+        })
+        .directive('resetButton', function(){
+            return function(scope, element, attrs) {
+                element.bind('click', function(e) {
+                    var target = this.getAttribute('target');
```
trang, line 75: You can use `attrs.target` instead of `this.getAttribute('target')`.
The `attrs` parameter will contain all the attributes of the element.

jiru:
@gioele-santi Thanks for your pull request. I tested your changes and it works! :blush:

In terms of visual aspect, I think there is room for improvement though. The first thing I noticed is that the hovering shadow overflows on the horizontal line, which I find a little bit inelegant in my opinion.
overflow

But the most troubling part to me is the alignment of the X icon with the text. If I compare with the top-bar search field, it looks like your X icon is a little bit down:
text_field_alignment

gioele-santi:
@jiru I think I found a solution for the second problem just by editing the custom css added for md-button. Using `top: -8px;` instead od -4 px fixes the alignment. For the background shadow I need to figure out what property I should address in css.

trang:
@gioele-santi 

For the button background shadow, that is because the button is defined as a normal button. Every `md-button` will have this background hover effect.

There is another type of button, which is icon buttons. These buttons are used for instance on the new sentence design menu. There is no background effect and the size of the buttons is also different.

You can see the demo here: https://material.angularjs.org/latest/demo/button

To use icon buttons instead of normal buttons, you simply have add the class `md-icon-button` on the button (`<md-button class="md-icon-button">`). 

That's why in my CSS sample, I defined `.md-button-right .md-icon-button` instead of `.md-button-right .md-button`.

gioele-santi:
I have updated buttons and also directive access to attributes as I noticed I didn't see a previous advice.

jiru:
Thank you and good job with your pull request, Gioele! :blush:

Issue #2326, opened 2020-05-15
In the new design, when the menu is expanded, the username is no longer visible


trang:
Reported by [AlanF_US and CK on the Wall](https://tatoeba.org/eng/wall/show_message/35160#message_35160):

> the user name disappears in "display more" mode, and I don't want to have to keep switching modes. 

> While we see the the owner's name in the "show less" version, when browsing while showing all the things we need using the "show more" version, we don't see the owners' names, which means we still need to constantly toggle between the two.

Yorwba:
This was fixed in #2855.

trang:
Yes, thanks for the reminder :)

Issue #2280, opened 2020-04-14
"Add sentence to list" checkbox affects unrelated sentences on the same page


jiru:
Reported by arh, when trying to create a list of sentences to record.

> First, I used the advanced search to select those of my sentences not yet recorded, and got a total of 298 results, in three pages.
>
> To better control them, I created 3 lists: arh-1, arh-2 and arh-3, so that I could add a whole page of sentences to each one of them.
>
> To mark the sentences to be added to a given list, I keep their frame (check box) expanded, as this helps me visualize which ones I have already marked and not skip any of them as I go on adding new ones.
>
> But then I noticed that, whenever I tick any of the sentences to add it to the list, all of the other sentences appear also marked; similarly, if I unmark any of them, the rest also appear unmarked, as if I had already selected all of them as a whole. On the other hand, whenever I expand a new sentence to mark it, the ones previously marked turn blank, as if I had unmarked them.
>
> At the view of this behaviour I thought that, perhaps, I was able to mark/unmark the whole page, which would save a lot of time, but I checked my lists a couple of times to make sure and, indeed, only those which I had actually ticked are added to the list, whether or not they appear marked on the screen.
>
> This makes the process a bit confusing, as I don’t have a way to differentiate which ones have I actually marked.
>
> I attach a few screenshots to help you better understand the issue.
>
> ![01-frases-3](https://user-images.githubusercontent.com/5107734/79218772-9131d880-7e83-11ea-8ac1-231193164140.PNG)
> ![02-listas-3](https://user-images.githubusercontent.com/5107734/79218775-93943280-7e83-11ea-87d5-debb2b17ec69.PNG)
> ![03-lista-arh-3](https://user-images.githubusercontent.com/5107734/79218780-94c55f80-7e83-11ea-91ab-655efea5edf2.PNG)

Issue #2232, opened 2020-03-26
Closing the sentence edition form changes the ownership icon


jiru:
Using the new design, I go to a sentence page which I own. I click on the unadopt button. The icon updates to show that the sentence is now orphan. Then, I click on the edit button. Then, I decide not to edit anything so I click Close. Bug: the ownership icon now shows as if the sentence is owned.

Issue #2210, opened 2020-03-17
Displaying only 5 translations is not enough for some users


trang:
## Problem

Reported initially by [deniko on the Wall](https://tatoeba.org/eng/wall/show_message/34487#message_34487):

> By default, some translations are hidden in the new design and I need to click "Show N more translations" every time in my search results:
> 
> https://i.imgur.com/YyBRn8s.png
> 
> It is very annoying and very counterproductive.

Several others users felt the same frustration at having to expand the translations in the new sentence design.

## Possible solutions

The solution advocated by the users was to implement the option to have all translations displayed in the new sentence design. This option exists in the old sentence design, but in reverse: by default all translations are shown and there is an option `Display a link to expand/collapse translations when there are too many translations`.

The problem with that is that we cannot display an infinite amount of translations. There's a point where having too many translations is counterproductive so having an option to display all translations would not be a long-term solution. We have to set a limit, but a limit of 5 translations is too low for the more "hardcore" users.

So my preferred solution would be to introduce a new option `Number of translations displayed by default`, with the values: 5, 10, 20, 50. The default would be 5. This mimics the option `Number of sentences per page`.

agrodet:
I think the solution you mentioned is only solving a part of the problem you described. In my opinion, that problem can be broken down to three pieces (at least).

Besides your "preferred solution", that is letting the choice to the user to display a different amount of translations, I think the problem won't be solved as long as the two following points aren't solved either.  

- Add a setting or another way to decide "my preferred languages for translations", that would decide what languages should be on top of the translations list. In my opinion, that is the most important point. Basically, that is solving #1907.
- Add the possibility to display all translations for a particular sentence. That is, solving #1157.

trang:
#1907 is more about the default order of indirect translations (which we can extend to direct translations as well). What if a user is not authenticated or has yet not indicated their language preferences, what should be the default order of our translations? The way they are currently ordered is a bit questionable. We order them by language name but there's really no reason for it besides the fact that it was easy and quick to order them that way. I think that's the issue that #1907 should address.

Having a "preferred languages for translations" is the next step. It's basically addressing the issue of users not being satisfied with the defaults. I agree it's also a must have to solve the problem expressed by our users in the Wall thread but I'll create a separate issue for it because it can be implemented completely independently.

PR #2351, opened 2020-05-28
Use new sentence design as default

trang:
This PR sets the new sentence design as the default one. Along the way, it fixes some of the remaining issues that were reported so far:
- #2326
- #2210
- #2232
- #2280

The condition to display sentences with the new design is no longer based on the `use_new_design` setting but on a new setting, `disable_new_design`. 

The option to use the new design is now reading from this new setting. It is enabled by default (similarly to the random sentence on the homepage) and was moved out of the "Experimental" section into the "Main options" because it's not experimental (although it is temporary).

The setting `use_new_design` was kept so that we can display the announcement message only to those who had not switch to the new design. People who have switched do not need to see this message.

Wall post asking for tests: 
https://tatoeba.org/eng/wall/show_message/35389#message_35389

### DIFF
--- a/src/Controller/SentencesController.php
+++ b/src/Controller/SentencesController.php
@@ -476,7 +476,7 @@ class SentencesController extends AppController
             $numberOfTranslations = $this->request->getQuery('numberOfTranslations');
             $includeTranslations = $translationLangFilter == 'und';
             $sentence = $this->Sentences->getSentenceWith($sentenceId, ['translations' => $includeTranslations]);
-            $sentence->extraTranslationsCount = $numberOfTranslations + 1 - SentencesTable::MAX_TRANSLATIONS_DISPLAYED;
+            $sentence->extraTranslationsCount = $numberOfTranslations + 1 - $sentence->max_visible_translations;
 
             $this->loadComponent('RequestHandler');
             $this->set('sentence', $sentence);
--- a/src/Controller/SentencesController.php
+++ b/src/Controller/SentencesController.php
@@ -402,16 +402,17 @@ class SentencesController extends AppController
     private function renderAdopt($id)
     {
         $acceptsJson = $this->request->accepts('application/json');
-        $sentence = $this->Sentences->get($id, [
-            'contain' => ['Users' => ['fields' => ['username']]]
-        ]);
 
         if ($acceptsJson) {
+            $sentence = $this->Sentences->getSentenceWith($id);
             $this->loadComponent('RequestHandler');
-            $this->set('user', $sentence->user);
-            $this->set('_serialize', ['user']);
+            $this->set('sentence', $sentence);
+            $this->set('_serialize', ['sentence']);
             $this->RequestHandler->renderAs($this, 'json');
         } else {
+            $sentence = $this->Sentences->get($id, [
+                'contain' => ['Users' => ['fields' => ['username']]]
+            ]);
             $this->set('sentenceId', $id);
             $this->set('owner', $sentence->user);
             $this->viewBuilder()->setLayout('ajax');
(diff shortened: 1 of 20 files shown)


AndiPersti:
You should validate the user input for `max_visible_translations` because by simply modifying the select control with the browser's dev tools it is possible to set it to any value you want:
````
root@envy:~# echo 'select settings from users where username = "rumpelstilzchen2";' | mysql tatoeba
settings
{"is_public":"","lang":null,"use_most_recent_list":"","collapsible_translations":"","show_transcriptions":"","sentences_per_page":10,"max_visible_translations":"200","users_collections_ratings":"","native_indicator":"","hide_random_sentence":"0","use_new_design":"0","disable_new_design":"0","default_license":"CC BY 2.0 FR","can_switch_license":false,"new_terms_of_use":"1","license_switch_list_id":null,"hide_new_design_announcement_2":false}
````

trang:
Thanks @AndiPersti! I added the validation.

githubshrek:
Hi @trang do you have plans to push this further along any time soon? I'm guessing not, since it has been in the current status for approaching one year.

trang:
@githubshrek I'm aware this has been on standby for a while now, but I haven't given up on this :) 

This merge request is however blocked by another ticket: #2364 (to order translations in a more meaningful way). We need to solve that ticket first because there has been some reluctance from veteran members of Tatoeba in regards of the new design. It is much less compact than the old design and feels, for some, less efficient to use.

I actually have already started to work on #2364 months ago, I just need to find time where I have a bit of peace of mind to continue it. Then once that is live, we'll have to see if users have any other major issue with the UX of the new design, and if not, we can proceed to switch to the new design as default, and eventually, completely getting rid of the old design and fully focus on improving the new one.

githubshrek:
I think that your trying to overcomplicate the dependencies. The problem expressed with the new design is that it is much less compact than the old design. The solution therefore is simple: make the new design more compact.

When I compare the new design with the old, I see that the new design is a lot more spaced out. There is, for example, on my screen 14 mm between each flag in the new design and only 10 mm between each flag in the old design (i.e. 40% more spaced out).

New design:
image

Old design:
image

The solution should be to fix the new design, not to fix some completely different part of the system. Sure, I understand how solving "Translations are not ordered in a helpful way" would keep things more compact.

However, I think the "new sentence design" is an epic ticket, and "Translations are not ordered in a helpful way" is a totally different epic ticket. They should not be coupled together, because there is really no genuine dependency.

I would encourage you to make some small easy changes to the new design (e.g. make the flags 10mm spaced rather than 14mm) and in order to get it into production, rather than link it to some other large epic, which no doubt will have its own problems.

The aim should be to make the changes smaller (e.g. adding a "Translations" and "Translations of translations" title to the design) and the dependencies less, to try and get small and steady progress.

jiru:
> I actually have already started to work on #2364 months ago

@trang Whoops, I also already started working on this months ago. :hand_over_mouth: I just pushed that branch and created a draft PR #2791.

2colours:
@jiru @trang sorry to bother. What is the state of this PR? Is there any point in keeping it around?

trang:
@2colours Indeed, it's probably best to close it at this point. We still want to transition to the new design, but better do it in a new PR with fresh code and a fresh mind :)

Just a few notes about the status:

1) The suggestion to reduce the spacing between the translations has already been deployed on prod (54a391e586c730cab20d65d830f96354333517f8). I think we cannot make the new design much more compact at this point.

2) The new design is already set as default for any new registered user since May 2022 (f08e0bbf3fa77663490ac89b4154e01b34eb5203). Older members only had an announcement telling them that there's a new design and that they have the option to switch to it, if they wanted to. Some have switched, some haven't.

3) The goal now is enforce the new design and there could be several ways to approach this, but I guess we can talk about it when we are ready to work again on this topic.

PR #2362, opened 2020-05-31
Encode tag suggestion query before sending it to the server

AndiPersti:
`#` and `?` at the start of a tag query need to be encoded in order to not interpret them as fragment or query string respectively.


jiru:
Thanks. :+1:

Issue #2237, opened 2020-03-31
Show the most-recently tagged sentences at the top the "/tags/show_sentences_with_tag/" pages


ckjpn:
## I'd like to see this be the default sort for tags.

Show the most-recently tagged sentences at the top.

https://tatoeba.org/eng/tags/show_sentences_with_tag/276/eng?sort=added_time&direction=desc

## The Reason

This is a good way to monitor which sentences are getting added to the "tag" lists.

If I remember correctly, this is also how it used to be.

## This now takes 3 clicks, plus the time it takes to load in the pages.

I would assume this also puts an extra unnecessary load on the server.

1. Go to the Proverb tags

https://tatoeba.org/eng/tags/show_sentences_with_tag/407

2. Click "date of tag"

https://tatoeba.org/eng/tags/show_sentences_with_tag/407?sort=added_time&direction=asc

3. Click "date of tag" again to reverse the sort

https://tatoeba.org/eng/tags/show_sentences_with_tag/407?sort=added_time&direction=desc

## I don't think this is just my own preference.

Should I or someone else put a Wall post about this to solicit other members' opinions?

agrodet:
It seems indeed logical that on a page focused on tags, the default order should relates to tags. 

In a more general level, #2161 will redesign the sort options.

AndiPersti:
Fixed in #2367.

PR #2367, opened 2020-06-01
Issue 2237 resolved

bakananbanjin:
I think this is all that needed to be changed to resolve the issue. Slowly I get the hang of it, after a few more tutorials about php and cake it will be a piecse of cake XD

### DIFF
--- a/src/Controller/TagsController.php
+++ b/src/Controller/TagsController.php
@@ -207,7 +207,7 @@ class TagsController extends AppController
                 'contain' => ['Sentences' => $contain],
                 'conditions' => $conditions,
                 'limit' => CurrentUser::getSetting('sentences_per_page'),
-                'order' => ['sentence_id' => 'DESC']
+                'order' => ['added_time' => 'DESC']
             ];
             $this->paginate = $pagination;
 


trang:
Well that was fast 😄 

Until we figure out your next challenge, here's a few GitHub tips:
- When you reference to an issue, you should add a hashtag in front of it (`#2237` instead of just `2237`).
- In your pull request description, you can write something like "This PR solves #2237". You'll see that #2237 is clickable and has a tooltip, so someone who reads your PR can easily check what is the issue. But also, if you go to #2237, you'll see there's a link back to the PR. It can help trace which PR is related to which issue.
- In your commit message, you can also reference to an issue the same way, by adding the issue number with a hashtag in front of it. This helps for traceability too, because your commit will be automatically mentioned in the related issue.
- With some [special keywords](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword), you can get an issue to be automatically closed when your pull request is merged.

bakananbanjin:
thx I will try to remember to use # ( #2123   and keywords like closes #2237 I called it learning while doing wwww

AndiPersti:
Congrats.

The PR looks good to me. :+1:

Issue #2319, opened 2020-05-11
Add "UTC" to tooltip displayed when hovering over a time


alanfgh:
I was unsure as to whether the times displayed for my contributions corresponded to UTC or to my own time zone. The only way I could figure it out was to add comments on the dev site (for instance, "I wrote this comment at 7:23 p.m. on Saturday, 2020-03-14, my time"), and then come back after enough time had passed that they were no longer displayed as "36 minutes ago" but rather as "March 14, 2020 at 11:23 PM". By matching the time that I wrote in the comment with the time as displayed, I could see that the times corresponded to UTC, not to my time zone. It would probably be overkill for the actual text for times to include "UTC" at the end to indicate that all times correspond to Universal Coordinated Time, but it would be nice if the hover text included this.

trang:
> It would probably be overkill for the actual text for times to include "UTC" at the end to indicate that all times correspond to Universal Coordinated Time, but it would be nice if the hover text included this.

+1. This is the way GitHub does it and I think it makes sense.

Related: #1125

polkadotazn:
Got it! Looks like it's in this page here? https://github.com/Tatoeba/tatoeba2/blob/49ea1514135aeeaddb6290cbb22b54e4b6f15c38/src/Template/Element/sentence_comments/comment.ctp

jiru:
@polkadotazn Yes, that’s a good starting point! If you look closely, you’ll see that `$dateLabel` is created from `$this->Date->getDateLabel()`, which is located in the DateHelper under `src/View/Helper/DateHelper.php`.

polkadotazn:
I have just made this PR: https://github.com/Tatoeba/tatoeba2/pull/2368

jiru:
@polkadotazn Your enhancement  is now running on tatoeba.org. Thanks @polkadotazn and @alanfgh. :smile_cat:

PR #2368, opened 2020-06-02
add UTC to tooltip datetime

polkadotazn:
I added UTC as part of the `nice(date)` method. I updated the Date Helper tests as well.

### DIFF
--- a/src/View/Helper/DateHelper.php
+++ b/src/View/Helper/DateHelper.php
@@ -58,9 +58,9 @@ class DateHelper extends AppHelper
         if (empty($date) || $date == '0000-00-00 00:00:00') {
             return __('date unknown');
         } elseif ($date instanceof DateTimeInterface) {
-            return $date->nice();
+            return "{$date->nice()} UTC";
         } else {
-            return $this->Time->nice($date);
+            return "{$this->Time->nice($date)} UTC";
         }
     }
 
--- a/tests/TestCase/View/Helper/DateHelperTest.php
+++ b/tests/TestCase/View/Helper/DateHelperTest.php
@@ -94,19 +94,19 @@ class DateHelperTest extends TestCase {
             'created within 30 days eng' =>
             ['{createdDate}, edited {modifiedDate}', '2018-09-29 09:12:34', '2018-09-29 09:12:34', false, 'en', '25&nbsp;days ago'],
             'created within 30 days tooltip eng' =>
-            ['{createdDate}, edited {modifiedDate}', '2018-09-09 09:12:34', '2018-09-09 09:12:34', true, 'en', 'September 9, 2018 at 9:12 AM'],
+            ['{createdDate}, edited {modifiedDate}', '2018-09-09 09:12:34', '2018-09-09 09:12:34', true, 'en', 'September 9, 2018 at 9:12 AM UTC'],
             'created and modified within 30 days eng' =>
             ['{createdDate}, edited {modifiedDate}', '2018-09-29 09:12:34', '2018-10-10 01:23:45', false, 'en', '25&nbsp;days ago, edited 14&nbsp;days ago'],
             'created and modified within 30 days tooltip eng' =>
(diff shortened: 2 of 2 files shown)


RyckRichards:
Thanks for this PR, @polkadotazn

trang:
@polkadotazn Congratulations on your first pull request 🥳

Tip: you don't need to comment on the associated issue every time you open a PR. Do mention which issue your PR is solving in your description though. GitHub will take care of creating links between the issue and the PR. And if you use some [special keywords](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) before the issue reference (for instance "This PR fixes #2319"), it can automatically close the associated issue once your PR is merged.

We'll get back to you soon for the next issue :)

jiru:
@polkadotazn Nice! And good job with updating the tests too. :blush:

There is just one problem with your implementation. It doesn’t allow "UTC" to be translated. In other words, it is alright when the website is displayed in English, but other languages might look weird with that English acronym thrown inside.

Localization of dates and time is actually a tedious problem. That’s why we rely on special methods like [`Time::nice()`](https://book.cakephp.org/3/en/core-libraries/time.html#Cake\I18n\Time::nice). The format used by `nice()` is defined at the end of `config/bootstrap.php`:
https://github.com/Tatoeba/tatoeba2/blob/bb839f64b7693e9eaa419b7d5e8b3812bea7cda2/config/bootstrap.php#L272

That array consists of two constants: one for the format of the date, and the other for the format of the time. These are [constants from the ICU library](https://www.php.net/manual/en/class.intldateformatter.php#intl.intldateformatter-constants), which job is to localize dates while abstracting out all the language specifics. So ideally, we want to tell the ICU library to just "format the datetime including the timezone" and let it do the job. According to the documentation on php.net, it looks like you could use `IntlDateFormatter::FULL` to get the timezone included. Could you try that?

AndiPersti:
> So ideally, we want to tell the ICU library to just "format the datetime including the timezone" and let it do the job. According to the documentation on php.net, it looks like you could use IntlDateFormatter::FULL to get the timezone included.

Unfortunately there are a few more things to consider:
1. Most languages use `GMT` instead of `UTC`
2. All ICU format levels from `MEDIUM` upwards include the seconds in the time (which we currently do not).
3. The current version of `DateHelper` uses the same `nice` format for the text and the tooltip.

Example (using `IntlDateFormatter::FULL`):
date

jiru:
What about using `IntlDateFormatter::FULL` only in tooltips?

polkadotazn:
Ok definitely! Thanks for the feedback!! I will try that. Unfortunately I spilled a bit of water on my laptop yesterday and I am leaving it off for a couple days to dry out 😨

jiru:
Oh, no! :hushed: I hope your laptop will survive. You can also try to completely cover your laptop components with (uncooked) rice. Over time, the rice should suck up all the moisture.

Yorwba:
FWIW, the people at [iFixit recommend using isopropyl alcohol to displace the water instead](https://www.ifixit.com/Wiki/Don%27t_Put_Your_Device_in_Rice._Here%27s_Why...) of letting the device dry (in rice or otherwise), because the water can keep corroding components inside while you wait.

AndiPersti:
That's what I would suggest too.

But `TimeHelper::nice()` doesn't take a format and just uses the one defined in `bootstrap.php`. So the date formatting for tooltips needs to use [`TimeHelper::format()`](https://api.cakephp.org/3.7/class-Cake.View.Helper.TimeHelper.html#_format).

jiru:
Maybe you can use `TimeHelper::i18nFormat()` instead of nice() for the tooltip.

polkadotazn:
ok sounds good! I'll look into these options. thanks everyone for these suggestions and also the computer tips :D

jiru:
> Example (using `IntlDateFormatter::FULL`):
> date

@polkadotazn That screenshot is actually using `IntlDateFormatter::LONG`. `IntlDateFormatter::FULL` spells out the whole words of the timezone.

jiru:
See #2368.

PR #2373, opened 2020-06-03
Issue#895

### LINKED ISSUE #895: Make it easier to remove a sentence from a list
By [cueyayotl](https://tatoeba.org/eng/user/profile/cueyayotl)

> A way to remove a single sentence from a list, without having to browse through the list.
> 
> We now can see which lists the sentence belongs to. How about a link to remove the sentence from the list (if the list is ours, or is collaborative).
> 
> Also, language filter for lists.

bakananbanjin:
this should close #895 

I also added a  new line between tags in Module (i thought it looks better)

other than that wrote a new functions in ListHelper:
 displayListsModuleWithRemove which is basicly the same function as the old  displayListsModule just that it adds an X behind the list and if the sentences is not in a List the List DIV container still gets shown(I didnt want to change the old function in case you need it again) 

_displayRemoveLink creates the link as well as X button

i see Issue #2237 changes are also included

if any questions you know how to contact me 

looking forward to the next task and hope this is correct and as you wanted it


AndiPersti:
> I also added a new line between tags in Module (i thought it looks better)

What do you mean?
I can only see that the space above the `Lists` heading got bigger which I don't think is a good idea because now it doesn't match the other modules in the sidebar any more.


> other than that wrote a new functions in ListHelper:
> displayListsModuleWithRemove which is basicly the same function as the old displayListsModule just that it adds an X behind the list and if the sentences is not in a List the List DIV container still gets shown(I didnt want to change the old function in case you need it again)

Since that function is only used in `Template/Sentences/show.ctp` I don't think it's necessary to have two almost identical functions so I would suggest to add the delete button to the already existing function.

> i see Issue #2237 changes are also included

That's not good IMHO. I think when you've created the branch for this PR you were still in the branch for the other one so the branch for this issue is now based on the other one. But you should base all new branches on the `dev` branch (preferably from its latest version).

I suggest to do the following:
````
git checkout -b issue-895 dev
git cherry-pick 27d63dc
````
The first git command creates a new branch `issue-895` based on the current `dev` branch and switches to it.
The second git command takes your last commit from this PR and puts it on the current branch (i.e. the new `issue-895`).
So you end up with a new branch containing only the changes related to this PR. I think it is best to open a new PR from that branch.

About the file modes changes also included in your commit: I think you should follow the instructions from the [Imouto README](https://github.com/Tatoeba/imouto#editing-the-source-code) but maybe @Trang knows more about it.

bakananbanjin:
sry still didnt get it how to update the pull request so i think i send a new one XD
other than that all infos on the next pull request.
sry

Issue #1800, opened 2019-02-26
Show list information on sentence pages for public lists


soliloquist-tatoeba:
If a sentence belongs to a list, on the sentence page, only the creator of the list can see that information regardless of the privacy setting of the list. 

![index](https://user-images.githubusercontent.com/47579687/61997682-ecc9f500-b0ad-11e9-8e8a-4dd830183e10.png)
https://prnt.sc/mq613p

That's how it should be for private and hidden lists, but for public lists, anyone should be able to see it. Just like tags. It would help users to discover other sentences on the same subject.

trang:
For this we will need to split the "public" option into two options:
1) The list is listed on the index page only.
2) The list is listed on the index page and on the sentence's page.

Some users may not want to have their lists listed on the sentence's page, but still listed on the index page:
https://tatoeba.org/eng/wall/show_message/25227#message_25227

soliloquist-tatoeba:
Thanks for letting me know about the previous discussion. I personally don't see much point showing a list on the index page and hiding it on sentence pages, but apparently that's what some users want. Your suggestion will surely solve the issue.

jiru:
I think that putting more options is generally a bad idea.

@soliloquist-tatoeba Just to make the context clear. Unlisted lists are a somewhat new feature (see #677). Before, all the lists were public and listed, but none of the lists were listed on the sentence page. As a result, lists didn’t have that much exposure and members created collaborative lists about topics they didn’t necessarily want to be displayed on the sentence page. Consider this [previously "collaborative" list](https://tatoeba.org/sentences_lists/show/992). There is a huge difference between showing *every visitor of the sentence page* the way *only you and some other members* are classifying sentences, and having the your list listed on the *much harder to find* "list of lists" page.

soliloquist-tatoeba:
Thanks for the information.

> Consider this [previously "collaborative" list](https://tatoeba.org/sentences_lists/show/992). 

Yes, I know how messy that list is. Actually, there's always this risk for collaborative lists. They're very easy to be spoiled and removing sentences is the other problem. Many active users prefer private lists to avoid that. That's why I requested a limited-collaborative privacy setting for lists before: #1704

And yet I think this is more reason to show list information on sentence pages. This way, it becomes easier to notice if a list is spoiled or misused. Displaying misinformation (spoiled collaborative lists) on the index page isn't less harmful than on the sentence page, I believe.

[...]

PR #2379, opened 2020-06-04
issue-#895

### LINKED ISSUE #895: Make it easier to remove a sentence from a list

bakananbanjin:
this should close #895
its me again, I made some more adjustments. For starts I show now all public lists and own lists instead of just the users own lists. I think its for the user more interesting to see in which (public) lists the sentences is. he only can remove them if he got the remove right. also the options to remove sentences are only visible for members.

other than that no changes in the old functionality. I changed according to your @AndiPersti suggestions. 
except the first because i didnt knew what you ment

> What do you mean?
I can only see that the space above the Lists heading got bigger which I don't think is a good idea because now it doesn't match the other modules in the sidebar any more.

ok still looking forward to hear from you guys what you think.

(and one more question how does i update a pull request if i make further changes? just press the button again?)

everyone who goes to bed now like me I wish a good night everyone else a nice day morning and evening XD


AndiPersti:
Thanks for cleaning up the PR :+1: 

> For starts I show now all public lists and own lists instead of just the users own lists.

I'm not sure whether that's a good idea. #1800 discusses some privacy issues related to showing public lists on the sentence page.

> except the first because i didnt knew what you ment

In your old version you've added an additional `div` which added some whitespace above the heading `Lists`. This isn't the case anymore so you can ignore this comment.

But I can see now, that you've actually changed the design of the tag module by listing each tag on its own line. As with the displaying of all public lists, I think that needs further discussion.

In general I would suggest to focus on the specific issue and don't mix in further changes if they aren't necessary to fix the problem.
Especially changing the UI is a sensitive issue because it will quite often annoy some users. :smile:

> (and one more question how does i update a pull request if i make further changes? just press the button again?)

You just need to `git push` the changes to your branch. The PR will be updated automatically.

bakananbanjin:
> I'm not sure whether that's a good idea. #1800 discusses some privacy issues related to showing public lists on the sentence page.

I see, I still think if someone has put his list on public its public information. But I am not here to decide it. Just in my opinion for usersit could be beneficial to see all list with this sentences.

> But I can see now, that you've actually changed the design of the tag module by listing each tag on its own line. As with the displaying of all public lists, I think that needs further discussion.

Same argument then above, but also your call. I think you will discuss it with the other developers. Just let me know.

The changes are all minor, so if you have decided how to proceed, i will change it. Other then that looking forward for your feedback and for more work to do XD

src/View/Helper/ListsHelper.php
```diff
@@ -535,33 +578,45 @@ public function displayAddSentenceForm($listId)
...
                     array(
                         'controller' => 'sentences_lists',
                         'action' => 'show',
                         $list['id']
                     )
                 );
+                 if(CurrentUser::isMember())
+                 {
+                    echo $this->_displayRemoveLink($list['id'], $sentenceId, $list['name']);
+                 }
```
AndiPersti, line 609: As written, the remove button is displayed after each list even if the current user is not the list owner.

You should only display the button if the list belongs to the current user.
Currently you don't have this information available but if you modify `SentencesSentencesListsTable::getOwnAndPublicListsForSentence()` to include the list owner's id, the correct condition would be `CurrentUser::get('id') == $list->user_id`.

bakananbanjin:
@AndiPersti Thanks for the time your time to look over my code and all the advises. The simple ones I already changed. (PR request if I done everything, maybe tomorrow) I added some few questions for clarification. 
For now I sign off, cu

AndiPersti:
> I added some few questions for clarification.

I don't see any questions.

bakananbanjin:
@AndiPersti hi i meant I commented on your comments. wwww

bakananbanjin:
Ok now most should be resolved. (the space between x and listname i forgot, but push it in a min). 
1. 
I decided for now that if the list is public+ and its editable by anyone that i still show the x. But if the list is public+ and only editable by its creator I dont show the delete button. But if you think I should only show it by the users list, let me know and I change it:
```
`if($list['user_id'] == $currentUserId || $list['editable_by'] == 'anyone')
                 {
                    echo $this->_displayRemoveLink($list['id'], $sentenceId, $list['name']);
                 }
```
2. 
If the sentence is in no List i still show the UI Module, because I wasnt sure if you think it should get deleted. w8 for your response. 

> This changes the UI by displaying now an empty Lists block on all sentence pages when the sentence doesn't belong to any list. I'm not sure if that is what we want

3. Ok now we come to the main problem, the migration file. I am quite sure that I could write it from hand, but wanted to use 
cake migrations create MyNewMigration
but somehow I can not access the cake files
i tryed;
bin/cake
cake
cake migrations create MyNewMigration
bin$ phinx create MyNewMigration

nothing seems to work
image`
image

I will further try it. For now just wanted to give an update on my workflow. And sry for spelling errors or style errors in messages or code. still learning XD 

@AndiPersti did I miss something from your comments before? If yes sry, you have to remind me again

src/Controller/SentencesListsController.php
```diff
@@ -323,9 +324,18 @@ public function remove_sentence_from_list($sentenceId, $listId)
...
         }
+                
+        if($origin == 0)
+        {
+            return $this->redirect([
+                'controller' => 'sentences',
+                'action' => 'show',
+                $sentenceId
+            ]);
+        }
```
AndiPersti, line 335: You don't need to add an additional parameter to find out where the request comes from. The request contains a `Referer` header with all the necessary information. In a controller the referer string is available through `$this->referer()`.

So what I've had in mind was 
````php
if (strpos($this->referer(), 'sentences/show')) {
    return $this->redirect($this->referer());
}
````

And I think you could also move the condition up just after the call to `$this->SentencesLists->removeSentenceFromList` because in the case where the request comes from `sentences/show` there's nothing else to do.

**Comment by AndiPersti** [comment]
> I decided for now that if the list is public+ and its editable by anyone that i still show the x.

Yes, that's a good catch because I've forgot about collaborative lists. :+1: 

But there's still another little issue: If a user is the creator of a list but the list is editable by no one, there should be no remove button.

So we have two cases when the button should be displayed:
1. The list is editable by anyone
2. The current user is the creator of the list and the list is editable by the creator.

>  Ok now we come to the main problem, the migration file. I am quite sure that I could write it from hand, but wanted to use
>        cake migrations create MyNewMigration
>        but somehow I can not access the cake files

It looks like you run the command from the `Tatoeba/bin` directory but you should call it from `Tatoeba`. If you run `cd ~/Tatoeba && cake` do you get the help text for the `cake` command?

But the code you've posted is a good starting point.

**Comment by bakananbanjin** [comment]
> It looks like you run the command from the Tatoeba/bin directory but you should call it from Tatoeba. If you run cd ~/Tatoeba && cake do you get the help text for the cake command?

I tryed it from every location, even updated migration 
I also dont get the cake help page, dont know why
image

tried backslash slash..
https://book.cakephp.org/3/en/quickstart.html



anyway other then that I fixed this:

> You don't need to add an additional parameter to find out where the request comes from. The request contains a Referer header with all the necessary information. In a controller the referer string is available through $this->referer().

and this

> But there's still another little issue: If a user is the creator of a list but the list is editable by no one, there should be no remove button.

File: src/View/Helper/ListsHelper.php [hunk 3]
```diff
@@ -604,11 +604,13 @@ public function displayListsModule($allListsArray, $sentences)
...
-                 {
-                    echo $this->_displayRemoveLink($list['id'], $sentenceId, $list['name']);
-                 }
-                echo '</li>';
+                if(currentUser::isMember()){
+                    if(($list['user_id'] == $currentUserId && $list['editable_by'] != 'no_one') || $list['editable_by'] == 'anyone')
+                    {
+                       echo $this->_displayRemoveLink($list['id'], $sentenceId, $list['name']);
+                    }
+                   echo '</li>';
```
AndiPersti, line 612: Having this line inside the if-block will result in an unclosed `li` element for guests. I guess most browsers will still be able to render the list correctly because they are used to fixing broken HTML code but I don't think we should rely on that.

config/Migrations/20200610160031_AddPublicplus.php
```diff
@@ -0,0 +1,28 @@
...
+    $table->changeColumn('visibility', 'enum', [
+        'after' => 'modified',
+        'default' => 'unlisted',
+        'null' => false,
+        'values' =>[
+            'private', 
+            'unlisted', 
+            'public', 
+            'public+'] 
+        ])->save();
```
AndiPersti, line 26: The [Phinx docs](https://book.cakephp.org/phinx/0/en/migrations.html#the-change-method) recommend using `update()` instead of `save()` when changing a column inside the `change` method.

And I'm not sure whether `public+` is really a good name for the option. But @trang and @jiru will have the final call about this.

src/Controller/SentencesListsController.php
```diff
@@ -306,26 +306,33 @@ public function add_sentence_to_list($sentenceId, $listId)
      *
      * @param int $sentenceId Id of sentence to be removed from list.
      * @param int $listId     Id of list that contains the sentence.
+     * @param int $origin     From where the function is called from list page (default 1) or from Sentence page.
```
AndiPersti, line 309: This outdated documentation should be removed.

docs/database/tables/sentences_lists.sql
```diff
@@ -26,7 +26,7 @@ CREATE TABLE `sentences_lists` (
   `numberOfSentences` mediumint(8) unsigned NOT NULL DEFAULT '0',
   `created` datetime DEFAULT NULL,
   `modified` datetime DEFAULT NULL,
-  `visibility` enum('private', 'unlisted', 'public') NOT NULL DEFAULT 'unlisted',
+  `visibility` enum('private', 'unlisted', 'public', 'public+') NOT NULL DEFAULT 'unlisted',
```
jiru, line 29: Now that you added the migration script, you can remove the public+ from the sql file. This is because we currently initialize the database with these sql scripts, and then run all migrations on the top of that.

src/View/Helper/ListsHelper.php
```diff
@@ -337,9 +337,17 @@ public function displayVisibilityOption($listId, $value)
...
             <md-radio-group ng-controller='optionsCtrl' ng-model='visibility' ng-change='visibilityChanged()'>
-                <md-radio-button value='public' class='md-primary'>
+                <md-radio-button value='public+' class='md-primary' title='The list will be shown on the sentences page'>
                     <?php /* @translators: visibility option of a list */ ?>
-                    <?=  __('Public') ?>
+                    <?=  __('Public+'); 
+                    echo $this->Html->image(IMG_PATH .'new1.png');
+                    ?>
+                </md-radio-button>
+                <md-radio-button value='public' class='md-primary' title='The list can be found only over Browse -> Browse by list, not at the sentence page'>
```
jiru, line 346: The text inside the `title` attributes is in English and cannot be displayed into other languages (when one selects a different UI language using the top-right dropdown). To make it translatable, use:
```html
<md-radio-button ... title='<?= h(__("blah blah blah")) ?>'>
```

The [`h()`](https://book.cakephp.org/3/en/core-libraries/global-constants-and-functions.html#h) function is a convenience wrapper for PHP’s `htmlspecialchars()`.

jiru:
This is all good, but I think there is room for improvement in the naming:
* "public+" is not very self-explicit. The very fact that you need to add a tooltip to explain what it is shows that this wording is not easy to understand.
* old/new is relative to the present time. In a few months or years, everything will be old already, but both the UI and the filenames `new1.png` and `old1.png` will still use the wording "new" and "old" which wouldn’t make much sense any more.

How about renaming "public" into "listed", and "public+" into "public"?

bakananbanjin:
@jiru thanks for you comments. changed it accordingly. one problem is, we have to change all public lists into listed lists. after we migrate. I am not sure if we can change all values with a migration file. havent found out how to do it. but something like that:

`$table->update()
       ->set(['visibility' => 'listed'])
       ->where(['visibility' => 'public'])
       ->execute();`


other than that 

> old/new is relative to the present time. In a few months or years, everything will be old already, but both the UI and the filenames new1.png and old1.png will still use the wording "new" and "old" which wouldn’t make much sense any more. I removed them for now and remove the images as well. (they didnt look that well either wwwww , I draw with MS paint)

I also mentioned that sometimes ago, just thought for the users it would be a small visual signal that we had some changes. we should have removed the images after a while.

as always thanks for taking the time to check it

trang:
FYI I have sent a private message to some people who used public list to see what they say about the rewording of "Public" into "Listed". I'll let you know if anyone finds it confusing.

jiru:
@bakananbanjin About the update query, have a look at this [example in the documentation](https://book.cakephp.org/phinx/0/en/migrations.html#creating-an-update-query). You need to use `getQueryBuilder()` and also `->set('visibility', 'listed')`.

I see. I think it’s reasonable to have the "new" indicator if we don’t forget to remove it later. You could display it as a superscript using `<sup>`. Make sure to make the string "new" translatable too using `__()`.

Also, it would be good to add an `@translators` note in the line preceding `__("new")`. The `@translators` notes are automatically extracted from the source code and displayed to UI translators on Transifex when they are about to translate that particular string. The note should give them some contextual information about the string "new": where it is displayed in the website and what "new" is referring to. It’s important because UI translators would otherwise have a hard time figuring out how to translate a single word without context. Have a look at the existing `@translators` notes in the file.

bakananbanjin:
@jiru thanks for the comments, hints and explanations. I hope its now OK. One question is, will the migration file run multiple times? If yes we have to remove it after the first running, otherwise the new public lists will be set to listed everytime we migrate.

bakananbanjin:
@jiru @AndiPersti @trang One more question, because we "change" the visibility of lists from public to listed, should we include a feature  "set all my lists public" because its quite tedious to reset my lists manually all the time and my db just has 10 lists. Maybe some users have more than that.

jiru:
> will the migration file run multiple times?

No. Once you run it once, CakePHP remembers it and won’t run it again even if you execute `cake migrations migrate` again.

> should we include a feature "set all my lists public"

I think we can figure this out later from the feedback of users. Let’s do one thing at a time. :slightly_smiling_face:

src/View/Helper/ListsHelper.php
```diff
@@ -337,9 +337,23 @@ public function displayVisibilityOption($listId, $value)
...
                     <?php /* @translators: visibility option of a list */ ?>
-                    <?=  __('Public') ?>
+                    <?=  __('Public'); 
+                    /* @translators: visibility option of a list *                  
+                    /*should be removed after 2020*/
+                    echo '<span style="color:red">' . __(' new');                    
+                    /* @translators: notice of new visibility feature of a list */
+                    
+                    ?>
+                    </span>
```
jiru, line 350: This code works, but I think it would be easier to read if you leave out the markup out of the PHP `<? ?>` tags. For example:
```html
<?= __('Public') ?>
<span style="color:red"><?= __('new') ?></span>
```

jiru: So you included a space before the string "new". Although it looks okay, we’d like to avoid that. Instead, try to use CSS positioning to move the string where you want it to be. This is because in HTML, the contents (characters, including the space character) should be clearly [separated from the presentation](https://en.wikipedia.org/wiki/Separation_of_content_and_presentation).

jiru: About the `@translators` note, I think it’s not exactly doing what you think it is. Only the note *above* the `__()` call is taken into account, not underneath. You can see by yourself how the note is interpreted by running the strings extraction script `./tools/generate_pot.sh` and then `git diff` to check how the file `src/Locale/default.pot` was modified. With your current code, the string is extracted like this (this is an excerpt from `src/Locale/default.pot` after running `./tools/generate_pot.sh`):
```pot
#. visibility option of a list *
#. /*should be removed after 2020
#: github.com/Tatoeba/tatoeba2/tree/dev/src/View/Helper/ListsHelper.php#L346
msgid " new"
msgstr ""
```
So we end up with this note:
```
visibility option of a list *
/*should be removed after 2020
```

AndiPersti:
While I agree that "public" and "listed" are better names, just renaming the options is not enough IMHO.

Old public lists were also listed in the select control of the advanced search form, they were listed in the select control of `exports/index` (not sure whether this endpoint is still needed; AFAIK nothing links to it, so it is only reachable by entering the URL) and they were displayed on `sentences_lists/index`, `sentences_lists/collaborative` and `sentences_lists/of_user`  for everyone (members and guests).

So I guess these parts should also be modified otherwise there won't be any(?) difference between "listed" and "unlisted" lists after merging this PR:
- https://github.com/Tatoeba/tatoeba2/blob/119c63774b3e9ee62eec77c02f8d7b442d8d35f6/src/Form/SentencesSearchForm.php#L346
- https://github.com/Tatoeba/tatoeba2/blob/119c63774b3e9ee62eec77c02f8d7b442d8d35f6/src/Model/Table/SentencesListsTable.php#L105
- https://github.com/Tatoeba/tatoeba2/blob/119c63774b3e9ee62eec77c02f8d7b442d8d35f6/src/Controller/SentencesListsController.php#L109-L110
- https://github.com/Tatoeba/tatoeba2/blob/119c63774b3e9ee62eec77c02f8d7b442d8d35f6/src/Controller/SentencesListsController.php#L126-L127
- https://github.com/Tatoeba/tatoeba2/blob/119c63774b3e9ee62eec77c02f8d7b442d8d35f6/src/Controller/SentencesListsController.php#L355-L356

(BTW: I think it would also be nice to have some tests added to this PR. @bakananbanjin  would you like to work on that?)

src/Controller/SentencesController.php
```diff
@@ -195,11 +195,12 @@ public function show($id = null)
             $tagsArray = $this->Sentences->getAllTagsOnSentence($id);
             $this->loadModel('SentencesSentencesLists');
             $listsArray = $this->SentencesSentencesLists->getListsForSentence($id);
```
AndiPersti, line 197: `$listsArray` is never used so that line is unnecessary.

But since your `getOwnAndPublicListsForSentence` is just a modified copy of `getListsForSentence` and that function was only used for getting the lists to show on the sentence page and so would be obsolete after merging the PR, I think you can simply include your changes in the old version.

src/Model/Table/SentencesSentencesListsTable.php
```diff
@@ -56,6 +56,27 @@ public function getListsForSentence($sentenceId)
...
+    {
+        return $this->find()
+            ->where([
+                'sentence_id' => $sentenceId,
+                'user_id' => CurrentUser::get('id')
+            ])
+            ->orwhere([
+                'sentence_id' => $sentenceId,
+                'visibility' => 'public'
+            ])
```
AndiPersti, line 69: I've just noticed that [`orwhere` is deprecated since 3.5.0](https://api.cakephp.org/3.7/class-Cake.Database.Query.html#_orWhere) so this should better be written as:
````php
->where([
    'sentence_id' => $sentenceId,
    'OR' => ['user_id' => CurrentUser::get('id'), 'visibility' => 'public']
])
````

**Comment by trang** [comment]
Some feedback on this:

> FYI I have sent a private message to some people who used public list to see what they say about the rewording of "Public" into "Listed". I'll let you know if anyone finds it confusing.

We can go ahead with renaming the current "public" option to "listed" and use "public" for lists that are shown on the sentence's page. There were users who found the wording confusing but that was because they thought that public lists were already showing up for everyone on the sentence's page.

**Comment by bakananbanjin** [comment]
@jiru I see will update after i finished @AndiPersti comments as well. 
@AndiPersti I see you are right atm are listed and unlisted lists the same. havent thought of that XD

`(BTW: I think it would also be nice to have some tests added to this PR. @bakananbanjin would you like to work on that?)`
I dont know what you mean by this?  I think I would do it, just need a few directions what a test is and how to do it

@all thanks and we slowly get this done

**Comment by bakananbanjin** [comment]
image
image

The public and listed lists are now shown also in the advanced search. I think this should be fine now.

Still some issues but I am working on it

**Comment by bakananbanjin** [comment]
` public function getPaginatedLists(
        $search = null, $username = null, $visibility = null, $editableBy = null
    ) {
        $conditions = null;
        if (!empty($search)) {
            $conditions['SentencesLists.name LIKE'] = "%$search%";
        }
        if (!empty($username)) {
            $userId = $this->Users->getIdFromUsername($username);
            $conditions['SentencesLists.user_id'] = $userId;
        }
        if (!empty($visibility)) {
            **$conditions['SentencesLists.visibility'] = $visibility;**
        }
        if (!empty($editableBy)) {
            $conditions['SentencesLists.editable_by'] = $editableBy;
        }

        return [
            'conditions' => $conditions,
            'contain' => [
                'Users' => [
                    'fields' => ['username']
                ]
            ],
            'order' => ['created' => 'DESC'],
            'limit' => 50
        ];
    }`

ok I am a little stucked, how can I change the visibility condition, that it takes an OR value 

SELECT * FROM sentences_lists WHERE visibility = "listed" OR visibility = "public";

As I understand it, I get a String in our case the word 'public' so I can not add an other condition for visibility in this construct? hints would be appreciated @AndiPersti

**Comment by AndiPersti** [comment]
>> (BTW: I think it would also be nice to have some tests added to this PR. @bakananbanjin would you like to work on that?)

> I dont know what you mean by this? I think I would do it, just need a few directions what a test is and how to do it

We use [PHPUnit](https://phpunit.de/)[*] for [unit-testing](https://en.wikipedia.org/wiki/Unit_testing) the code (mostly controllers and models). You can find these tests in `tests/TestCase`.

For example, your PR will change `src/Form/SentencesSearchForm::getSearchableLists`. There are [already tests for this method](https://github.com/Tatoeba/tatoeba2/blob/119c63774b3e9ee62eec77c02f8d7b442d8d35f6/tests/TestCase/Form/SentencesSearchFormTest.php#L254-L308) in `tests/TestCase/Form/SentencesSearchFormTest.php`. So you would add some new entries to `tests/Fixture/SentencesListsFixture.php` (We do not use the "real" database for testing but instead use a test database. All the tables and their contents for that test database are defined in the "fixture" files and will be newly created for each test case.) and then change the tests accordingly.

You can than test the code locally running `phpunit` from the root directory. (See also https://github.com/Tatoeba/imouto#tests)

[*] The [CakePHP docs](https://book.cakephp.org/3/en/development/testing.html) have more information how to write tests and sometimes it also helps to look at the [PHPUnit docs](https://phpunit.de/manual/6.5/en/index.html) (we still use the outdated version 6.5!)

File: src/Controller/SentencesController.php [hunk 11]
```diff
@@ -195,9 +195,8 @@ public function show($id = null)
...
-
+
```
AndiPersti, line 198: You have some unnecessary spaces here.

(In order to detect such unnecessary whitespace changes, you could run `git diff` before `git add`ing or `git diff --staged` before `git commit`ting)

src/Form/SentencesSearchForm.php
```diff
@@ -344,6 +344,9 @@ public function getSearchableLists($byUserId) {
                 'OR' => [
                     'user_id' => $byUserId,
                     'visibility' => 'public',
+                    'OR' => [
+                        'visibility' => 'listed',
+                    ]
```
AndiPersti, line 349: How about 
```php
'OR' => [
    'user_id' => $byUserId,
    'visibility IN' => ['listed', 'public']
]
```

src/Model/Table/SentencesListsTable.php
```diff
@@ -103,6 +103,10 @@ public function getSearchableLists()
                 'OR' => [
                     'user_id' => CurrentUser::get('id'),
                     'visibility' => 'public',
+                    'OR' =>[
+                        'visibility'=> 'listed',
+                    ]
```
AndiPersti, line 108: See above.

src/View/Helper/ListsHelper.php
```diff
@@ -337,9 +337,21 @@ public function displayVisibilityOption($listId, $value)
             ?>
             <input type="radio"  name="visibility" data-list-id='<?= $listId ?>'  value="{{visibility}}" checked hidden ng-init="visibility = '<?= $value ?>';"/>
             <md-radio-group ng-controller='optionsCtrl' ng-model='visibility' ng-change='visibilityChanged()'>
-                <md-radio-button value='public' class='md-primary'>
+                <md-radio-button value='public' class='md-primary' title=
+                                 '<?=h(__("The list will be shown on the sentences page"))?>'>
```
AndiPersti, line 341: I think that tooltip should rather be more like `The list is displayed everywhere for everyone`, i.e. we shouldn't tie ourselves to future problems where we may want to display the list on pages other than the sentence page too.

src/View/Helper/ListsHelper.php
```diff
@@ -337,9 +337,21 @@ public function displayVisibilityOption($listId, $value)
...
+                    <?=  __('Public');?>      
+                    
+                    <?php
+                    /*should be removed after 2020*/
+                    /* @translators: notice of new visibility feature of a list */?>
+                    <span style="color:red"><?= __('new')?>
+                    </span>
+                </md-radio-button>
+                <md-radio-button value='listed' class='md-primary' title=
+                                 '<?= h(__("The list can be found over Browse -> Browse by list, not at the sentence page"))?>'>
```
AndiPersti, line 352: Maybe this tooltip also needs some rewording?

AndiPersti:
> ok I am a little stucked, how can I change the visibility condition, that it takes an OR value
> 
> SELECT * FROM sentences_lists WHERE visibility = "listed" OR visibility = "public";
> 
> As I understand it, I get a String in our case the word 'public' so I can not add an other condition for visibility in this construct?

You could just change the condition on [line 217](https://github.com/Tatoeba/tatoeba2/blob/119c63774b3e9ee62eec77c02f8d7b442d8d35f6/src/Model/Table/SentencesListsTable.php#L217) in the model to use the `IN` operator and then pass the array `['listed', 'public']` in the controllers for the `visibility` parameter.

And this won't break calling `getPaginatedLists` with a string parameter, because whenever you use the `IN` operator in the `where` clause in combination with a string value instead of an array value like `where(['visibility IN' => 'public'])` CakePHP will do the right thing and construct the SQL so that it will look like `visibility in ('public')`.

(A rather crude way to test this is adding the line 
````php
debug($this->SentencesLists->find()->where(['visibility' => 'public']));
````
to one of the actions in the `SentencesListsController` and load that page.

You should see in the debugging output the resulting SQL query.

Compare that with changing the debugging line to
````php
debug($this->SentencesLists->find()->where(['visibility' => ['listed', 'public']]));
````

There is also a debugging plugin called [DebugKit](https://book.cakephp.org/debugkit/3/en/) which would simplify debugging and e.g. would show you all SQL queries for a page but you need to install it in your local environment (and also modify some parts in the code; see the docs).)

**Comment by jiru** [comment]
> sry somehow the code dosnt look that well

@bakananbanjin You can use [triple backquotes](https://help.github.com/en/github/writing-on-github/creating-and-highlighting-code-blocks) to format code blocks.

**Comment by bakananbanjin** [comment]
ok again an other update  :)
sry was a little bussy the last days but still working on the issue. 
as @AndiPersti mentioned we still have to do some rewording see his comments. Also I have to create some test for the code. (still pending xD)

other than that I tryed to fix all mentioned issues

File: src/View/Helper/ListsHelper.php [hunk 16]
```diff
@@ -480,6 +492,41 @@ private function _displayRemoveButton($sentenceId) {
...
+
+        $removeSentenceFromListImg =  $this->Html->image(
+            IMG_PATH . 'close.png',
+            array(
+                "class" => "removeFromListButton",
+                "id" => 'deleteFromListButton'.$listId,
+                "alt" => $removeSentenceFromListAlt
+            )
+        );
+        // X link to remove tag from sentence
```
jiru, line 513: This comment is talking about tags, it should be about lists.

src/View/Helper/ListsHelper.php
```diff
@@ -558,10 +608,21 @@ public function displayListsModule($listsArray)
                         $list['id']
                     )
                 );
+                if(currentUser::isMember()){
```
jiru, line 611: `currentUser` should be capitalized into `CurrentUser`. The code works because PHP is not case sensitive, but let’s try to keep case consistent. :slightly_smiling_face:

src/View/Helper/ListsHelper.php
```diff
@@ -558,10 +608,21 @@ public function displayListsModule($listsArray)
...
+                }
                 echo '</li>';
             }
             echo '</ul>';
             echo '</div>';
+        } else {
+            echo '<div class="section md-whiteframe-1dp">';
+            /* @translators: header text on the sidebar of a sentence page */
+            echo $this->Html->tag('h2', __('Lists'));
+            echo '</div>';
```
jiru, line 625: These 4 lines are always executed on the other side of the if branch too, so you can take them out of this if-else block.

src/View/Helper/TagsHelper.php
```diff
@@ -81,7 +81,7 @@ public function displayTagsModule($tagsArray, $sentenceId = null, $sentenceLang
                     $this->displayTag(
                         $tagName, $tagId, $sentenceId, $userId, $username, $date, $sentenceLang
                     );
-
+                    echo '<br>';
```
jiru, line 84: I believe this shouldn’t be part of your pull request.

jiru:
Hello :blush: I wondered what’s the status of this pull request, so I took a closer look and added a few (minor) comments.

@bakananbanjin Did you have time to think about a good wording for the tooltip? What about the tests you mentioned you wanted to write? Do you need some help from us, or you just need some more time?

bakananbanjin:
@jiru 
thx for the update I changed everything according to your comments. the rewording I still wait for your reply.

>  AndiPersti 13 days ago Member
>
>Maybe this tooltip also needs some rewording?
>@bakananbanjin
>bakananbanjin Author Member Pending
>
>@AndiPersti maybe something like: The list is listed and can be searched.
>
>?

And to be honest I thought the test case would be the next assignment after we finished this one (I looked a little bit in to test cases of tatoeba, but I got confused because it was just datasets). 
I am also sry atm our corona restrictions got cut back, so I have to work on weekdays again, so I only have the weekends for tatoeba XD.

But i really would love to get this issue resolved soon! We spent now exactly 1 month on it, I was thinking this would be a job for a day or 2.

greetings

@AndiPersti 
@trang 
@bakananbanjin 
@jiru

alanfgh:
In general, when questions about wording come up, please add me (@alanfgh) to the discussion. 

> "The list can be found over Browse -> Browse by list, not at the sentence page"

This wording does need to be improved, but I'd like to understand the context better. Could you please explain when this tooltip would appear?

AndiPersti:
> This wording does need to be improved, but I'd like to understand the context better. Could you please explain when this tooltip would appear?

This PR adds a new option to the sidebar on the list page where you can choose the visibility of the list. The tooltip is displayed for the old `Public` option which was renamed to `Listed` and should indicate that the list is only visible on the list page (and not on the sidebar of a sentence page; that's what the new `Public` option is used for):
tooltip

AndiPersti:
> (I looked a little bit in to test cases of tatoeba, but I got confused because it was just datasets).

Do you mean the test fixtures inside `tests/Fixture`? These are used to setup the database for testing.

The code for the actual tests is inside `tests/TestCase`. This directory is structured like the `src` directory, i.e. you find for example the tests for `src/Controller/SentencesController.php` in `tests/TestCase/Controller/SentencesControllerTest.php`.

bakananbanjin:
image

thats new too. :(
other than that how is the issue going? still w8 for the right wording. other than that is the code ok?
Its weekend again and the weather is nice XD
and my back hurts www
sorry that was offtopic

@AndiPersti
@trang
@bakananbanjin
@jiru
@alanfgh

AndiPersti:
> thats new too. :(

This is unrelated to your PR. See #2438.

> other than that how is the issue going? still w8 for the right wording. other than that is the code ok?

All in all I think it looks ok.

Are you still trying to add some tests?

trang:
Here's my suggestion for the tooltips of each visibility option.

@alanfgh, do you confirm this is okay? (see @AndiPersti screenshot above for where these tooltips will be displayed)

**Private**
The list is accessible only to you.

**Unlisted**
The list is accessible by anyone but is not listed on the "Browse by list" page.

**Listed**
The list is accessible by anyone and is listed on the "Browse by list" page.

**Public**
The list is listed on the "Browse by list" page and is displayed on the sentence page.

alanfgh:
Looks good. I would just standardize on "accessible to" rather than "accessible by" and maybe make the last option more explicit:

_**Private**
The list is accessible only to you.

**Unlisted**
The list is accessible to anyone but is not listed on the "Browse by list" page.

**Listed**
The list is accessible to anyone and is listed on the "Browse by list" page.

**Public**
The list is accessible to anyone and is listed on the "Browse by list" page, as well as on the sentence page for every sentence it contains._

But you might find the wording for the last item too long.

jiru:
@bakananbanjin Good job with this PR, it is finally ready for merge :+1: I will publish it on tatoeba.org in one week so that UI translators have some time to translate the new strings. Thank you! :blush:

ckjpn:
```
Public
The list is accessible to anyone and is listed on the "Browse by list" page, as well as on the sentence page for every sentence it contains._

But you might find the wording for the last item too long.
```

This is a bit shorter.  Maybe seeing this will trigger an even more concise wording in someone's mind.

The list is accessible to anyone and is listed on the "Browse by list" page, as well as on each listed sentence's page.

ckjpn:
I was testing this on dev.tatoeba.org and noticed that the "Public (new)" doesn't actually get set.
I can click it, but when I return to the list, it reverts back to what it was.

Screen Shot 2020-07-30 at 13 19 40

ckjpn:
Just a quick follow up.  I checked to make sure that it doesn't show on a non-CK account, and it doesn't.
(This account isn't the owner of the list.)

Screen Shot 2020-07-30 at 13 18 03

AndiPersti:
> I was testing this on dev.tatoeba.org and noticed that the "Public (new)" doesn't actually get set.
> I can click it, but when I return to the list, it reverts back to what it was.

Thanks for testing.
I've just pushed a fix and deployed it on the dev server. It should work now.

PR #2381, opened 2020-06-07
Added localization for private messages (Issue 2199)

### LINKED ISSUE #2199: Insert a comma when sent or messages in inbox > 1,000
It occurs to me that on https://tatoeba.org/eng/private_messages/folder/Sent (and on the Inbox folder too) a comma is not inserted when it's > 1,000.

https://prnt.sc/rg12j8

image

The same happened with the advanced search too not long ago.

RyckRichards:
This PR closes #2199.  It allows localization for numbers above 1,000 (displayed like "1000" before).

Thanks a lot for @AndiPersti  and @trang for all the help here and all the members of @Tatoeba/kodoeba :)

### DIFF
--- a/src/Template/PrivateMessages/folder.ctp
+++ b/src/Template/PrivateMessages/folder.ctp
@@ -66,7 +66,7 @@ $this->set('title_for_layout', $this->Pages->formatTitle(
                 <h2 flex>
                     <?php
                     $n = $this->Paginator->param('count');
-                    echo format(__n('{folderName} ({n}&nbsp;message)',
+                    echo $this->Number->format($__n('{folderName} ({n}&nbsp;message)',
                                     '{folderName} ({n}&nbsp;messages)',
                                     $n, true),
                                 compact('folderName', 'n'));
--- a/src/Template/PrivateMessages/folder.ctp
+++ b/src/Template/PrivateMessages/folder.ctp
@@ -66,11 +66,11 @@ $this->set('title_for_layout', $this->Pages->formatTitle(
                 <h2 flex>
                     <?php
                     $n = $this->Paginator->param('count');
-                    $this->Number->format($n)
                     echo format(__n('{folderName} ({n}&nbsp;message)',
                                     '{folderName} ({n}&nbsp;messages)',
                                     $n, true),
-                                    ['folderName'=> $folderName, 'n'=> $this->Number->format($n)];
+                                ['folderName'=> $folderName, 'n'=> $this->Number->format($n)]
+                                );
                     ?>
                 </h2>
 
--- a/src/Template/PrivateMessages/folder.ctp
+++ b/src/Template/PrivateMessages/folder.ctp
@@ -66,10 +66,11 @@ $this->set('title_for_layout', $this->Pages->formatTitle(
                 <h2 flex>
                     <?php
                     $n = $this->Paginator->param('count');
-                    echo $this->Number->format($__n('{folderName} ({n}&nbsp;message)',
+                    $this->Number->format($n)
+                    echo format(__n('{folderName} ({n}&nbsp;message)',
                                     '{folderName} ({n}&nbsp;messages)',
                                     $n, true),
-                                compact('folderName', 'n'));
+                                    ['folderName'=> $folderName, 'n'=> $this->Number->format($n)];
                     ?>
                 </h2>
 


src/Template/PrivateMessages/folder.ctp
```diff
@@ -66,7 +66,7 @@ $this->set('title_for_layout', $this->Pages->formatTitle(
                 <h2 flex>
                     <?php
                     $n = $this->Paginator->param('count');
-                    echo format(__n('{folderName} ({n}&nbsp;message)',
+                    echo $this->Number->format($__n('{folderName} ({n}&nbsp;message)',
                                     '{folderName} ({n}&nbsp;messages)',
                                     $n, true),
                                 compact('folderName', 'n'));
```
trang, line 72: With your modifications, if you try to go to the [private messages locally](http://localhost:8080/eng/private_messages/folder/Inbox), you should see some error in your browser. The page will not be displayed as it should.

1. There isn't supposed to be any `$` after `format(`.
2. You are here mixing the `format()` function with `$this->Number->format()`. These are two different functions. 
    - The `format()` function should not be touched (that is, the line that you modified should not be changed).
    - You want to use `$this->Number->format()` around the number only. Not around the whole text. The number is `$n`.

Now this piece of code is actually a little bit more advanced and probably harder to understand for a beginner because there are functions inside of functions. The functions involved are:

**format()**
This function replaces placeholders inside a string by the value of the associated variable. For instance, if you have `$n = 3` and you do `echo format('I have {n} apples', ['n' => $n])`, it will print `I have 3 apples`.

**__n()**
This function converts the English text into other languages when the UI language is different than English. It handles plural forms of a text.

**compact()**
This function is a short way to create an array. Here for instance, `compact('folderName', 'n')` is just a shortcut for writing `['folderName' => $folderName, 'n' => $n]`.

My hint is: 
1. Change temporarily `$n = $this->Paginator->param('count');` by `$n = 1234`. This will force the text to display "1234 messages" in the private messages page.
2. Replace `compact('folderName', 'n')` by `['folderName' => $folderName, 'n' => $n]`. You should see now difference on the web page.

From there, try to figure out what is left to do so that it displays `1,234 messages` instead of `1234 messages`.
Once you figure it out, don't forget to revert `$n = 1234` back to `$n = $this->Paginator->param('count');`.

RyckRichards:
Uhm... I got this

Error in: ROOT\src\Template\PrivateMessages\folder.ctp, line 69
If you want to customize this error message, create src\Template\Error\error500.ctp

Going to work on it, thank you :)

src/Template/PrivateMessages/folder.ctp
RyckRichards: Uhm... I got this

Error in: ROOT\src\Template\PrivateMessages\folder.ctp, line 69
If you want to customize this error message, create src\Template\Error\error500.ctp

Going to work on it, thank you :)

src/Template/PrivateMessages/folder.ctp
```diff
@@ -66,10 +66,11 @@ $this->set('title_for_layout', $this->Pages->formatTitle(
                 <h2 flex>
                     <?php
                     $n = $this->Paginator->param('count');
-                    echo $this->Number->format($__n('{folderName} ({n}&nbsp;message)',
+                    $this->Number->format($n)
```
AndiPersti, line 69: You don't need this line, so just remove it.

src/Template/PrivateMessages/folder.ctp
```diff
@@ -66,10 +66,11 @@ $this->set('title_for_layout', $this->Pages->formatTitle(
...
+                    echo format(__n('{folderName} ({n}&nbsp;message)',
                                     '{folderName} ({n}&nbsp;messages)',
                                     $n, true),
-                                compact('folderName', 'n'));
+                                    ['folderName'=> $folderName, 'n'=> $this->Number->format($n)];
```
AndiPersti, line 73: You are missing a closing `)` between `]` and `;` at the end of the line.

RyckRichards: Thanks, Andi! I'm getting this now

> syntax error, unexpected ''{folderName} ({n}&nbsp;messag' (T_CONSTANT_ENCAPSED_STRING), expecting ',' or ')'

> Error in: ROOT\src\Template\PrivateMessages\folder.ctp, line 70


Line 70:

`$this->Number->format '{folderName} ({n}&nbsp;messages)',
                                    $n, true),
                                ['folderName'=> $folderName, 'n'=> $this->Number->format($n)]);`

AndiPersti: The whole block should look like
````php
<?php
    $n = $this->Paginator->param('count');
    echo format(__n('{folderName} ({n}&nbsp;message)',
                    '{folderName} ({n}&nbsp;messages)',
                    $n, true),
                ['folderName'=> $folderName, 'n'=> $this->Number->format($n)]
    );
?>
````

The line starting with `$n = ...` gets the number of messages.

The `echo format(...` outputs the string. It uses the CakePHP helper function `__n()` to get the correct translation. So the result of this function will be either `{folderName} ({n}&nbsp;message)` or `{folderName} ({n}&nbsp;messages)` depending on the number of messages, i.e. the value stored in `$n`.
`{folderName}` and `{n}` are placeholders which we want to replace with the correct values. So we use another helper function `format` to do this. `folderName` will be replaced with the value stored in `$folderName` and `$n` will be replaced with the localized value stored in `$n`.

RyckRichards: Oh, I got it now... How silly I am!

The final result should be localized, not the "processing result". Thanks a lot, @AndiPersti 

I committed the changes - https://github.com/Tatoeba/tatoeba2/pull/2381/commits/5b7d9c11c61a558beae9491023adc773658c9b81

PR #2382, opened 2020-06-08
Fix expand label when adding translations while no translations are displayed

### LINKED ISSUE #2380: Adding a translation shows the button to display more translations but the button does nothing
1. Look for sentences in English without French translation
2. Translate the sentence

image

(2.5 You can see that only my French translation is displayed)
3. Click the button to show more translations
4. Nothing happens although the button label is modified to "Show less translations"

This may have been reported before. If not, it's related to another issue (that I forgot the number) explaining that the "additional translations" shouldn't be loaded anyway, because my search criterion is to show translations only in French. These additional translations appears when one add a translation that is duplicate, do they? Maybe not... 
[EDIT] The same thing happen if I link an indirect translation. For example, a French indirect translation is displayed (due to my search criterion), I link it, the "Show 24 more" button is displayed, clicking doesn't do anything.

Anyway, that's even weirder in this case since the buttons do happen but has no effect, even if we see its label changing.

trang:
This PR fixes #2380.

### DIFF
--- a/src/Model/Table/SentencesTable.php
+++ b/src/Model/Table/SentencesTable.php
@@ -684,7 +684,7 @@ class SentencesTable extends Table
             ];
         }
 
-        if (isset($what['translations'])) {
+        if (isset($what['translations']) && $what['translations']) {
             $translationFields = [
                 'id', 'text', 'lang', 'correctness', 'script',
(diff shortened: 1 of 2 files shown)


jiru:
Thank you! :+1:

Issue #1880, opened 2019-05-02
Make the search engine treat the letters U+06D5 as U+0647 and U+06AD as U+0643 when searching in Ottoman Turkish


soliloquist-tatoeba:
https://www.fileformat.info/info/unicode/char/06d5/index.htm

https://www.fileformat.info/info/unicode/char/0647/index.htm

I tried to explain the reason on #1872.

https://github.com/Tatoeba/tatoeba2/issues/1872#issuecomment-488438111

@alanfgh said this might be possible so I'm opening the issue.

jiru:
Yes, it’s possible.

soliloquist-tatoeba:
I have some more requests:

ڭ (U+06AD) ⟶ ك (U+0643)

ڭ is a rarely used accented letter. ك is usually used instead of it.

It would be better if the search engine treated ڭ as ك.

https://www.fileformat.info/info/unicode/char/06ad/index.htm

https://www.fileformat.info/info/unicode/char/0643/index.htm


I also noticed that the Arabic diacritics aren't ignored in searches. I think they should be.

https://tatoeba.org/eng/sentences/search?query=%22%D8%B3%D9%8F%D9%88%D9%8A%D8%B3%D9%92%D8%B1%D9%8E%D8%A7%22&from=und&to=und

https://tatoeba.org/eng/sentences/search?query=%22%D8%B3%D9%88%D9%8A%D8%B3%D8%B1%D8%A7%22&from=und&to=und

Both searches above are "Switzerland" in Arabic. Wouldn't it be better if both results were combined? I'm not an Arabic speaker so I can't speak for sure about it, but for Ottoman Turkish, it would be better if the following diacritics were ignored in searches.

https://www.fileformat.info/info/unicode/char/0650/index.htm
https://www.fileformat.info/info/unicode/char/0651/index.htm
https://www.fileformat.info/info/unicode/char/0652/index.htm
https://www.fileformat.info/info/unicode/char/064e/index.htm
https://www.fileformat.info/info/unicode/char/064f/index.htm
https://www.fileformat.info/info/unicode/char/064b/index.htm
https://www.fileformat.info/info/unicode/char/064c/index.htm
https://www.fileformat.info/info/unicode/char/064d/index.htm

jiru:
@soliloquist-tatoeba The things you requested in this issue have been applied on the production server. You can try on https://tatoeba.org/.

soliloquist-tatoeba:
[...]

PR #2384, opened 2020-06-10
Update Manticore configuration for Arabic script (closes #1595, #1880)

### LINKED ISSUE #1595: Search engine should ignore Arabic diacritics
As reported by OsoHombre on the wall: 

> Can Tatoeba's search engine just ignore Arabic diacritics? Because the search engine takes into account Arabic diacritics, countless sentences with diacritics are ignored in the search results when a word is searched without diacritics, and vice versa: when a word is search with diacritics, countless sentences that contain the same word but without diacritics, are ignored in the search results.
> 
> For example, when I search the word "الشعب" (without the 'shadda' diacritic that indicates gemination), this sentence is ignored in the search results:
> 
> https://tatoeba.org/eng/sentences/show/668466
> 
> This is a serious flaw on Tatoeba as far as the Arabic language is concerned. Because I will soon be adding literally tens of thousands of Arabic sentences to this website (almost exclusively translations of my own English sentences), I would love to see this issue resolved as soon as possible.
> 
> Thank you

https://tatoeba.org/ita/wall/show_message/29404#message_29404

Yorwba:
Vowel marks in Arabic script are optional, so someone searching for a word without diacritics would expect to also get results that have them and vice versa. (See issue #1595.) Adding those characters to `ignore_chars` makes that possible. The wrinkle with that approach is that words with _different_ vowel marks in the same position become indistinguishable. For Arabic, we can do better by _not_ ignoring diacritics, since the stemmer already removes them. Then someone who wants to find an exact match can simply turn off the stemmer with `=`.

I also unified the graphical variants of each of the letters Heh, Yeh, and Kaf and added the Arabic-Indic Digits and their extended variants.

Specifically for Ottoman Turkish, I also implemented the character equivalences requested in #1880.

These additions to the configuration file create some lines that are rejected by Manticore as too long. I already kicked this can down the road once in 7b8077897 but that only bought us a bit of breathing room. I added a step for breaking up lines if they get too long, which should solve the problem once and for all.

(Finally some additional progress on issue #1970.)

### DIFF
--- a/src/Shell/SphinxConfShell.php
+++ b/src/Shell/SphinxConfShell.php
@@ -721,10 +721,32 @@ EOT;
         return $conf;
     }
 
+    /**
+     * Manticore has an internal buffer limit of 8192 bytes:
+     * https://github.com/manticoresoftware/manticoresearch/blob/69c389347d44f136cd93b899640d7f4f4a6ce750/src/sphinxutils.cpp#L1201
+     * To ensure that the configuration stays below the limit, we need to add
+     * a backslash-escaped newline into overly long lines.
+     */
+    public function escape_long_lines($conf, $limit = 8192) {
+        $limit -= 3; // we need enough room for the escape characters + \0
+        $lines = explode("\n", $conf);
+        $conf = "";
+        foreach ($lines as $line) {
+            for ($i = 0; $i < strlen($line); $i += $limit) {
+                $conf .= substr($line, $i, $limit);
+                if ($i + $limit < strlen($line)) {
+                    $conf .= "\\\n";
+                }
+            }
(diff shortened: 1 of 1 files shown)


jiru:
Thank you Yorwba! :v: 

I tried to guess which language are using the Arabic script and I reindexed the following languages on the production server: ara arz acm arq apc ary afb ota chg pnb pes pus snd uig urd

Yorwba:
There are some other languages that contain Arabic characters in some sentences: deu eng hun kab kas kur swe tat

Some of these are due to direct quotes, others may be classification errors.

jiru:
Thanks. I reindexed these too.

PR #2388, opened 2020-06-11
Add stemming for Indonesian

Yorwba:
I noticed that there were more stemmers listed in the output of `strings /usr/bin/searchd | grep UTF_8_stem$` than appeared in the `$morphology` array. It was a bit difficult to see what was missing because the array was listing languages in no particular order, but after sorting it alphabetically, it was easy to see that Snowball also supports the Porter stemmer (for English) and Indonesian.

This PR enables the Indonesian stemmer.


jiru:
Thanks, Yorwba :sunglasses: 
It looks like Indonesian stemming is now working.

PR #2397, opened 2020-06-15
Issue 2325

ftumsh:


jiru:
Hi @ftumsh!

As you guessed, the export script is trying to read the SQL scripts from /var/www-prod/ but this is not working. We should fix this, but for the time being here is a tip. You can create a symbolic link to make the script work out of the box, like this:
```sh
```

jiru:
I tried your code and it works great! :+1:

However, it only produces the file and there is no mention of its existence on the website. This means only curious people looking at the [server listing](https://downloads.tatoeba.org/exports/) may discover your new `sentences_based_on_id.tar.bz2`. Most people use the [Downloads page](https://tatoeba.org/downloads) instead. So it would be great if you could update the Downloads page to present your file in the same way other files are.

In the past, I added transcriptions to our exports in #2249. You are welcome to get inspiration from that.

src/Template/Pages/downloads.ctp
```diff
@@ -247,6 +250,34 @@ $transcriptionsOptions = $this->Downloads->createOptions('transcriptions');
...
+            <h2><?= __('Sentences Based On ID') ?></h2>
+            <dl>
+                <dt><?= $filename ?></dt>
+                <?= $this->element(
+                    'per_language_files',
+                    [
+                        'model' => 'sentences',
+                        'options' => $sentencesBasedOnIdOptions
+                    ]
+                ) ?>
```
jiru, line 265: The `per_language_files` element is meant to be used for export files that are split by languages. For example, the file `sentences.csv` is further proceeded by the [export script](https://github.com/Tatoeba/tatoeba2/blob/dev/docs/cron/export.sh) to produce multiple files inside the [per_language directory](https://downloads.tatoeba.org/exports/per_language/). `$this->element('per_language_files', ...)` allows to download these individual files with a dropdown box.

Since your export is only consisting of a single file `sentences_based_on_id.tar.bz2`, you should only display a single link to it. You can get inspiration from the way `links.tar.bz2` is displayed on the Downloads page.

src/Template/Pages/downloads.ctp
```diff
@@ -247,6 +250,34 @@ $transcriptionsOptions = $this->Downloads->createOptions('transcriptions');
...
+                <dt><?= $description ?></dt>
+                <dd>
+                    <?= format(
+                        __(
+                            'Contains all the sentences in the selected language. '.
+                            'Each sentence is associated with a unique id and the id'.
+                            ' of the sentence it is based on.'
```
jiru, line 272: It’s a bit more complicated actually. The value for `based_on_id` can be 3 different things:
* **a number greater than zero**: gives the sentence number it’s based on;
* **zero**: means the sentence is original, not based on any sentence;
* **`\N`** (which is how NULL is displayed in the CSV file): we don’t know. This is actually quite rare but we have a handful of sentences for which we don’t know whether they have been added as translations or not.

src/Template/Pages/downloads.ctp
```diff
@@ -46,6 +46,8 @@ $description = __('File description');
 // Field names
 /* @translators: field name in Fields and structure on Downloads page */
 $sentence_id = __('Sentence id');
+/* @translators: field name in Fields and structure on Downloads page */
+$sentence_based_on_id = __('Sentence Id of base sentence');
```
AndiPersti, line 50: Style nitpick: All the other field name descriptions use `id` (lowercase letters) so I suggest using `Sentence id of base sentence`.

jiru:
I postponed the milestone to give some time to UI translators to translate the new strings.

PR #2400, opened 2020-06-16
Added draft button for sort list

gioele-santi:
This is the base to start adding a button to show the list.


gioele-santi:
I still have to figure out different things like hot to show the list and the icon, then when it works I will change all other sortable lists in the website.

src/Template/SentencesLists/index.ctp
```diff
@@ -61,31 +61,190 @@ $this->set('title_for_layout', $this->Pages->formatTitle($title));
...
+                                        else {
+                                            echo 'blank';
+                                        }
+                                    ?>
+                                </md-icon>
+                                <span style="padding-right: 12px">
+                                    <?php echo $this->Paginator->sort('name', __('list name'), ['direction'=>'asc', 'lock'=>true]); ?>
+                                </span>
+                            </md-button>
+                        </md-menu-item>
```
trang, line 95: You could consider creating an [element](https://book.cakephp.org/2/en/views.html#elements) for these `<md-menu-item>`.

Then here you would just have something like this:
```
echo $this->element('sort_option', [
    'param' => 'name',
    'direction' => 'asc',
    'label' => __('list name')
]);
````

Comment (gioele-santi) [reply src/Template/SentencesLists/index.ctp:95]: Yes I thought my solution was not the most elegant. 
Now I made two of these Elements as the first page I made has a couple of lines with different code, but since it is used inly twice, I could change it and add a parameter to make only one.

File: src/Template/Element/sort_option.ctp [hunk 2]
```diff
@@ -0,0 +1,17 @@
+<md-menu-item>
+    <md-button>
+        <md-icon>
+            <?php 
+                if (strcmp($this->Paginator->sortKey(),$param) == 0 && strcmp($this->Paginator->sortDir(),$direction) == 0) {
```
AndiPersti, line 5: Why do you use `strcmp` instead of a "simple" string comparison `s1 == s2`?

src/Template/Users/all.ctp
```diff
@@ -89,22 +89,54 @@ $this->set('title_for_layout', $this->Pages->formatTitle(__('Members')));
...
+                            'label' => __('Newest first')
+                    ));?>
+
+                    <?php echo $this->element('sort_option', array(
+                            'param' => 'since',
+                            'direction' => 'asc',
+                            'label' => __('Oldest first')
+                    ));?>
+
+                    <?php echo $this->element('sort_option_role', array(
```
AndiPersti, line 123: The only difference between `sort_option` and `sort_option_role` is the creation for the pagination link. But `Pagination->sortForRole` is only necessary for displaying the correct arrow. Since your PR gets rid of the arrows, that method isn't useful anymore and you can just use `Paginator->sort()` (i.e. the `sort_option` element) as with the other options (and also remove `Element/sort_options_role.ctp`).

src/Template/Element/sort_option.ctp
```diff
@@ -0,0 +1,17 @@
...
+                    echo 'check'; 
+                }
+                else {
+                    echo 'blank';
+                }
+            ?>
+        </md-icon>
+        <span style="padding-right: 12px">
```
AndiPersti, line 13: Maybe move the CSS styling to `webroot/css/layouts/elements.css`?

src/Template/Reviews/of.ctp
```diff
@@ -103,10 +103,58 @@ $this->set('title_for_layout', $this->Pages->formatTitle($title));
...
+                                'label' => __('Newest sentences')
+                        ));?>
+
+                        <?php echo $this->element('sort_option', array(
+                                'param' => 'sentence_id',
+                                'direction' => 'asc',
+                                'label' => __('Oldest sentences')
+                        ));?>
+                            
+                            <!-- Status, check key name ans label name -->
```
AndiPersti, line 149: I guess this comment can be removed.

AndiPersti:
Currently when I open a page with your new sort menu, the options will be displayed in their "raw" (i.e. unprocessed by the AngularJS compiler) form for a short time. You can prevent this by adding `ng-cloak` to an HTML element (see the [AngularJS docs](https://docs.angularjs.org/api/ng/directive/ngCloak)).
But I'm not sure where to place it: on `md-menu`, on `div class="md-toolbar-tools"`, or on `md-toolbar`?

src/Template/Element/sort_option.ctp
gioele-santi: Defensive programming. In another project I was working I had many issues because of string comparison as some inputs where not strings and this prevent many errors. So I do like this because I am used to.

AndiPersti:
While having an element for outputting a single menu item helps reducing code repetition, after applying my suggested changes you'll end up with a single `sort_option` element.
Now there is still a lot of repetition in the main templates because you need to call the element for each menu item.
So how about creating a `sort_menu` element instead which gets passed an array containing all necessary information for all the menu items and in the `sort_menu` you iterate through the array and output each item. Then I think you won't need the `sort_option` element anymore because the code for a single menu item will only exist once inside the loop.

src/Template/Element/sort_option.ctp
gioele-santi: Anyway I will change this while taking care of the other changes listed below.

AndiPersti: If you want to be sure that both operands are strings, you should use `===`, i.e. strict comparison. (Actually we should probably use always `===` when testing whether two strings are equal, see https://stackoverflow.com/questions/3333353/string-comparison-using-vs-strcmp)
I think using `strcmp` is only useful if you want to know which string is greater than the other.

src/Template/Reviews/of.ctp
```diff
@@ -103,10 +103,58 @@ $this->set('title_for_layout', $this->Pages->formatTitle($title));
...
+
+                        </md-menu-content>
+                    </md-menu>
+
             </div>
         </md-toolbar>
 
-        <div class="sortBy">
+        <!-- <div class="sortBy">
```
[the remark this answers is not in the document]
gioele-santi: Removed. I forgot to remove after testing the page.

src/Template/SentencesLists/of_user.ctp
```diff
@@ -71,27 +71,46 @@ $this->set('title_for_layout', $this->Pages->formatTitle($title));
...
+                                    'label' => __('Newest first')
+                            ));?>
+
+                            <?php echo $this->element('sort_option', array(
+                                    'param' => 'created',
+                                    'direction' => 'asc',
+                                    'label' => __('Oldest first')
+                            ));?>
+                            
+                            <!-- Status, check key name ans label name -->
```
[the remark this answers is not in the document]
gioele-santi: Removed.

src/Template/Reviews/of.ctp
gioele-santi: Removed.

gioele-santi:
> Currently when I open a page with your new sort menu, the options will be displayed in their "raw" (i.e. unprocessed by the AngularJS compiler) form for a short time. You can prevent this by adding `ng-cloak` to an HTML element (see the [AngularJS docs](https://docs.angularjs.org/api/ng/directive/ngCloak)).
> But I'm not sure where to place it: on `md-menu`, on `div class="md-toolbar-tools"`, or on `md-toolbar`?

Yes I did notice that but I was not sure about the cause. I will check what's the better place for `ng-cloak` while making and testing the new all in one `sort_menu`.

src/Template/Element/sort_option.ctp
gioele-santi: Ok I changed to `===`

src/Template/SentencesLists/index.ctp
```diff
@@ -61,31 +61,70 @@ $this->set('title_for_layout', $this->Pages->formatTitle($title));
...
+                                'label' => __('Highest number of sentences')
+                        ));?>
+
+                        <?php echo $this->element('sort_option', array(
+                                'param' => 'numberOfSentences',
+                                'direction' => 'asc',
+                                'label' => __('Lowest number of sentences')
+                        ));?>
+
+                        <?php echo $this->element('sort_option_defaults', array(
```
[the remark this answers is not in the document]
gioele-santi: Merged in unique sort menu element

src/Template/Users/all.ctp
gioele-santi: As above, I merged in new sort menu element

src/Template/Element/sort_option.ctp
gioele-santi: I forgot this tomorrow I will change it.

gioele-santi:
I created a unique sort menu element that works for all cases. I still have to fix the css styling in html tag and one small issue I noticed: I don't know why but in `show_angular.ctp` and `sentences/of_user.ctp` when parameters are `modified` and `desc`, it does not add the check mark. I wonder if it has to do with defaults or something else, as the generated sort link seems different from others (no get parameters).

AndiPersti:
> I don't know why but in `show_angular.ctp` and `sentences/of_user.ctp` when parameters are `modified` and `desc`, it does not add the check mark.

The reason is that the corresponding controller actions for both templates set the default sort key using an explicit table prefix:
https://github.com/Tatoeba/tatoeba2/blob/49f922be68cb56298306e9c481605a6655c938e5/src/Controller/SentencesController.php#L700
https://github.com/Tatoeba/tatoeba2/blob/49f922be68cb56298306e9c481605a6655c938e5/src/Controller/SentencesListsController.php#L180
So the key returned by `Paginator->sortKey()` will be `Sentences.modified` and not `modified`.

I think you can safely remove the table prefix because the paginator will always use the columns from the main table. If we would ever want to use a column from an associated table for sorting we would need to set a whitelist for the sort keys (see [Control which Fields Used for Ordering](https://book.cakephp.org/3/en/controllers/components/pagination.html#control-which-fields-used-for-ordering)).

src/Template/Element/sort_option.ctp
gioele-santi: I added a sortOption class to css file.

gioele-santi:
I fixed the default prefix and moved styling in css file. It should be ok to be reviewed now. I hope I didn't forget to delete comments added during work.

jiru:
Hello Gioele and thanks for working on this! :slightly_smiling_face: I am going to review your code now you set the PR as ready for review.

I found a bug. From the user menu, I go to "My review", then click on "Sort by" → "Oldest first". Nothing happens.

src/Template/Element/sort_menu.ctp
```diff
@@ -0,0 +1,35 @@
+<md-menu md-offset="5 50" md-position-mode="target-right target">
+    <md-button ng-click="$mdOpenMenu($event)">
+        <md-icon>sort</md-icon> Sort by
```
jiru, line 3: The string "Sort by" is part of the UI so it should be translatable. Please [put in into a `__()` call](https://book.cakephp.org/3/en/core-libraries/internationalization-and-localization.html).

src/Template/Element/sort_menu.ctp
```diff
@@ -0,0 +1,35 @@
...
+    </md-button>
+    <md-menu-content ng-cloak>
```
jiru, line 5: You set `ng-cloak` for the menu content, but the sort button and its icons also need to be hidden during page load. See how it looks during page load:

image


And then after load:

image

src/Template/Element/sort_menu.ctp
```diff
@@ -0,0 +1,35 @@
...
+                            if ($this->Paginator->sortKey() === $option['param'] && $this->Paginator->sortDir() === $option['direction']){
+                                echo 'check'; 
+                            }
+                            else {
+                                echo 'blank';
+                            }
+                        ?>
+                    </md-icon>
+                    <span class="sortOption">
+                        <?php echo $this->Paginator->sort($option['param'], $option['label'], ['direction'=>$option['direction'], 'lock'=>true]); ?>
```
jiru, line 25: I’ve checked the documentation of the Paginator helper, but I still don’t understand the meaning of the `lock` parameter. Do you mind explaining why it’s needed?

src/View/Helper/PaginationHelper.php
```diff
@@ -144,13 +144,13 @@ public function display($extramParams = array(), $extraNumbersOptions = array())
      *
      * @return string
      **/
-    public function sortForRole() {
+    public function sortForRole($label = 'Member', $options = array()) {
```
jiru, line 147: It looks like this function is not needed any more, so you might as well remove it altogether.

src/Template/Element/sort_menu.ctp
gioele-santi: OK I am updating the code.

src/Template/Element/sort_menu.ctp
gioele-santi: I see. Maybe my browser cached the pages so I could not see this detail and the other `ng-cloak` seemed to work enough.

I will move `ng-cloak` in the top `md-menu` and see if one is enough for all.

src/Template/Element/sort_menu.ctp
gioele-santi: That was suggested to me. But I checked myself.
Basically if`locked = false`, when you click on the link, the link in reloaded page is generated in the other direction. 
So an `asc` link when clicked will be a `desc` link in new page.
It was useful with the old sort links as it was posible to manage both directions with a single link, but in the requirements I had to generate a link for each direction option.

AndiPersti: The old version displayed a single link (e.g. "created") and clicking it toggled the direction (that's how `Paginator->sort()` usually works).

But in this new version we display two links for each sort key (one for ascending order and one for descending order). Without `lock`, clicking on the current sort option would toggle the direction which is not what we want.

So strictly speaking `lock` is only necessary for the current sort option because all the other options still work for the first click which will reload the page.

An alternative would be to not create a link for the current sort option, i.e. clicking on it would do nothing.

src/View/Helper/PaginationHelper.php
gioele-santi: I was not completely sure, but now I removed it.

src/Template/Element/sort_menu.ctp
gioele-santi: I will make a small change, I already fixed other small issues.
I think I will use the same check for the mark to find out if the link needs to be created.

jiru: Thank you both for clarifying :pray:

AndiPersti:
> I found a bug. From the user menu, I go to "My review", then click on "Sort by" → "Oldest first". Nothing happens.

I guess you didn't actually click on the link. :smile: 

The problem is that `md-button` is used without an `href` attribute. So clicking on it won't do anything. Only clicking exactly on the `a` element inside it will work.

@gioele-santi Instead of the `a` element inside `md-button` you need to provide the URL for the button in the `href` attribute. In addition you can't use `Paginator->sort()` for creating the URL because that method creates an `a` element. You need to use `Paginator->generateUrl()`.

src/Template/Element/sort_menu.ctp
gioele-santi: I modified it it should be ok now.

src/Template/Element/sort_menu.ctp
gioele-santi: I changed the menu with new code to work using `md-button` anf `generateUrl` the `lock` option is not used anymore as url is not generated for current option-

gioele-santi:
I updated the code that generates the menu and now buttons are generated instead of links inside buttons. This way every click produces an effect. And code looks much cleaner.

src/Template/Sentences/of_user.ctp
```diff
@@ -105,26 +105,21 @@ $this->set('title_for_layout', $this->Pages->formatTitle($title));
...
         <md-content layout-padding>
-        <div class="sortBy">
-            <strong><?php echo __("Sort by:") ?> </strong>
-            <?php
-            if ($onlyOriginal) {
-                $urlOptions = $this->Paginator->generateUrlParams(
-                    array('?' => array('only_original' => ''))
-                );
-                $this->Paginator->options(array('url' => $urlOptions));
-            }
```
AndiPersti, line 120: On the page for the sentences of an user, there is an option in the sidebar which allows you to only show original sentences. The old behavior was that this option kept its status when you changed the sort mode. So I think you shouldn't remove this part but move it before the `$options` array (line 110).

AndiPersti:
Your PR removed all the translator's notes for the sort options, i.e. the special comments starting with `/* @translators: ` above each line where a translatable string is used. These notes are visible for the translators and give them hints to where/how the string is used.

I guess we want to have them also with the new options.

src/Template/Sentences/of_user.ctp
gioele-santi: I restored those lines. I hope those options are also applid to generateUrl function.

gioele-santi:
> Your PR removes all the translator's notes for the sort options, i.e. the special comments starting with `/* @translators: ` above each line where a translatable string is used. These notes are visible for the translators and give them hints to where/how the string is used.
> 
> I guess we want to have them also with the new options.

Should I add them for each option or just one special comment is enough for each page?

I mean, this way?
```
/* @translators: sort option in the "Sentences of user" page */
array( 'param' => 'modified', 'direction' => 'desc', 'label' => __('Most recently updated')),
/* @translators: sort option in the "Sentences of user" page */
array( 'param' => 'modified', 'direction' => 'asc', 'label' => __('Least recently updated')),
```

Or this way?
```
/* @translators: sort option in the "Sentences of user" page */
array( 'param' => 'modified', 'direction' => 'desc', 'label' => __('Most recently updated')),
array( 'param' => 'modified', 'direction' => 'asc', 'label' => __('Least recently updated')),
```

AndiPersti:
> Should I add them for each option or just one special comment is enough for each page?

You need to put them before each option. One comment only works for the immediately following translation function (`__()`, `__n()`, ...).

src/Template/Sentences/of_user.ctp
AndiPersti: Yes, `Paginator->options()` sets the default options for every Paginator method called afterwards.

gioele-santi:
I restored translator comments and also added in a couple of pages where originals didn't. I hope those comments are ok.

jiru:
@gioele-santi Yes, it’s perfect! :+1: Thank you very much, I think we are good to go now. :slightly_smiling_face:

Issue #205, opened 2014-04-14
Stop storing IP of contributors in the logs


jiru:
I tried to access a local tatoeba installation via IPv6. While going on the login page, the debug showed the following error. The IPv6 network Tatoeba is running on is `fd01:2345:6789:abcd::/64`.

```
Warning (512): SQL Error: 1062: Duplicate entry 'fd01:2345:6789:' for key 'ip' [CORE/cake/libs/model/datasources/dbo_source.php, line 526]
Query: INSERT INTO `visitors` (`timestamp`, `ip`) VALUES (1397479495, 'fd01:2345:6789:abcd:XXXX:XXXX:XXXX:XXXX') 
```

Analysis : The length of the field `visitors.ip` in the database is too short to hold an IPv6, resulting in truncation while insterting values. The length of the fields holding IPs should be extended to at least 39 characters to fit the maximum size of an IPv6. This include : `visitors.ip`, `contributions.ip` and `last_contributions.ip`

jiru:
Note: HTTP_X_FORWARDED_FOR may contain a comma-separated list of IPs. In this case, because we use it as-is to fill the field *ip*, we end up having the same duplicate error.

halfdan:
What's the status of this? As it stands today tatoeba.org is not reachable via IPv6

```
✗ host -t AAAA tatoeba.org
tatoeba.org has no AAAA record
```

jiru:
Hello halfdan, nice to meet you and welcome to the project. Note that I opened this ticket to make Tatoeba work with IPv6, while enabling IPv6 on the machine that currently hosts tatoeba.org is another problem. Since you’re mentioning it, it seems that [our host](firstheberg.com) is [providing IPv6 on their VPSes](https://twitter.com/FirstHeberg/status/673893054819016705) while I don’t know about dedicated servers like us. I asked them and I’m waiting for their reply.

Another option is to get an IPv6 connectivity using a tunnel.

jiru:
We got IPv6 connectivity from our host and the production server now has IPv6 connectivity. Once this issue is fixed (among with other similar ones), we can enable IPv6 on tatoeba.org.

halfdan:
Just revisiting this: IP addresses should not be stored in a VARCHAR field anyway. They should be converted with `inet_pton` and stored in a VARBINARY field (see also http://stackoverflow.com/questions/6964143/storing-ipv6-addresses-in-mysql). 

[...]

PR #2401, opened 2020-06-18
Issue #205: Stop storing IP of contributors in the logs

jpear1:
Someone will have to run `bin/cake migrations migrate` on the production server for this to take effect.

### DIFF
--- a/src/Model/Table/ContributionsTable.php
+++ b/src/Model/Table/ContributionsTable.php
@@ -84,7 +84,6 @@ class ContributionsTable extends Table
                 'sentence_id' => $sentenceId,
                 'user_id' => CurrentUser::get('id'),
                 'datetime' => date("Y-m-d H:i:s"),
-                'ip' => CurrentUser::getIp(),
                 'action' => $created ? 'insert' : 'update',
                 'type' => 'license',
                 'text' => $data['license'],
@@ -274,7 +273,6 @@ class ContributionsTable extends Table
             'text' => $text,
             'user_id' => CurrentUser::get('id'),
             'datetime' => date("Y-m-d H:i:s"),
-            'ip' => CurrentUser::getIp(),
             'type' => 'sentence',
             'action' => $action
         ]);
@@ -300,7 +298,6 @@ class ContributionsTable extends Table
             'translation_id' => $translationId,
             'user_id' => CurrentUser::get('id'),
             'datetime' => date("Y-m-d H:i:s"),
-            'ip' => CurrentUser::getIp(),
             'type' => 'link',
             'action' => $action
         ]);
@@ -317,18 +314,6 @@ class ContributionsTable extends Table
 
         return $query;
     }
-
-    public function getLastContributionOf($userId)
-    {
-        return $this->find()
-            ->select(['ip', 'count' => 'count(*)'])
-            ->where(['user_id' => $userId])
-            ->group(['group' => 'ip'])
-            ->orderDesc('count')
-            ->limit(10)
-            ->all();
-    }
-
 
     public function getOriginalCreatorOf($sentenceId)
     {
--- a/config/Migrations/20200618002139_RemoveIPsFromContributions.php
+++ b/config/Migrations/20200618002139_RemoveIPsFromContributions.php
@@ -1,0 +1,23 @@
+<?php
+use Migrations\AbstractMigration;
+
+class RemoveIPsFromContributions extends AbstractMigration
+{
+    /**
+     * Change Method.
+     *
+     * More information on this method is available here:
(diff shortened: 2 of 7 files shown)


jiru:
@jpear1 Good job with your first pull request :+1:

There are a few things that can be improved though.

1. You also need to update the `last_contributions` table. This table is holding a copy of the last entries of `contributions`. It has exactly the same structure, so just repeating the same schema update command should be enough. `last_contributions` is created and updated with the help of SQL code in `docs/database/triggers/last_contributions_triggers.sql`. This code needs to be updated accordingly.

2. You also need to update the test fixtures in `tests/Fixture/ContributionsFixture.php` (and `tests/Fixture/LastContributionsFixture.php`). I don’t know if you are familiar with [unit testing](https://en.wikipedia.org/wiki/Unit_testing), but we have a test suite located in the `tests` folder. The test suite is executed by PHPunit, which creates a dummy database (called `tatoeba_test`), fills it with dummy values, and use these values as a base to check if the code in `src` is working. These dummy values are what we call test fixtures, and we need to keep them synchronized with the actual database. So for example, the file `tests/TestCase/Model/Table/SentencesTableTest.php` is testing the code located in `src/Model/Table/SentencesTable.php`. You can run all or some specific tests like this:
    ```sh
    vagrant ssh
    cd Tatoeba/
    # run all the test suite
    phpunit
    # run a specific test file
    phpunit tests/TestCase/Model/Table/SentencesTableTest.php
    # run a specific test in a specific file
    phpunit tests/TestCase/Model/Table/SentencesTableTest.php --filter testLogsSentenceDeletionOnDelete
    ```
    We use these tests to make sure any change does not break existing functionality. The little green check mark next to your commit on Github shows that your modifications did not break anything so far.

3. You should remove all the code related to the storage of IPs, even though just removing the field is enough to get things working. For example, the `saveSentenceContribution()` method in `src/Model/Table/ContributionsTable.php` is still trying to store the IP address:
    https://github.com/Tatoeba/tatoeba2/blob/dc86c5e96a3047721d5b5ee3ccab25edec5357fd/src/Model/Table/ContributionsTable.php#L267-L283
    Similarly, `CurrentUser::getIp()` can be safely removed because it’s not used anywhere else.

jpear1:
I found a usage of contributions' IP addresses being used and I'm not sure how to safely remove it:

https://github.com/Tatoeba/tatoeba2/blob/119c63774b3e9ee62eec77c02f8d7b442d8d35f6/src/Controller/Component/MailerComponent.php#L47-L68

It seems like `Email` requires IP addresses... but it's a CakePHP class and IP addresses aren't mentioned in the docs. Does `viewVars` actually create new members of the `Email` class that are used somewhere else in the program?

AndiPersti:
> It seems like `Email` requires IP addresses... but it's a CakePHP class and IP addresses aren't mentioned in the docs. Does `viewVars` actually create new members of the `Email` class that are used somewhere else in the program? Maybe when generating the text of the email?

The `Email` class allows you to use a template for the email body (see [line 52](https://github.com/Tatoeba/tatoeba2/blob/119c63774b3e9ee62eec77c02f8d7b442d8d35f6/src/Controller/Component/MailerComponent.php#L52)). And in order to pass variables to that template you can use the `viewVars` method. So all the members in the array are variable names that are available in the template `src/Template/Email/html/blocked_or_suspended_user.ctp`.

I don't think having the IPs in the email is necessary because the recipients are community admins that don't have access to the server in order to e.g. block these IPs. And a sysadmin can get the IPs through other means.

Thus, I would simply remove the IPs from the email message.

jiru:
@jpear1 Good catch! Yes, you can just remove the part that lists IPs from the email template.

jiru:
@jpear1 I see that you removed the IPs from the list of variables passed to the `blocked_or_suspended_user` template. That all good, but you also need to modify that template to remove the part that mentions the IP addresses. Otherwise, that will produce "undefined variable" warnings.

src/Model/Table/ContributionsTable.php
```diff
@@ -321,9 +318,8 @@ public function excludeBots($query)
     public function getLastContributionOf($userId)
     {
         return $this->find()
-            ->select(['ip', 'count' => 'count(*)'])
+            ->select(['count' => 'count(*)'])
             ->where(['user_id' => $userId])
-            ->group(['group' => 'ip'])
             ->orderDesc('count')
             ->limit(10)
             ->all();
```
jiru, line 325: This method is not used anywhere else, so you could as well remove it altogether.

jpear1: Alright, done.

jpear1:
Added suggested changes.

jiru:
Thank you, it’s all good! :+1:

PR #2402, opened 2020-06-20
issue 2325, downloads page

ftumsh:
Added the sentences_based_on_id file to the downloads page. On my machine clicking the file gives an error. However, all the other files do too and I have zero php foo so Im unable to comment further.


AndiPersti:
I don't see any changes in your commits because you have used your `dev` branch for opening this PR. So I'm closing it because there is already #2397 which is based on your `issue-2325` branch and contains the new commit which modifies the Downloads page.

Issue #1832, opened 2019-03-17
"Cache" the search we do in "advanced search"


RyckRichards:
**The problem**

Everytime we access https://tatoeba.org/eng/sentences/advanced_search, a "white" form is shown to us to fill in. Sometimes it's annoying because you may 

→ do the same research again
→ do a "likely" research "most of the parametres are the same but some not - like "exclude/include" sentences that ... / show translations in ...
→ share the search for a friend (not the final result)

**Possible solution**

Add a "do you want to cache this search to use later?" checkbox in https://tatoeba.org/eng/sentences/advanced_search .

trang:
What is the difference if you simply bookmark the URL of your search?

RyckRichards:
I can bookmark the result of my search but not the search itself.

Let's say I want sentences in Portuguese, show translations in English, include sentences that have indirect Portuguese links /random. Ok, I can bookmark the result. However, if I want Spanish instead of English, I'd have to do the advanced search again. If I want to get fewest word first, another advanced search. and so on

And worse: Besides having to do a lot of advanced searches, I'd have a billion bookmarks ....

trang:
I see, so you want to saved a template of criteria.

jiru:
Duplicate/related: #1652

trang:
@RyckRichards We now have a feature that addresses (at least partially) your issue.

As an example, I can share a search template I just created:
https://tatoeba.org/eng/sentences/advanced_search?query=this+is+an+example&from=eng&to=por&user=&orphans=no&unapproved=no&has_audio=&tags=&list=&native=yes&trans_filter=limit&trans_to=und&trans_link=&trans_user=&trans_orphan=&trans_unapproved=&trans_has_audio=&sort=random&sort_reverse=

Please try to create a search template on your side and let us know if this feature fits your needs.

Issue #1833, opened 2019-03-21
Caching works a little differently now than it used to.


ckjpn:
Caching works a little differently now than it used to.

Perhaps this is one reason the site isn't getting as many contributions as we used to, since this likely not only slows down the perceived speed, but would likely slow down the whole site since people would be downloading more from the site.

I noticed it this way.

Before when I'd remove a sentence from a sentences page, using a JavaScript bookmarklet, and then returned to the page, it was still listed as being on the list on the side of the page, even though it wasn't.  I would have to reload the page if I wanted to verify that it wasn't on the list anymore.

Now, using the same method, when I go back to the page, the page completely reloads, clearly without the list name being shown.

Note that we have only had one day with over 2,000 new sentences since January 18, 2019.

http://tatoeba.ueuo.com/timeline/

Unless interest has dropped off drastically, it's likely that members are finding the website to be much slower.

jiru:
I, too, noticed that pages are reloaded when clicking on the "back" button of my browser. It’s probably because CakePHP 3 has different defaults regarding caching headers. From a user’s point of view, I welcome this change because I am not showed invalid pages anymore, after logging in for example.

I don’t think this change makes the website any slower. Note that assets, which account for a great deal of the page loading time, are still being properly cached. So I don’t think it’s the reason.

Even if there is evidence that the website is slower that before, it doesn’t mean that’s the reason why we’re getting less contributions, because [correlation does not imply causation](https://en.wikipedia.org/wiki/Correlation_does_not_imply_causation).

ckjpn:
It's OK with me if you close this.  I just thought you might want to be aware of the change and that it might possibly be part of why the website has seemed slow recently.

Issue #1498, opened 2017-08-29
Search terms reappear after a page refresh when cleared


Ppjet6:
### Steps to reproduce

- Search something and submit
- Clear the search with the "X" icon
- Refresh the page (or go somewhere else on the website)

### What happens

The searched words reappear in the input field.

### What do you expect?

The input field shouldn't be repopulated.

Probably what contains the data is never cleared.

jiru:
It’s not a bug, it’s a [feature](https://github.com/Tatoeba/tatoeba2/blob/1e5435a3a5f91376fd4d4a19189adb4d80a2382c/app/Controller/SentencesController.php#L621).

Ppjet6:
Sarcasm? :-°

When I click on a "clear" button I would expect a field to be cleared. If we don't want to clear it maybe we can remove that button instead?

trang:
I think perhaps @jiru missed out on the part where you click on "X".

If you search something and just refresh the page, the search string should stay there indeed.
If you reset the input by clicking on the "X", I agree the search should remain empty.

From my tests though, the search does remain empty after refreshing. It reappears only when navigating to another page.

trang:
I remember the topic as well. I can't find the Wall thread either but let's go ahead and remove this feature.

jiru:
Whoops, now the search query is cleared on the results page too.

PR #2403, opened 2020-06-20
Issue #1832: "Cache" the search we do in "advanced search"

jpear1:
I'm opening this PR to start a discussion about how to implement a solution to #1832.

Here's my thoughts so far:

Search caching should be the default, a user shouldn't have to do anything to enable it. So if a user does a search then presses "back", or just refreshes the page, their search should be just as they left it. Basically just copy the behavior of Google's advanced search.

If people sometimes use the existing behavior, we could add a "reset all" button on the form.

This would solve two of the use cases from #1832: 
* *do the same research again*
* *do a "likely" research "most of the parametres are the same but some not - like "exclude/include" sentences that ... / show translations in ...*

The third use case, "*share the search for a friend (not the final result)*" could be implemented separately, perhaps via `$_GET`. I believe this would also allow bookmarking a search.

I did some googling and it looks like there's a couple of options for caching the search: cookies and sessions.

By my understanding, cookies are easier to use, but sessions prevent hackers from being able to access data they shouldn't be able to.

Any and all concerns/advice welcome. Especially on cookies vs sessions and how they would be implemented CakePHP/Angular.


jiru:
Sorry for the late reply! :pray: 

> Search caching should be the default, a user shouldn't have to do anything to enable it. So if a user does a search then presses "back", or just refreshes the page, their search should be just as they left it. Basically just copy the behavior of Google's advanced search.

What you describe as the default behaviour of keeping form values isn’t so easy to do because the caching behaviour differs among browsers. Some might keep the values while others won’t. That said, right now we are explicitly telling browsers not to cache the page (using HTTP headers Cache-Control and Pragma), so users have no chance to see their search criteria back when pressing the back button:

```
$ curl -v https://tatoeba.org/jpn/sentences/advanced_search 2>&1 >/dev/null | grep -i cache
< cache-control: no-store, no-cache, must-revalidate
< pragma: no-cache
```

For the record, this wasn’t the default behaviour until we switched from Cakephp 2 to CakePHP 3 last year, and [somebody noticed it](https://github.com/Tatoeba/tatoeba2/issues/1833). So I think a good first step would be to disable caching on the advanced search page so that users have a chance to get back the data they entered when pressing the "back" button.

From there, since the restoring the values might still not work on some browsers, so it’s best to store the data in session/cookies as well to make it bulletproof.

> I did some googling and it looks like there's a couple of options for caching the search: cookies and sessions.
> 
> By my understanding, cookies are easier to use, but sessions prevent hackers from being able to access data they shouldn't be able to.

It depends on the context, because after all, what we call a session is actually a cookie containing an identifier granting you access to the data remotely. So in both cases, you need to access a cookie first in order to get access to the data. In your case, giving how large the amount of data you want to store is (all the search fields), it makes sense to use [sessions](https://book.cakephp.org/3/en/development/sessions.html#accessing-the-session-object).

> The third use case, "*share the search for a friend (not the final result)*" could be implemented separately, perhaps via `$_GET`. I believe this would also allow bookmarking a search.

Yes, that’s a good idea. Something like https://tatoeba.org/jpn/sentences/advanced_search?query=foobar. See the documentation about [Modelless forms](https://book.cakephp.org/3/en/core-libraries/form.html). This should be easy to implement as we already have a form for the search in `src/Form/SentencesSearchForm.php`. You could add a "permalink to this search criteria" link/button on the form. I think the best way to do that is to use the [`formaction`](https://developer.mozilla.org/en/docs/Web/HTML/Element/input/submit#formaction) HTML5 attribute on a new button.

So I identified 3 things you can do so far to improve the situation:
1. Enable caching of the advanced  search page
2. Add a "permalink to this search criteria" link/button
3. Store field values in session

Of course any other suggestion is welcome. Feel free to open a new PR for each of these. I think 1. has the best value/effort ratio so I suggest starting with this if you are familiar with HTTP headers and browser caching. 2. should be easy to implement as well if you are familiar with front-end code. 3. is certainly the hardest.

Here are some pointers to where is what in the code:
* Searches are performed by the `search()` method in `src/Controller/SentencesController.php`.
* The advanced search page is first handled by the `advanced_search()` method in `src/Controller/SentencesController.php`, and then the form is displayed by `src/Template/Sentences/advanced_search.ctp`
* The actual search form is displayed by `src/Template/Element/advanced_search_form.ctp`. This code is used both in the Advanced search page and on the right side of the search results.

trang:
> Search caching should be the default, a user shouldn't have to do anything to enable it. So if a user does a search then presses "back", or just refreshes the page, their search should be just as they left it. Basically just copy the behavior of Google's advanced search.

For this specific use case, if the user wanted to perform another search with just one different criteria, technically they wouldn't really need to press "back" in order to readjust their search. We display the advanced search form along with the search results so they can make adjustments from there.

I'm not sure if we have enough evidence to assume that when a user who goes back to the advanced search form, they will generally prefer to see their latest search rather than an empty form.

It was mentioned in #1498 that some users have been annoyed/confused about the fact that the inputs in the green search bar were kept in session. Perhaps we should put aside the idea of storing the search criteria in session, or at least until we can justify that no one wants to see an empty form by default?

Event though the title of the #1832 mentions "cache", I think the need is not to cache but to be able to load *quickly* search criteria. It is possible to bookmark and load predefined search criteria, by bookmarking search results, but loading these predefined criteria currently comes with the cost of performing an unnecessary search. In some cases, waiting for the results may just take as much time as filling up manually an empty form.

I personally would suggest to first only implement this "permalink" that can be bookmarked/shared and see if is sufficient to the users needs.

jpear1:
Alright, makes sense. Adding permalinking does seem like the least objectionable change. I'll start there.

One reason I think it may be useful to store the search parameters in session is that the advanced search page is a lot easier to read than the sidebar you see beside your search results. Not all of the parameters fit on the screen (you have to scroll down), and some of the sentences are kind of a pain to read. Ex. "Limit to sentences having translations that match all the following criteria" takes three lines. When I first started using Tatoeba, I tried to go back and modify my search on the original advanced search page for just this reason.

trang:
I'm closing this PR since there's another one open.

PR #2404, opened 2020-06-21
Filter duplicate charset_table declarations

Yorwba:
Manticore doesn't define what happens if charset_table contains multiple declarations for the same character, and it appears that the result is not the one we want (see [comment on issue #1880](https://github.com/Tatoeba/tatoeba2/issues/1880#issuecomment-646845312)).

This PR adapts the array_filter conditions to remove all duplicate declarations.


jiru:
@Yorwba Thanks. I reindexed the following languages: lat tur ota

PR #2405, opened 2020-06-21
Add Serbian as UI

RyckRichards:
This PR adds Serbian (ISO 639-3 [српски]) as User Interface.
Lazar has been doing a great job and keeps translating every day.

### DIFF
--- a/config/app_local.php.template
+++ b/config/app_local.php.template
@@ -203,6 +203,7 @@ return [
             ['ron', null, 'Română'],
             ['rus', null, 'Русский'],
             ['spa', null, 'Español'],
+            ['srp', null, 'српски'],
             ['swe', null, 'Svenska'],
             ['tgl', null, 'Tagalog'],
             ['tur', null, 'Türkçe'],


jiru:
As much as I’d like to make Serbian available, I don’t think it’s ready yet:
image

That said, we can still merge your PR without enabling Serbian. This way, we’ll have the configuration ready for when the Serbian translation will be complete enough.

trang:
It's definitely not ready for prod but I think the main goal here is to enable it on the dev website, so that the translators can get to see the result of their translations.

RyckRichards:
That too. Also, I think people will notice that there's a "Serbian UI" already but it's not fully translated and so, they might get motivated to start helping us.

jiru:
Thank you, @RyckRichards :smile:

jiru:
Serbian is now available on tatoeba.org as a UI language.

RyckRichards:
Thank you!!!!! More languages are coming :)

PR #2406, opened 2020-06-21
Fixed naming for Show Transcriptions on Settings (issue 2353)

### LINKED ISSUE #2353: The option "Always show transcriptions..." is confusing
On both the product site:

transcriptions_prod

and the dev site:

transcriptions

I see transcriptions even though I've deselected "Always show transcriptions and alternative scripts" in both cases:

always_show_transcriptions

I'm a little confused about the interaction between deselection and the use of the word "always". If the wording were simply "Show transcriptions and alternative scripts", I would expect them to be displayed when it's checked, and hidden when it's unchecked. But "always" makes me wonder whether some condition I don't know about affects whether they're shown when the option is deselected. In other words, does "not always" == "never" for this option?

RyckRichards:
This PR resolves #2353 by replacing "Always show transcriptions and alternative scripts" to "Always show unreviewed transcriptions and alternative scripts".

### DIFF
--- a/src/Template/User/settings.ctp
+++ b/src/Template/User/settings.ctp
@@ -159,7 +159,7 @@ $this->set('title_for_layout', $this->Pages->formatTitle(__('Settings')));
                     ng-init="showTranscriptions = <?= $showTranscriptions ?>"
                     class="md-primary">
                 </md-checkbox>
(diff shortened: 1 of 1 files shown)


jiru:
Well, your pull request is okay so I’ll just merge it. :smile:

PR #2408, opened 2020-06-22
Fix wrong usage of asort in tests

AndiPersti:
Some tests used the return value of `asort` in the assertions. But `asort` sorts the array in place and just returns a boolean that indicates succes/failure. Thus these tests always succeeded although they should have failed with the current code version.

While here I've also refactored the test for `Sentences::findFilteredTranslations`.


jiru:
Thank you Andreas :blush:

PR #2411, opened 2020-06-25
made the same as the links code, expanded description and changed Id to id

ftumsh:


AndiPersti:
This was just a test PR.

PR #2412, opened 2020-06-25
Add language to the database without specifying a list

AndiPersti:
[Point 3](https://github.com/Tatoeba/tatoeba2/wiki/Adding-a-new-language#3-run-the-mysql-add_new_language-procedure) on the wiki page for adding a new language says that one should use the procedure `add_new_language()`. A comment in the script file mentions a way to [add a language without specifying a list](https://github.com/Tatoeba/tatoeba2/blob/109f993bd4d0286be94ac1aa68941eea2ccc9133/docs/database/procedures/add_new_language.sql#L9-L10).

But this currently doesn't work:
```sql
MariaDB [tatoeba]> call add_new_language('abcd', 0);
+-------------------------------------------------------------+
| CONCAT('There are no sentences in list ', list_id_for_lang) |
+-------------------------------------------------------------+
| There are no sentences in list 0                            |
+-------------------------------------------------------------+
1 row in set (0.035 sec)

Query OK, 2 rows affected (0.037 sec)

MariaDB [tatoeba]> select * from languages where code = 'abcd';
Empty set (0.023 sec)
```

Since that is a reasonable use case (especially in a development environment) this PR fixes the script.

In addition I've also implemented the following changes:
- Added a better label for the warning message when the list is empty.
- Avoid running the list-specific code when no valid list id is specified.
- Avoid checking for existence of the language since we call the insert statement anyways in order to pass the exception to the caller.

### DIFF
--- a/docs/database/procedures/add_new_language.sql
+++ b/docs/database/procedures/add_new_language.sql
@@ -24,21 +24,23 @@ END IF;
 -- to occur so the caller will catch it.
 INSERT INTO languages (code) VALUES (lang_iso_code);
 
-SELECT COUNT(*) INTO @sentences_in_list FROM sentences, sentences_sentences_lists
-    WHERE sentences_list_id = list_id_for_lang AND sentences.id = sentence_id;
+IF (list_id_for_lang > 0) THEN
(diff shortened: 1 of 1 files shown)


jiru:
Thanks :slightly_smiling_face:

PR #2417, opened 2020-06-27
Fix vocabulary count when adding new items

### LINKED ISSUE #2350: Requested vocab listed as matching "1000+" sentences instead of 2
When I added the Russian vocabulary word велеть, it was listed as matching 1000+ sentences:
vocab_request_0

But in fact, there are only 2:

vocab_request_1

AndiPersti:
This PR should fix #2350.

There were actually two problems:
1. After the introduction of the `Search` model, the query should be included in the search engine options. Thus we searched for all sentences in a language because the query was empty. That's why there were `1000+` results.

2. The caching in `SphinxBehavior` assumed there was only one search query per request. But on the vocabulary page we use several queries in a single request. Thus the result from the first query was also used for all other queries.

In addition I moved the static function from `App\Utility\Search` to `App\Model\Search` and removed some obsolete commented code.


src/Model/Table/VocabularyTable.php
```diff
@@ -125,15 +125,13 @@ private function _getNumberOfSentences($lang, $text)
...
-        $index = array($lang . '_main_index', $lang . '_delta_index');
-        $sphinx = array(
-            'index' => $index,
-            'matchMode' => SPH_MATCH_EXTENDED2
-        );
-        $query = Search::exactSearchQuery($text);
+        $search = new Search();
+        $search->filterByLanguage($lang);
+        $search->filterByQuery(Search::exactSearchQuery($text));
+        $search->sort('random');
```
jiru, line 131: Why randomize the results?

src/Model/Search.php
```diff
@@ -290,4 +290,11 @@ public function getTranslationFilters() {
             function ($v) { return !is_null($v); }
         );
     }
+
+    public static function exactSearchQuery($text) {
```
jiru, line 294: Why make it a static method?

src/Model/Table/VocabularyTable.php
AndiPersti: This is a workaround to set the `SPH_MATCH_EXTENDED2` match mode. Without it the default match mode would be `SPH_MATCH_ALL`.
So if there are two sentences `He goes.` and `She goes and he runs.`, on the vocabulary page there will be 2 sentences found for the vocabulary item `he goes`. But after clicking on the link, there will only be one sentence found because the search page uses the extended mode.
(Since we are only interested in the count the order isn't relevant.)

Should I add a new method to the `Search` model which allows to set the match mode independent of the sorting? Or should I use another sorting order?

src/Model/Search.php
AndiPersti: This function is also used in two other places that do not build a query for the search engine:
- https://github.com/Tatoeba/tatoeba2/blob/8d702ba27136e45a949a15fd1478ee8e53d53799/src/Model/Table/VocabularyTable.php#L106
- https://github.com/Tatoeba/tatoeba2/blob/8d702ba27136e45a949a15fd1478ee8e53d53799/src/View/Helper/VocabularyHelper.php#L54

To me it is comparable to the static functions in `CurrentUser` but if you prefer having it as a method I will change it.

src/Model/Table/VocabularyTable.php
jiru: I see! For now, the Search model implementation kinda hides the match mode thing, so I think we can set it to `SPH_MATCH_EXTENDED2` regardless of `Search->sort` being set or not. From the caller’s perspective, it doesn’t make sense to filter exact matches with `="..."` and actually not get exact matches in the results.

src/Model/Search.php
jiru: Then it’s okay to keep it as static. I was just being curious. :wink:

PR #2420, opened 2020-07-02
Report Manticore errors

jiru:
Currently, when Manticore returns a search error, there is no way to see the actual error message. The search error page just shows a generic error, and the logs do not contain the error. Setting debug to true doesn’t help.

This pull request adds a new component to log Manticore search errors into `logs/error.log`, along with context information.

### DIFF
--- a/src/Controller/Component/ErrorComponent.php
+++ b/src/Controller/Component/ErrorComponent.php
@@ -29,4 +29,18 @@ class ErrorComponent extends Component
             return $message;
         }
     }
+
+    protected function generateNewCode() {
+        return uniqid();
+    }
+
+    public function traceError($message, $code = null) {
+        if (is_null($code)) {
+            $code = $this->generateNewCode();
+        }
+        $message = $this->format("[$code] $message");
+        $this->_registry->getController()->log($message);
+
+        return $code;
+    }
 }
--- a/src/Controller/Component/ErrorComponent.php
+++ b/src/Controller/Component/ErrorComponent.php
@@ -1,0 +1,33 @@
+<?php
+namespace App\Controller\Component;
+
+use Cake\Controller\Component;
+use Psr\Log\LogLevel;
+
+class ErrorComponent extends Component
+{
+    private function infoLine($infoKey, $infoValue) {
+        $info = '';
+        if ($infoValue) {
+            $info = "$infoKey: $infoValue\n";
+        }
+        return $info;
+    }
+
+    public function log($message, $level = LogLevel::ERROR, $context = []) {
+        if (strlen($message)) {
+            $message .= "\nRequest URL: " . $this->request->getRequestTarget() . "\n";
+
+            $referer = $this->request->getEnv('HTTP_REFERER');
+            $message .= $this->infoLine('Referer URL', $referer);
+
+            $clientIp = $this->request->clientIp();
+            $message .= $this->infoLine('Client IP', $clientIp);
+
+            $agent = $this->request->getEnv('HTTP_USER_AGENT');
+            $message .= $this->infoLine('User Agent', $agent);
+
+            return parent::log($message);
+        }
+    }
+}
--- a/src/Controller/Component/ErrorComponent.php
+++ b/src/Controller/Component/ErrorComponent.php
@@ -2,7 +2,6 @@
 namespace App\Controller\Component;
 
 use Cake\Controller\Component;
-use Psr\Log\LogLevel;
 
 class ErrorComponent extends Component
 {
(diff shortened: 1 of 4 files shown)


AndiPersti:
:+1: for adding more logging.

> Currently, when Manticore returns a search error, there is no way to see the actual error message. The search error page just shows a generic error, and the logs do not contain the error. Setting debug to true doesn’t help.

Well, the Manticore API never throws an exception or triggers an error but provides `GetLastError` and `GetLastWarning` instead. So that's why there's no log message from PHP and setting debug doesn't help. And it's our fault if "there is no way to see the actual error message". :smiley: 
We only trigger a warning when we get a result but there was still a warning reported by Manticore.[*]

> This pull request adds a new component to log Manticore search errors into `logs/error.log`, along with context information.

If the goal is to have additional information in the log message why not let the component just create the message and continue using CakePHP's logging mechanism? Something like:
```php
$this->loadComponent('LogMessage');
$message = $this->LogMessage->create('...');
$this->log($message);
// or
$this->log($message, 'debug');
// or
Cake\Log\Log::warning($message);
```
At least testing the component would be much simpler because you don't have to create a temporary log file.

[*] https://github.com/Tatoeba/tatoeba2/blob/e107584d05bdadfb8fb503cd28efda8a709a0b5f/src/Model/Behavior/SphinxBehavior.php#L143-L144
I suggest to change the `trigger_error` call to a log call. The result will be the same (a new entry in `debug.log`) but in the dev environment the app won't crash (try searching without specifying a language to see what I mean).

jiru:
@AndiPersti I’ve been thinking about all this. All things considered, `trigger_error()` might be better.

The downside of using `Cake\Log\Log` compared to `trigger_error()` is that we loose contextual information (Request URL, client IP etc.) in the logs. As you can see in `vendor/cakephp/cakephp/src/Error/BaseErrorHandler.php`, the error handler gets the request context from within a model using `Router::getRequest()`. This makes it able to retrieve contextual information from pretty much anywhere. Such information can be really useful to debug problems, especially when users are reporting bugs without being able to provide much context.

> try searching without specifying a language to see what I mean

If I’m looking at the same thing as you, the app crashes because the warning message includes contextual debug information, which includes a copy of the Response object, which includes the page body, which includes AngularJS expressions such as `{{native ? "yes" : ""}}`, which breaks AngularJS. I’m sure there are ways around this without suppressing the warning message altogether. I don’t know how you go about it, but I tend to rely on such output to spot notices and warnings during development. On the other hand, some of these messages, such as messages sent from page redirections or ajax responses, might go unnoticed. Is it a bad habit?

AndiPersti:
> The downside of using `Cake\Log\Log` compared to `trigger_error()` is that we loose contextual information (Request URL, client IP etc.) in the logs.

While in general I agree that losing contextual information is bad, for this specific problem (i.e. warnings/errors from Manticore) I don't think the additional info we get when using `trigger_error()` is of much value. (At least not for me personally: I know exactly where the problem happens so I don't need a stack trace. And the only useful information from the request object would be the request URL (specifically the query string). But logging the contents of e.g. `$options[sphinx]` would be as helpful.)

> If I’m looking at the same thing as you, the app crashes because the warning message includes contextual debug information, which includes a copy of the Response object, which includes the page body, which includes AngularJS expressions such as `{{native ? "yes" : ""}}`, which breaks AngularJS. I’m sure there are ways around this without suppressing the warning message altogether.

Thanks for pushing me into the right direction. (For some reason I've never really looked for the reason why the app crashes. Probably because I've noticed that problem already very early when I started contributing and I just got used to not using the search feature without selecting a language :facepalm: )
I've just found out that setting the output format for the [`Debugger`](https://api.cakephp.org/3.7/class-Cake.Error.Debugger.html#_setOutputFormat) to `html` instead of the default `js` won't result in a crash.[*] So for me that particular problem is solved.  :bouquet: 

PS: If we ever update Manticore to 3.4.0 we [get rid of the `query word(s) mismatch` warning](https://github.com/manticoresoftware/manticoresearch/commit/95628c9b9b8f16fe1bd18599a84e9673575adf1a) which is by far the most logged warning we get from Manticore.

[*] Alternatively one could also add `ng-non-bindable` to the outer `pre` element of the `error` template of the `js` format but this would be more "hackish" because one would need to literally [copy the template string](https://github.com/cakephp/cakephp/blob/1e89aa0399ce4c0f9af764e3b61aa24826278feb/src/Error/Debugger.php#L120-L128).

jiru:
> While in general I agree that losing contextual information is bad, for this specific problem (i.e. warnings/errors from Manticore) I don't think the additional info we get when using `trigger_error()` is of much value. (At least not for me personally: I know exactly where the problem happens so I don't need a stack trace. And the only useful information from the request object would be the request URL (specifically the query string). But logging the contents of e.g. `$options[sphinx]` would be as helpful.)

I still think using `trigger_error()` is probably better in general to issue various warnings around.

As for search errors, I implemented a new system to trace back errors using a code displayed to users. Let me know what you think! :slightly_smiling_face:

AndiPersti:
> As for search errors, I implemented a new system to trace back errors using a code displayed to users. Let me know what you think! slightly_smiling_face

Looks good to me.

Issue #1891, opened 2019-05-14
Make search bar appear faster


jiru:
Related thread: https://tatoeba.org/wall/show_message/31798#message_31798

I used [this simple method](https://stackoverflow.com/questions/36979907) to measure the loading time of the Wall page. I found out that removing jQuery quite improved the loading time (from about 500 ms to about 300 ms). I think we can remove jQuery on pages it’s not needed or otherwise defer its loading in the background.

Issue #74, opened 2014-02-17
make tatoeba interface support right to left languages


loolmeh:
Basically this will need the following, which is simple but tedious. For a contributor's sentences, all tags where there's text should have a "dir" [directionality] attribute which will be set to "rtl" or "ltr" when changing tatoeba's interface languages. To do this: add a little mechanism (in the same place as the mechanism to change language interface) to set a constant to "rtl" or "tlr" and then have a $dir everywhere there's interface text in the html. so, easy ticket

jiru:
For the record, `dir=""` is currently set everywhere except in `<html>`. I made it so because the current interface is not designed for RTL languages.

trang:
Re-opening this issue since @SafaAlfulaij is working on redesigning the UI for RTL.

SafaAlfulaij:
Some screenshots of progress:
![screenshot_ _](https://cloud.githubusercontent.com/assets/3509617/17426588/15c995b6-5ae2-11e6-967e-5cfd203c5cec.png)
![screenshot_ _](https://cloud.githubusercontent.com/assets/3509617/17426590/15ce9e76-5ae2-11e6-8ac7-f66d88ae8376.png)
![screenshot_ _](https://cloud.githubusercontent.com/assets/3509617/17426589/15cd2384-5ae2-11e6-9303-66cea947a095.png)

SafaAlfulaij:
I pushed (for reference) some commits that implements the RTL layout in the [issue-74](https://github.com/SafaAlfulaij/tatoeba2/tree/issue-74) branch.
The UI now is under redesigning and I'm not sure that the fixes will still long.

trang:
@SafaAlfulaij, I finally got the time to look at your commits, the results look pretty good =)

I think a lot of the CSS you've added will be obsolete once the corresponding pages will be redesigned with Angular Material, because (to my surprise) the mirrorring is handled by the framework. 

The Material Design spec actually has guidelines about RTL user interface:
https://material.google.com/usability/bidirectionality.html

But anyway, you can open a pull request for this. There's probably a few more things to tweak but I hope your work can be merged and released some time in September.

PR #2429, opened 2020-07-04
Responsive landing page

trang:
This pull request makes the landing page responsive. 

I took this occasion to introduce support for the right-to-left user interface (#74) as well as starting to migrate away from jQuery (#1891).

**Top navigation menu**

- Opening a sub-menu no longer happens on mouseover but on click.
- Opening a sub-menu and selecting an item in the sub-menu is possible with the keyboard.
- For non-authenticated users, the item "Contribute" was removed. 
  - The main reason is to ensure that all the items can be displayed in one row in all the languages.
  - But another reason is that this section is not really useful for non-authenticated users.
- For authenticated users, the user menu was redesigned.
  - The profile picture is displayed next to the username. They are encapsulated in a button which will open the user menu on click.
  - The UI language selection and the log out button were moved into the user menu.
  - The items "My profile", "Language", "Settings" and "Log out" have an icon next to them.
  - The quick links to the user content ("My sentences", "My vocabulary", "My reviews", etc) are displayed at the end, below a separator.
- The menu items "Browse", "Contribute" and "Community" will disappear on medium and small screens, and a menu icon will appear next to the logo instead. Clicking on this icon opens a side nav menu that contains the same items as the horizontal menu.
- The login box was redesigned into a dialog. It takes the full screen on screen size medium and small.
- The UI language selector was also redesigned into a button with a globe icon. This button opens a dialog that displays all the possible languages on 4 columns in large screens, 2 columns in medium screens, and 1 column in small screens.

**Content**

- The tagline text in the green background is now centered rather than aligned on the left.
- The search section was redesigned to adapt better on smaller screens.
  - The language selection was removed because it was taking too much space and is not the most essential.
  - The position of the "Advanced search" link, the "More tips" link, and the search button was readjusted to be aligned.
- In the sentence component, the "Copy sentence" and "Go to sentence page" buttons are no longer displayed on small screens.
  - They will appear or disappear when tapping/clicking on the sentence.
  - They will appear next to the audio button, making the space to display the sentence text shrink. The text will therefore not be nicely displayed, but I didn't have time yet to implement a good design for displaying these buttons differently.
- The blocks "Want to help?" and "Stats" will be displayed below each other on small screens.

**Footer**

- The UI language button was added to the footer for in case people somehow miss it at the top.
  - I've been considering moving this button to the footer instead of duplicating it, but in the end, I think it's best to have it on both places to increase as much as possible the chances for the user to figure out how to change the language in case the UI is not in a language they understand.
- The footer sections will be displayed below each other on small screens.

**Right-to-left support**

This is, for now, a quick implementation and I'm aware of several things that would need to be fixed still, but nothing that makes the RTL interface worse than it is now (I think) so they can be fixed in another iteration.

**Removal of jQuery where possible**

- The announcement message was migrated to Angular and also redesigned a bit along the way.
- jQuery is bundled with the old sentence component javascript so jQuery is now only included when the old sentence component is used.
- Everywhere else, the code was rewritten to use pure Javascript.

--- 

Related Wall thread: https://tatoeba.org/eng/wall/show_message/35464


src/Template/Element/short_description.ctp
```diff
@@ -42,7 +42,7 @@
         </div>
     </md-toolbar>
 
-    <div class="container">
+    <div class="container" id="langing-page-search" ng-cloak>
```
AndiPersti, line 45: The id isn't used anywhere else but nevertheless I think there is a typo and it should be `landing-page-search`.

src/Template/Element/sentences/sentence_icons.ctp
```diff
@@ -0,0 +1,27 @@
...
+</div>
+
+<md-button class="md-icon-button" ng-class="<?= $responsiveNgClass ?>" ngclipboard data-clipboard-text="{{<?= $angularVar ?>.text}}">
+    <md-icon>content_copy</md-icon>
+    <md-tooltip><?= __('Copy sentence') ?></md-tooltip>
+</md-button>
+
+<?= $this->element('sentence_buttons/audio', ['angularVar' => $angularVar]); ?>
+
+<md-button class="md-icon-button" ng-class="<?= $responsiveNgClass ?>" href="<?= $sentenceBaseUrl ?>/{{<?= $angularVar ?>.id}}">
```
AndiPersti, line 24: Since you use interpolation markup (`{{}}`) it would be safer to use `ng-href` here.

AndiPersti:
The star indicating nativeness is now placed apart from the other icons directly next to the sentence. Is this on purpose?
star

trang:
> The star indicating nativeness is now placed apart from the other icons directly next to the sentence. Is this on purpose?

Not on purpose. I've repositioned it as it was before.

I wanted to put the star next to the username but then I realized the Settings label would need to be updated so we'll leave that for another time.

AndiPersti:
I've found some more problems:

1. If I'm logged in and use my 7'' tablet in landscape mode, the height of the menu that opens when I click on my username is too big and doesn't fit on the screen. But the opened menu disables scrolling so I cannot select menu options at the bottom. The screenshot is from Firefox's Responsive Design Mode where the problem is also reproducible:
landscape

2. Click on `BROWSE` to open the menu. Then click on `COMMUNITY`.
I would expect that the `BROWSE` menu closes and the `COMMUNITY` menu opens. But instead only the `BROWSE` menu closes and the `COMMUNITY` button just gets highlighted. I have to click a second time on `COMMUNITY` to open the menu.

3. Logged out use your phone/tablet in portrait mode on the landing page and open the UI language  selection. The dialog fills most of the screen. Now rotate the device into landscape mode. Again the dialog is responsive and fills most of the screen. Rotate the device back into portrait mode. The dialog now only covers about half the screen.
I'm not sure whether this is a problem with our code or the framework. In the browser console I get a warning whenever I open the dialog:
```
$mdDialog: using `<md-dialog ng-cloak >` will affect the dialog opening animations. <div class="md-dialog-container" tabindex="-1"> layout.js:21:1904
```

config/auth_actions.php
```diff
@@ -59,13 +59,15 @@
...
             'for_language',
+            'login_dialog_template',
         ],
         'vocabulary' => [ 'of' ],
         'wall' => [
             'index',
             'show_message',
             'messages_of_user',
         ],
+        'angular_templates' => '*',
```
jiru, line 70: Could you write this as `'angular_templates' => [ '*' ]` instead? This is because newer versions of the cake-auth-actions plugin will require this syntax. See the yet-to-be-merged commit 266c5162f8e402e906736d7da4a5149ea1ea9f6f.

src/Template/AngularTemplates/interface_language.ctp
```diff
@@ -0,0 +1,47 @@
...
+}
+
+usort(
+    $languages,
+    function($a, $b) {
+        return strnatcmp($a['text'], $b['text']);
+    }
+);
+
+$languagesJSON = htmlspecialchars(json_encode($languages), ENT_QUOTES, 'UTF-8');
```
jiru, line 27: I think you can use the CakePHP shorthand `h()` here.
```php
$languagesJSON = h(json_encode($languages));
```

webroot/js/elements/login-dialog.ctrl.js
```diff
@@ -0,0 +1,29 @@
...
+            var vm = this;
+
+            vm.showDialog = showDialog;
+            
+            ///////////////////////////////////////////////////////////////////////////
+
+            function showDialog(url) {
+                $mdDialog.show({
+                    controller: DialogController,
+                    templateUrl: get_tatoeba_root_url() + '/users/login_dialog_template?redirect=' + url,
```
jiru, line 16: The redirection URL is not escaped. This is a problem if the URL contains `&`.

For example, if I open https://dev.tatoeba.org/eng/sentences_lists/index?sort=name&direction=asc as a guest, and login from there, I am redirected to https://dev.tatoeba.org/eng/sentences_lists/index?sort=name (without &direction=asc).

config/auth_actions.php
trang: I'm getting a `403 Forbidden` if I put the wildcard in an array. I guess that change will have to be done at the same time as upgrading to the newer version?

jiru: I see. My bad :pray:

jiru:
Others suggest [using `md-menu-bar` and `md-toolbar`](https://stackoverflow.com/questions/35959169/md-menu-behaviour-in-angular-material). The linked codepen is quite laggy however.

After all, can’t we just keep the existing CSS-based menu when the "hamburger" is not displayed?

jiru:
In the new footer, the text of the links are centered. I think it makes them much harder to read because the eye has to jump to random locations when processing the next line.

image

trang:
> Others suggest using md-menu-bar and md-toolbar. The linked codepen is quite laggy however.

The md-toolbar is also too buggy on mobile for what I had tried.

> After all, can’t we just keep the existing CSS-based menu when the "hamburger" is not displayed?

Yes, I had been considering it too and after some thoughts, I went for that option. I didn't exactly reuse the old code because it didn't work out of the box anymore. I just implemented a new CSS.

The `md-menu` allowed to select the menu items with the keyboard, which was nice I think. That's something we don't have with the CSS menu. But this can wait for another time.

src/Template/AngularTemplates/interface_language.ctp
```diff
@@ -24,7 +24,7 @@ usort(
     }
 );
 
-$languagesJSON = htmlspecialchars(json_encode($languages), ENT_QUOTES, 'UTF-8');
+$languagesJSON = h(json_encode($languages), ENT_QUOTES, 'UTF-8');
```
jiru, line 27: You don’t need to provide the extra parameters `ENT_QUOTES, 'UTF-8'`. That’s the neat thing about `h()` :blush:

jiru:
Neat! Now it really has the look and feel of a menu. :smile:

jiru:
@trang When I open the UI language dialog, I cannot close it by clicking outside it. I think mobile users will expect this behavior.

trang:
> @trang When I open the UI language dialog, I cannot close it by clicking outside it. I think mobile users will expect this behavior.

Indeed. I did it for the login dialog but forgot it for the UI language dialog.

PR #2432, opened 2020-07-05
Add log configuration for Queue plugin

### LINKED ISSUE #2423: Configure logging for the queue runworker
Andreas wrote:

> we run the cake queue runworker in verbose mode so every ten seconds it will output
> 
>> Looking for job ... Nothing to do, sleeping.
>
> I would suggest to run it in silent mode and configure a logger in order to still get the info messages:
> https://github.com/dereuromark/cakephp-queue/tree/master/docs#logging

AndiPersti:
This PR closes #2423.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -346,6 +346,14 @@ return [
             'url' => env('LOG_QUERIES_URL', null),
             'scopes' => ['queriesLog'],
         ],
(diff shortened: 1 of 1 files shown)


jiru:
Thanks :+1:

PR #2434, opened 2020-07-06
Added Bulgarian as a User Interface (UI) language

RyckRichards:
This PR enables Bulgarian as a language interface on Tatoeba.
The translator asked me to enable it on dev Tatoeba to make sure how his translations are rendering and make his job easier.
He's aware that Bulgarian will be added on prod once it gets more translations.

### DIFF
--- a/config/app_local.php.template
+++ b/config/app_local.php.template
@@ -155,6 +155,7 @@ return [
             ['bel', null, 'Беларуская'],
             ['ben', null, 'বাংলা'],
             ['bre', null, 'Brezhoneg'],
+            ['bul', null, 'Български'],
             ['cat', null, 'Català'],
             ['ces', null, 'Čeština'],
             ['cmn', 'Hans', '中文', ['chi']],


jiru:
Thank you :+1:

RyckRichards:
@jiru I've got the same error at the time I've added #2405 

Instead of https://dev.tatoeba.org/bul, I get https://dev.tatoeba.org/bul/bul/ and therefore, it's displayed:

 
image

jiru:
I just fixed it now.

RyckRichards:
I believe we can add it on prod

image

image

RyckRichards:
I believe they would love to have that on prod since they've been working on that. Here we have the updated status

image

trang:
@RyckRichards The default.pot is still only 40% translated (that's only +2% compared to last time). Was there a specific push from them to have it Bulgarian UI on production already?

We know that it's not a good experience for users to have a UI only half translated so we should perhaps wait until the default.pot is ~80% translated. But if having Bulgarian activated on prod motivates the translators to keep translating, we can of course find a compromise.

AndiPersti:
I agree that a new language should have a higher percentage of `default.pot` translated.
But in addition I think there also should be complete translations of the main pages available, e.g. homepage for guests and members, sentence page, search (result page and advanced search form).

PR #2437, opened 2020-07-06
Make Uyghur available as UI

RyckRichards:
This PR enables Uyghur as a language interface.
It has been requested to help translators to check how their translations on Transifex are rendering.


jiru:
Thank you very much as always :+1:

jiru:
Now available on dev: https://dev.tatoeba.org/uig/

PR #2442, opened 2020-07-10
Add Filipino (fil) as UI language

RyckRichards:
This PR adds Filipino as a UI language.
It has been asked from Filipino translators to check how their translations render online.


jiru:
Thank you :+1:

PR #2450, opened 2020-07-14
Add Slovak as UI language

RyckRichards:
This PR adds Slovak (slk) as UI language on Tatoeba.
It has been requested from the translators since they can't check how their translations are rendering.


RyckRichards:
@jiru I believe Slovak can be add on prod, too.

image

jiru:
@RyckRichards Thanks for letting me know. I will make Slovak available on tatoeba.org during the next update.

jiru:
Slovak now available on tatoeba.org :tada:

PR #2451, opened 2020-07-14
Fix vocabulary number

AndiPersti:
The primary reason for this PR (the first commit) is to get rid of a warning in the debug log:
```
2020-07-14 15:39:54 Notice: Notice (8): A non well formed numeric value encountered in [/var/www-prod-d6245ffc91bdc31f5b4b27199099359fc1920c4a/vendor/cakephp/cakephp/src/I18n/PluralRules.php, line 182]
Request URL: /pol/vocabulary/of/Amastan?page=2
```
The problem is that the third parameter of the translation helper function `__n()` should be a number but when the number of sentences containing a vocabulary item is equal to the total limit for a search query (currently 1000) we pass a string:
https://github.com/Tatoeba/tatoeba2/blob/d6245ffc91bdc31f5b4b27199099359fc1920c4a/src/View/Helper/VocabularyHelper.php#L39-L47
This affects mainly languages that have more than one plural form like Polish, Russian or Ukrainian because they use a different code path in `I18n/PluralRules.php`.

The other commits are some refactoring:
- Using a configuration option instead of a constant for the total limit
- Putting the constructing of the label string into its own function
- Moving code for building the request body from the controller to the view

I'm not sure about the test for `SentencesController::show_all_in()` because it depends strongly on the resulting HTML code. But if it is ok I could add similar tests for `SentenceCommentsController::index()` and `ContributionsController::of_user()`

### DIFF
--- a/src/Controller/VocabularyController.php
+++ b/src/Controller/VocabularyController.php
@@ -141,23 +141,6 @@ class VocabularyController extends AppController
 
         $result = $this->Vocabulary->addItem($lang, $text);
 
-        $numSentences = $result['numSentences'];
-
-        if (is_null($numSentences)) {
-            $numSentencesLabel = __('Unknown number of sentences');
-        } else {
-            $numSentences = $numSentences == 1000 ? '1000+' : $numSentences;
-            $numSentencesLabel = format(
-                __n(
-                    '{number} sentence', '{number} sentences',
-                    $numSentences,
-                    true
-                ),
-                array('number' => $numSentences)
-            );
-        }
-        $result['numSentencesLabel'] = $numSentencesLabel;
-
         $this->set('result', $result);
         $this->viewBuilder()->setLayout('json');
     }
--- a/src/Template/Vocabulary/save.ctp
+++ b/src/Template/Vocabulary/save.ctp
@@ -1,3 +1,7 @@
 <?php
+if ($result) {
+    $numSentences = $result['numSentences'];
+    $result['numSentencesLabel'] = $this->Vocabulary->sentenceCountLabel($numSentences);
+}
 echo json_encode($result);
 ?>
--- a/src/View/Helper/VocabularyHelper.php
+++ b/src/View/Helper/VocabularyHelper.php
@@ -29,22 +29,28 @@ class VocabularyHelper extends AppHelper
         'Html', 'Url'
     );
 
+    /**
+     * Create the label for the link to the search page
+     *
+     * @param integer $numSentences Number of sentences containing the vocabulary item
+     *
+     * @return string
+     */
+    public function sentenceCountLabel($numSentences) {
+        if (is_null($numSentences)) {
+            return __('Unknown number of sentences');
+        } else {
+            return format(
+                __n('{number} sentence', '{number} sentences', $numSentences),
+                ['number' => $numSentences == 1000 ? "1000+" : $numSentences]
+            );
+        }
+    }
+
     public function vocabulary($vocab) {
         $lang = $vocab['lang'];
         $text = $this->_View->safeForAngular($vocab['text']);
-        $numSentences = $vocab['numSentences'];
-        if (is_null($numSentences)) {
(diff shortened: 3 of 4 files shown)


jiru:
Thank you. It looks all good but I just wonder if it’s necessary to allow modifying the limit from the configuration file (as opposed to hardcoding it as a constant). Will we ever need to change this value?

AndiPersti:
> Thank you. It looks all good but I just wonder if it’s necessary to allow modifying the limit from the configuration file (as opposed to hardcoding it as a constant). Will we ever need to change this value?

I agree, that we probably don't need to change this value.
My reasons for making it a configuration were mainly:
1. The 'DEFAULT' in the name suggests it is changeable.
2. It enabled testing the pagination for the pages that use `paginateLatest`.
3. I generally prefer configurable settings instead of hardcoded values.

But 1 is easy to change, 3 is highly subjective :smile: and the more I think about 2, the more I doubt the usefulness for these tests.

So, I've reverted the change but moved the constant into the bootstrapping phase. Otherwise I can't think of a good way to use it in the helper. Do you have any suggestions/good solutions?

AndiPersti:
Well, I've just realized that the `1000` in the `VocabularyHelper` isn't controlled by the constant but is the default limit for a Manticore search (set by `SetLimits`). So using the constant in the helper is actually wrong. :facepalm: 

So should I
- just fix the bug (i.e. only keep the first commit)
- remove all the changes involving the constant but keep the refactoring
- make the Manticore search limit more explicit, i.e. change [line 121 in `SphinxBehavior`](https://github.com/Tatoeba/tatoeba2/blob/58ad65f02cfd06c3d3ccb155dc9632fec6317668/src/Model/Behavior/SphinxBehavior.php#L121) to `$sphinx->SetLimits(($page - 1) * $limit, $limit, PAGINATION_TOTAL_LIMIT);`

?

jiru:
Wait, aren’t we talking about two separate limits that happen to be equal to 1000?
1. One is the pagination limit used in `AppController::paginateLatest()` that prevents users from browsing pages of high number because of slow performance. The effect of that limit is that a `where id >= N` clause is added to the pagination query, *N* being calculated using a 1000 limit value.
2. One comes from Manticore only returning the first 1000 results. This is the limit your initial commit is dealing with. Note that this limit is controlled by `SetLimits()` from the API, but also capped to the `max_matches` Manticore configuration option.

AndiPersti:
> Wait, aren’t we talking about two separate limits that happen to be equal to 1000?

Yes, I realized too late that the 1000 for the vocabulary page is different from the pagination limit. Sorry for the confusion.

So I guess I should just leave the 1000 as a magic number for the vocabulary, shouldn't I?

jiru:
Well, can’t we figure out whether the result is "more than *n*" (as opposed to "exactly *n*") by comparing the ["real total" and the "returned total"](https://github.com/Tatoeba/tatoeba2/blob/58ad65f02cfd06c3d3ccb155dc9632fec6317668/src/Model/Behavior/SphinxBehavior.php#L167-L175) values returned by Manticore? This way, we can get rid of the magic value.

AndiPersti:
This would definitely be the best solution.

But I'm afraid that we cannot easily (re)use these two methods.
The both depend on the cached result we got from an earlier Manticore query. For the vocabulary numbers we call Manticore from [`_getNumberOfSentences`](https://github.com/Tatoeba/tatoeba2/blob/58ad65f02cfd06c3d3ccb155dc9632fec6317668/src/Model/Table/VocabularyTable.php#L123-L135) so we would need to get the real total there. But that method currently returns an int while the new implementation would return an array of two ints (the returned total and the real total) and so all the other methods in the call graph would need to change too.
Furthermore I don't think using `Sentences->find('withSphinx)` is the most efficient solution for getting the count for a vocabulary item because this also queries unnecessarily the MariaDB database.

All in all I think there is more refactoring necessary to get rid of the magic value. So for this PR I would suggest to just fix the bug.

PR #2454, opened 2020-07-16
Add Chinese Jinyu as UI language

RyckRichards:
This PR adds Chinese Jinyu on Tatoeba in order to help translators to check how their translations are rendering.

### DIFF
--- a/config/app_local.php.template
+++ b/config/app_local.php.template
@@ -159,6 +159,7 @@ return [
             ['bul', null, 'Български'],
             ['cat', null, 'Català'],
             ['ces', null, 'Čeština'],
+            ['cjy', null, '简体中文'],
             ['cmn', 'Hans', '中文', ['chi']],
             ['cym', null, 'Cymraeg'],
             ['dan', null, 'Dansk'],


jiru:
@RyckRichards I am not a native speaker, but I don’t think the language name 简体中文 means "Chinese Jinyu". "简体中文" means "Chinese using simplified characters" (that is to say Mandarin). The UI is already translated in Chinese (using simplified characters) using the ISO code is cmn (zh_CN on Transifex) and the language name 中文 ("Chinese"). Shouldn’t "Chinese Jinyu" be something like 晋语/晉語?

RyckRichards:
I asked that to a translator who is translating Tatoeba UI into Chinese Jinyu and he gave me that. I'll make a "deeper" research.

jiru:
I just hope that the translator didn’t re-translate the UI in Mandarin…

RyckRichards:
Goot catch! I'll compare both translations to make sure everything is OK in there

RyckRichards:
Since the translators and native speakers replied to [this query](https://github.com/Tatoeba/tatoeba2/pull/2454#issuecomment-659715788), I'm closing it.

jiru:
@RyckRichards I’m confused. What was their reply?

RyckRichards:
Sorry, I wasn't clear. I've noticed that now.
No one has replied to this query.

PR #2457, opened 2020-07-17
Add advanced search permalink option

jpear1:
Partially addresses #1832 as discussed in #2403.

### DIFF
--- a/src/Controller/SentencesController.php
+++ b/src/Controller/SentencesController.php
@@ -555,7 +555,14 @@ class SentencesController extends AppController
     public function advanced_search() {
         $search = new SentencesSearchForm();
 
-        $search->setData([]);
+        $data = $this->request->getParam('?');
+
+        if ($data) {
+            $search->setData($data);
+        } else {
+            $search->setData([]);
+        }
+
         $this->set($search->getData());
 
         $searchableLists = $search->getSearchableLists(CurrentUser::get('id'));
--- a/src/Controller/SentencesController.php
+++ b/src/Controller/SentencesController.php
@@ -556,8 +556,9 @@ class SentencesController extends AppController
         $search = new SentencesSearchForm();
 
         $search->setData($this->request->getQueryParams());
+        $usesTemplate = $search->getData() != SentencesSearchForm::DEFAULT_CRITERIA;
 
-        $this->set($search->getData());
+        $this->set($search->getData() + compact('usesTemplate'));
 
         $searchableLists = $search->getSearchableLists(CurrentUser::get('id'));
         $this->set(compact('searchableLists'));
--- a/src/Controller/SentencesController.php
+++ b/src/Controller/SentencesController.php
@@ -556,7 +556,7 @@ class SentencesController extends AppController
         $search = new SentencesSearchForm();
 
         $search->setData($this->request->getQueryParams());
-        $usesTemplate = $search->getData() != SentencesSearchForm::DEFAULT_CRITERIA;
+        $usesTemplate = !$search->isUsingDefaultCriteria();
 
         $this->set($search->getData() + compact('usesTemplate'));
 
--- a/src/Controller/SentencesController.php
+++ b/src/Controller/SentencesController.php
@@ -555,13 +555,7 @@ class SentencesController extends AppController
     public function advanced_search() {
         $search = new SentencesSearchForm();
 
-        $data = $this->request->getParam('?');
-
-        if ($data) {
-            $search->setData($data);
-        } else {
-            $search->setData([]);
-        }
+        $search->setData($this->request->getQueryParams());
 
         $this->set($search->getData());
 
--- a/src/Template/Element/advanced_search_form.ctp
+++ b/src/Template/Element/advanced_search_form.ctp
@@ -378,7 +378,7 @@ echo $this->Form->create('AdvancedSearch', [
             </md-button>
             <span>
                 <md-icon>help</md-icon>
-                <md-tooltip md-direction="top">
+                <md-tooltip class="multiline" md-direction="top">
                     <?= __('Use this button to use the currently selected criteria as a base for other searches. '
                           .'You can also bookmark/share the search template with someone else.') ?>
                 </md-tooltip>
--- a/src/Template/Element/advanced_search_form.ctp
+++ b/src/Template/Element/advanced_search_form.ctp
@@ -30,15 +30,17 @@ echo $this->Form->create('AdvancedSearch', [
 ?>
 
 <?php if ($this->request->getParam("?")): ?>
-    <i><?= __("This form has been filled by a template.")?></i>
-    <md-button class="md-primary md-raised"
-                href="<?= $this->Url->build([
-                    "controller" => "Sentences",
-                    "action" => "advanced_search",
-                    "?" => [],
-                ]);?>">
-        <?= __('Clear Form') ?>
-    </md-button>
+    <div id="template-notif">
+        <i><?= __("This form has been filled by a template.")?></i>
+        <md-button class="md-primary md-raised"
+                    href="<?= $this->Url->build([
+                        "controller" => "Sentences",
+                        "action" => "advanced_search",
+                        "?" => [],
+                    ]);?>">
+            <?= __('Clear Form') ?>
+        </md-button>
+    </div>
     <md-divider></md-divider>
 <?php endif; ?>
 
--- a/src/Template/Element/advanced_search_form.ctp
+++ b/src/Template/Element/advanced_search_form.ctp
@@ -32,14 +32,14 @@ echo $this->Form->create('AdvancedSearch', [
 <div layout="column" ng-app="app" ng-cloak>
     <?php if ($this->request->getParam("?") && !(isset($isSidebar) && $isSidebar)): ?>
         <div id="template-notif">
-            <i><?= __("This form has been filled by a template.")?></i>
+            <i><?= __("This form is pre-filled.")?></i>
             <md-button class="md-primary md-raised"
                         href="<?= $this->Url->build([
                             "controller" => "Sentences",
                             "action" => "advanced_search",
                             "?" => [],
                         ]);?>">
-                <?= __('Clear Form') ?>
+                <?= __('Clear form') ?>
             </md-button>
         </div>
         <md-divider></md-divider>
@@ -362,7 +362,7 @@ echo $this->Form->create('AdvancedSearch', [
         </div>
     </div>
 
-    <div class="buttons" layout="<?= $layout ?>" style="align-items:<?= (isset($isSidebar) && $isSidebar) ? 'stretch' : 'center'?>">
+    <div class="buttons" layout="<?= $layout ?>">
         <md-button type="submit" class="md-primary md-raised">
             <?php /* @translators: search form submit button (verb) */ ?>
             <?= __x('button', 'Search') ?>
@@ -378,7 +378,7 @@ echo $this->Form->create('AdvancedSearch', [
             </md-button>
             <span>
                 <md-icon>help</md-icon>
-                <md-tooltip md-direction="top" md-delay="1000">
+                <md-tooltip md-direction="top">
                     <?= __('Use this button to use the currently selected criteria as a base for other searches. '
                           .'You can also bookmark/share the search template with someone else.') ?>
                 </md-tooltip>
--- a/src/Template/Element/advanced_search_form.ctp
+++ b/src/Template/Element/advanced_search_form.ctp
@@ -347,7 +347,7 @@ echo $this->Form->create('AdvancedSearch', [
         </div>
     </div>
 
-    <div class="buttons" layout="<?= $layout ?>">
+    <div class="buttons" layout="<?= $layout ?>" style="align-items:center">
         <md-button type="submit" class="md-primary md-raised">
             <?php /* @translators: search form submit button (verb) */ ?>
             <?= __x('button', 'Search') ?>
@@ -357,15 +357,22 @@ echo $this->Form->create('AdvancedSearch', [
                    href="http://en.wiki.tatoeba.org/articles/show/text-search">
             <?= __('More search options') ?>
         </md-button>
-        <?php if ($layout == 'row'): ?>
+        <?php if (!(isset($isSidebar) && $isSidebar)): ?>
             <md-button type="submit" class="md-primary" formaction="">
-                <?= __('Permalink to search') ?>
+                <?= __('Create a search template') ?>
             </md-button>
+            <span>
+                <md-icon class="unsure">help</md-icon>
+                <md-tooltip md-direction="right" md-delay="1000">
+                    <?= __('Use this button to use the currently selected criteria as a base for other searches.
+                        You can also bookmark/share the search template with someone else.') ?>
+                </md-tooltip>
+            </span>
         <?php endif; ?>
     </div>
 
     <?php if ($this->request->getParam("?")): ?>
-        <md-divider style="text-align:left">
+        <md-divider>
             <md-button class="md-primary md-raised"
                         href="<?= $this->Url->build([
                             "controller" => "Sentences",
--- a/src/Template/Element/advanced_search_form.ctp
+++ b/src/Template/Element/advanced_search_form.ctp
@@ -31,7 +31,7 @@ echo $this->Form->create('AdvancedSearch', [
 
 <div layout="column" ng-app="app" ng-cloak>
     <?php if (!(isset($isSidebar) && $isSidebar) && $usesTemplate): ?>
-        <div id="template-notif">
+        <div>
             <i><?= __("This form is pre-filled.")?></i>
             <md-button class="md-primary"
                         href="<?= $this->Url->build([
--- a/src/Template/Element/advanced_search_form.ctp
+++ b/src/Template/Element/advanced_search_form.ctp
@@ -378,7 +378,7 @@ echo $this->Form->create('AdvancedSearch', [
             </md-button>
             <span>
                 <md-icon>help</md-icon>
-                <md-tooltip md-direction="right" md-delay="1000">
+                <md-tooltip md-direction="top" md-delay="1000">
                     <?= __('Use this button to use the currently selected criteria as a base for other searches. '
                           .'You can also bookmark/share the search template with someone else.') ?>
                 </md-tooltip>
--- a/src/Template/Element/advanced_search_form.ctp
+++ b/src/Template/Element/advanced_search_form.ctp
@@ -362,7 +362,7 @@ echo $this->Form->create('AdvancedSearch', [
         </div>
     </div>
 
-    <div class="buttons" layout="<?= $layout ?>" style="align-items:center">
+    <div class="buttons" layout="<?= $layout ?>" style="align-items:<?= (isset($isSidebar) && $isSidebar) ? 'stretch' : 'center'?>">
         <md-button type="submit" class="md-primary md-raised">
             <?php /* @translators: search form submit button (verb) */ ?>
             <?= __x('button', 'Search') ?>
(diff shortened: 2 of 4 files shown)


src/Controller/SentencesController.php
```diff
@@ -555,7 +555,14 @@ public function search()
...
         $search = new SentencesSearchForm();
 
-        $search->setData([]);
+        $data = $this->request->getParam('?');
+
+        if ($data) {
+            $search->setData($data);
+        } else {
+            $search->setData([]);
+        }
```
jiru, line 564: I think this code can be simplified. There is another way to get the query parameters: [getQueryParams()](https://api.cakephp.org/3.7/class-Cake.Http.ServerRequest.html#_getQueryParams). This method will return an empty array if there are no parameters.

src/Template/Element/advanced_search_form.ctp
```diff
@@ -357,6 +357,10 @@ echo $this->Form->create('AdvancedSearch', [
                    href="http://en.wiki.tatoeba.org/articles/show/text-search">
             <?= __('More search options') ?>
         </md-button>
+
+        <md-button type="submit" class="md-primary" formaction="/eng/sentences/advanced_search">
+            <?= __('Permalink to search') ?>
```
jiru, line 362: I don’t think this button should appear on the right pane of the search results.

In terms of UI, I’m not sure a button fits the use case after all. This button forces the user to reload the page (by clicking it) in order to access the actual link. I’d personally rather have a link on which I can either click, or right-click → copy URL.

I wonder what would be the best wording. It refers to the *search criteria*, not the *search results*, so "search" alone may be too ambiguous. How about "Permalink to this form"? Maybe we can go for something totally different like "Bookmark this search"; but the action of the button is not exactly bookmarking. @alanfgh any suggestions?

jiru:
@jpear1 Thank your for your pull request. I tried it out and it works great! :+1: I think there is still room for improvement so I left a few comments.

jpear1:
@jiru Thanks!

src/Controller/SentencesController.php
jpear1: Ah was thinking that too. Will fix.

src/Template/Element/advanced_search_form.ctp
jpear1: How about just "create link"? It's appearing right next to two search buttons so I think it'll be clear that it's referring to the search.

src/Template/Element/advanced_search_form.ctp
```diff
@@ -357,6 +357,10 @@ echo $this->Form->create('AdvancedSearch', [
                    href="http://en.wiki.tatoeba.org/articles/show/text-search">
             <?= __('More search options') ?>
         </md-button>
+
+        <md-button type="submit" class="md-primary" formaction="/eng/sentences/advanced_search">
```
jpear1, line 361: Whoops, just realized this is hard-linked to English.

src/Template/Element/advanced_search_form.ctp
jiru: If I’d see a "Create link" button, I’d ask to myself: "what link?" I understand it’s related to search, but it’s still rather unclear. From there, the result of clicking on this "Create link" button is that I am brought to the same page. Only if I am curious enough to pay attention to the URL (which would probably be hidden on mobile), I may understand that I can re-use that URL to open the form with pre-filled values.

I think we may be able to make the whole thing easier to understand by showing more clearly what happens *after* clicking on the button. For example, you could display a flash message (after the page reloads) that says something among the lines "You can bookmark this page or copy its URL to load the search form with pre-filled values."

jpear1: Makes sense. How does this sound:

Label the button "create link" (or something else concise and somewhat vague)

When the button is clicked:
Use a little Javascript to copy the link to the clipboard without reloading, then display a small popup above the button that reads "Bookmark or share this link to load the search form with the values you've already filled in"

src/Template/Element/advanced_search_form.ctp
jiru: That’s right. You should be able to create the correct link using the [URL helper](https://book.cakephp.org/3/en/views/helpers/url.html#generating-urls).

src/Template/Element/advanced_search_form.ctp
jiru: Thanks for your feedback :smile: In your suggestion, the UI mentions a link, but that link is not visible anywhere (because it’s directly copied to the clipboard), so one might still ask "What link?"

Compare for example with how Youtube does it. They have a "share" button that opens a dialog containing a text field containing a link, with a "copy" button next to it.

Looking back to #1832, having the link automatically copied to the clipboard makes it easier to "share the search to a friend", but harder to "do the same research again" or "do a similar research" because the user will have to paste the link around instead of just clicking. However, I think only the latter was the main point of the ticket. I can see a lot more use cases for a pre-filled search form one can immediately re-use than for pasting the link somewhere or to somebody. I might be wrong/biased, but I can imagine how one would want do a lot of repeated searches, while I cannot really see why somebody would want to do a lot of repeated pasting.

The more I think about it, the more I think we shouldn’t mess up with the clipboard. I think a "Create a search template" button might be the best way to go because it can easily extend to "Create a search template that I can share" into one’s mind, whereas "Bookmark/share" won’t easily extend to "Bookmark/share a search template".

jpear1: I agree, redoing a similar/identical seems like a far more common use case, which a "copy link to clipboard" button wouldn't facilitate very well. Keeping things as is does make redoing research a little easier.

My one concern is that this may be a little non-intuitive/unusual. I don't know of any buttons on any site that store form data in the URL. It always seems to be stored in session (hence my earlier suggestion in #2403 to separate sharing via url from redoing research). From my very little UI design experience, I believe it's better to duplicate behaviors users have learned in other programs. However, since trang thinks that the default behavior should be to not save search criteria and I don't have a strong reason to believe otherwise, this may be the best compromise. So, I think that some explanation would be helpful for users who aren't watching the address bar like a hawk. A popup explaining the button every time you create a template would be annoying, so maybe we could put some alt text on it?

jiru: Alright. The problem with alt text (I assume you mean a tooltip) is that it won’t show up on mobile. An [md-tooltip](https://material.angularjs.org/latest/api/directive/mdTooltip) will, but only if you tap the thing it is applied to, so it’s no use to apply it to the button. So I suggest putting the tooltip but on a "?" icon next to the button.

I still have one concern. Since there no difference between the pre-filled search form page and the empty search form page (except the URL), the user might get confused. Let’s say I use a templated search at some point, and then I just go on doing other unrelated searches and I forget that the form is pre-filled. Soon, I might get frustrated because the form values keep being pre-filled with unwanted values, even after reloading the page. To avoid this, I think it might be a good idea to have some kind of indication on the pre-filled form that the user is browsing a pre-filled form. It could be as simple as having a "This form is pre-filled. [Clear]" text on the top. Having such text would make the "Your search will remain when you refresh the page or go back to the page" part of the tooltip unnecessary.

About the text, I suggest avoiding the technical word "URL". How about: "Use this button to use the currently select criteria as a base for other searches. You can also bookmark/share the search template with someone else."

jpear1: Okay I like this. I'll rename the button to "Create a search template", add a tooltip, and a little notice that tells you that you're using a template, with a button that will clear the form (just link back to the starting advanced search page).

src/Template/Element/advanced_search_form.ctp
```diff
@@ -357,6 +357,11 @@ echo $this->Form->create('AdvancedSearch', [
                    href="http://en.wiki.tatoeba.org/articles/show/text-search">
             <?= __('More search options') ?>
         </md-button>
+        <?php if ($layout == 'row'): ?>
```
jiru, line 360: Your code would be easier to understand if you used `$isSidebar` instead of `$layout == 'row'`.

jpear1:
I initially added the "this is a pre-filled form" message to the bottom of the form because I thought it would be annoying for users who regularly used a bookmarked template, then moved it to the top in a separate commit for comparison.

If you think it's better on the bottom, I can remove that last commit.

jiru:
@jpear1 Thanks for the update. :slightly_smiling_face: I tried it out and I found out a few issues.
1. The message *This form has been filled by a template.* is displayed on the right pane of the search results.
2. A link such as [this one](http://localhost:8080/fra/sentences/advanced_search?query=) or [that one](http://localhost:8080/fra/sentences/advanced_search?foobar=) triggers the message even though the form contains no pre-filled values.

src/Template/Element/advanced_search_form.ctp
```diff
@@ -29,6 +29,21 @@ echo $this->Form->create('AdvancedSearch', [
...
+        <md-button class="md-primary md-raised"
+                    href="<?= $this->Url->build([
+                        "controller" => "Sentences",
+                        "action" => "advanced_search",
+                        "?" => [],
+                    ]);?>">
+            <?= __('Clear') ?>
+        </md-button>
+        <i><?= __("This form has been filled by a template.")?></i>
+    </md-divider>
```
jiru, line 43: I thought the `<md-divider>` tag wasn’t supposed to contain anything. :thinking:

src/Template/Element/advanced_search_form.ctp
```diff
@@ -29,6 +29,21 @@ echo $this->Form->create('AdvancedSearch', [
...
+<?php if ($this->request->getParam("?")): ?>
+    <md-divider>
+        <md-button class="md-primary md-raised"
+                    href="<?= $this->Url->build([
+                        "controller" => "Sentences",
+                        "action" => "advanced_search",
+                        "?" => [],
+                    ]);?>">
+            <?= __('Clear') ?>
+        </md-button>
```
jiru, line 41: I would have put the button *after* the text. I think it makes it clearer what the button is clearing. You could as well make the button say "Clear form" for more clarity.

src/Template/Element/advanced_search_form.ctp
```diff
@@ -357,6 +372,18 @@ echo $this->Form->create('AdvancedSearch', [
                    href="http://en.wiki.tatoeba.org/articles/show/text-search">
             <?= __('More search options') ?>
         </md-button>
+        <?php if (!(isset($isSidebar) && $isSidebar)): ?>
+            <md-button type="submit" class="md-primary" formaction="">
+                <?= __('Create a search template') ?>
+            </md-button>
+            <span>
+                <md-icon class="unsure">help</md-icon>
```
jiru, line 380: I don’t think you need to use the `unsure` class here.

src/Template/Element/advanced_search_form.ctp
```diff
@@ -357,6 +372,18 @@ echo $this->Form->create('AdvancedSearch', [
...
+                <md-tooltip md-direction="right" md-delay="1000">
+                    <?= __('Use this button to use the currently selected criteria as a base for other searches.
+                        You can also bookmark/share the search template with someone else.') ?>
```
jiru, line 383: I’m not exactly sure why, but I have trouble seeing the tooltip. It keeps blinking every second. I think you can reproduce the problem if you shrink your window enough so that the text overflows the browser window.

Another (probably unrelated) problem is that the string contains a "line return" character as well as lots of spaces before the word "You". This is likely to show up strangely for UI translators. You can break the string into multiple chunks using the dot operator to keep a nice indentation like this:
```php
                    <?= __('This is a string '    // notice the final space here
                          .'on multiple lines '   // and here
                          .'nicely indented.') ?>
```

jiru:
@jpear1 About the message on the top vs. bottom, I think top is fine. Maybe it’s just me, but I feel like two horizontal lines are a bit too much compared to how the rest of the form is displayed. A single divider on the bottom may be enough. Alternatively, you could put the whole text + button into a box with a light background color or something.

src/Template/Element/advanced_search_form.ctp
```diff
@@ -29,6 +29,21 @@ echo $this->Form->create('AdvancedSearch', [
...
+                        "controller" => "Sentences",
+                        "action" => "advanced_search",
+                        "?" => [],
+                    ]);?>">
+            <?= __('Clear') ?>
+        </md-button>
+        <i><?= __("This form has been filled by a template.")?></i>
+    </md-divider>
+    <md-divider></md-divider>
+<?php endif; ?>
```
jiru, line 45: This block is visible as unstyled text during page load. We can use the HTML attribute `ng-cloak` to [prevent this](https://www.wikitechy.com/angularjs/directives/angularjs-ngcloak). So I suggest you move this code inside the div below `<div layout="column" ng-app="app" ng-cloak>`.

src/Template/Element/advanced_search_form.ctp
jpear1: That does read more naturally. I thought that having a big green button just floating in the middle of the page looked a little odd, so I tried right aligning it. Compare 9c1f780 and d99b11f.

src/Template/Element/advanced_search_form.ctp
```diff
@@ -347,7 +362,7 @@ echo $this->Form->create('AdvancedSearch', [
         </div>
     </div>
 
-    <div class="buttons" layout="<?= $layout ?>">
+    <div class="buttons" layout="<?= $layout ?>" style="align-items:center">
```
jpear1, line 365: This seems to be causing the search and more search options buttons to display incorrectly on the search results screen.

src/Template/Element/advanced_search_form.ctp
jpear1: Looks like this is partially fixed by changing the tooltip direction to "top". It still runs off the page, but it doesn't flash.

I looked into putting newlines or doing automatic wrapping in md-tooltips and it doesn't seem like they're supported. Would love to be wrong about this though.

jpear1:
Sorry for the force-pushing mess. 😅

src/Template/Element/advanced_search_form.ctp
```diff
@@ -362,7 +362,7 @@ echo $this->Form->create('AdvancedSearch', [
         </div>
     </div>
 
-    <div class="buttons" layout="<?= $layout ?>" style="align-items:center">
+    <div class="buttons" layout="<?= $layout ?>" style="align-items:<?= (isset($isSidebar) && $isSidebar) ? 'stretch' : 'center'?>">
```
jiru, line 365: You can put this style CSS files instead. There are two files involved:
* `webroot/css/sentences/advanced_search.css`
* `webroot/css/sentences/search.css`

`advanced_search.css` is imported on the advanced search page. It also includes `search.css` with an `@import` CSS directive. The search results page only includes `search.css`.

It looks like you can just set `align-items` to `center` in `advanced_search.css` and leave it unset on the search results page.

src/Template/Element/advanced_search_form.ctp
jiru: Good job fixing the blink problem. :+1:

I’ve looked into how to make a multiline tooltip that automatically wraps. I was able to get something working with the following code:
```html
<!-- src/Template/Element/advanced_search_form.ctp -->
<md-tooltip class="multiline" ...>
```

```css
/* webroot/css/sentences/advanced_search.css */
md-tooltip.multiline {
    /* Only wrap tooltip if it takes more than 90% of screen width. */
    max-width: 90%;
}

md-tooltip.multiline .md-content {
    height: auto;
    white-space: normal;
    /* The original md-tooltip use line-height to pad, but we want a nice
     * multi-line text, so we need to use line-height instead. */
    line-height: normal;
    padding: 5px 8px;
}
```

90% is not enough to make the text wrap using the current text length, but you can see the result by lowering this value. This should make the tooltip responsive-proof and long-UI-translations-proof.

src/Template/Element/advanced_search_form.ctp
```diff
@@ -357,6 +372,18 @@ echo $this->Form->create('AdvancedSearch', [
                    href="http://en.wiki.tatoeba.org/articles/show/text-search">
             <?= __('More search options') ?>
         </md-button>
+        <?php if (!(isset($isSidebar) && $isSidebar)): ?>
+            <md-button type="submit" class="md-primary" formaction="">
+                <?= __('Create a search template') ?>
+            </md-button>
+            <span>
+                <md-icon>help</md-icon>
+                <md-tooltip md-direction="top" md-delay="1000">
```
jiru, line 381: I’ve tried "tapping" the tooltip using the in-browser mobile emulation (Ctrl+Shift+M) and it looks like `md-delay` has the annoying effect of delaying the tooltip appearance by 1 second after tapping the icon. So I suggest removing this.

src/Template/Element/advanced_search_form.ctp
jiru: Having it on the right looks odder to me than having a floating button. Maybe you can lower the attractiveness of the button by removing `md-primary` and/or `md-raised` and/or replacing it by a text link.

src/Template/Element/advanced_search_form.ctp
```diff
@@ -30,6 +30,21 @@ echo $this->Form->create('AdvancedSearch', [
 ?>
 
 <div layout="column" ng-app="app" ng-cloak>
+    <?php if ($this->request->getParam("?") && !(isset($isSidebar) && $isSidebar)): ?>
+        <div id="template-notif">
+            <i><?= __("This form has been filled by a template.")?></i>
```
jiru, line 35: I wonder if this sentence is very clear. The way I see it, the pre-filled form *is* the template. I suggest *This form is prefilled.*

src/Template/Element/advanced_search_form.ctp
```diff
@@ -30,6 +30,21 @@ echo $this->Form->create('AdvancedSearch', [
...
+            <md-button class="md-primary md-raised"
+                        href="<?= $this->Url->build([
+                            "controller" => "Sentences",
+                            "action" => "advanced_search",
+                            "?" => [],
+                        ]);?>">
+                <?= __('Clear Form') ?>
```
jiru, line 42: About the button label, we use sentence case in the rest of the website, so please use `Clear form` instead.

webroot/css/sentences/advanced_search.css
```diff
@@ -26,6 +26,24 @@
     margin: 18px 0;
 }
 
+.buttons {
+    align-items: center;
+}
+
+md-tooltip.multiline {
+    /* Only wrap tooltip if it takes more than 90% of window width. */
```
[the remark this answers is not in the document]
jpear1: Oh, good to know!

src/Template/Element/advanced_search_form.ctp
```diff
@@ -30,6 +30,21 @@ echo $this->Form->create('AdvancedSearch', [
 ?>
 
 <div layout="column" ng-app="app" ng-cloak>
+    <?php if ($usesTemplate && !(isset($isSidebar) && $isSidebar)): ?>
```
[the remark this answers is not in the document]
jpear1: Whoops, forgot to test the search screen. Thanks!

src/Template/Element/advanced_search_form.ctp
```diff
@@ -30,6 +30,21 @@ echo $this->Form->create('AdvancedSearch', [
 ?>
 
 <div layout="column" ng-app="app" ng-cloak>
+    <?php if (!(isset($isSidebar) && $isSidebar) && $usesTemplate): ?>
+        <div id="template-notif">
```
jiru, line 34: You can remove this id now.

src/Controller/SentencesController.php
```diff
@@ -555,8 +555,10 @@ public function search()
     public function advanced_search() {
         $search = new SentencesSearchForm();
 
-        $search->setData([]);
-        $this->set($search->getData());
+        $search->setData($this->request->getQueryParams());
+        $usesTemplate = $search->getData() != SentencesSearchForm::DEFAULT_CRITERIA;
```
jiru, line 559: @jpear1 Thanks answering my comments! I saw you changed the default criteria into a const. I’d like to suggest a more object-oriented approach. You can keep this attribute private and add a new method to `SentencesSearchForm` to check if the form is using default values or not. If you choose a good name for that method, the above line could be easier to understand. For example:
```php
$usesTemplate = !$search->isUsingDefaultCriteria();
```

jiru:
@jpear1 Thanks for answering all our comments and good job with this PR! :+1:

PR #2458, opened 2020-07-17
Add Turkmen as UI language

RyckRichards:
This PR adds Turkmen as a UI language in order to help translators to check how their translations are rendering.


jiru:
Thanks :cat:

Issue #1667, opened 2018-09-11
Results in random order should be displayed without replacement


alanfgh:
When I do a search for English-Russian sentence pairs where the English is "push":

["push" in Russian](https://tatoeba.org/eng/sentences/search?query=push&from=eng&to=rus&orphans=no&unapproved=no&user=&tags=&list=&has_audio=&trans_filter=limit&trans_to=rus&trans_link=&trans_user=&trans_orphan=&trans_unapproved=&trans_has_audio=&sort=random)

the same sentence:

[sentence 252087](https://tatoeba.org/eng/sentences/show/252087)

shows up on multiple pages of the result. It can be seen on [page 2](https://tatoeba.org/eng/sentences/search/page:2?query=push&from=eng&to=rus&orphans=no&unapproved=no&user=&tags=&list=&has_audio=&trans_filter=limit&trans_to=rus&trans_link=&trans_user=&trans_orphan=&trans_unapproved=&trans_has_audio=&sort=random), [page 3](https://tatoeba.org/eng/sentences/search/page:3?query=push&from=eng&to=rus&orphans=no&unapproved=no&user=&tags=&list=&has_audio=&trans_filter=limit&trans_to=rus&trans_link=&trans_user=&trans_orphan=&trans_unapproved=&trans_has_audio=&sort=random), and probably more. It would be better if a result only showed up on one page, since displaying it on multiple pages is inefficient and distracting.

jiru:
Well, when the order is random, it is not kept among different pages. Going to a different page shows the results in a different random order, so the same sentence may show up again. To put it differently, reloading page 2 will always show different results. I understand this is not very intuitive though.

I can think of a way to solve this problem. We could keep the order among pages of a given randomly sorted result set by using Sphinx’s `rand_seed` option:
1. When clicking on the search button, generate a random value and give it to Sphinx as `rand_seed`.
2. Add this value to the GET parameters of all the page links.
3. Run the Sphinx query with this value, if provided as GET parameter.

With this design, when clicking on the search button, you’ll see the results of page 1 randomly sorted. If you click again, or reload the page using F5, it will show different results. From there, if you go to page 2 (or 3, 4 etc.), sentences of page 1 won’t show up, and if you reload page 2, it will show the same sentences again. From there, if you go to page 1 again, you will see the same results as when you last clicked on the search button. Reloading it won’t load different results. The only way to bring a new random order is to click on the search button.

What do you think?

alanfgh:
[...]

PR #2460, opened 2020-07-19
Make randomly-sorted result sets consistent over pages

jiru:
This PR solves #1667 by adding a random seed value mechanism to keep randomly-sorted result sets consistent over pages.

New behaviour: when a search is performed with random order, a seed value is randomly generated and the request is redirected to the same URL with an additional `rand_seed` GET parameter containing the seed value encoded in base64. This seed value will stick on other pages of the result. Reloading the page won’t show a different order any more; only clicking the Search button again will. This change might confuse some users.

Random searches will be slower (to some extent) because of the redirection, but I think this implementation is simpler and more solid.

### DIFF
--- a/src/Form/SentencesSearchForm.php
+++ b/src/Form/SentencesSearchForm.php
@@ -44,6 +44,8 @@ class SentencesSearchForm extends Form
         'rand_seed' => '',
     ];
 
+    private $paramsOrder = [];
+
     public function __construct(EventManager $eventManager = null) {
         parent::__construct($eventManager);
         $this->search = new Search();
@@ -277,6 +279,9 @@ class SentencesSearchForm extends Form
 
     public function setData(array $data)
     {
+        /* Remember params order */
+        $this->paramsOrder = array_keys($data);
+
         /* Convert simple search to advanced search parameters */
         if (isset($data['to']) && !isset($data['trans_to'])) {
             $data['trans_to'] = $data['to'];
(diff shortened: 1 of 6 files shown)


src/Form/SentencesSearchForm.php
```diff
@@ -268,6 +307,33 @@ public function setData(array $data)
...
+        $data = parent::getData($field);
+        if (is_null($field)) {
+            uksort($data, function($a, $b) {
+                return $this->paramIndex($a) <=> $this->paramIndex($b);
+            });
+        }
+        return $data;
+    }
+
+    public function generateRandomSeedIfNeeded() {
```
AndiPersti, line 328: Should the check for the `rand_seed` parameter only work for the pagination links or also for the sidebar search form?

I.e. if I go to the advanced search page and set a query string, a language and random order, I get redirected to the result page and the search form in the sidebar is prefilled with the search settings. If I then change another parameter but keep the random order should I get the result in the same random order or not?

If yes, then that currently doesn't work because the advanced form never sets the random seed. (You would need to set a hidden field in the form.)

If no, I wonder whether it makes sense to have the option for the reversed order when the current sort order is random. After submitting the form, a new random seed will be generated and so the result isn't in reversed order but in a new random order.

AndiPersti:
From the commit message:
> That URL, generated by the Router::url() method, should be as close as possible to the original requested URL. That’s why we want to keep the same array order.

Out of curiosity, is there a technical reason why the parameter order should be as close as possible to the original order?

AndiPersti:
Sorry, the closing was accidental.

src/Form/SentencesSearchForm.php
jiru: > I then change another parameter but keep the random order should I get the result in the same random order or not?

Currently, you can only get a new random order. Maybe I can add a new option in the sort dropdown: "Random (keep current order)" and "Random (generate new order)". These options would only show up in the side bar of a randomly sorted search.

> I wonder whether it makes sense to have the option for the reversed order when the current sort order is random

This is already a problem and this PR does not aim at solving it.

jiru:
> Out of curiosity, is there a technical reason why the parameter order should be as close as possible to the original order?

No. I just want to make it easy to manually edit URLs. I am used to edit parameters such as `from` and `to`. So my goal is to have the same URL with an extra `&rand_seed=...` parameter. It is still not perfect because empty parameters are stripped out by `Router::url()`.

Issue #2462, opened 2020-07-20
Unable to add profile language without language details


jiru:
Steps to reproduce:
1. Go to profile page
2. Click on the Add language button
3. Select a language you don’t have in your profile yet
4. Select any level
5. Leave the "details" field blank
6. Submit
7. Error: "This language has already been added to your profile."

**Analysis**

The actual error is: `SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'details' cannot be null`

Commit 403a56ac36f5c4eb8f5f68ab12b611c12057acb0 changed something in the way the [request is constructed](https://github.com/Tatoeba/tatoeba2/blob/9168aee044c64aa4084081faae65ae5e6e3eb34d/webroot/js/user/language.ctrl.js#L68-L75). In the past, the request would produce a JSON containing an empty *details* key, whereas now, it doesn’t contain the *details* key any more.

AndiPersti:
The problem is that `$httpParamSerializerJQLikeProvider` isn't really jQuery-like because undefined values won't be included in the serialized parameter string.

The easiest fix would probably be to update AngularJS to at least 1.7.0 where this was fixed: https://github.com/angular/angular.js/commit/301fdda648680d89ccab607c413a7ddede7b0165

trang:
I'll take this.

I'm currently checking the upgrade of AngularJS Material to 1.1.5 because I found out that earlier versions won't allow to set the cursor position on mobile (see https://github.com/angular/material/issues/10301). So I might as well upgrade AngularJS to 1.7.0 at the same time.

AndiPersti:
I've already updated locally to 1.8.0 (the newest and also last version) and haven't found any issues yet. Looking at the [changelog](https://github.com/angular/angular.js/blob/master/CHANGELOG.md) there also isn't nothing mentioned which should break our code.

PR #2463, opened 2020-07-20
Upgrade AngularJS and AngularJS Material

trang:
This pull request upgrades AngularJS to v1.8.0. This solves #2462.

It also upgrades AngularJS Material to v1.1.5 to fix a bug where mobile users cannot position the cursor in text inputs.


jiru:
Thank you very much! :blush:

PR #2469, opened 2020-07-22
Replace jQuery with pure Javascript

### LINKED ISSUE #2468: Audio player quit working on this page: https://tatoeba.org/eng/audio/index
I tried this.  I also cleared my cache and tried it again.

https://tatoeba.org/eng/audio/index
Screen Shot 2020-07-22 at 15 02 04

The audio player doesn't play the audio.

However, if I go to the sentence's page, it plays.

Screen Shot 2020-07-22 at 15 02 14

AndiPersti:
This PR fixes #2468 

The proposed fix replaces the jQuery code with a pure Javascript implementation.

### DIFF
--- a/webroot/js/sentences.play_audio.js
+++ b/webroot/js/sentences.play_audio.js
@@ -16,13 +16,23 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-$(document).ready(function () {
-  $(document).watch('addrule', function () {
-    $('.audioAvailable').off();
-    $('.audioAvailable').click(function () {
-      var audioURL = $(this).attr('href');
-      var audio = new Audio(audioURL);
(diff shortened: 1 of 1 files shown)


jiru:
Thanks :rocket: 

So you tried it out on IE?

AndiPersti:
> So you tried it out on IE?

Not yet. (I don't have easy access to IE.)
I've just used the information on [caniuse.com](https://caniuse.com/#feat=mdn-api_nodelist_foreach) and [MDN](https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach#Polyfill) so it should work for IE 9+.

trang:
I've tested on IE 11 and it works fine there.

Thanks @AndiPersti, as always :)

PR #2471, opened 2020-07-24
change bootstrap frozen time for #2319

polkadotazn:
Hi, I'm really sorry it has taken so long. I ended up making a new branch for this issue. Please let me know how it looks. @trang @jiru @AndiPersti

### DIFF
--- a/src/View/Helper/DateHelper.php
+++ b/src/View/Helper/DateHelper.php
@@ -124,7 +124,7 @@ class DateHelper extends AppHelper
         $diff = Time::fromNow($dateObj);
 
         if ($diff->days > 30) {
-            $formattedDate = $dateObj->nice();
+            $formattedDate = $dateObj->i18nFormat([\IntlDateFormatter::LONG, \IntlDateFormatter::SHORT]);
 
             if ($alone) {
                 return $formattedDate;
--- a/src/View/Helper/DateHelper.php
+++ b/src/View/Helper/DateHelper.php
@@ -124,7 +124,7 @@ class DateHelper extends AppHelper
         $diff = Time::fromNow($dateObj);
 
         if ($diff->days > 30) {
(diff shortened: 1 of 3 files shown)


jiru:
@polkadotazn Thank you :+1: I tried it out and it works great!
ok

However, when the date is more than 30 days in the past, both the tooltip and the date show the same long date:
ko

As you can see, it is so long that it got ellipsed out. In that case, I think the date displayed on the comment block should be more concise (for example only day, month and year), while the tooltip can show the full format including the time and "UTC". You can try this case by manually changing the date of a comment with a SQL request such as `update sentence_comments set created = '2010-10-10 01:22:33' where id = 2;`.

polkadotazn:
I see. good catch, thanks! I updated the past 30 days dates! I'm not sure if this was the best way to do it, so please let me know if there's a better method :)

src/View/Helper/DateHelper.php
```diff
@@ -124,7 +124,7 @@ public function ago($date, $alone=true)
         $diff = Time::fromNow($dateObj);
 
         if ($diff->days > 30) {
-            $formattedDate = $dateObj->nice();
+            $formattedDate = $dateObj->i18nFormat([\IntlDateFormatter::LONG, \IntlDateFormatter::SHORT]);
```
jiru, line 127: This code results in the following:
image

I think you could as well remove the time:
```
$formattedDate =  $dateObj->i18nFormat([\IntlDateFormatter::LONG, \IntlDateFormatter::NONE]);
```
So that we get instead:
image


The rationale is that we are not displaying the time for dates between 1 and 30 days ago, so we shouldn’t display it for dates older than 30 days ago too.

polkadotazn: that makes sense!

polkadotazn:
alrighty how's it looking now? @jiru

jiru:
@polkadotazn Perfect! :+1: Thanks again :smile:

PR #2473, opened 2020-07-25
Responsive search results

trang:
This pull request makes the search results responsive. This applies only for non-authenticated users.

- On desktop mode, the search results shouldn't look very different from before.
- On mobile mode, the search bar has been simplified to only have the query input. Other options have to be specified from the advanced search form, which is hidden by default and displayed through the three-dot icon at the end of the search input.
- The advanced search form has been slightly redesigned so that the different sections ("Sentences", "Translations" and "Sort") are a bit more clearly separated.
- The top menu had to be redesigned for small screens (below 600px width). The problem was that for some users, the "Tatoeba" name and the language selection button were expanded beyond the screen. As a side effect, the button to open the advanced search would not be visible at first sight. For this reason, the following has been changed:
  - The language selection button only shows the icon (not the language name) on small screens.
  - The login and register items have been moved into the sidenav menu.

Related Wall thread: https://tatoeba.org/eng/wall/show_message/35638


jiru:
I found a bug: in desktop mode, using the old design, the audio icon is pushed to the next line:
image

src/Template/Layout/default.ctp
```diff
@@ -70,7 +74,7 @@ $htmlDir = LanguagesLib::getLanguageDirection($lang);
     <link rel="search" type="application/opensearchdescription+xml"
           href="/opensearch.xml" title="Tatoeba" />
     
-    <?php if ($controller === 'pages' && $action === 'index' && !CurrentUser::isMember()) { ?>
+    <?php if ($isHomepage && !CurrentUser::isMember() || $isSearch) { ?>
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
```
jiru: Since we are going to need that viewport meta code on more and more pages, I think you could make good use of [extended views](https://book.cakephp.org/3/en/views.html#extending-views) here. I’ve never used that myself, but it should allow `layout.ctp` to display variables that are defined by child views.

src/Template/Element/search_bar.ctp
```diff
@@ -28,17 +28,51 @@
...
+
+    <div layout="row" layout-align="center center" flex>
+        <md-input-container class="md-accent" flex md-no-float>
+            <md-icon>search</md-icon>
+            <input name="query" 
+                accesskey="4" 
+                dir="auto" 
+                ng-model="ctrl.searchQuery" 
+                ng-init="ctrl.searchQuery = <?= $searchQuery ?>"
+                placeholder="<?= __('Search') ?>"/>
```
jiru, line 62: About the string "Search": you should add a context with `__x()` and a translators’ note like below.

webroot/css/sentences/search.css
```diff
@@ -16,6 +16,31 @@
...
  */
 
+@media only screen and (max-width: 960px) {
+    .search_bar {
+        display: none;
+    }
+
+    #search-bar {
+        display: flex;
+    }
```
jiru, line 26: I think there is room for improvement in the naming of these class and id. :smile: 
How about: search-bar-full and search-bar-shrinkable?
It looks like both could be made ids.

src/Template/Sentences/search.ctp
```diff
@@ -165,7 +165,15 @@ if (!isset($results)) {
             md-component-id="advanced-search"
             md-disable-scroll-target="body"
             md-is-locked-open="$mdMedia('gt-sm')">
-    <md-subheader><?php echo __('More search criteria'); ?></md-subheader>
+    <md-toolbar>
+        <div class="md-toolbar-tools" ng-controller="SidenavController">
+            <h2 flex><?= __('Search criteria'); ?></h2>
```
jiru, line 170: Could you add a translator note to this string?

src/Template/Element/search_bar.ctp
```diff
@@ -28,17 +28,51 @@
...
+        </md-input-container>
+        <md-button class="md-icon-button" ng-click="toggle('advanced-search')">
+            <md-icon>more_vert</md-icon>
+        </md-button>
```
jiru, line 66: I’m not sure this "three-dots" button has enough visibility. Maybe this could be a button labelled "Narrow down" placed next to the search results title.

Besides, using small screens, I guess the only way to start a search after typing keywords is to use the Enter key. This might not be intuitive enough for many users. I think we should keep a "search" button.

Here is a draft of the two above suggestions using small screen:
```
≡          Tatoeba          🌐
[cat_dog________________] [🔎]
1234 results [Narrow down]
```

Of course the "Narrow down" button would allow more than that: modify results order, see more results and modify criteria. We could think about a different naming. But I think narrowing down the results currently really is the primary use case because most users will see too much results at first.

src/Template/Element/header.ctp
```diff
@@ -60,12 +60,13 @@ use Cake\Core\Configure;
...
         ]
     );
     ?>
 
-    <?= $this->element('ui_language_button', [
-        'displayOption' => 'hide-gt-xs'
-    ]); ?>
+    <md-button hide-gt-xs class="md-icon-button" ng-click="showInterfaceLanguageSelection()">
+        <md-icon>language</md-icon>
+    </md-button>
```
jiru, line 71: This code duplicates the contents of the `ui_language_button` element. Isn’t there a way to factorize this with its `hide-xs` counterpart in `src/Template/Element/top_menu.ctp`?

src/Template/Element/login.ctp
```diff
@@ -35,7 +35,7 @@ $registerUrl = $this->Url->build([
 ]);
 ?>
 
-<div ng-controller="LoginDialogController as vm" layout="row" layout-align="center center" flex>
+<div ng-controller="LoginDialogController as vm" layout="row" layout-align="center center" flex hide-xs hide-sm>
```
jiru, line 38: `hide-xs` displays when width < 600px.
`hide-sm` displays when 600px <= width < 960px.
I think their combination can be replaced by `show-gt-sm`, which displays when width >= 960px.

src/Template/Element/top_menu.ctp
```diff
@@ -226,6 +229,32 @@ $sidenavPosition = $htmlDir == 'rtl' ? 'md-sidenav-right' : 'md-sidenav-left';
...
+                    ?>
+                    </p>
+                </md-list-item>
+
+                <md-list-item ng-controller="LoginDialogController as vm" ng-click="vm.showDialog('<?= $this->Pages->currentPageUrl() ?>')">
+                    <p>
+                    <md-icon>chevron_right</md-icon>
+                    <?php 
+                    /* @translators: link to open the Login box in the top bar (verb) */
+                    echo __('Log in');
```
jiru, line 249: These two translators notes are incorrect. We are not in the top bar but the navigation side menu.

jiru:
Good job! :+1: It’s very encouraging to see progress on the responsive part. The search feature could be revamped in many ways to better fit mobile users, but this can be done later. My above comment about the "three dots" button could be addressed later as well.

src/Template/Layout/default.ctp
trang: Actually there won't be the need to use extended views here. I just realized we can already use in the layout variables set in the views. We did this with `title_for_layout`.

But thanks for mentioning this mechanism from CakePHP! That can help a lot for future redesign 😄

src/Template/Element/login.ctp
trang: The `show` directive is only negating `hide`. So `show-gt-sm` alone will still show the element even if the width is in the `xs` or `sm` range. To show *only* when width >= 960px, it would have to be `hide show-gt-sm`.

src/Template/Sentences/search.ctp
```diff
@@ -86,6 +86,19 @@ if ($ignored) {
...
 <md-content class="md-whiteframe-1dp" flex>
+
+<md-toolbar class="md-hue-2" hide-gt-xs ng-cloak ng-controller="SidenavController">
+    <div class="md-toolbar-tools">
+        <h2 flex><?= $this->Pages->formatResultCount($this->Paginator, $real_total); ?></h2>
+
+        <md-button ng-click="toggle('advanced-search')">
+            <md-icon>filter_list</md-icon>
+            <?php /* @translators: button to open the advanced search option sidebar on mobile */ ?>
+            <?= __('Narrow search') ?>
```
jiru, line 97: That looks pretty cool! :sunglasses: But now I can’t see how to open the search criteria sidebar when the width is between 600 and 960 pixels. :thinking:

src/Template/Element/ui_language_button.ctp
```diff
@@ -4,8 +4,13 @@ use Cake\Core\Configure;
...
-$label = $this->Languages->getInterfaceLanguage();
+if (!isset($class)) {
+    $class = '';
+}
+if (!isset($label)) {
+    $label = $this->Languages->getInterfaceLanguage();
+}
 ?>
-<md-button <?= $displayOption ?> ng-click="showInterfaceLanguageSelection()">
+<md-button <?= $displayOption ?> class="<?= $class ?>" ng-click="showInterfaceLanguageSelection()">
```
jiru, line 14: Pro tip: you can remove the above `isset` checks and then use the [null coalescing operator](https://www.php.net/manual/en/language.operators.comparison.php#language.operators.comparison.coalesce):
```html
<md-button <?= $displayOption ?? '' ?> class="<?= $class ?? '' ?>" ...>
```

jiru:
The old design is still problematic when the screen width smaller than 628px.

Using 930px:
Screen Shot 2020-08-02 at 17 28 02

Using 600px:
Screen Shot 2020-08-02 at 17 25 48

Using 470px:
Screen Shot 2020-08-02 at 17 29 19

src/Template/Element/top_menu.ctp
```diff
@@ -178,14 +187,8 @@ $sidenavPosition = $htmlDir == 'rtl' ? 'md-sidenav-right' : 'md-sidenav-left';
...
-                $isOnLoginPage = ($controller == 'Users' && $action == 'login');
-
-                if (!$isOnLoginPage) {
-                    echo $this->element('login');
-                }
+            if (!$session->read('Auth.User.id') && !$isOnLoginPage) {
+                echo $this->element('login');
             } else {
                 echo $this->element('space', ['htmlDir' => $htmlDir]);
             }
```
AndiPersti, line 194: This change is the reason for #2487.

The problem is that there are three cases and not just two:
1. User is logged in -> display the `space` element in the top menu
2. User is not logged in and is on the login page -> display neither the `space` element nor the `login` element
3. User is not logged in and on any other page -> display the `login` element

So I guess you want something like
```php
if ($session->read('Auth.User.id')) {
    echo $this->element('space', ['htmlDir' => $htmlDir]);
} elseif (!$isOnLoginPage) {
    echo $this->element('login');
}
```

src/Template/Element/sentences/sentence_and_translations.ctp
```diff
@@ -83,7 +83,7 @@ $sentenceUrl = $this->Url->build([
     </div>
     <div layout="column">
         <div layout="row" class="header">
-            <md-subheader flex class="ellipsis">
+            <md-subheader flex>
```
AndiPersti, line 86: Removing the `ellipsis` class causes the sentence number and owner written out vertically on the left border of the sentence box when the sentence menu is expanded. You can see this on the dev site.

I guess you need to replace the `ellipsis` class with `ng-if="!vm.isMenuExpanded"`.

src/View/Helper/PagesHelper.php
```diff
@@ -29,35 +29,43 @@ public function formatTitle($pageTitle) {
...
-                                '{title} ({thousand}&nbsp;results out of {n}&nbsp;occurrences)',
-                                $real_total, true),
-                            array(
-                                'title' => $title,
-                                'thousand' => $this->Number->format($n),
-                                'n' => $this->Number->format($real_total)
-                            )
+            /* @translators: this formats the number of results on pages
+               that shows a list of sentences (search, browse by language,
+               adopt sentences…). Note the use of &nbsp; which is a non-breaking space. */
```
jiru, line 54: Did you remove on purpose the part about "thousand" in the translators note? I think it’s still relevant.

jiru:
I think this pull request is more or less ready for merge. Though I’d like to point out something (that is probably outside of the scope of this PR).

When logged-in, on smallest screens, the "Sentence #{n} - belongs to {username}" string now wraps in an ugly way.
Compare the existing code (we have to scroll right to see the rest of the sentence block):


After merging your PR (no need to scroll any more, but it creates blank space that makes the sentence block taller):

jiru:
@trang I tried from a real mobile phone on dev.tatoeba.org as you suggested. While the screen width is a bit larger than on my previous comment, the same problem can happen when the username and sentence id are long enough:

trang:
@jiru On the dev, I added a viewport meta tag with a fixed width for non-responsive pages.

```
    <?php if (isset($isResponsive) && $isResponsive) { ?>
        <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <?php } else { ?>
        <meta name="viewport" content="width=970">
    <?php } ?>
```

Could you check if the search page for authenticated users now shows up the same way as on desktop but in miniature? If that's the case, then I'll push this change.

It seems the behavior is not consistent between devices/browsers. Perhaps nowadays some phones/browsers will render a page as `width=device-width, initial-scale=1.0` if there's no viewport meta tag.

Below is how it is expected to render on mobile. That is a screenshot from my phone, *without* the `<meta name="viewport" content="width=970">`.

Screenshot_2020-08-03-17-03-18-185_org mozilla firefox

jiru:
> Could you check if the search page for authenticated users now shows up the same way as on desktop but in miniature? If that's the case, then I'll push this change.

The `<meta name="viewport" content="width=970">` didn’t change anything. But now you mentioned there are inconsistencies between browsers, I think the problem comes from my outdated Android default browser. Using Firefox for Android, it works as expected (regardless of `<meta name="viewport" content="width=970">`): the search results display as responsive for guests and as miniature for logged-in users.

So I think we can say it was my outdated/buggy browser’s fault.

src/View/Helper/PagesHelper.php
trang: Was not on purpose. I pushed a commit to dev to re-add it: d6013bebed1f6fbebb482bcbc47c11247f6673c5.

PR #2475, opened 2020-07-26
Issue 2120

ftumsh:
migration for sentences_lists.name to utf8mb4_general_ci


jiru:
Good job with the migration! It just tested it and it works great. :+1:

I can merge your pull request already. Or, if you’d like to learn about [unit testing](https://en.wikipedia.org/wiki/Unit_testing), you could also add a unit test to make sure the bug won’t happen again in the future. Don’t feel like you have to do this; it’s more for educational purpose than anything else.

PR #2477, opened 2020-07-28
Improve error output for development environments

AndiPersti:
This PR prevents a crash in the development environment (`debug = true`) whenever an error or warning output contains `{{`. See also the discussion on #2420.

Since AngularJS shouldn't interpret the error messages, one possible solution is to add `ng-non-bindable` to the HTML element created by CakePHP's error handler. Unfortunately the template string for the error message isn't easily readable so we need to use PHP's Reflection API in order to access it.


jiru:
Thanks :smiley_cat:

PR #2478, opened 2020-07-29
Really fix audio in the old design

### LINKED ISSUE #2476: Audio player quit working on this page: https://tatoeba.org/eng/sentences/add
** Problem **

When contributing a sentence that already exists that also has audio, we used to be able to click the audio player button to listen to the audio file.

You can test is with this sentence.
Tell me a little more about that.

https://tatoeba.org/eng/sentences/add
Screen Shot 2020-07-27 at 14 50 29

** Likely Related **

https://github.com/Tatoeba/tatoeba2/issues/2468

AndiPersti:
This PR closes #2476 

Simply adding an audio element whenever we display an audio button in the old design requires only minimal Javascript code in the `onclick` handler in order to play audio files.


jiru:
What a smart way to fix the problem! :+1:

There is just a small problem though. I can see an ugly selection box after clicking the audio button.
On Firefox:
audio

And on Chrome:
audio_chrome

Compared to the existing implementation, it’s no longer possible get a crowd of Toms speaking together by jitter clicking the audio button. :laughing: I suppose it’s a usability improvement, but some people could have been having fun with it…

AndiPersti:
> There is just a small problem though. I can see an ugly selection box after clicking the audio button.

For some reason I don't see the dotted outline on my Firefox version (78 on Linux).

I've added two CSS rules which should remove it (for the Firefox-specific rule see [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#Firefox)).
I've tested on Firefox and Chromium on Linux and the native Android browser and two Firefox variants (IceCat and Firefox Preview) on an Android tablet.

AndiPersti:
(I've thought rebasing my branch on the current `dev` will pass the tests but it looks like df83982bd132b3f9054f5a531b7a36cb7bb6feac broke some tests for the sentence lists.)

PR #2483, opened 2020-08-02
Blank page no vocabulary items are found

speedysera:
This pull request addresses issue Tatoeba#2273 - when filtering by language, there is no indication why the page is blank when there are no vocabulary items found.

### DIFF
--- a/src/Template/Vocabulary/add_sentences.ctp
+++ b/src/Template/Vocabulary/add_sentences.ctp
@@ -27,10 +27,15 @@
 ?>
 <?php
 $this->Html->script('/js/vocabulary/add-sentences.ctrl.js', ['block' => 'scriptBottom']);
-
-$title = __('Vocabulary that needs sentences');
-
-$this->set('title_for_layout', $this->Pages->formatTitle($title));
+    if (empty($langFilter)) {
+        $title = __('Vocabulary that needs sentences');
+        $this->set('title_for_layout', $this->Pages->formatTitle($title));
+    } else {
+        $title = format(
+            __('Vocabulary that needs sentences in {language}'),
+            array('language' => $this->Languages->codeToNameToFormat($langFilter))
+        );
+    }
 ?>
 
 <div ng-cloak id="annexe_content">
@@ -53,6 +58,13 @@ $this->set('title_for_layout', $this->Pages->formatTitle($title));
                 'Only vocabulary items that match fewer than 10 sentences are '.
                 'displayed here.'
             )
+            ?>
+        </div>
+        <div class="empty-info-text">
+            <?php
+                if ($vocabulary->count() == 0) {
+                    echo "There are no requests. ";
+                }
             ?>
         </div>
 
--- a/webroot/css/layouts/default.css
+++ b/webroot/css/layouts/default.css
@@ -796,3 +796,10 @@ html[dir="rtl"] .md-button-right .md-button {
     left: -8px;
     right: initial;
 }
+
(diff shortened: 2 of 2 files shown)


AndiPersti:
A few comments:

- As mentioned by Trang on the issue I think it would be better to move the language information into the title. See for example the page for orphan sentences: https://github.com/Tatoeba/tatoeba2/blob/c3764aa0b0f00aa9aa8f1855ca5e19c32700441d/src/Template/Activities/adopt_sentences.ctp#L29-L36
You can also see there how to use `format()` to create a string with variable parts.
- I would move the check for existing vocabulary items further up right after `md-content`:
```php
<md-content>
    <?php if ($vocabulary->count() == 0) { ?>
          ...
    <?php } else { ?>
          ...
    <?php } ?>
<md-content>
```
So when there are no requests we only output the info text, otherwise we output everything else as we've done before.

jiru:
@jpear1 That’s a good start! :+1: Also don’t forget to handle the case when there is [no language filtering](https://tatoeba.org/eng/vocabulary/add_sentences) (select "All languages" in the dropdown).

speedysera:
Hi Andreas, 
    Good day to you! Thanks for the tip regarding _format.
    I have a clarification, as I was checking the code and the screenshot from Trang:
In add_sentences.ctp
$title = __('Vocabulary that needs sentences');
$this->set('title_for_layout', $this->Pages->formatTitle($title));
 <md-toolbar class="md-hue-2"> <div class="md-toolbar-tools"> <h2><?= $title ?></h2> </div> </md-toolbar>
 <md-content>


In this case, the "Vocabulary that needs sentences in Korean" is displayed in the md-toolbarsection, is it alright to put it there instead?
Thanks,H    On Sunday, August 2, 2020, 08:48:50 PM GMT+9, Andreas Perstinger <notifications@github.com> wrote:  
 
 


A few comments:
   
   - As mentioned by Trang on the issue I think it would be better to move the language information into the title. See for example the page for orphan sentences: https://github.com/Tatoeba/tatoeba2/blob/c3764aa0b0f00aa9aa8f1855ca5e19c32700441d/src/Template/Activities/adopt_sentences.ctp#L29-L36   
You can also see there how to use format() to create a string with variable parts.
   - I would move the check for existing vocabulary items further up right after md-content:
<md-content>
    <?php if ($vocabulary->count() == 0) { ?>
          ...
    <?php } else { ?>
          ...
    <?php } ?>
<md-content>
So when there are no requests we only output the info text, otherwise we output everything else as we've done before.

speedysera:
Hi @AndiPersti !

  Good day to you! Thanks for the tip regarding _format.

    I have a clarification, as I was checking the code and the screenshot from trang:

In **add_sentences.ctp**

```
$title = __('Vocabulary that needs sentences');

$this->set('title_for_layout', $this->Pages->formatTitle($title));

        <md-toolbar class="md-hue-2">
            <div class="md-toolbar-tools">
                <h2><?= $title ?></h2>
            </div>
        </md-toolbar>

        <md-content>
```

image


In this case, the "Vocabulary that needs sentences in Korean" is displayed in the `md-toolbar` section, is it alright to put it there instead?

AndiPersti:
> In this case, the "Vocabulary that needs sentences in Korean" is displayed in the `md-toolbar` section, is it alright to put it there instead?

Yes, it should be displayed in the toolbar. We usually use the same text for the page title and the toolbar.

jiru:
Closing this in favour of #2510.

Issue #2364, opened 2020-05-31
Translations are not ordered in a helpful way


trang:
Translations are currently ordered alphabetically by language name. When a sentence has a lot of translations and the user is interested in languages that are at the bottom of the list, they have to expand the translations and scroll a lot.

This issue has been intensified with the [transition towards the new sentence design](https://tatoeba.org/eng/wall/show_message/35389#message_35389) because each translation takes twice the space vertically.

Related: #2354

alanfgh:
Is there any chance that this can be implemented before the old sentence design is eliminated? Has there been a discussion of possible approaches? How much effort would be required?

I guess the following existing data could be taken into account:
- the user's profile languages
- the list of languages under Settings -> Options

But this might require some discussion, since the list of languages under Settings -> Options is currently used as a filter (only sentences in languages that appear in the list are displayed), and here we'd be using it as a sorting guide (translations in languages that appear in the list are displayed at the top of the lists of translations, but translations in other languages are displayed as well).

trang:
> Is there any chance that this can be implemented before the old sentence design is eliminated?

Yes. Keep in mind that the old design won't disappear as soon as the new design becomes the default one. It will still be possible to revert to the old design for a certain period of time after the new design has become the default one. 

I don't think we will implement this before we set the new design as default, but there's definitely a chance that this gets implemented before we remove the old design completely.

> Has there been a discussion of possible approaches?

There hasn't yet, but as with many features, we will implement something, let the community test, then readjust the implementation based on the feedback.

The most basic implementation would be to reuse the languages in the Settings and add an option "Show translations only in my languages". This option would be true by default. People who are currently using the languages option would then have no impact. People who want only to prioritize languages but still see translations in all languages would uncheck this option.

Alternatively, if we implement #263, we may not need to have this extra option. We could consider always hiding translations in other languages and whenever the user wants to see translations in all languages, they would use the button to show all translations.

> How much effort would be required?

The most basic implementation isn't very high effort. It's basically 3 things:
- Adding a new option in the settings.
- Taking into account this option when fetching translations.
- Sorting the translations based on the order of the languages in the settings.

alanfgh:
That all sounds fine to me. There's another situation to consider: the case where the user has specified that they want to see only sentences with translations in language X, but hasn't limited display of the translations to language X. In this case, it seems pretty clear to me that translations in language X should be listed first. Currently, they are not.

gioele-santi:
[...]

PR #2485, opened 2020-08-02
Made a draft of needed changes

gioele-santi:
I added a few notes where code should be changed to support new sorting of translations.

### DIFF
--- a/src/Model/Entity/User.php
+++ b/src/Model/Entity/User.php
@@ -62,6 +62,7 @@ class User extends Entity
         'new_terms_of_use' => '1',
         'license_switch_list_id' => null,
         'hide_new_design_announcement' => false,
+        'show_known_languages_first' => true, //this may be set to false as default, true for tests
     );
 
     private $settingsValidation = array(
--- a/src/Model/Table/SentencesTable.php
+++ b/src/Model/Table/SentencesTable.php
@@ -791,6 +791,11 @@ class SentencesTable extends Table
      */
     public function getSentenceWith($id, $what = [], $translationLang = null)
     {
+        //get current user (if logged) and its setting for sort
+        $userLangFirst = false; //see if option breaks everything
+        $userLangCodes = array();
+        //UsersLanguagesTable->getLanguagesOfUser($userId)
+
         return $this->find('filteredTranslations', [
                 'nativeMarker' => CurrentUser::getSetting('native_indicator'),
                 'hideFields' => $this->hideFields(),
@@ -799,6 +804,7 @@ class SentencesTable extends Table
             ->where(['Sentences.id' => $id])
             ->contain($this->contain($what))
             ->select($this->fields($what))
+            //->order(['FIND_IN_SET(Language.code, \'' . implode(',', $userLangCodes) . '\')'], $userLangFirst) //no Language.code column
             ->first();
     }
 


gioele-santi:
Hi, I made a draft. I found where I have to make the main changes for the sorting of translations but I need to figure out how languages are organised: as you can see I used `Language.code` as I found it referenced in other files, but I guess I should look better for the id used in the table.
I hope the structure is good at least. Tomorrow I will go on.

AndiPersti:
`SentencesTable::getSentenceWith()` will only return one sentence (the `WHERE` clause in the resulting SQL query will look like `WHERE Sentences.id = 123`) so adding an ordering to the query won't do anything.

You want to order the translations and not the sentence.
We use a [custom finder method](https://book.cakephp.org/4/en/orm/retrieving-data-and-resultsets.html#custom-finder-methods) called [`findFilteredTranslations`](https://github.com/Tatoeba/tatoeba2/blob/f10d7d8cbf0b43021c29805b1c5230dd0ddd8eaa/src/Model/Table/SentencesTable.php#L470) for fetching the translations and for post-processing the results. So this is the place where you need to add the ordering.

gioele-santi:
I am sorry, I thought the problem was actually related to a sentence page and the loaded translations in its page. I will check the other method.

gioele-santi:
So `findFilteredTranslations` should be called when performing 
```
Sentence->find('filteredTranslations', $findOptions)
``` 
right? 
I suppose we should pass the new user options using `$findOptions` 
or should I check current user preferences from inside the filter itself thus expanding the else at line 473-474 in this where user's languages are loaded? https://github.com/Tatoeba/tatoeba2/blob/f10d7d8cbf0b43021c29805b1c5230dd0ddd8eaa/src/Model/Table/SentencesTable.php#L470-L474

Since I should take care of translation sorting I wonder if I should just change function `sortOutTranslations` called at line 478.

AndiPersti:
Yes, I think you need to modify `sortOutTranslations`. The translations are currently sorted alphabetically: https://github.com/Tatoeba/tatoeba2/blob/9d5f05b5e17dc00f85988a357ba04d7331416f3f/src/Model/Table/SentencesTable.php#L463-L464

It would also be possible to get sorted translations from the database (see [Sorting Contained Associations](https://book.cakephp.org/3/en/orm/retrieving-data-and-resultsets.html#sorting-contained-associations)) but I'm not sure whether the post-processing inside `sortOutTranslations` keeps the order. But you can try it :smile:.

gioele-santi:
I am sorry but I don't seem to understand how to figure out this. In the original [issue page](https://github.com/Tatoeba/tatoeba2/issues/2364) it seemed to me that the problem was loading translations in sentence page and bring up user languages to make page more useful, especially for sentences with many translations. 

So following suggestions there I edited by adding the suggested **FIND_IN_SET** trick in 
https://github.com/Tatoeba/tatoeba2/blob/c8c6854e7fba3deae34cc408c6ec8b445745b450/src/Model/Table/SentencesTable.php#L792

but I couldn't make it work as I have problems identifying the id used for languages (in other tables I saw `code` but it generates an error).

Now I checked for a couple fo days this other suggested change but in `sortOutTranslations` it looks like other languages are filtered out...
https://github.com/Tatoeba/tatoeba2/blob/c8c6854e7fba3deae34cc408c6ec8b445745b450/src/Model/Table/SentencesTable.php#L455-L456

so it should be outside of the scope if this issue.

If you have any other suggestions I will be glad because I am quite lost.

AndiPersti:
Ok, I'll try to give you a step-by-step explanation:

1. The first thing you need to understand is that whether you load just a single sentence + its translations (like on the page for a sentence) or several sentences + their translations (like for the "Show all in" pages) you'll end up calling the same methods which talk to the database.
2. CakePHP's `find` method is a high level abstraction for building the actual SQL query that is finally sent to the database. Its first parameter is a "finder", i.e. a method that adds commonly used options to the query (e.g. where clauses, order clauses, ...) and/or post-processes the result.
3. `getSentenceWith` uses such a custom finder (`$this->find('filteredTranslations', ...`) to retrieve the data from the database. The name of the finder method is `find + <custom_name>` where `custom_name` is the value of the first parameter (here `filteredTranslations`, thus the name of the method CakePHP uses is `findFilteredTranslations` [notice the capitalization of the first letter of the parameter value]). You can pass options to the finder method using `find`'s second parameter which should be an array.
4. In order to fetch the translations we use the `contain` method which basically adds joins to the query, i.e. you pass it a list of tables and it will figure out how to join them together. We construct this list of tables using `$this->contain`, i.e. a method with the same name but which belongs to `SentencesTable`.
5. After the other two methods `where` and `select` modified the query it finally gets passed to our finder `findFilteredTranslations`. Here we first determine whether the translations should be filtered (using either the passed option or the user setting) and then the query is sent to the database because `$query->formatResults` forces evaluation of the query, i.e. it is executed. The returned set is then post-processed.
6. `formatResults` takes a callback in order to format the results. In this callback we call `sortOutTranslations` for each row (i.e. sentence) and then return the complete result.
7. `sortOutTranslations`' job is
   - to flatten part of the result array (you will notice in `$this->contain` that indirect translations are nested inside the direct translations),
   - to filter out indirect translations which are also direct translations (E.g. if a sentence A has two direct translations B and C and C itself is also a direct translation of B it means that C is also an indirect translation of A. In that case we don't want that C appears twice, i.e. both as direct and as indirect translation)
   - to prevent that an indirect translation appears multiple times and
   - to filter the languages if needed.

Does this overview help you a little bit?
I would also recommend to read the [database chapter in the CakePHP cookbook](https://book.cakephp.org/4/en/orm.html).

trang:
@gioele-santi As agreed, I'm closing your pull request. Thanks for the attempt nonetheless! 🙂

PR #2500, opened 2020-08-08
Revert removing jQuery from layout.js

### LINKED ISSUE #2497: You can no longer right-click the audio player and download the MP3 file, or open the link into a new tab.
If this was purposely done for some reason, then this is not a bug.
However, it is inconvenient.

AndiPersti:
This PR closes #2497.

dcdab7f was to optimistic in the assumption that jQuery is only needed for displaying the full sentence block (with all the menu buttons). But we have still pages where we display a sentence in a minimal version of
the old design (just the sentence + the copy and audio button). These pages (e.g. all the audio pages) still require jQuery (and `watch.js`) so we need to load it even if we don't need `sentences-block-for-members.js`

There were attempts to fix the regressions caused by dcdab7f but they were all not succesful:
- Both 1df8c35 and 0b25a32 just caused further regressions
- c8cc8aa88fd62dc55d814fae1e16650f2635f966 was incomplete because the copy button in the old design still doesn't work on pages like `audios/index`, i.e. pages that display a minimum version of the sentence block.

In addition there is an undocumented regression on the `My favorites` page where the remove button doesn't work any more.

So all in all I think we still cannot remove jQuery from `layout.js` and I suggest to revert the changes.

### DIFF
--- a/config/asset_compress.ini
+++ b/config/asset_compress.ini
@@ -27,6 +27,7 @@ files[] = layouts/default.css
 files[] = layouts/elements.css
 
 [layout.js]
+files[] = jquery-1.11.3.min.js
 files[] = angular/angular.min.js
 files[] = angular/angular-animate.min.js
 files[] = angular/angular-aria.min.js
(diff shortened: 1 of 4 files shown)


jiru:
I agree that it’s best to revert the changes until we can come up with a fully working implementation. These front-end issues are hard to figure out because there is no test suite for that.

Thanks for the PR. :+1:

Issue #2273, opened 2020-04-12
Blank page no vocabulary items are found


jiru:
## Problem

When filtering by language, often the page is blank with no idication why it is blank.

https://tatoeba.org/eng/vocabulary/add_sentences/kor

![screen shot 2016-07-06 at 08 21 39](https://cloud.githubusercontent.com/assets/7934273/16602954/3866dd82-4353-11e6-9bf6-0eb29fbd71fa.png)

## Possible Solution

Put a message saying something like the following.

There are no requests for Korean.
There are no requests for this language.

There are no requests for Korean that do not already have at least 10 sentences.
There are no requests for this language that do not already have at least 10 sentences.

_Originally posted by @ckjpn in https://github.com/Tatoeba/tatoeba2/issues/1148#issuecomment-230632296_

speedysera:
Hi! Regarding the styling, would something like this be acceptable?

![tatoeba-message-no-requests](https://user-images.githubusercontent.com/43659877/85184506-fdd5b300-b2ca-11ea-8ae7-15896c4ea052.png)

ckjpn:
If you are referring to the thin light green text on a block of darker green, I find that a lot more difficult to read than the current black font on a white background as shown in the screenshot in the first post in this discussion.

I suspect this goes against accessibility guidelines that a lot of websites follow to help those with vision problems.

speedysera:
Thanks for the insight @ckjpn ! It's good to consider accessibility.

It seems the green header also exists in the case where there are sentences to display:

![tatoeba-message-header](https://user-images.githubusercontent.com/43659877/85186480-329a3800-b2d4-11ea-896c-1e8dbffd3aa1.png)

@trang are we going to change the display header anytime soon for accessibility? 


Regarding the styling question, I was wondering about this section in particular: 

![Screen Shot 2020-06-20 at 9 00 42](https://user-images.githubusercontent.com/43659877/85186528-8d339400-b2d4-11ea-9841-1a039c2647c1.png)

Do we want to display "There are no requests for this language that do not already have at least 10 sentences." along with the existing header as shown in the screenshots, or as a separate entry with different styling? 

I am open to suggestions :)

AndiPersti:
> Also, is there a function that would allow us to query the current language being featured, so that we can display "There are no requests for that do not already have at least 10 sentences."

In the controller we pass already `langFilter` which contains the 3-letter-code for the language and is currently not used in the template:
[...]

PR #2508, opened 2020-08-15
Blank page when no vocabulary items are found

speedysera:
This pull request addresses issue #2273 - when filtering by language, there is no indication why the page is blank when there are no vocabulary items found.

### DIFF
--- a/src/Template/Vocabulary/add_sentences.ctp
+++ b/src/Template/Vocabulary/add_sentences.ctp
@@ -27,9 +27,14 @@
 ?>
 <?php
 $this->Html->script('/js/vocabulary/add-sentences.ctrl.js', ['block' => 'scriptBottom']);
-
-$title = __('Vocabulary that needs sentences');
-
+if (empty($langFilter)) {
+    $title = __('Vocabulary that needs sentences');
+} else {
+    $title = format(
+        __('Vocabulary that needs sentences in {language}'),
+        array('language' => $this->Languages->codeToNameToFormat($langFilter))
+    );
+}
 $this->set('title_for_layout', $this->Pages->formatTitle($title));
 ?>
 
@@ -55,6 +60,12 @@ $this->set('title_for_layout', $this->Pages->formatTitle($title));
             )
             ?>
         </div>
+        <?php
+            if ($vocabulary->count() == 0) {
+                $noRequestsMessage = __("There are no requests.");
+                echo "<div class='empty-info-text'> $noRequestsMessage </div>";
+            }
+        ?>
 
         <?php
         $this->Pagination->display();
--- a/webroot/css/vocabulary/add_sentences.css
+++ b/webroot/css/vocabulary/add_sentences.css
@@ -42,3 +42,10 @@ md-input-container .md-errors-spacer {
 .new-sentences {
     margin-bottom: 5px;
 }
+
+.empty-info-text {
+    text-align: center;
+    font-size: 2em;
+    color: #666666;
+    margin: 50px 0 50px 0;
+}


src/Template/Vocabulary/add_sentences.ctp
```diff
@@ -55,6 +60,13 @@ $this->set('title_for_layout', $this->Pages->formatTitle($title));
             )
             ?>
         </div>
+        <div class="empty-info-text">
+            <?php
+                if ($vocabulary->count() == 0) {
+                    echo "There are no requests. ";
```
jiru, line 66: This string include an unnecessary space at the end. It should also made translatable using `__("your translatable string here")`, otherwise it will stay in English regardless of the UI language.

src/Template/Vocabulary/add_sentences.ctp
```diff
@@ -55,6 +60,13 @@ $this->set('title_for_layout', $this->Pages->formatTitle($title));
             )
             ?>
         </div>
+        <div class="empty-info-text">
```
jiru, line 63: When there is at least one request, this div will show up as empty, potentially creating a blank space. I think it would be better to make it *not show up at all* by putting it *inside* the `if()` branch.

webroot/css/layouts/default.css
```diff
@@ -795,4 +795,11 @@ html[dir="rtl"] .md-button-right input {
 html[dir="rtl"] .md-button-right .md-button {
     left: -8px;
     right: initial;
+}
+
+.empty-info-text {
+    text-align: center;
+    font-size: 2em;
+    color: #666666;
+    margin: 50px 0 50px 0;
```
jiru, line 804: `default.css` contains rules that are potentially applied on any page, such as style of re-usable blocks. As such, it is included on *every page* and we want to keep it free of unnecessary rules. Your rule `.empty-info-text` however is only used by the page `vocabulary/add_sentences`. For such rules, we use per-page CSS file named after the page name: `webroot/css/vocabulary/add_sentences.css`. You can move it there. :slightly_smiling_face:

jiru:
@speedysera Thanks for the pull request. I tested it and it works great! :+1: Just some implementation details need to be fixed. Have a look at my comments. :full_moon_with_face:

src/Template/Vocabulary/add_sentences.ctp
```diff
@@ -27,10 +27,15 @@
 ?>
 <?php
 $this->Html->script('/js/vocabulary/add-sentences.ctrl.js', ['block' => 'scriptBottom']);
-
-$title = __('Vocabulary that needs sentences');
-
-$this->set('title_for_layout', $this->Pages->formatTitle($title));
+    if (empty($langFilter)) {
+        $title = __('Vocabulary that needs sentences');
+        $this->set('title_for_layout', $this->Pages->formatTitle($title));
```
AndiPersti, line 32: This line should be moved after the if-block. Right now the page title is only set when there is no language filter.

And I suggest to remove the indentation for the whole if-block.

src/Template/Vocabulary/add_sentences.ctp
speedysera: thanks, learned something new about php today!

src/Template/Vocabulary/add_sentences.ctp
speedysera: good point!

src/Template/Vocabulary/add_sentences.ctp
speedysera: nice catch!

speedysera:
Addressed comments and closing this branch in favor of https://github.com/Tatoeba/tatoeba2/pull/2510
@jiru @AndiPersti please take a look when you have time

PR #2510, opened 2020-08-16
Blank page when no vocabulary items are found

speedysera:
This pull request addresses issue #2273 - when filtering by language, there is no indication why the page is blank when there are no vocabulary items found.

### DIFF
--- a/src/Template/Vocabulary/add_sentences.ctp
+++ b/src/Template/Vocabulary/add_sentences.ctp
@@ -27,9 +27,14 @@
 ?>
 <?php
 $this->Html->script('/js/vocabulary/add-sentences.ctrl.js', ['block' => 'scriptBottom']);
-
-$title = __('Vocabulary that needs sentences');
-
+if (empty($langFilter)) {
+    $title = __('Vocabulary that needs sentences');
+} else {
+    $title = format(
+        __('Vocabulary that needs sentences in {language}'),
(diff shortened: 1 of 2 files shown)


src/Template/Vocabulary/add_sentences.ctp
```diff
@@ -55,6 +60,12 @@ $this->set('title_for_layout', $this->Pages->formatTitle($title));
             )
             ?>
         </div>
+        <?php
+            if ($vocabulary->count() == 0) {
+                $noRequestsMessage = __("There are no requests.");
+                echo "<div class='empty-info-text'> $noRequestsMessage </div>";
+            }
+        ?>
```
jiru, line 68: This totally works, but here is an alternative way to put it:
```html
<?php if ($vocabulary->count() == 0): ?>
    <div class="empty-info-text">
        <?= __("There are no requests.") ?>
    </div>
<?php endif; ?>
```

jiru: Another, more CakePHP-ish way to do it, using the [Html helper](https://book.cakephp.org/3/en/views/helpers/html.html):
```php
if ($vocabulary->count() == 0) {
    echo $this->Html->div('empty-info-text', __('There are no requests.'));
}
```

jiru:
@speedysera Thanks again :slightly_smiling_face:

PR #2512, opened 2020-08-17
Concatenate multiline strings

AndiPersti:
Two multiline strings used "implicit" newlines instead of concatenating. As a consequence one of them appeared twice in `default.pot`.

If the intention really was to start a new line we would need to us the `br` tag because the newline character (i.e. `\n`) is ignored by browsers.

@jiru Since this change would require translators to unnecessarily "retranslate" the strings should I just remove the newline characters instead? (Some translators [kept](https://github.com/Tatoeba/tatoeba2/blob/0fe0273fcd5fee6d2be334e7e9708f534f9dd264/src/Locale/fr/default.po#L1222) the newline characters, some [removed](https://github.com/Tatoeba/tatoeba2/blob/0fe0273fcd5fee6d2be334e7e9708f534f9dd264/src/Locale/de/default.po#L1221) them.)

### DIFF
--- a/src/Template/Element/announcement.ctp
+++ b/src/Template/Element/announcement.ctp
@@ -48,9 +48,9 @@ if (!CurrentUser::hasAcceptedNewTermsOfUse()) {
     ?>
     <p>
     <?= format(
-        __('We have updated our <a href="{termsOfUse}">Terms of Use</a>.
-        By closing this announcement, you agree with the new Terms of Use.
-        If you have any question, feel free to <a href="{contact}">contact us</a>.'),
+        __('We have updated our <a href="{termsOfUse}">Terms of Use</a>. ' .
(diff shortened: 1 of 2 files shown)


jiru:
@AndiPersti If you want to remove the newlines yourself, you’d have to do it from Transifex, not the PO files, because they get overwritten by Transifex’s data. But do you have write permissions on every language on Transifex? Since translators will only have one string to retranslate, I think it’s reasonable to let them handle it. Transifex should find the old string from its translation memory as a near-100% match.

On a side note, in the past, when we moved from % placeholders to {} placeholders in #517, I wrote the script `docs/upload-pot.py` for this exact purpose: uploading PO files to Transifex while applying some regexps so that translators don’t have to retranslate just because of format changes. I’d be surprised if the script still works though. :smile:

PR #2513, opened 2020-08-17
Fix HTML in translation files

AndiPersti:
Some translation files contained HTML errors (e.g. unclosed tags, interrupting whitespace).

There is still one unclosed `a` tag in the Korean `default.po` but since I don't speak Korean I don't know where to put the closing tag:
https://github.com/Tatoeba/tatoeba2/blob/0fe0273fcd5fee6d2be334e7e9708f534f9dd264/src/Locale/ko/default.po#L631


kunigundeskugel:
I put the closing tag: #2514

jiru:
Thanks for your pull request, but it would be better to fix these strings from Transifex, because we regularly import the PO files from Transifex into this repository. Because of this, any change solely made from here would be overwritten next time we pull translations from Transifex.

PR #2516, opened 2020-08-18
Add language filter to Reviews page

### LINKED ISSUE #806: Allow users to "Filter by language" on the "reviews" page
img

AndiPersti:
This PR closes #806.

Like on other pages I've added the language filter module.

There was a small complication because `ReviewsController::of()` had two optional parameters (`$correctnessLabel` and `$lang`) and the language filter expects that the language is the last parameter. So `$correctnessLabel` always needs to be set (`reviews/of/user/eng` doesn't work).

### DIFF
--- a/src/Template/Reviews/of.ctp
+++ b/src/Template/Reviews/of.ctp
@@ -87,6 +87,8 @@ $this->set('title_for_layout', $this->Pages->formatTitle($title));
         }
         ?>
     </md-list>
+
+    <?php $this->CommonModules->createFilterByLangMod(3); ?>
 </div>
 <?php endif; ?>
 
--- a/src/Template/Reviews/of.ctp
+++ b/src/Template/Reviews/of.ctp
@@ -34,8 +34,19 @@ $categories = array(
     'outdated' => ['keyboard_arrow_right', __("Outdated reviews")]
 );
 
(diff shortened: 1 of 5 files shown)


src/Template/Reviews/of.ctp
```diff
@@ -34,8 +34,19 @@ $categories = array(
...
+    'ok' => __('Sentences in {language} marked as "OK"'),
+    'unsure' => __('Sentences in {language} marked as "unsure"'),
+    'not-ok' => __('Sentences in {language} marked as "not OK"'),
+    'all' => __('All sentences in {language}'),
+    'outdated' => __('Outdated reviews for sentences in {language}')
+];
+
+if ($lang) {
+    $category = format(
+        $categoriesWithLang[$correctnessLabel],
```
jiru, line 47: This produces an `Undefined index` notice if `$correctnessLabel` is set to an unexpected value (e.g. http://localhost:8080/fra/reviews/of/contributor/blah)

jiru: We should probably redirect to `all` directly from the controller if the `$correctnessLabel` parameter is invalid, instead of validating it in the view.

jiru:
Other than that, it looks good! :+1:

jiru:
Thank you, Andreas. :slightly_smiling_face:

Issue #458, opened 2014-10-15
Don’t lose contents when a POST request fails


jiru:
Currently, when you post a comment or a wall post, the contents of the message is lost when it can’t be send. It can be very frustrating to just loose what you take minutes or more to type in, so I set the priority to medium.

I know two scenarios when the message is lost.
1. The reply doesn’t come for some reason, like network problem or #459. When replying to a wall post, the loading.gif icon animates forever. The contents of the message is replaced by that icon.
2. Your session has expired or you logged yourself out. When replying a wall post, the contents of the message is replaced by a “your session has expired” message. When adding a comment to a sentence, or adding a new wall thread, you get redirected to the login page.

trang:
I think this is obsolete now. With the redesign of the Wall, the reply form remains displayed while the message is being saved. If the user sees that it's loading for too long, they will have the possibility to copy-paste the content.

jiru:
This issue is still actual for private messages.

trang:
> This issue is still actual for private messages.

Then also for comments and new Wall messages.

Possibly related: #293

PR #2531, opened 2020-08-28
Store unsubmitted form data in session storage

### LINKED ISSUE #293: Keep track of message contents even if you change from one page to another
wallebot says that other applications manage to remember the text you've typed into a message field even if you momentarily leave the page (for instance, by pressing the "Back" button) or are taken away from it (for instance, if you type the wrong message recipient), but Tatoeba doesn't.

AndiPersti:
This PR closes #293 and #458.

This approach uses the browser's session storage to save unsubmitted form data.

It adds an AngularJS service (`storageService`) that manages the session storage and an attribute directive (`resumable`; I'm open for a better name) that marks an input control as resumable, i.e. whenever the input control is shown, it is pre-filled with the last known value in the session storage. The session storage is updated after each change to the input control.

When the server succesfuly saved the data, the data is removed from the storage. For AJAX-calls this is done in the success callback. For traditional forms the server should send back a cookie named `submitted` whose value is the key for the session storage. Whenever the `storageService` is initialized, it looks for such a cookie, removes the key from the session storage and removes the cookie.

For a start, I made the Wall "resumable", i.e. the forms for adding a new message, editing a message and adding a response will keep their value.

If this approach is ok, I will make the other forms mentioned in the issues "resumable"

### DIFF
--- a/src/Template/Wall/index.ctp
+++ b/src/Template/Wall/index.ctp
@@ -38,6 +38,8 @@
 /* @translators: title of the Wall page */
 $this->set('title_for_layout', $this->Pages->formatTitle(__('Wall')));
 
+echo $this->Html->script('services/storage.srv.js', ['block' => 'scriptBottom']);
+echo $this->Html->script('directives/resumable.dir.js', ['block' => 'scriptBottom']);
 echo $this->Html->script('wall/wall.ctrl.js', ['block' => 'scriptBottom']);
 
 ?>
--- a/src/Template/Wall/index.ctp
+++ b/src/Template/Wall/index.ctp
@@ -35,7 +35,7 @@
  * @link     https://tatoeba.org
  */
 
-echo $this->AssetCompress->script('wall.js', ['block' => 'scriptBottom']);
+$this->AssetCompress->script('wall.js', ['block' => 'scriptBottom']);
 /* @translators: title of the Wall page */
 $this->set('title_for_layout', $this->Pages->formatTitle(__('Wall')));
 ?>
--- a/src/Template/Wall/index.ctp
+++ b/src/Template/Wall/index.ctp
@@ -35,13 +35,9 @@
  * @link     https://tatoeba.org
  */
 
+echo $this->AssetCompress->script('wall.js', ['block' => 'scriptBottom']);
 /* @translators: title of the Wall page */
 $this->set('title_for_layout', $this->Pages->formatTitle(__('Wall')));
-
-echo $this->Html->script('services/storage.srv.js', ['block' => 'scriptBottom']);
-echo $this->Html->script('directives/resumable.dir.js', ['block' => 'scriptBottom']);
-echo $this->Html->script('wall/wall.ctrl.js', ['block' => 'scriptBottom']);
-
 ?>
 
 <h2>
--- a/src/Template/Wall/show_message.ctp
+++ b/src/Template/Wall/show_message.ctp
@@ -35,13 +35,10 @@
  * @link     https://tatoeba.org
  */
 
+echo $this->AssetCompress->script('wall.js', ['block' => 'scriptBottom']);
 $this->set('title_for_layout', $this->Pages->formatTitle(
     format(__('Thread #{number}'), array('number' => $message->id))
 ));
-
-echo $this->Html->script('services/storage.srv.js', ['block' => 'scriptBottom']);
-echo $this->Html->script('directives/resumable.dir.js', ['block' => 'scriptBottom']);
-echo $this->Html->script('wall/wall.ctrl.js', ['block' => 'scriptBottom']);
 ?>
 <div id="annexe_content">
     <div class="module">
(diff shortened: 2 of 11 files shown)


jiru:
Good job dealing with these issues. :+1: The approach looks reasonable.

I’m concerned however about leaving data in the session storage unattended. What if the data is not removed from the session storage for some reason? It could have severe privacy implications if the device is shared: an unrelated user could be presented with pre-filled data from other users using the same browser even after switching Tatoeba accounts. We could also be just bloating the session storage with every field the user filled in but eventually decided not to send. It could also be just very annoying for a user to see a field pre-filled with an old message he or she intended to discard by closing the tab.

src/Template/Wall/show_message.ctp
```diff
@@ -35,13 +35,10 @@
  * @link     https://tatoeba.org
  */
 
+echo $this->AssetCompress->script('wall.js', ['block' => 'scriptBottom']);
```
jiru, line 38: No need to `echo` here because the script is stored in the `scriptBottom` block instead of being returned.

src/Template/Wall/index.ctp
```diff
@@ -35,13 +35,9 @@
  * @link     https://tatoeba.org
  */
 
+echo $this->AssetCompress->script('wall.js', ['block' => 'scriptBottom']);
```
jiru, line 38: No need to `echo` here because the script is stored in the `scriptBottom` block instead of being returned.

AndiPersti:
> I’m concerned however about leaving data in the session storage unattended. What if the data is not removed from the session storage for some reason? It could have severe privacy implications if the device is shared: an unrelated user could be presented with pre-filled data from other users using the same browser even after switching Tatoeba accounts.

I've thought about some of these issues, but I don't think there is a high risk.
The data in `window.sessionStorage` is only valid for a single browser window/tab (in contrast to [`window.localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#Basic_concepts)) and closing it will clean the storage (unless there is a bug in the browser which is very unlikely for common browsers). I suppose any privacy-aware user would always close the browser on a shared device before handling it to someone else.
And since the data in the session storage is synced to the current value in the textbox, even just clearing the text box will remove it from the session storage. So this would only be a problem for "forgotten" text fields.

> We could also be just bloating the session storage with every field the user filled in but eventually decided not to send.

Possible enhancements:
- Add an expiration date for the data. But then the question is how long should it be stored and how to communicate that period to the user?
- Clear the storage when the user logs out.

> It could also be just very annoying for a user to see a field pre-filled with an old message he or she intended to discard by closing the tab.

As explained above, closing a browser tab will clean the storage.
But there is a somehow related issue I've noticed while implementing the changes for the wall. Currently if I open the reply form for a wall message, type something into the textbox, click on `Cancel` and then reopen the same reply form again, the old text is still there. I've not changed that behavior yet (and I've copied this behavior to the edit form, i.e. if you cancel an edit message and reopen the page for editing the same message, the form will be pre-filled from the session storage). But I would expect that I get a blank form if after clicking on `Cancel` I reopen the same reply form.

At the end I think it is a compromise between security and usability. We need to somehow store the data on the client in order to prevent any data loss when the transmission to the server fails

jiru:
> Clear the storage when the user logs out.

This sounds like a must-wanted enhancement to prevent this kind of attack:

1. User A logs-in and types in a private message.
2. She decides not to send it and click the logout button. She is therefore brought to the login page.
3. The device is then left unattended while the tab is still open.
4. Since all of the above happened in the same tab, an attacker with physical access to the device can easily access the unsent message by re-opening the same page under her own account or directly from the browser console.

> Add an expiration date for the data. But then the question is how long should it be stored and how to communicate that period to the user?

Well, since you mentioned that the data is cleared when the tab is closed, then there is little chance that this would bloat the sessionStorage. The user has "control" over the stored data by leaving or closing the tab, I think that’s better than a timeout.

> But there is a somehow related issue I've noticed while implementing the changes for the wall. Currently if I open the reply form for a wall message, type something into the textbox, click on Cancel and then reopen the same reply form again, the old text is still there. I've not changed that behavior yet (and I've copied this behavior to the edit form, i.e. if you cancel an edit message and reopen the page for editing the same message, the form will be pre-filled from the session storage). But I would expect that I get a blank form if after clicking on Cancel I reopen the same reply form.

I think the benefits (potentially recovering a lost message) outweigh the trouble (having to clear the field before entering a new message).

AndiPersti:
>     2. She decides not to send it and click the logout button. She is therefore brought to the login page.

But clearing the storage should only happen if I actively log out by clicking the button, right?
If I get logged out due to session expiration, I should still be able to recover the unsubmitted data, shouldn't I?

jiru:
Yes.

AndiPersti:
Closing because I need to redesign/rewrite most parts.

PR #2532, opened 2020-08-30
Replace most language selection dropdowns with the one supporting autocompletion (fixes #2328)

### LINKED ISSUE #2328: The language selection dropdown on some pages is not searchable
The source and target languages in the search bar can be selected by typing part of the language's name and choosing from a filtered list. However that is not the case for the advanced search and a few other pages I checked (e.g. the "browse by language" page), so it's necessary to scroll through the full list to find a language that's not part of the profile languages.

Yorwba:
The commits in this PR come in two parts: one big rewrite of the `language_dropdown` element to use a template for the directive, which allows using isolated scope and accessing the selected language via bindings instead of the previous implementation using events. The rest is just mechanically replacing a bunch of plain `select`s with the `language_dropdown` element.

I didn't change the `select` elements in `src/View/Helper/SentencesHelper.php`, `src/View/Helper/SentenceButtonsHelper.php` and `src/Template/Element/sentences/add_sentences_jquery.ctp` because the first two are for setting the language of a sentence/translation where autocomplete seems like overkill and the last one has `jquery` in the name, so I guess I'm not supposed to stick AngularJS in it.

Many of the `select`s I replaced were using `onchange` to set `window.location.href` and reload the page when a different language is selected. AngularJS strongly discourages global objects and evaluates expressions in a scope where `window` is unavailable. I worked around that by passing window to the expression given in `on-selected-language-change`, which is probably not the proper AngularJS way of doing things, but I don't know what the proper way is supposed to be.

Edit: GitHub doesn't seem to do anything with issue numbers in PR titles, so repeating it here: This PR fixes #2328.


jiru:
Well done! :+1: The code looks good, but UX-wise the dropdown field is now appearing with some delay and pushes around other elements during page load. For example, on sentence pages, the page first display with previous/random/next buttons taking the whole space, and only after the page is fully loaded, they shrink down when the language dropdown appears.

Yorwba:
I'm not sure what the best practice for that is in AngularJS (I only read barely enough of the documentation to get this PR done) but I guess embedding the template as a string literal instead of using `templateURL` should eliminate the delay?

jiru:
I’m not sure if that’s the way to go, but you can try. Have a look at what Andreas did for the reviews in c10821f7cfd87316868aa3e8664647537fec1108.

Yorwba:
Thanks or pointing me at Andreas' commit, it was a good example to learn from.

I prevented the creation of multiple identical copies of the template by checking whether the ID already exists in the `scriptBottom` block before appending to it; I'm not sure whether that isn't a bit too hackish.

jiru:
It’s a bit hackish indeed. Maybe you can create a helper function/class that takes care of adding angular templates. It should keep track of which block has been added so far based on the `id` attribute of the element, and call `scriptBlock()` only once.

Yorwba:
I added a new helper to take care of deduplication for AngularJS templates and converted the existing script blocks to use the helper instead. (The template for editing reviews was also being included multiple times.)

trang:
There's one issue regarding the on-select behavior. It will reload the page even if not selecting anything. For instance:
- Go "Browse by language".
- Click on the language selector but don't select any language.
- Click outside of the language selector to remove the dropdown list.
- The page is reloaded.

There's another issue that is not very critical but would be good to solve before merging this. Ever since we upgraded to a more recent version of AngularJS Material, the language selector looks different after interacting with it. The border disappears and the X button gets a grey background and border.

Screenshot_2020-09-08 Advanced search - Tatoeba(2)

I'm not sure why this happens though. It's not too weird in the search bar because the search bar has a background color which compensates for the disappearance of the selector's border. But when the background is white, it looks quite strange.

jiru:
By the way, #2294 is likely to create a conflict with your branch because of the way it sets `md-min-length`. Do you mind if I merge it first and let you take care of adapting your code?

Yorwba:
@jiru Sure, go ahead, I'll take care of any conflicts.

@trang I'll add a check to prevent reloading when the selected language isn't changed.

Regarding the appearance of the button, I noticed this a few weeks ago with the search bar on prod. Here's how it looks in my Firefox on Linux:
Unstyled button in search bar
I originally planned to fix that as part of this PR, but then the problem didn't happen while testing locally, so I forgot about it.

The reason why the button ends up looking like that is that the CSS relies on the class `.md-show-clear-button`, which is removed when the button is hidden while the dropdown is active, but doesn't get added back after selecting a language. Since adding and removing that class is handled by AngularJS Material, I thought it was a bug on their side. It looks fixed on dev with AngularJS Material 1.1.12 (instead of 1.1.5 on prod) and I thought that's also why I didn't see it locally, but just now I discovered that it seems related to asset compression:

If I run
```bash
bin/cake asset_compress build -f && sudo systemctl reload php7.0-fpm
```
in imouto, the button appears buggy, but if I delete all the generated files
```bash
bin/cake asset_compress build -f && rm webroot/cache_js/*.js* && sudo systemctl reload php7.0-fpm
```
the bug is fixed. Very strange. There are some other problems with asset compression for which I'll open a new issue.

Yorwba:
I added the check to prevent reloading when the selected language isn't changed.

Judging by [this comment](https://github.com/Tatoeba/tatoeba2/issues/2549#issuecomment-689719980), the button styling should be fixed by `0b7ee77` once #2294 is merged. (I haven't tested this yet, because it looks like I need to update my VM to install the updated dependencies required for that branch.)

jiru:
> (I haven't tested this yet, because it looks like I need to update my VM to install the updated dependencies required for that branch.)

I think you only need to run `composer install` after switching to/from that branch.

Yorwba:
> I think you only need to run `composer install` after switching to/from that branch.

I tried that, but most of the dependencies require a more up-to-date PHP. The one in my VM is version `7.0.33`.

trang:
@Yorwba I guess you would need to `vagrant box update` but this will re-create the VM entirely, so make sure to backup anything you need to restore after the new VM is created (cf. https://github.com/orgs/Tatoeba/teams/kodoeba/discussions/14).

AndiPersti:
> Regarding the appearance of the button, I noticed this a few weeks ago with the search bar on prod. Here's how it looks in my Firefox on Linux:

I can not reproduce this any more after rebasing your branch on the current `dev`. So I'm pretty sure https://github.com/Tatoeba/tatoeba2/commit/0b7ee77af87236d6812ba03ecb0da5abe4633d36 fixed it this too.

src/Template/Element/language_dropdown_angular.ctp
```diff
@@ -0,0 +1,28 @@
+<div class="language-dropdown-container"
+     title="{{selectedLanguage.name}}">
```
AndiPersti, line 2: This title attribute causes two tooltips on the language dropdown in the navigation bar:
tooltip

I don't think that the title is necessary here, because it only repeats the current value of the dropdown.

But after removing it there is still a problem because the tooltip for the navigation bar will hide the first entry of the dropdown after selecting it and not moving the mouse. So I would suggest to add `md-direction='top'` to the `md-tooltip` directive in `src/Template/Element/sentences/navigation.ctp`:
https://github.com/Tatoeba/tatoeba2/blob/98269d3af50a4b45e0c75456307bc13cbb5a9a7e/src/Template/Element/sentences/navigation.ctp#L41-L43

Yorwba: Good suggestion. I implemented it.

src/Template/Element/language_dropdown_angular.ctp
```diff
@@ -1,5 +1,4 @@
 <div class="language-dropdown-container"
-     title="{{selectedLanguage.name}}">
```
AndiPersti: You've also removed the closing `>` for the `div` tag which breaks the HTML code.

Yorwba: Thanks for pointing that out. My editor actually highlighted the following `<input` in red, but somehow I failed to notice that...

PR #2534, opened 2020-08-30
Properly separate paramerters by slashes in createFilterByLangMod

Yorwba:
While working on #2532, I noticed that the language filter on `sentences/of_user/<user>/<lang>` was actually broken, because when constructing the URL for a new language, the slashes were put after the parameters like so: `sentences/of_user<user>//<lang>`.


jiru:
Sorry, I introduced this regression! :disappointed:
Thanks for fixing it. :smile:

PR #2535, opened 2020-08-31
Fix wrong license for translations

### LINKED ISSUE #2533: Some translations are wrongly licensed as CC0
**To Reproduce**

A user who recently got granted the permissions to use the license CC0 reported that their translations were set with the license CC0.

Example: https://tatoeba.org/eng/sentences/show/9004276

I haven't checked the exact steps to reproduce but I suspect this is happening only with the new sentence design.

**Expected behavior**

In principle, a translation of a CC BY sentence should not be licensed as CC0. But also, we have not yet enabled CC0 on translations so every translation should be created under CC BY.

**Additional information (if relevant)**

Related bug: #2322.

It seems we have 5700+ translations that shouldn't be CC0.
```
MariaDB [tatoeba]> select count(*) from sentences where license = 'CC0 1.0' and based_on_id > 0;
+----------+
| count(*) |
+----------+
|     5705 |
+----------+
1 row in set (6.065 sec)
```

AndiPersti:
This PR closes #2533.

The license for a translation was never explicitly set and so it was set in `SentencesTable::beforeSave` to the user's default license which in the past was always `CC BY`.

But c5191c87e124079dfb3b27fe2989029ba27c3f72 enabled users to change their default license in the settings. So when a user changed their default license to `CC0`, the license of a translation was always incorrectly set to that license.

In addition to setting the license explicitly in `SentencesTable::saveTranslation` I've also added a check in `SentencesTable::beforeSave`.

### DIFF
--- a/src/Model/Table/SentencesTable.php
+++ b/src/Model/Table/SentencesTable.php
@@ -191,7 +191,7 @@ class SentencesTable extends Table
                 $user = $this->Users->get($entity->user_id);
                 if ($user) {
                     $userDefaultLicense = $user->settings['default_license'];
-                    $entity->license = $entity->based_on_id == 0 ?
+                    $entity->license = $entity->based_on_id === 0 ?
                                        $userDefaultLicense :
                                        'CC BY 2.0 FR';
                 }
--- a/src/Model/Table/SentencesTable.php
+++ b/src/Model/Table/SentencesTable.php
@@ -191,7 +191,9 @@ class SentencesTable extends Table
                 $user = $this->Users->get($entity->user_id);
                 if ($user) {
                     $userDefaultLicense = $user->settings['default_license'];
-                    $entity->license = $userDefaultLicense;
+                    $entity->license = $entity->based_on_id == 0 ?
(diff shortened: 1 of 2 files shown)


src/Model/Table/SentencesTable.php
```diff
@@ -191,7 +191,9 @@ public function beforeSave($event, $entity, $options)
                 $user = $this->Users->get($entity->user_id);
                 if ($user) {
                     $userDefaultLicense = $user->settings['default_license'];
-                    $entity->license = $userDefaultLicense;
+                    $entity->license = $entity->based_on_id == 0 ?
+                                       $userDefaultLicense :
+                                       'CC BY 2.0 FR';
```
jiru, line 196: If `$entity->based_on_id` is null, this will set the license to `$userDefaultLicense`. It’s not supposed to happen on sentence creation, but who knows. Could you change this into triple equal just in case?

AndiPersti: Sure.

jiru:
Thank you very much! :smile:

PR #2548, opened 2020-09-08
Improve sphinx shell

AndiPersti:
I've noticed that the delta index update shell always tries to update all languages that were changed since the last merge even if there was no change in the last 15 minutes (i.e. since the delta update before the current one). Last sunday evening (CEST) the script even ran longer than 15 minutes, so that the next scheduled run had no chance to do its work.
Similar, the merging also tried to merge the indexes for every language, even for languages that had no change in the last 24 hours.

While here, I've also changed some more stuff, that I found a little bit annoying:
- The old version changed the user very early and run everything as `manticore`. This caused problems with file permissions in some directories (see also https://github.com/Tatoeba/imouto/issues/71). Running only the `indexer` command as `manticore` (i.e. `sudo -u manticore indexer ...`) on the one hand avoids these problems and on the other hand allows non-root users to execute the shell as long as they have the `sudo` permissions (e.g. in the dev environment, the `vagrant` user can now run the shell commands).
- The output/logs from the cron jobs weren't really useful (just a sequence of timing results in the `/tmp` directory without further information). After the changes get merged I would redirect the output into files in the `/var/log` directory on the servers and put them under logrotation as we do with most of the other log files.
- All our CakePHP CLI utilities run with the default locale `en_US` but in the Web UI we use as default locale `en`. Thus whenever a CLI tool would have to use cached translation files it would create new ones although the cached translations for `en` would have been already there. As a solution I propose to set the locale to `en` for CLI utilities too.

### DIFF
--- a/config/bootstrap_cli.php
+++ b/config/bootstrap_cli.php
@@ -14,6 +14,7 @@
  */
 use Cake\Core\Configure;
 use Cake\Core\Exception\MissingPluginException;
+use Cake\I18n\I18n;
 use Cake\Core\Plugin;
 
 /**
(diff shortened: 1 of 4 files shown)


jiru:
What great improvements! All the changes look fine to me. :+1:

Issue #2544, opened 2020-09-05
Initial UI language does not honor browser settings properly


jiru:
1. Configure your browser preferred languages to "fr-FR".
2. Open a new private browsing session.
3. Go to https://tatoeba.org
4. Bug: Tatoeba displays in English.

Alternatively, using the command line:

```
$ curl -s -D - 'https://tatoeba.org/' -H 'Accept-Language: fr-FR' | grep -i ^location:
location: https://tatoeba.org/eng/
```

jiru:
In a addition, if the user first arrives on Tatoeba on a page that contains a URL code (that is to say, anything but https://tatoeba.org/), then the browser language settings are not honored:

```shell
$ curl -s -D - 'https://tatoeba.org/eng/' -H 'Accept-Language: fr' | grep -i ^location:
# No redirection!
$ 
```

This explains why in [UX test 1](https://en.wiki.tatoeba.org/articles/show/ux-test-1#exploration) the user ended up using Tatoeba in English even though she had proper browser language settings. She looked up "Tatoeba" on Google and clicked on the first link, https://tatoeba.org/eng. She ended up on the English homepage. Then, she went on using the website in English because she understands it, however the test proven that using English introduced problems for her in understanding some words.

AndiPersti:
I've noticed that ignoring the language code in the URL path has a side-effect on webcrawlers.

For example here are the number of responses with status code 200 and 301 for `Googlebot`:

date|200|301
-|-|-
27/Sep/2020| 68,710|  82,423
26/Sep/2020|  61,806 |  74,301
25/Sep/2020|  64,113|  85,712
24/Sep/2020| 76,429 | 89,408
23/Sep/2020| 63,881 | 76,820
22/Sep/2020| 82,687 | 84,816
21/Sep/2020| 98,305 | 67,047
20/Sep/2020| 96,138 | 9,090
19/Sep/2020|132,951| 7,403
18/Sep/2020|114,733| 2,269
17/Sep/2020| 76,451|1,922
16/Sep/2020|116,342| 4,223

As you can see the number of 301 responses have risen sharply since we introduced this change[*]. And I guess in the long term this will hurt the ranking for the localized pages in search results (see #507).

[*] `Googlebot`'s `Accept-Language` header seems to be set to `en` because all pages get redirected to the `eng` prefix. On the other hand `YandexBot` is obviously configured to act as a Russian client, because all its requests are redirected to the `rus` prefix.

[...]

PR #2553, opened 2020-09-14
Fix initial language redirection

jiru:
This pull request solves #2544.


src/Controller/AppController.php
```diff
@@ -323,12 +317,14 @@ public function getSupportedLanguage()
...
 
-        if (isset($_SERVER['HTTP_ACCEPT_LANGUAGE'])) {
+        $header = $this->request->getHeaderLine('Accept-Language');
+        if (!empty($header)) {
 
-            $browserLanguages = explode(',', $_SERVER['HTTP_ACCEPT_LANGUAGE']);
+            $browserLanguages = explode(',', $header);
 
             foreach ($browserLanguages as $browserLang) {
                 $browserLangArray = explode(';', $browserLang);
```
AndiPersti, line 326: There's also [`Cake\Http\ServerRequest::acceptLanguage()`](https://api.cakephp.org/3.7/class-Cake.Http.ServerRequest.html#_acceptLanguage).

For a header line like `Accept-Language: en,de-AT;q=0.8,en-US;q=0.5,fr-FR;q=0.3` it would return
```
2020-09-18 13:16:43 Debug: Array
(
    [0] => en
    [1] => de-at
    [2] => en-us
    [3] => fr-fr
)
```

src/Controller/AppController.php
```diff
@@ -153,14 +152,12 @@ public function beforeFilter(Event $event)
...
-                $this->Cookie->write('CakeCookie.interfaceLanguage', $langInCookieAlias, false, "+1 month");
                 $langInCookie = $langInCookieAlias;
             }
             $lang = $langInCookie;
-        } else if (!empty($langInURL)) {
-            $lang = $langInURL;
-            $this->Cookie->write('CakeCookie.interfaceLanguage', $lang, false, "+1 month");
         }
+
+        $this->Cookie->write('CakeCookie.interfaceLanguage', $lang, false, "+1 month");
```
AndiPersti, line 160: Is the `CakeCookie` namespace really necessary?

CakePHP will send this cookie as `CakeCookie={"interfaceLanguage":"eng"}`, i.e. the browser stores it with the key `CakeCookie` and a JSON object as the value.

But on the Javascript side (when using the language chooser dialog) we [set the cookie as `CakeCookie[interfaceLanguage]=eng`](https://github.com/Tatoeba/tatoeba2/blob/98269d3af50a4b45e0c75456307bc13cbb5a9a7e/webroot/js/responsive/app.module.js#L101), i.e. the browser stores it as another cookie although CakePHP recognizes it for the same cookie (and even prefers this version).

But this is the only cookie for which we use the `CakeCookie` namespace so I wonder whether we could simply just use `interfaceLanguage` as the cookie key and thus avoid having two cookies stored for the same value.

jiru: For historical reasons, but also because the namespace CakeCookie is configured as unencrypted. I think we should get rid of that JS code that sets the cookie as `CakeCookie[interfaceLanguage]=eng`, and have it set by the server using an action. It would avoid having to maintain two implementations to set that cookie.

src/Controller/AppController.php
jiru: Right! :smile:

src/Controller/AppController.php
jiru: And it would allow to saving the UI language as a user setting. Anyway I think this is all well out of the scope of this pull request. :smile:

AndiPersti: > Anyway I think this is all well out of the scope of this pull request. smile

Agreed. I was just reminded of this duplication when I've reviewed the code.

PR #2564, opened 2020-09-26
Rewrite tag deletion

### LINKED ISSUE #2563: Sentences with tags still appear in custom queries even the tag is removed
As reported by Pfirsichbaeumchen on the wall:

> I removed a large number of obsolete "@needs native check" tags, but the sentences I removed them from are still found when I search for sentences with that tag even though the tag is gone.
> 
> Is this a bug, or is it just taking unusually long for the server to know that the tags are gone? 

https://tatoeba.org/ita/wall/show_message/35983#!#message_35983

AndiPersti:
This PR closes #2563.

The problem was that the table method for removing a tag used `deleteAll()` which doesn't trigger the `afterDelete` event. But we use this event to update the attributes of the search index.

### DIFF
--- a/src/Model/Table/TagsSentencesTable.php
+++ b/src/Model/Table/TagsSentencesTable.php
@@ -85,11 +85,23 @@ class TagsSentencesTable extends Table
             ->toList();
     }
 
-    public function removeTagFromSentence($tagId,$sentenceId) {
-        $this->deleteAll([
-            'tag_id' => $tagId,
-            'sentence_id' => $sentenceId
-        ]);
+    /**
+     * Remove a tag from a sentence
+     *
+     * @param int $tagId       Id of the tag
+     * @param int $sentenceId  Id of the sentence
+     *
+     * @return boolean
+     */
+    public function removeTagFromSentence($tagId, $sentenceId) {
+        $entity = $this->find()
+                       ->where(['tag_id' => $tagId, 'sentence_id' => $sentenceId])
+                       ->first();
+        if ($entity) {
+            return $this->delete($entity);
+        } else {
+            return false;
+        }
     }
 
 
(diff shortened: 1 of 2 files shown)


src/Model/Table/TagsSentencesTable.php
```diff
@@ -85,11 +85,23 @@ public function getAllTagsOnSentence($sentenceId)
...
+     *
+     * @param int $tagId       Id of the tag
+     * @param int $sentenceId  Id of the sentence
+     *
+     * @return boolean
+     */
+    public function removeTagFromSentence($tagId, $sentenceId) {
+        $entity = $this->find()
+                       ->where(['tag_id' => $tagId, 'sentence_id' => $sentenceId])
+                       ->first();
```
jiru, line 99: This won’t work for tags that have been applied more than once on the same sentence. As far as I remember, this only happens when Horus merges two sentences having the same tag. See by yourself:
```sql
select count(*) as c, tag_id, sentence_id from tags_sentences group by tag_id, sentence_id having c > 1;
+---+--------+-------------+
| c | tag_id | sentence_id |
+---+--------+-------------+
| 2 |      2 |      482049 |
| 2 |      5 |     2337281 |
| 2 |      5 |     3214286 |
...
+---+--------+-------------+
4024 rows in set (1.481 sec)
```

jiru: Just to explain this mess. At the time Horus was created, we were aware of these "duplicate" tags and we (well, I) chose to adapt the way tags are displayed (in the sentence page and in the "sentences tagged as x" listing) to hide these duplicates (as opposed to deleting them) in order not to loose the information that two users had tagged that sentence. Looking back, I think it was a mistake because it produced a number of new bugs and added no real value in the end.

AndiPersti: Thanks for reminding me of this Horus bug.

PR #2565, opened 2020-09-28
Fix language count on "Browse by language" page

AndiPersti:
The language count on the "Browse by language" page was too high because it used the recursive mode of `count()` which also includes the elements on a higher level if the array is a multidimensional array, i.e.
`count([[1, 2], [3, 4]], COUNT_RECURSIVE)` will return 6 and not 4.


jiru:
It looks alright. :slightly_smiling_face:

PR #2566, opened 2020-09-28
Fix warnings in debug log

AndiPersti:
This PR fixes some `Trying to get property '...' of non-object` warnings in the debug log due to sentence comment or wall message authors who do not exist any more in the database.

For comments, the non-existing username wasn't displayed at all, e.g. https://tatoeba.org/eng/sentences/show/1998393#comment-213608:
comment

For wall messages, the current username is displayed for the non-existing user, e.g. https://tatoeba.org/eng/wall/show_message/8319#message_8319:
wall

In both cases the icon for sending a PM was also displayed.

My proposed solution is to remove the PM icon and to display `Former member` for the username but I'm open for other suggestions:
new

### DIFF
--- a/src/Template/Element/sentence_comments/comment.ctp
+++ b/src/Template/Element/sentence_comments/comment.ctp
@@ -117,7 +117,7 @@ if ($sentenceOwnerLink) {
                 <a href="<?= $userProfileUrl ?>"><?= $username ?></a>
             </span>
         <?php else: ?>
-            <i><?= __('Former member') ?></i>
+            <i><?= h(__('Former member')) ?></i>
         <?php endif; ?>
             <span class="md-subhead ellipsis">
                 <?= $dateLabel ?>
--- a/src/Template/Element/sentence_comments/comment.ctp
+++ b/src/Template/Element/sentence_comments/comment.ctp
@@ -2,8 +2,15 @@
 use App\Lib\LanguagesLib;
 use App\Model\CurrentUser;
 
-$username = $comment->user->username;
-$avatar = $comment->user->image;
+$username = $comment->user->username ?? null;
+if ($username) {
+    $userProfileUrl = $this->Url->build(array(
+        'controller' => 'user',
+        'action' => 'profile',
+        $username
+    ));
+}
+$avatar = $comment->user->image ?? null;
 $createdDate = $comment->created;
 $modifiedDate = $comment->modified;
 $commentId = $comment->id;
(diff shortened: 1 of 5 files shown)


trang:
> My proposed solution is to remove the PM icon and to display `Former member` for the username

I think this solution is good.

jiru:
Right, this is exactly what we used to do for sentence comments since ee0a0cfa2a1c1795ebcf48c28dcfd7a2d3865afe and until 319a880d24dbe46ecfc8a5a3688fc533f81efcb7 (and until later on for other comment-like things like private messages in 684c005f39d721720f75bbd0d1653789273a2784).

src/Template/Element/sentence_comments/comment.ctp
```diff
@@ -110,9 +112,13 @@ if ($sentenceOwnerLink) {
             <?= $this->Members->image($username, $avatar, array('class' => 'md-user-avatar')); ?>
         </md-card-avatar>
         <md-card-header-text>
+        <?php if ($username): ?>
             <span class="md-title">
                 <a href="<?= $userProfileUrl ?>"><?= $username ?></a>
             </span>
+        <?php else: ?>
+            <i><?= __('Former member') ?></i>
```
jiru, line 120:
```suggestion
            <i><?= h(__('Former member')) ?></i>
```

src/Template/Element/wall/message.ctp
```diff
@@ -60,9 +62,13 @@ $canReply = false;
             <?= $this->Members->image($username, $avatar, array('class' => 'md-user-avatar')); ?>
         </md-card-avatar>
         <md-card-header-text>
+        <?php if ($username): ?>
             <span class="md-title">
                 <a href="<?= $userProfileUrl ?>"><?= $username ?></a>
             </span>
+        <?php else: ?>
+            <i><?= __('Former member') ?></i>
```
jiru, line 70:
```suggestion
            <i><?= h(__('Former member')) ?></i>
```

src/Template/Element/sentence_comments/comment.ctp
AndiPersti: Do you suggest to add `h()` to every translatable string in general?

jiru: I think we should do so. Of course it’s not always required as sometimes the function you give the translatable string to escapes it for you (such as methods in HtmlHelper). Until we go through all the strings, I try to escape new ones at least.

Issue #2567, opened 2020-09-29
Pop-up menus don't scroll vertically on mobile


Yorwba:
**To Reproduce**
1. Perform a search using a mobile browser.
2. Click on "Narrow Search" to open the "Search critera" sidebar.
3. Attempt to scroll down.
4. Page doesn't move.

The same happens with the hamburger menu in the top-left corner. Horizontal scrolling appears unaffected.

**Expected behavior**
It should be possible to reach parts of the pop-up menus that are initially below the fold.

**Additional information**
- O.S.: Android 10
- Browsers: Firefox 81.1.1, Chrome 85.0.4183.127

trang:
Thanks for creating this bug report, @Yorwba! I noticed this too the other day, when checking #2561.

I looked further into it and found out that this happens because of `height: auto` on the `body`, introduced in e66fafcf6a095ac32821e569f4d9ea7718603d54. This `height: auto` makes the height of the sidenav extend to the whole content of the page rather than just limit it to height of the browser, that's why there is no more scroll.

I guess one solution would be to add media query so that we set `height: auto` only for width>960px.

PR #2568, opened 2020-10-04
Make sidenav scrollable again on mobile

trang:
This pull request solves #2567.

The issue happens because the `body` has the CSS rule `height: auto`. This rule was introduced because opening the autocomplete dropdown would remove the vertical scrollbar and therefore displace the content of the page. Having `height: auto` prevents the scrollbar from disappearing when using the autocomplete suggestions but has the side effect of expanding the height of the sidenavs, making them no longer scrollable.

The fix in this PR is to set `height: auto` on the body only when the screen width is 960px or more, which is the breakpoint for desktop mode. The scrollbar disappearing is not an issue on mobile so not having `height: auto` for smaller screens should be okay.

### DIFF
--- a/webroot/css/layouts/default.css
+++ b/webroot/css/layouts/default.css
@@ -119,8 +119,10 @@ html, body {
     min-width: 100%;
 }
 
-body {
-    height: auto;
+@media only screen and (min-width: 960px) {
+    body {
+        height: auto;
+    }
 }
 
 textarea {


jiru:
> The scrollbar disappearing is not an issue on mobile

Are you sure? What about mobile devices having a screen larger than 960px?

trang:
The scrollbar disappearing is not an issue on mobile because mobile scrollbars are designed differently from desktop. They don't take any space, they just float over the content. So even if they disappear when opening the autocomplete dropdown, it doesn't create this annoying "jumping" effect that happens on desktop.

If a mobile device has a screen larger than 960px, then the current change will not have any effect. The website will be displayed like it is displayed on desktop. There will be `height: auto` on the body, which makes the content of the sidenav non-scrollable, but there will be no need to have the sidenav content scrollable anyway because the sidenav will be displayed in "desktop mode". It will be integrated in the page content instead of being some modal popup that shows up on top of the content. And since the content is scrollable, that's okay.

On the other hand, if a desktop user has a screen smaller than 960px then the current change will actually bring back the "jumping" effect when interacting with the autocomplete because `height: auto` will not be set on the body. But this case is rare enough that I think it's okay if we don't provide an optimal experience for desktop users with very small screens.

PR #2569, opened 2020-10-05
Clean messages helper

jiru:
This is a pure code-cleaning PR with no functional change.

We have a helper called MessagesHelper for historical reasons. Private messages, sentence comments and Wall messages used to share the same rendering functions. However, with the introduction of the new design, Trang decided to break these apart. But MessagesHelper still had leftovers from this past era. The PR removes methods that are no longer used and moves methods specific to Wall and PM into their own helpers.


AndiPersti:
Why is this PR marked as a draft? Do you plan to add more commits to it?

jiru:
@AndiPersti Yes. It is now complete. I updated the description.

AndiPersti:
Looks good to me.

Issue #2570, opened 2020-10-13
Some final adjustments for the Kurdish languages


soliloquist-tatoeba:
I'm creating this issue upon [the request](https://github.com/Tatoeba/tatoeba2/issues/2540#issuecomment-707509680) of Trang.

1- The qualifier could be removed from Southern Kurdish. The name of Xwarîn seems to be unknown to even some Kurds. Unlike Northern and Central Kurdish, Southern Kurdish probably doesn't have a standardized name. Other names like '[Feyli](https://www.youtube.com/watch?v=pFbF9E-3dIo)' or '[Pehlewani](https://tatoeba.org/eng/sentences/show/5050741#comment-824764)' are used, too. Perhaps it would be better to drop it to avoid confusion.

2- All Kurdish (KUR) sentences could be moved to Northern Kurdish (KMR). I've already moved the others. The rest are all in Northern Kurdish.

3- The writing direction of Central and Southern Kurdish should be RTL. Please see [this](https://tatoeba.org/eng/sentences/show/8996029) example. The Kurdish sentences aren't displayed correctly. They should be right-aligned like the indirectly-linked Persian sentence.

4- A Kurdish [user](https://tatoeba.org/eng/user/profile/Sarchia) suggested using the Persian font for Central and Southern Kurdish. Currently, they are in the Arabic font.

5- Another user insists on using [this flag](https://en.wikipedia.org/wiki/File:Flag_of_Rojava.svg) instead of [the current flag](https://commons.wikimedia.org/wiki/File:Flag_of_Kurdistan.svg) for Northern Kurdish, but I'm not sure if we should apply that change without informing other Kurdish contributors and discussing it. I leave the decision up to the admins.

RyckRichards:
@trang How should I do # 2 ? 
>> All Kurdish (KUR) sentences could be moved to Northern Kurdish (KMR). I've already moved the others. The rest are all in Northern Kurdish.

AndiPersti:
> All Kurdish (KUR) sentences could be moved to Northern Kurdish (KMR). I've already moved the others. The rest are all in Northern Kurdish.

I guess nobody wants to manually change the language for the remaining 1300 sentences, so my idea is to adapt/copy the existing Cake command for mass-editing the license.

> A Kurdish [user](https://tatoeba.org/eng/user/profile/Sarchia) suggested using the Persian font for Central and Southern Kurdish. Currently, they are in the Arabic font.

[...]

PR #2571, opened 2020-10-13
Some final adjustments for the Kurdish languages

RyckRichards:
This PR solves part of #2570 

- The qualifier has been removed from Southern Kurdish.
- Corrected the writing direction of Central and Southern Kurdish

### DIFF
--- a/src/Lib/LanguagesLib.php
+++ b/src/Lib/LanguagesLib.php
@@ -617,7 +617,7 @@ class LanguagesLib
                 'nus' => __d('languages', 'Nuer'),
                 'ckb' => __d('languages', 'Central Kurdish (Soranî)'),
                 'kmr' => __d('languages', 'Northern Kurdish (Kurmancî)'),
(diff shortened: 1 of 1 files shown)


src/Lib/LanguagesLib.php
```diff
@@ -617,7 +617,7 @@ public static function languagesInTatoeba()
                 'nus' => __d('languages', 'Nuer'),
                 'ckb' => __d('languages', 'Central Kurdish (Soranî)'),
                 'kmr' => __d('languages', 'Northern Kurdish (Kurmancî)'),
-                'sdh' => __d('languages', 'Southern Kurdish (Xwarîn)'),
+                'sdh' => __d('languages', 'Xwarîn'),
```
AndiPersti, line 620: This should be the other way around: we just want `Southern Kurdish` as the language name.

RyckRichards: Oh, thanks!

PR #2580, opened 2020-11-15
Add Zaza (Kirmancki) as UI language

RyckRichards:
According to https://www.transifex.com/tatoeba/tatoeba_website/dashboard/, as for November 14th, 2020, Zaza's default pot got 100% translated

image

The translator told me that the most preferable name for this language is Kirmancki. That's the one I added here

https://en.wikipedia.org/wiki/Zaza_language

### DIFF
--- a/config/app_local.php.template
+++ b/config/app_local.php.template
@@ -208,6 +208,7 @@ return [
             ['uzb', null, 'Oʻzbekcha'],
             ['vie', null, 'Tiếng Việt'],
             ['zsm', null, 'Bahasa Melayu'],
+            ['zza', null, 'Kirmancki'],
         ],
     ],
 


AndiPersti:
Looks good to me.

The only remark I have is that Zaza (`zza`) is [considered a macrolanguage by ISO 639-3](https://iso639-3.sil.org/code/zza) but the translator named it Kirmancki which sounds like the [name of one of the individual languages](https://iso639-3.sil.org/code/kiu). So I guess the three letter code we currently use (`zza`) is probably wrong and it should rather be `kiu`, shouldn't it?
But we currently don't support `kiu` so I think that's something to consider in the future.

BTW I've already enabled it on dev.

RyckRichards:
Thanks, @AndiPersti !

trang:
Now that we have dealt with splitting Kurdish into the appropriate individual languages, I suppose we could try to deal with Zaza next. It should go a bit smoother now that we've had some experience with removing Kurdish :)

@RyckRichards Can you start contacting the Zaza contributors to ask which language exactly they used, then create the necessary ticket(s)?

RyckRichards:
Sorry for the late reply.

Yes, for sure! I just need a list of Zaza contributors.

Ricardo

AndiPersti:
@RyckRichards I've sent you a PM.

PR #2582, opened 2020-11-26
Improve UI language selection

### LINKED ISSUE #725: Allow to remove UI languages
There are UI languages that needs to be removed because they are not maintained. However, just removing a language from the config file will make all the links using with this language code return 404. We need to have a redirection mechanism that takes browser’s preference and fall back to English, just like going to http://tatoeba.org/.

AndiPersti:
Originally my plan was to fix #725 with this PR but as I have [commented on that issue](https://github.com/Tatoeba/tatoeba2/issues/725#issuecomment-734437143) I think this is already solved.

Nevertheless I think this part of the code can be improved:

1. As discussed already on #2544 we should still use the language in the URL in some situations.
2. Keeping all the information about the UI languages in an array of arrays makes it usage harder than necessary. Since we almost always know the ISO code, I think having a map from ISO code to language data makes it easier for further processing
3. Currently all the language selection happens in the `AppController`. But I think it would be better to move it into a middleware class. Thus whenever we redirect to a new language we can do it immediately after the request passed through the router and avoid continuing starting up the app.
4. The handling of the cookie is a little bit broken. We actually use two different cookies for storing the language (one is  `CakeCookie[interfaceLanguage]=eng` and the another one is `CakeCookie={"interfaceLanguage":"deu"}`). The reason is that the handling of cookies changed from CakePHP 2 to 3 (and even from earlier 3.x versions).

This PR is still a work in progress (especially it needs more tests for the middleware) but I still would appreciate it if you could give me some feedback for the general idea. :smile: 
If you want to test it locally don't forget to back up your `config/app_local.php` since you need to change it for running this version.


AndiPersti:
I've switched to this branch on the dev server and I haven't found any issues yet.

jiru:
@AndiPersti Good job moving the code into a new middleware! :slightly_smiling_face:
It looks all nice and clean as usual. :sparkles: It’s very fortunate that you are around taking care of Tatoeba.

I’m surprised the new `interface_language` cookie is working out of the box. If I remember correctly, when we last migrated CakePHP, I had to explicitly [disable cookie encryption](https://github.com/Tatoeba/tatoeba2/blob/8bc30db05b852fa34011c90f9502e8c6612cc460/src/Controller/AppController.php#L107). That line will certainly become obsolete after your code is deployed for more than a month anyway.

When a UI language is removed, wouldn’t it be better to redirect to the browser’s preferred language instead of English? Redirecting to English is a welcomed improvement though.

AndiPersti:
> @AndiPersti Good job moving the code into a new middleware! slightly_smiling_face
> It looks all nice and clean as usual. sparkles It’s very fortunate that you are around taking care of Tatoeba.

Thanks.

> I’m surprised the new `interface_language` cookie is working out of the box. If I remember correctly, when we last migrated CakePHP, I had to explicitly [disable cookie encryption](https://github.com/Tatoeba/tatoeba2/blob/8bc30db05b852fa34011c90f9502e8c6612cc460/src/Controller/AppController.php#L107). That line will certainly become obsolete after your code is deployed for more than a month anyway.

That's because you've used the Cookie Component at that time which is deprecated since 3.5.0. Creating a new cookie using `Cake\Http\Cookie\Cookie` (which is accessible from the request object) will create an [unsecure cookie by default](https://api.cakephp.org/3.7/class-Cake.Http.Cookie.Cookie.html#___construct). For secure cookies there is now a [Encrypted Cookie Middleware](https://book.cakephp.org/3/en/controllers/middleware.html#encrypted-cookie-middleware).

> When a UI language is removed, wouldn’t it be better to redirect to the browser’s preferred language instead of English? Redirecting to English is a welcomed improvement though.

With my refactoring of the UI languages array, an unmaintained language can be set to `null` and so will follow the standard order (cookie language -> browser language -> URL language -> English).
You can try it on dev:
```
# grep -E "'(lat|lit)'" config/app_local.php
            'lat' => null, # ['Latina', null],
            'lit' => null, # ['Lietuvių', null],
```

Issue #2586, opened 2020-12-04
It's difficult for admins to handle large amounts of undesirable sentences


trang:
**Story**

When a contributor creates many undesirable sentences, admins have the possibility to block the contributor from adding more sentences. If the contributor had added a dozen of such sentences, these sentences can be handled manually (either marking them as unreliable, or deleting them, or fixing them if that's possible). However, there has been several occasions where admins blocked accounts with hundreds of undesirable sentences.

A recent case was with these accounts:
- https://tatoeba.org/deu/user/profile/Tagawawt
- https://tatoeba.org/deu/user/profile/Tilelli
- https://tatoeba.org/deu/user/profile/Tamazight

In such cases, someone who has access to the database has to make an intervention to handle because there is no possibility from the UI to delete all sentences from a given user, or mark all the sentences of a given user as unreliable. It would not be fun for admins to handle them one by one from the UI.

I'm most of the time that "someone" and as far as I'm concerned, I've always mark the sentences rather than deleting them.

```
UPDATE sentences SET correctness = -1 WHERE user_id IN (...);
```

I prefer not to delete because I don't want to take the risk of making of a mistake and deleting the wrong sentences. It wouldn't be easy to revert if that happened. Also, the sentences are often not all bad to the point where they must all be deleted. It is quite a burden to be the one who choses to delete a large amount of sentences, even more when it comes to deleting sentences in a language I don't understand.

**Idea**

As a minimum, we could add a button "Mark as unreliable" on the sentences page of each user. This button would mark all the sentences of the given user as unreliable. It would be available only to admins, and only for accounts that are blocked or suspended.

HelioStrike:
@trang Hello! If the issue isn't already being worked on, I'd like to take it up.

AndiPersti:
@HelioStrike Welcome and thank you for working on that issue.

If you have any questions, just ask.

PR #2591, opened 2020-12-10
Allow admins to mark a blocked/suspended user's sentences as unreliable

HelioStrike:
This PR is for solving #2586.

This is what the UI currently looks like (look at the bottom right to see the "Mark Unreliable" button).

Screenshot_2020-12-10 cerrax's sentences - Tatoeba

## Some things to be looked into:
- 2 new UI strings have been used in the code. 'Mark Unreliable' (button text) and 'Are you sure you want to mark user as unreliable?' (confirmation text). Should I replace these with anything else? Also, I think we'll need to add these strings to the admin files of each language. Are there any docs I can refer to, that can help me do this the right way?
- How do I improve the button UI?

### DIFF
--- a/src/Controller/SentencesController.php
+++ b/src/Controller/SentencesController.php
@@ -29,6 +29,7 @@ namespace App\Controller;
 use App\Controller\AppController;
 use App\Form\SentencesSearchForm;
 use App\Model\CurrentUser;
+use App\Model\Entity\User;
 use App\Model\Table\SentencesTable;
 use App\Lib\LanguagesLib;
 use App\Lib\SphinxClient;
@@ -677,6 +678,9 @@ class SentencesController extends AppController
             return;
         }
 
+        $user = $this->Users->getUserById($userId);
+        $this->set("unreliableButton", CurrentUser::canMarkSentencesOfUser($user));
+
         $this->set("userExists", true);
 
         $onlyOriginal = array_key_exists('only_original', $this->request->getQueryParams());
@@ -828,6 +832,38 @@ class SentencesController extends AppController
         }
     }
 
+    
+    /**
+     * Mark all sentences of a user as incorrect.
+     *
+     * @param string $username User name of the user. 
+     * 
+     * @return void
+     */
+    public function mark_unreliable($username)
+    {   
+        $marked = $this->Sentences->markUnreliable($username);
+        
+        if($marked) {
+            $this->Flash->set(format(
+                __d('admin', 'Marked all sentences added by {username} as unreliable.'),
+                ['username' => $username]
+            ));
+        } else {
+            $this->Flash->set(format(
+                __d('admin', 'Error: Sentences added by {username} could not be marked as unreliable.'),
+                ['username' => $username]
+            ));
+        }
+        
+        $this->redirect(
+            array(
+                "controller" => "sentences",
+                "action" => "of_user",
+                $username
+            )
+        );
+    }
 
     public function edit_audio()
     {
--- a/src/Model/Table/SentencesTable.php
+++ b/src/Model/Table/SentencesTable.php
@@ -1167,6 +1167,35 @@ class SentencesTable extends Table
         return $this->save($sentence);
     }
 
+    /**
+     * Marks all sentences of the user as incorrect.
+     * Only admins can perform this action.
+     *
+     * @param string $username User name of the user.
+     *
+     * @return bool
+     */
+    public function markUnreliable($username)
+    {
+        $user = $this->Users->getInformationOfUser($username);
+        if(empty($user) || !CurrentUser::canMarkSentencesOfUser($user)) {
+            return false;
+        }
+
+        $sentences = $this->find('all')
+            ->where(['user_id' => $user['id'], 'correctness' => 0])
+            ->select(['id', 'correctness'])
+            ->toList();
+
+        $editValues = array();
+        foreach($sentences as $sentence) {
+            array_push($editValues,  [ 'id' => $sentence['id'], 'correctness' => -1 ]);
+        }
+
+        $this->patchEntities($sentences, $editValues);
+        return $this->saveMany($sentences);
+    }
+
     public function getSentencesLang($sentencesIds) {
         if (empty($sentencesIds)) {
             return [];
--- a/config/auth_actions.php
+++ b/config/auth_actions.php
@@ -108,6 +108,7 @@ $config = [
             'change_language'       => User::ROLE_CONTRIBUTOR_OR_HIGHER,
             'edit_audio'            => [ User::ROLE_ADMIN ],
             'edit_correctness'      => [ User::ROLE_ADMIN ],
+            'mark_unreliable'      => [ User::ROLE_ADMIN ],
         ],
         'sentences_lists'      => [ '*' => User::ROLE_CONTRIBUTOR_OR_HIGHER ],
         'tags'                 => [ '*' => User::ROLE_ADV_CONTRIBUTOR_OR_HIGHER ],
--- a/src/Model/CurrentUser.php
+++ b/src/Model/CurrentUser.php
@@ -365,4 +365,11 @@ class CurrentUser
             return self::isTrusted() && $userAccountDeactivated;
         }
     }
+
+    public static function canMarkSentencesOfUser($user)
+    {
+        $userBlocked = $user->level == -1;
+        $userSuspended = $user->role == User::ROLE_SPAMMER;
+        return (CurrentUser::isAdmin() && ($userBlocked || $userSuspended));
+    }
 }
--- a/src/Template/Sentences/of_user.ctp
+++ b/src/Template/Sentences/of_user.ctp
@@ -71,7 +71,30 @@ $this->set('title_for_layout', $this->Pages->formatTitle($title));
             </md-checkbox>
         </div>
         <?php
-     } ?>
+    }
+
+    if ($unreliableButton) {
+        ?>
+        <md-button class="md-warn md-raised"
+                   aria-label="<?= __d('admin', 'Mark as unreliable') ?>" >
+            <?php
+            echo $this->Html->link(
+                __d('admin', 'Mark as unreliable'),
+                [
+                    'controller' => 'sentences',
+                    'action' => 'mark_unreliable',
+                    $userName
+                ],
+                [
+                    'confirm' => __d('admin', 'Are you sure you want to mark all the sentences of this member as unreliable?'),
+                    'style' => 'color: white;'
+                ]
+            );
+            ?>
+        </md-button>
+        <?php
+    }    
+    ?>
 </div>
 <?php endif; ?>
 
--- a/tests/TestCase/Controller/SentencesControllerTest.php
+++ b/tests/TestCase/Controller/SentencesControllerTest.php
@@ -379,7 +379,7 @@ class SentencesControllerTest extends IntegrationTestCase {
     }
 
     public function testEditCorrectness_asCorpusMaintainer() {
-        $this->logInAs('advanced_contributor');
+        $this->logInAs('corpus_maintainer');
         $this->post('/jpn/sentences/edit_correctness', ['id' => '1', 'correctness' => '-1']);
         $this->assertRedirect('/');
     }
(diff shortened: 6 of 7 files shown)


trang:
@HelioStrike Welcome and thank you for this pull request! I must say I'm super impressed, it's not often that we have pull requests with this level of difficulty being completed as first pull request :)

Here's my feedback.

1. I suggest to readjust the UI strings as followed:
    * `Mark Unreliable` → `Mark as unreliable`
    * `Are you sure you want to mark user as unreliable?` → `Are you sure you want to mark all the sentences of this member as unreliable?`

2. You will have to use the `__d()` function instead of `__()` for the strings to be extracted into the admin.pot file (see [examples in the code](https://github.com/Tatoeba/tatoeba2/search?q=__d%28%27admin%27)).

3. For the design of the button, you could use the class `md-warn` instead of `md-primary`, since it's an action that shouldn't be taken lightly and therefore a more "danger" look and feel would suit better. You can combine `md-warn` it with `md-raised` to have the button look like a button rather than a link. As for the position of the button, I think it's fine where you put it.

4. It would be great to write tests for your newly created functions, if you know how to write tests (or want to give it a try). You can check the `/tests` folder for examples. Otherwise it's no problem, your PR can be merged without the tests.

Persikarbeto:
Thanks for working on this admin feature, which will help a lot. Some feedback:

1. It would be good to be able to mark only the displayed sentences (either all, or all in a given language) as unreliable. It seems that right now the only option is to mark all sentences as unreliable, which may be good in some cases, but too coarse in others.
2. It would be good if it were possible to undo the process in the same way.

trang:
@Persikarbeto Thanks for your feedback :) I think it would be better to consider your suggestions as part of another iteration, and for this pull request to focus on the minimum requirement so that we can deploy this on production ASAP.

I would create separate tickets for your suggestions but I have some questions:

1. Do you have some concrete examples of users for whom you would not mark all sentences as reliable, but only the ones displayed on the current page? I suspect that for such cases, it might be more useful if there would be a button on the list of sentences which would allow you to mark as unreliable, similarly to how you would mark as "not OK" with the reviews. And if it happens that the majority of the sentences on the page need to be marked as unreliable, I think having a checkbox next to each sentence, with a "Check all" button would be better. This would allow the most flexibility: you could check all at first, then uncheck the few ones that are okay.

2. Being able to revert is definitely an important feature to have for any bulk actions generally speaking, but it's not very easy to implement. In this case, we could consider having a button "Mark as normal" (or something like that) that would set all sentences correctness back to 0. I think it would be simple enough to include it as part of this pull request, but it would not be the same as reverting because prior to marking all sentences as unreliable, there could have been already sentences that were set as unreliable individually, and that work would be erased. If you think it would still be useful to be able to mark all sentences as normal, then perhaps @HelioStrike will be kind enough to add this 2nd button :) Otherwise, to be able to properly revert, we would need to store somewhere the information of which sentences exactly have been marked as unreliable, so it gets a bit more complicated and it would be better to leave it for later. Let us know.

AndiPersti (changes requested):

config/Migrations/20200618002139_RemoveIPsFromContributions.php
```diff
@@ -17,7 +17,9 @@ public function change()
         $table->update();
 
         $table = $this->table('last_contributions');
-        $table->removeColumn('ip');
-        $table->update();
+        if ($table->getColumn('ip')) {
+          $table->removeColumn('ip');
+          $table->update();
+        }
```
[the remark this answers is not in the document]
HelioStrike: @AndiPersti I had bit of trouble working with imouto earlier today. I haven't had any trouble in a while so I probably just messed something up.

src/Controller/SentencesController.php
```diff
@@ -677,6 +678,18 @@ public function of_user($userName, $lang = null)
...
+        $this->set("userBlocked", false);
+        if($level == -1) {
+            $this->set("userBlocked", true);
+        }
+
+        $role = $this->Users->getRoleOfUser($userId);
+        $this->set("userSuspended", false);
+        if($role == User::ROLE_SPAMMER) {
+            $this->set("userSuspended", true);
+        }
```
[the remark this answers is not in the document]
HelioStrike: @AndiPersti I agree! I'll make the change. :)

src/Model/Table/SentencesTable.php
```diff
@@ -1167,6 +1167,23 @@ public function editCorrectness($sentenceId, $correctness)
...
+     * @return void
+     */
+    public function markUnreliable($userId)
+    {
+        $sentence = $this->query()
+            ->update()
+            ->set(['correctness' => -1])
+            ->where(['user_id' => $userId])
+            ->execute();
+    }
```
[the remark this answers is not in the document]
HelioStrike: @AndiPersti I tried to write code that does things this way, but can't seem to get it right.
```
public function markUnreliable($userId)
{
    $sentences = $this->find('all')
        ->where(['user_id' => $userId, 'correctness' => 0])
        ->toList();
    $this->patchEntities($sentences, ['correctness' => -1]);
    return $this->saveMany($sentences);
}
```
The problem is that `$sentences` isn't getting updated with the correctness value after `patchEntities` is called. I tried replacing `['correctness' => -1]` with `[['correctness' => -1]]`, `array(['correctness' => -1])` and a few other similar ones but they don't seem to do it. I feel like there might be something very simple I'm missing here. Do you notice anything? I'll give testing a look and get back to this issue.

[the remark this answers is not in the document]
AndiPersti: `patchEntities` merges both parameters, so the second parameter should also be an array of arrays. Each of these arrays should contain the primary key and the fields (with new values) that should change, i.e. it should look like
```php
[
    [0] => ['id' => 123, 'correctness' => -1],
    [1] => ['id' => 345, 'correctness' => -1],
    ...
]
```
`patchEntities` needs the primary key in order to match up the corresponding entities in the first parameter.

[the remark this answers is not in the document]
HelioStrike: @AndiPersti Works now. Thank you!

HelioStrike:
Some changes made since the last review:

- I changed the header from `/sentences/mark_unreliable/<username>` to `/sentences/mark_unreliable/<userId>`. It somehow felt more correct to have it this way. What do you think?
- Added tests for `mark_unreliable` (of `SentenceController`) in [SentencesControllerTest.php](https://github.com/Tatoeba/tatoeba2/pull/2591/files#diff-9faed9d37204cbed970f7012f2e5af4cfdaa482bae50bf0449e9a8252a7fb6b8R393), and `markUnreliable` (of `SentencesTable`) in [SentencesTableTest.php](`mark_unreliable` (of `SentenceController`) in [SentencesControllerTest.php](https://github.com/Tatoeba/tatoeba2/pull/2591/files#diff-9faed9d37204cbed970f7012f2e5af4cfdaa482bae50bf0449e9a8252a7fb6b8R393).
- Found a [couple' places](https://github.com/Tatoeba/tatoeba2/pull/2591/files#diff-9faed9d37204cbed970f7012f2e5af4cfdaa482bae50bf0449e9a8252a7fb6b8L382) in `SentencesControllerTest`, where the test needed the user to be logged in as `corpus_maintainer` but the user was logged in as `advanced_contribution`, and changed them.

src/Controller/SentencesController.php
```diff
@@ -828,6 +835,51 @@ public function edit_correctness()
         }
     }
 
+    /**
+     * Mark all sentences of a user as incorrect.
+     *
+     * @param string $username Username of the user.
```
AndiPersti, line 841: You should change `username` to `id` here too.

HelioStrike: Oops, got it.

AndiPersti:
> I changed the header from `/sentences/mark_unreliable/<username>` to `/sentences/mark_unreliable/<userId>`. It somehow felt more "correct" to have it this way. What do you think?

I thought about that too, so I agree.

> Added tests for `mark_unreliable` (of `SentenceController`) in [SentencesControllerTest.php](https://github.com/Tatoeba/tatoeba2/pull/2591/files#diff-9faed9d37204cbed970f7012f2e5af4cfdaa482bae50bf0449e9a8252a7fb6b8R393), and `markUnreliable` (of `SentencesTable`) in [SentencesTableTest.php](`mark_unreliable` (of `SentenceController`) in [SentencesControllerTest.php](https://github.com/Tatoeba/tatoeba2/pull/2591/files#diff-9faed9d37204cbed970f7012f2e5af4cfdaa482bae50bf0449e9a8252a7fb6b8R393).

Thanks for adding the tests.

> Found a [couple' places](https://github.com/Tatoeba/tatoeba2/pull/2591/files#diff-9faed9d37204cbed970f7012f2e5af4cfdaa482bae50bf0449e9a8252a7fb6b8L382) in `SentencesControllerTest`, where the test needed the user to be logged in as `corpus_maintainer` but the user was logged in as `advanced_contribution`, and changed them.

Good catches. :+1: 

This PR looks pretty good to me.
The only thing I don't really like is the duplicated code in the two Controller actions for testing whether the user is allowed to mark the sentences.
Thus I suggest the following:
- Move the check to a static method on the `CurrentUser` class like we do already for [`canRemoveSentence`](https://github.com/Tatoeba/tatoeba2/blob/cd1a87eba80a5f19ed25f7dd9803915345b3a7dc/src/Model/CurrentUser.php#L221), [`canEditLicenseOfSentence`](https://github.com/Tatoeba/tatoeba2/blob/cd1a87eba80a5f19ed25f7dd9803915345b3a7dc/src/Model/CurrentUser.php#L345), ... This method would take a `User` entity since we need several properties from it.
- In `SentencesController::of_user()` instead of getting just the id, load the whole entity and replace the checking code with `$this->set("unreliableButton", CurrentUser::canMarkSentencesOf($user))`.
- In `SentencesController::mark_unreliable()` just pass the id to the Model method and act on the return value (similar to `SentencesController::delete()`).
- In `SentencesTable::markUnreliable()` first check whether the action is allowed and then do the rest.

(`UsersTable::getRoleOfUser()` wouldn't be necessary with this solution.)
 
What do you think?

AndiPersti (approved):

HelioStrike:
@AndiPersti Will work on the changes! 

> In SentencesTable::markUnreliable() first check whether the action is allowed and then do the rest.

How should I be doing this? Since the method only gets `$userId`, do I fetch the user again?

AndiPersti:
In the Controller I would just do:
```php
{
    $result = $this->Sentences->markUnreliable($userId);
    // Use $result for further action
}
```
and in the Model:
```php
{
    $user = $this->Users->getUserById($userId);
    if(CurrentUser::canMarkSentencesOf($user)) {
        // admin
    } else {
        return false;
    }
}
```

HelioStrike:
@AndiPersti Done!

src/Controller/SentencesController.php
```diff
@@ -828,6 +833,61 @@ public function edit_correctness()
...
+                array(
+                    "controller" => "sentences",
+                    "action" => "of_user",
+                    $user['username']
+                )
+            );
+            return;
+        }
+
+        if (CurrentUser::canMarkSentencesOfUser($user)) {
```
AndiPersti, line 860: Sorry that I wasn't clear enough but you don't need all these lines since only admins are allowed to access this controller action (in `config/auth_actions.php`).

src/Controller/SentencesController.php
```diff
@@ -828,6 +833,61 @@ public function edit_correctness()
...
+            $marked = $this->Sentences->markUnreliable($userId);
+            
+            if($marked) {
+                $this->Flash->set(format(
+                    __d('admin', 'Marked all sentences added by {username} as unreliable.'),
+                    ['username' => $user->username]
```
AndiPersti, line 866: Since we want to give the admin feedback and we need the username I think after all it is better not to pass the id but pass the username as you did earlier. For the view and the model it doesn't really matter but here we would need to do an extra database fetch.

AndiPersti (changes requested):

src/Controller/SentencesController.php
HelioStrike: Ah, got you now.

src/Controller/SentencesController.php
HelioStrike: That makes sense, but that would mean that everything in `markUnreliable` would have to use the username to avoid that extra fetch, instead of using the ID. Is that alright? (else, we'll just be delaying the fetch)  Considering that usernames are unique it shouldn't be a problem.

HelioStrike: Took a look at the code again. If we pass username to `markUnreliable`, we would still need a query to get the ID, so that we can fetch the user to perform the security check, with the current code structure. (._.)

AndiPersti: In `markUnreliable` you load the user entity for the check, so you have access to all its properties there.
(For loading a user if you know the username you can use [`Users::getInformationOfUser()`](https://github.com/Tatoeba/tatoeba2/blob/cd1a87eba80a5f19ed25f7dd9803915345b3a7dc/src/Model/Table/UsersTable.php#L238).)

HelioStrike: @AndiPersti Ah, I didn't notice the method. Fixed it now.

AndiPersti (approved):

AndiPersti:
From my side it looks good. Thanks for your work.

If no one else has any remarks, I'll merge it at the beginning of next week and deploy it the following Sunday. Thus our translators have time to translate the new strings.

HelioStrike:
@AndiPersti That was a lot a revisions (so far)! Thank you for being so patient, and your very detailed code reviews. :)

PR #2590, opened 2020-12-07
Add language Southern Zaza (Dimli)

RyckRichards:
This PR closes #2587.
It part of the process to remove the macrolanguage Zaza from Tatoeba.

PR #2589, opened 2020-12-07
Add language Northern Zaza (Kirmanjki)

RyckRichards:
This PR closes #2588.
This is the first step to replace the Macrolanguage Zaza (zza) on Tatoeba.

PR #2592, opened 2020-12-10
Replace Zaza flag - neutral one

RyckRichards:
This PR changes the Zaza flag to a more neutral one.


AndiPersti:
It looks like you've added the wrong image because I just get an invisible flag:
zaza

trang:
For the record, ZZA is a macrolanguage and we're supposed to get rid of macrolanguages but in this case, it may not be easy to remove ZZA from our supported languages. It appears that some users may not identify their language to be DIQ nor KIU (which are the two individual languages under the ZZA macrolanguage in the ISO 639-3). 

The language icon we currently used appears to have no meaning to our ZZA contributors. We do not have any information on why this particular flag was chosen. We could not find a consensus on what would be a better icon for ZZA.

Since it may take several months before we come to a decision on whether or we keep or remove ZZA, we figured it's best to change the language icon to neutral one.

AndiPersti:
I see but wouldn't it be better to use the three letter code (`ZZA`) instead like it was suggested in the other two PR's (#2589 and #2590) for the individual languages.

This neutral/transparent flag looks a little bit strange IMHO.

(Maybe it's time to reconsider the [mixing of languages and flags/countries](http://www.flagsarenotlanguages.com/blog/why-flags-do-not-represent-language/) :smile: :stuck_out_tongue_winking_eye: )

trang:
Perhaps a bit of misunderstanding here. My reply wasn't trying to justify the transparent picture, just to bring additional info on why we're making this change despite the fact that ZZA is supposed to be removed :)

The language icon should have "Zazaki" written on it, with a white background. It's not supposed to be just a transparent icon. I'm not sure if having only "ZZA" instead of "Zazaki" on the icon was considered. All I can say is that having Zazaki was agreed with the various Zaza members involved. @soliloquist-tatoeba Perhaps you have more insight on this? But in any case, if it was agreed with "Zazaki", we don't need to drag the discussion longer, we can just go ahead with it.

As for using country flags for our language icons, (from my perspective) this is mostly because flags are often the first thing that comes to people's mind if they must think of a picture that would represent their language and on our side, we will usually just take whatever picture our users suggest as long as there's no conflict with other native speakers. But our instructions when it comes to language icons have never asked to use a flag, only to provide an image that can be used to visually represent the language and in the past few years, we made it more explicit that it does not need to be a country flag (https://en.wiki.tatoeba.org/articles/show/language-icons). We could take a more radical approach which is to completely reject country flags but I think that would be too extreme, and also that means a lot more design work on us (which we can't really afford for lack of resource and, I guess, lack of talent too).

soliloquist-tatoeba:
We are using the language name as the icon for Baluchi, and the language code for others like Kashmiri. When I informed the user who requested the flag change about both of the options, he preferred [the Baluchi style](https://tatoeba.org/eng/sentences/show_all_in/bal/none).

As for the sublanguages, there are various alternate names on [Ethnologue](https://user-images.githubusercontent.com/47579687/101938174-f77a2100-3bf3-11eb-99e6-9e3f93c690db.png), so using the codes might be a more neutral approach. We also prepared some custom icons that are visually more attractive, but if the user who requested the flag change doesn't approve them, we'll just use codes as the last resort. Ricardo will update the issues if that user approves them.

I may write a more detailed message later to provide background information on this issue for archival purposes, if needed.


As for the problematic or competing flags, instead of getting rid of flags altogether, supporting 4-letter language codes as separate categories might be considered. This way, languages that are spoken in different countries, are written in multiple alphabets or have competing flags can have their own multiple categories. So the users who contribute in those languages can add their sentences in the category they prefer comfortably without offending others or creating a mess. There will be likely no more flag issues as long as they don't break the French law.

These 4-letter categories then can be merged together into the standard 3-letter codes again in data exports. (they can also be offered to download separately for other purposes) The same merging operation can be applied to the search and even to Horus, too. I may create a new issue for this idea later if I refine and develop my thoughts about it.

RyckRichards:
For the records, some Zaza speakers have agreed to use to following flag instead - https://i.ibb.co/ZS3CZ5V/ZZA.png. I've uploaded it.

Issue #2581, opened 2020-11-19
Make it so if someone adds a sentence that is an orphan, the sentence automatically gets adopted by that username.


ckjpn:
**Story** or **Context**
For the example, I'll just include the private message that I received.

*Example:*

When I add a sentence which is the same as an orphan sentence, the sentence still is an orphan, the sentence doesn't become mine.

(1) In the case added as a translation:
The added sentence is only linked, and the owner of the sentence doesn't become ourselves.

(2) In the case a new sentence is added:
The following message is displayed.
・Your sentence was not added because the following already exists.

If I remember correctly, I think it used to be as follows.

In case (1); The added sentence was linked and the owner of the sentence was automatically myself.

In case (2); The message wasn't displayed and the owner of the sentence was automatically myself.

What do you think about those?

Annotation: https://tatoeba.org/jpn/sentences/show/9207530

https://tatoeba.org/eng/sentences/show/9207530
アドプトされていない文（孤児の文）と同じ文を追加した時に、自動でアドプトされないような気がします。

この文は、
#9199907 Will Tom come today?
に追加した後、意図的に文を手放しました。その後、
#9207428 Is Tom going to come today?
の文にコピペで貼り付けましたが、small_snowでアドプトされませんでした。

テスト環境がうまく使えなかったので、こっちで試しちゃったごめんなさい。
昨日も、
#202714 ちょっと聞いてもいい？
で同様のことが起こりました。掲示板に書けばいいのだろうけど、不確かなのでまずは、こちらにメモ書き。

HelioStrike:
@ckjpn I think the issue makes sense, but I think it was a conscious decision to make the change (from the earlier behavior to the present one). What do you think, @AndiPersti?

If the issue is valid, I'd like to work on this. I only discovered Tatoeba recently, and couldn't find a way to create orphan sentences. Could you please tell me how I could it, so I could try to replicate this behavior on my own system?

AndiPersti:
I think this issue is valid and the suggested solution sounds reasonable to me.

I don't know anything about the old behavior but maybe @trang does.

As far as I can tell the current behavior was introduced by #1311 (fix for #577) and nobody probably thought about duplicates of orphan sentences.

trang:
When the duplicate detection was implemented, I believe that indeed, no one thought about orphan sentences.

The suggestion is okay but we should limit it to only advanced contributors (who are allowed to adopt sentences) otherwise it creates a loophole.

[...]

Issue #1016, opened 2016-02-23
When the duplicate-merging script deletes a sentence in the process of merging sentences, automatically rate the sentence OK by the person who contributed the duplicate.


ckjpn:
This may be too difficult to program, but the following might be something to consider.

When the duplicate-merging script deletes a sentence in the process of merging sentences, automatically rate the sentence OK by the person who contributed the duplicate.

By contributing the duplicate sentence, the second contributor is saying that he thinks the sentence is good, isn't he?

Here's an example.

Guybrush88, a native Italian speaker, contributes a sentences that is a duplicate of a non-native speaker's contribution.

From a user's point of view, it would be nice to know that Guybrush88 thought that that Italian sentence was good, so rate the sentence OK by Guybrush88.

I doubt that anyone would complain about this.  Why would someone intentionally contribute a sentence that they didn't feel was good?

RyckRichards:
Actually, as far it's a collection of sentences that an user has,
maybe each user should "check" if they want this or not. What we can
do is to tag the sentence as OK since a native already posted. that
sentence.

On 2/23/16, ckjpn <notifications@github.com> wrote:
> This may be too difficult to program, but the following might be something
> to consider.
>
> When the duplicate-merging script deletes a sentence in the process of
> merging sentences, automatically rate the sentence OK by the person who
> contributed the duplicate.
>
> By contributing the duplicate sentence, the second contributor is saying
> that he thinks the sentence is good, isn't he?
>
> Here's an example.
>
> Guybrush88, a native Italian speaker, contributes a sentences that is a
> duplicate of a non-native speaker's contribution.
>
> From a user's point of view, it would be nice to know that Guybrush88
> thought that that Italian sentence was good, so rate the sentence OK by
> Guybrush88.
>
> I doubt that anyone would complain about this.  Why would someone
> intentionally contribute a sentence that they didn't feel was good?
>
>
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/Tatoeba/tatoeba2/issues/1016


-- 

Ricardo Vernaut Jr

Mobile // Celular: (5521) 967-640-668
Email: rvernautjr@gmail.com
<https://www.linkedin.com/in/vernautjr>

trang:
I've changed the title.

The idea itself is not a wrong idea, although it would need some discussion, but it's not the job of the duplicate-merging script and won't be implemented in the duplicate-merging script.

[...]

PR #2598, opened 2020-12-20
Users adopt orphan sentence on adding its duplicate

HelioStrike:
Solves #2581.

### What the PR does
```
If the sentence that was added already exists {
    If sentence has no owner {
        Make current user as sentence owner.
    } else {
        If the owner is a spammer or inactive, and current user is an advanced contributor (or higher) {
            Make current user as sentence owner.
        } else {
            Mark duplicate.
        }
    }
} else {
    Create sentence.
}
```

### Remaining work
- Add and/or refactor tests.


src/Model/Table/SentencesTable.php
```diff
@@ -944,7 +944,28 @@ public function saveNewSentence($text, $lang, $userId, $correctness = 0, $basedO
...
+        // Duplicate sentence found
+        if ($sentence != null) {
+            // If sentence is orphan
+            if(empty($sentence->user_id)) {
+                $sentence->user_id = $userId;
+                return $this->save($sentence);
+            } else {
+                // If sentence is owned by spammer/inactive user, and you are an advanced contributor (or higher)
+                if($this->setOwner($sentence->id, $userId, CurrentUser::get('role'))) {
+                    $sentence->user_id = CurrentUser::get('id');
```
AndiPersti, line 960: I think you could change `setOwner` to return a sentence entity instead of just a boolean:
https://github.com/Tatoeba/tatoeba2/blob/67fd7ccb648f8d23d8e4daaa7dd8b568a7e8deac/src/Model/Table/SentencesTable.php#L1042-L1045
`save` could return `false` in case of a failure so throwing away its return value and always return `true` isn't really failproof.

HelioStrike: @AndiPersti Done! I had to change the function to [get the entire sentence object](https://github.com/Tatoeba/tatoeba2/pull/2598/files#diff-43e566d2d517ab1d9e32da7a397e04ba620afa5dc3e614cd59415cc9888f055eR1058), but I think it feels more proper this way. 

I also made it so that if the sentence is duplicate, [it is also marked OK](https://github.com/Tatoeba/tatoeba2/pull/2598/files#diff-43e566d2d517ab1d9e32da7a397e04ba620afa5dc3e614cd59415cc9888f055eR1058). What do you think of this code?

Its pretty late here, so I'll try to think about the other comment tomorrow. And thanks for the neat review! :)

alanfgh:
> I also made it so that if the sentence is duplicate, it is also marked OK. What do you think of this code?

Are you talking about adding an OK tag or a positive review setting? I don't think we should do either, since it's possible that the sentence has issues. "OK" should really only be added by a human.

src/Model/Table/SentencesTable.php
```diff
@@ -944,7 +944,28 @@ public function saveNewSentence($text, $lang, $userId, $correctness = 0, $basedO
             return false;
         }
 
-        return $this->saveWithDuplicateCheck($newSentence);
+        $sentence = $this->find('all')
+        ->where(['text' => $newSentence->text, 'lang' => $lang])
+        ->first();
+
+        // Duplicate sentence found
+        if ($sentence != null) {
```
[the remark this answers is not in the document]
trang: Using a hash was a suggestion from sysko which I simply followed when implementing the duplicate detection. I wouldn't be able to elaborate on why exactly, in all honesty, I simply trusted him that it would most likely be one of the more optimal solutions to detect duplicates.

Just for the record, the hash mechanism was first implemented to avoid duplicates for the vocabulary feature (#1218). It was later on replicated into the sentences by another developer (#577).

> So maybe we should stop using a hash and rely on the database for uniqueness of a sentence? It should be at least as efficient (or even more efficient)

If it simplifies the code and is just as efficient, then yes, we could do that.

trang:
@alanfgh The idea is to automatically add a review "OK" on sentences that are were detected as duplicates (not on every sentence). This way, the effort spent on adding a sentence that ends up being a duplicate isn't completely wasted. I think it's not a bad idea but we could use a bit more time to think about it.

@HelioStrike It's probably best that you work on this in a separate pull request :) There might be more to do than just inserting a row in the users_sentences table. Note that there is a separate issue for this too: #1016.

alanfgh:
@trang In general, there is only one situation in which I believe that "OK" ratings, tags, or lists are beneficial, and that is when it is important to indicate that the quality of a sentence was previously in question, but no longer is (either because the sentence has been fixed or because someone has said that the sentence is fine as it is). Otherwise, due to the sheer volume of sentences in our corpus, "OK" ratings get preferentially applied to sentences that can be evaluated quickly because they are short, use common vocabulary, and have simple grammar. These are the same sentences 
towards which people are steered for translating and for recording audio. This process feeds on itself because people are then directed toward translating the same sentences that already have audio and the highest number of translations. As a result, sentences that are longer, contain less common words, and are more diverse receive less and less attention, relatively speaking. Even worse, searches for particular vocabulary words pull up pages' worth of boilerplate sentences ("I am <adjective>", "Tom is <adjective>"), making it harder to find unique context to help one understand or remember a particular word. One other problem is that there are a significant number of faulty sentences that are incorrectly marked as OK (which I mark or correct, but track).

Sentences that, as a result of correction, are now duplicates of existing sentences are a special case. If you think of them in isolation, then you could say that they fit the category of "previously questionable status, now approved". But in fact, they're now essentially assimilated into their existing "twins", whose status was never in question. So there doesn't seem to be a point in emphasizing that the previously correct twin has now been joined by a corrected twin. More importantly, these sentences are no more correct, and no more worthy of being marked correct, than other sentences in the corpus. Yet, since simple sentences are more likely to be duplicated, they would end up accumulating more "OK" ratings when the duplicates were merged.

I'm not sure I understand the "salvaging wasted time" argument. The best way not to waste one's time is to write sentences that are unlikely to be duplicates in the first place. As for correcting sentences that thereby become duplicates, the "win" in terms of consolidating them is self-evident.

trang:
@alanfgh Thanks for elaborating on this. I perfectly understand your concerns. I'll report your arguments into #1016 and continue the discussion over there, to not hijack this pull request with an unrelated topic.

src/Model/Table/SentencesTable.php
trang: > I also made it so that if the sentence is duplicate, it is also marked OK. What do you think of this code?

@HelioStrike Based on the comments from Alan, it would be best to remove this code from this PR and to not start anything until we've discussed this a bit more. We need to find a consensus on if/when to add automatically an "OK" review.

HelioStrike:
@trang @AndiPersti Removed the code for the marking OK bit. 

I probably should've consulted here, but I made some changes regarding the unique key discussion.
- Added [migration](https://github.com/Tatoeba/tatoeba2/pull/2598/files#diff-636646fa68debc1391c7d5c2f45ed684a8d072699e6aade57537327d4c5753a6) to add the unique key `(lang, text)`, remove the `hash` column, and the obsoleted keys.
- Removed/edited tests obsoleted by removal of the hash column.
- Edited [SentencesTable.php](https://github.com/Tatoeba/tatoeba2/pull/2598/files#diff-43e566d2d517ab1d9e32da7a397e04ba620afa5dc3e614cd59415cc9888f055e) to not use the hash column.

Is there anything else that needs to be changed?

src/Model/Table/SentencesTable.php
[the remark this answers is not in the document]
AndiPersti: Sorry for the late answer but I wanted to get some more data.

> Using a hash was a suggestion from sysko which I simply followed when implementing the duplicate detection. I wouldn't be able to elaborate on why exactly, in all honesty, I simply trusted him that it would most likely be one of the more optimal solutions to detect duplicates.

I guess that at that time you couldn't use the full `text` field as an index and thus he thought that using a hash would solve the problem. The name of the index in the database (`dedup_idx`) suggests that this index was added for Horus which uses its own API for accessing the database.

But after reading and thinking more about this problem, I think even using a shorter prefix index (e.g. `text(100)`) would have helped at that time because the main reason for using an index is to reduce the number of rows the database needs to search for the wanted row. And just using the first e.g. 100 bytes reduces the number of candidates to 1 row in almost all cases[*]. That's the same situation we have with the hash collisions because the length of the hash is also limited.

[*] I've searched the longest common prefixes in the whole corpus (including the unreliable sentences) and there are 1815 pairs with a common prefix of more than 100 bytes, 360 of them have more than 150 bytes in common and 124 have more than 200 bytes in common. (You can download all common prefixes greater than 50 bytes [here](https://downloads.tatoeba.org/index_test/common-prefix-50.bz2)).

I've also tried to determine how much space we could save with a shorter prefix. On the dev server I've changed the prefix length, run `optimize table sentences` afterwards and checked the index length with `select index_length from information_schema.tables where table_name = 'sentences'`. I got the following values:

prefix length|index length (in bytes)
-|-
10|1,276,887,040
20|1,360,904,192
100|1,547,550,720
200|1,556,987,904
767 (current setting)|1,559,085,056
no prefix (full column)|1,559,085,056

As you can see there is no difference between the current setting and the full column (I guess internally MariaDB also limits the length of the index hash) and the difference to a prefix 100 is relatively small (10MB). Only going down to a very small prefix would get us some space reduction. But all in all I think a index for the full column is more useful than one with a very short prefix and so I would opt for the full one.

Finally I've tried to measure the speed difference between the old version using `DuplicateBehavior` and the version suggested by HelioStrike using the database itself. Therefore I added some timing calls to the code and modified the method to randomly use one of the two versions (see https://github.com/Tatoeba/tatoeba2/compare/dev...AndiPersti:profiling). On my local machine I run the testsuite for the `SentencesTable` three times and on the dev server I've used three different index settings (full column, prefix 767, prefix 20) and added manually random sentences (duplicates and new ones; we don't have `phpunit` installed on the dev server). The average times (in ns) I got were (raw data [here](https://downloads.tatoeba.org/index_test/)):

Setting | `DuplicateBehavior` | Database
-|-|-
local | 2,337,677 | 2,354,544
dev, prefix 20 | 1,699,825 | 1,289,747
dev, prefix 767|1,472,743 |1,345,367
dev, full column|1,563,595|1,527,053

It looks like the current version is slightly slower.

> If it simplifies the code and is just as efficient, then yes, we could do that.

It will definitely simplify the code and taking all the above into account I'd say we should change the way we look for duplicates.

AndiPersti:
Just a general comment: Please do not force push every time you add a new commit. This makes it harder for us to review your PR or running it on our own computers. Just add new commits to your working branch.

If there are really a lot of commits with small minor changes (like a typo or some style issues) we can always squash the commits to clean up the commit history before merging.

HelioStrike:
@AndiPersti Ah, I think of the indexes ordering. You're definitely right about the speed! Also made all the other changes, Will avoid force-pushing everytime from now on.

config/Migrations/20201223141106_SentencesReplaceHashWithUniqueKey.php
```diff
@@ -0,0 +1,25 @@
...
+     * More information on this method is available here:
+     * https://book.cakephp.org/phinx/0/en/migrations.html#the-change-method
+     * @return void
+     */
+    public function change()
+    {
+        $table = $this->table('sentences');
+        $table->addIndex(['text', 'lang'], [
+            'unique' => true,
+            'name' => 'unique_text_lang'
```
AndiPersti, line 18: Sorry but I forgot that we can't use a unique index.

Whenever someone changes a sentence so that this sentence becomes a duplicate of another already existing sentence we will allow these duplicate pair to exist in the database. The two sentences will be merged by [Horus](https://github.com/Tatoeba/horus), a bot that runs every 15 minutes.

HelioStrike: @AndiPersti Done!

AndiPersti (commented):
In general, this PR looks good.

But due to the rather big changes to the database there is some more work to do (e.g. updating Horus) before it can be merged.

Issue #2597, opened 2020-12-19
Easier way to add vocabulary on "Add vocabulary items"


RyckRichards:
**Story** or **Context**
I was adding vocabulary items on https://tatoeba.org/eng/vocabulary/add to create and request sentences. However, each time I wanted to add a new vocabulary request, I did the following:

* I clicked on the box
* Typed the sentence
* Hit "Enter"
* Repeat the process

It would be good if I didn't have to click on the box again making it "typeable" again.

Image: Box after a vocabulary item is added

![image](https://user-images.githubusercontent.com/7658430/102683722-d1cfb780-41b1-11eb-9a8d-3e803b3c9d50.png)

AndiPersti:
@RyckRichards What browser do you use?

It looks like on Firefox the input field is always focused after pressing <Enter> while on Chromium (a Chrome variant) it isn't.

RyckRichards:
@AndiPersti Chrome

AndiPersti:
Fixed by #2602

PR #2602, opened 2020-12-23
Add vocabulary input auto-focuses on submitting

HelioStrike:
Solves #2597.

Clicking on the "Add vocabulary items" input, entering some data, and submitting, auto-focuses the input, eliminating the the need to click the input again.

### DIFF
--- a/webroot/js/vocabulary/add.ctrl.js
+++ b/webroot/js/vocabulary/add.ctrl.js
@@ -59,6 +59,7 @@
                     vm.vocabularyAdded.unshift(data);
                     vm.data.text = '';
                     vm.isAdding = false;
+                    setInterval(function() { document.getElementById('vocabulary-item').focus(); }, 1);
                 }
             );
         }
--- a/webroot/js/vocabulary/add.ctrl.js
+++ b/webroot/js/vocabulary/add.ctrl.js
@@ -20,17 +20,18 @@
 
     angular
         .module('app')
-        .controller('VocabularyAddController', ['$http', VocabularyAddController])
+        .controller('VocabularyAddController', ['$http', '$scope', VocabularyAddController])
         .directive(
             'focusInput', function($timeout) {
                 return {
+                    scope: { trigger: '=focusInput' },
                     link: function($scope, $element, $attrs) {
-                        $scope.$watch($attrs.focusInput, function(value) {
+                        $scope.$watch('trigger', function(value) {
                             if(value === true) { 
                                 $timeout(function() {
                                     $element[0].focus();
-                                    $scope[$attrs.focusInput] = false;
-                                }, 250);
+                                    $scope.trigger = false;
+                                });
                             }
                         });
                     }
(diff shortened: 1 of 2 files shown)


AndiPersti:
@HelioStrike What browser/OS do you use?

I've tried this on Firefox/Linux and the input field is always focused after pressing <Enter> regardless whether the `autofocus` attribute is set.

But on Chromium/Linux the input field is never focused after pressing <Enter> regardless whether the `autofocus` attribute is set.

Furthermore, [according to the specification](https://html.spec.whatwg.org/multipage/interaction.html#the-autofocus-attribute) `autofocus` would only focus the input field on page load and shouldn't have any influence on the field after some interactions.

HelioStrike:
@AndiPersti Ah, sorry, I tested it on Firefox/Ubuntu too, but am having the same behavior when using Chrome.

I could come up with a dirtier JS workaround. I'll try to see if I can find anything better.

(Sorry for the force-pushes, it was a 1-line change, so..)

webroot/js/vocabulary/add.ctrl.js
```diff
@@ -59,6 +59,7 @@
                     vm.vocabularyAdded.unshift(data);
                     vm.data.text = '';
                     vm.isAdding = false;
+                    setInterval(function() { document.getElementById('vocabulary-item').focus(); }, 1);
```
HelioStrike, line 62: `focus()` doesn't seem to work without the 1 millisecond delay. :/

HelioStrike: Ok, this is weird. It works with a 0 millisecond delay too, just needs `setInterval` to be there.

AndiPersti:
> I could come up with a dirtier JS workaround. I'll try to find something better.

Well, in an AngularJS controller you [should not manipulate the DOM](https://docs.angularjs.org/guide/controller) (and focusing an element is a DOM manipulation).
The "AngularJS" way would be to create a service or directive and use that inside the controller. See for example [the answers to a StackOverflow question](https://stackoverflow.com/q/14833326).

webroot/js/vocabulary/add.ctrl.js
AndiPersti: [AngularJS uses its own digest cycle](https://docs.angularjs.org/guide/scope#integration-with-the-browser-event-loop) and any manipulation done outside of AngularJS isn't visible to this digest cycle. That's why you have to artificially trigger a change so that AngularJS gets notified.

Furthermore using this approach it is now impossible to focus any other element (e.g. the language selection dropdown or the input field for the search query) because the set interval fires immediately and puts the focus back to the vocabulary input field.

HelioStrike:
@AndiPersti Thanks for all the helpful resources! I made the changes accordingly.

So the changes work for me on both Firefox and Chrome, but there seems to be a race-condition here. At 250 ms, it worked everytime for me on both browsers (of the ~10 times I tried). It seems that it I only pressed enter, and avoid clicking, I can reduce it to 200 ms even. When I used `trigger` as shown [here](https://stackoverflow.com/a/14837021/6731940), I had to push it to 400 ms. What do you think of this?

HelioStrike:
@AndiPersti Fixed the race condition!

AndiPersti:
Looks good. Thanks.

Issue #944, opened 2015-11-28
Tarifit (rif)


RyckRichards:
* Language name: Tarifit
* Language info: http://www-01.sil.org/iso639-3/...ion.asp?id=rif // https://en.wikipedia.org/wiki/Riffian_language
* ISO 639-3 code: rif
* Public List: https://tatoeba.org/eng/sentences_lists/show/4976
* Icon: https://upload.wikimedia.org/wikipedia/commons/thumb/6/61/Flag_of_the_Republic_of_the_Rif.svg/900px-Flag_of_the_Republic_of_the_Rif.svg.png
* Requested by: https://tatoeba.org/eng/user/profile/doemaar14

https://tatoeba.org/eng/wall/show_message/24984#message_24984

RyckRichards:
https://github.com/Tatoeba/tatoeba2/commit/e8e326ee3d7c1ebb8a711042728fdb8f4e627ce2

![screenshot_1](https://cloud.githubusercontent.com/assets/7658430/11457954/1648d4c4-969e-11e5-819d-2ce9921dd8de.png)

soliloquist-tatoeba:
Note: The original icon of the Tarifit language was changed from the flag of 'the Republic of the Rif' to the flag of 'the Kingdom of Morocco' when the icon of the Kabyle language was changed. It was probably a proactive step to avoid double standards and potential criticism from some Kabyle users. However, as the Berber flag (and not the flag of Algeria) is used for Kabyle and also will be used for Standard Moroccan Berber, the same would be a more standardized and balanced choice for Tarifit, too. Therefore, the Tarifit icon [will be changed](https://github.com/Tatoeba/tatoeba2/pull/2607) to the Berber flag with the 3-letter language code.

PR #2607, opened 2021-01-07
Changed flag for Tarifit

RyckRichards:
This PR is related to #944.
It replaces the flag for Tarifit in order to have all Berber languages using the same standard flag.


RyckRichards:
Thanks, @AndiPersti

PR #2608, opened 2021-01-07
Enhance display of multiline language icons

jiru:
On the ["browse by language" page](https://tatoeba.org/eng/sentences/index), when a language name spans over multiple lines, it feels like an icon is missing.

That was [reported by Trang](https://tatoeba.org/eng/wall/show_message/35895#message_35895) a while ago.

This commits adds a vertical gray line to fill the blank created by multiline languages.

pr_middle_line

### DIFF
--- a/webroot/css/sentences/index.css
+++ b/webroot/css/sentences/index.css
@@ -126,6 +126,10 @@ svg.language-icon {
 }
 .language-list li {
   display: block;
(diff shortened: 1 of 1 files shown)


jiru:
Note that I also came up with other designs but the middle line looks best to me.

Left line:
pr_flag_bar

Extended shadow:
pr_extended_shadow

Issue #1703, opened 2018-10-18
Implement proper email validation on registration


jiru:
Currently, when someone wants to register, we check his/her email address using a regular expression. This is a problem because:

1. The current regular expression doesn’t look good. For example, it doesn’t match the rather frequently used character `+` (hence the `bug` label of this issue).
2. Using a regular expression is not good in the first place, because there are no correct regular expressions for emails.

So: better send a confirmation link by email instead.

PR #2610, opened 2021-01-10
Use built-in email validation instead of overly-strict regexes

### LINKED ISSUE #2609: Some email addresses are incorrectly rejected on registration
**To Reproduce**
1. As a logged-out user...
2. Attempt to register.
3. Enter `john+doe@example.com` in the email field.
4. The field is marked red with the error message "Invalid email address"

**Expected behavior**
The email address is recognized as valid, because it is.

**Additional information**
Related: #1703, where the issue with `+` in email addresses was mentioned, but which is a bit wider in scope.

This issue was reported in the XMPP chat on December 23rd last year, but I didn't get a notification until I accidentally opened my XMPP client just now.

Yorwba:
This PR fixes #2609.

As it turns out, both the [AngularJS email input](https://docs.angularjs.org/api/ng/input/input%5Bemail%5D) and [CakePHP's email validator](https://api.cakephp.org/3.8/class-Cake.Validation.Validator.html#_email) were already being used to ensure users enter a valid email address during registration. The additional regex was just unnecessarily rejecting spec-compliant addresses that didn't fit its simplified pattern.

CakePHP supports additionally validating that the domain part of the email has valid MX records (otherwise it can't receive email). This is disabled by default (probably because it requires DNS queries, which might be slow), but I had to specify a value to be able to pass the optional error message as the third parameter. So I decided to enable it because I think it's a good idea and one check per registration attempt shouldn't be much of a burden for the server.

I tested this locally and it seems to work in the sense that emails that should pass validation (like `john+doe@example.com`) do pass it while others are rejected (like `john+doe@example`, because `example` isn't a registered domain with MX records, or `john+doe`, because the `@` is mandatory).

However, after passing validation, I was greeted with an error message `SQLSTATE[HY000]: General error: 1364 Field 'description' doesn't have a default value`. This also happens when using the current `master` and `dev` branches in a freshly-provisioned VM, so I think it's unrelated to this PR. It also doesn't happen on *dev.tatoeba.org*, so I guess my VM is misconfigured somehow.

### DIFF
--- a/src/Model/Table/UsersTable.php
+++ b/src/Model/Table/UsersTable.php
@@ -89,13 +89,9 @@ class UsersTable extends Table
             ->minLength('password', 6, __('Password must be at least 6 characters long'));
 
         $validator
-            ->email('email')
+            ->email('email', True /* Check MX records */, __('Invalid email address'))
             ->requirePresence('email', 'create')
-            ->notEmpty('email', __('Field required'))
-            ->add('email', 'email', [
-                'rule' => ['custom', '/^([a-zA-Z0-9_\.\-])+\@(([a-zA-Z0-9\-])+\.)+([a-zA-Z0-9]{2,4})+$/'],
-                'message' => __('Invalid email address'),
-            ]);
+            ->notEmpty('email', __('Field required'));
 
         $validator
             ->date('since');
--- a/src/Model/Table/UsersTable.php
+++ b/src/Model/Table/UsersTable.php
@@ -89,7 +89,7 @@ class UsersTable extends Table
             ->minLength('password', 6, __('Password must be at least 6 characters long'));
 
         $validator
-            ->email('email', True /* Check MX records */, __('Invalid email address'))
+            ->email('email', False /* Don't check MX records */, __('Invalid email address'))
             ->requirePresence('email', 'create')
             ->notEmpty('email', __('Field required'));
 
--- a/src/Template/Users/register.ctp
+++ b/src/Template/Users/register.ctp
@@ -160,7 +160,6 @@ $label = format(
         <label for="registrationEmail"><?php echo __('Email address'); ?></label>
         <md-icon>email</md-icon>
         <?php
-        $pattern = '/^([a-zA-Z0-9_\.\-])+\@(([a-zA-Z0-9\-])+\.)+([a-zA-Z0-9]{2,4})+$/';
         echo $this->Form->input(
             'email',
             array(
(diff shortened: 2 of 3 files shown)


jiru:
The SQL error you got looks like what Tatoeba/imouto@f5652bbc7b6a7ea9200e8626241eeeac354f2659 is solving.

jiru:
The code looks good! :slightly_smiling_face:

> I tested this locally and it seems to work

Could you add a unit test? It would be best to have the test *not* actually send out the MX query, so that the test suite works offline.

Yorwba:
> The SQL error you got looks like what Tatoeba/imouto@f5652bb is solving.

I think you're right, but I did use a freshly-provisioned VM built from Tatoeba/imouto@66729a9. The problem fixed itself when I ran `vagrant up` a second time today, so I think that just adding a file in `/etc/mysql/conf.d/` is not enough to change the configuration; `mysqld` also needs to read the file, which I suspect it only does during startup.

> Could you add a unit test? It would be best to have the test not actually send out the MX query, so that the test suite works offline.

I added two testcases. I'm not sure how to best disable the MX query, since the amount of indirection between configuring the validator and performing the actual check is rather large. Maybe there's actually an easy way to achieve it with `phpunit`s mocking capabilities, but I have never used them.

jiru:
> I added two testcases. I'm not sure how to best disable the MX query, since the amount of indirection between configuring the validator and performing the actual check is rather large. Maybe there's actually an easy way to achieve it with phpunits mocking capabilities, but I have never used them.

I had a look at the implementation of MX check. It uses [getmxrr()](https://www.php.net/manual/en/function.getmxrr.php) (or fallbacks on other methods if that function is not available, but we do have it). There is a note on that documentation page saying:

> This function should not be used for the purposes of address verification. Only the mailexchangers found in DNS are returned, however, according to » RFC 2821 when no mail exchangers are listed, hostname itself should be used as the only mail exchanger with a priority of 0.

I believe nowadays it’s quite unlikely to see an email server without MX records, but hey RFC are here to be followed. So I think we should rather disable this MX verification and send an actual email to verify the address as in #1703.

Yorwba:
OK, thanks for digging that up!
I flipped that configuration option and adjusted the test, since `example.invalid` is no longer recognized as an invalid domain.

Sending an actual verification email is definitely the right approach, but I'd rather leave that out of scope for this PR.

jiru:
Thank you! :+1:

PR #2612, opened 2021-01-17
Replace registration quiz with honeypot

jiru:
This PR replaces the registration form quiz with a honeypot (a hidden field spambots may be tempted to fill in) because the quiz has multiple problems.

We sometimes receive emails from users complaining that they cannot answer the quiz question correctly. The reason it that they omit punctuation characters in their answer. We already have an explanation text for that, but in the real world people don’t read carefully, so that doesn’t work.

In general, such captcha are not very UX-friendly because they force the user into performing a boring and seemingly useless action.

While I don’t have any number to estimate the performance of the quiz in preventing spambots from registering, I believe it’s about the same as the honeypot (that is to say pretty bad). But UX-wise the honeypot is definitely better; that’s the real improvement of this PR.

### DIFF
--- a/src/Controller/UsersController.php
+++ b/src/Controller/UsersController.php
@@ -287,8 +287,7 @@ class UsersController extends AppController
 
         $newUser = $this->Users->newEntity();
 
-        $correctAnswer = mb_substr($this->request->getData('email'), 0, 5, 'UTF-8');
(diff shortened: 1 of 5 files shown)


jiru:
I am merging this PR despite the lack of feedback because I believe it’s a rather trivial change.

PR #2615, opened 2021-02-05
Language dropdowns UX improvements

jiru:
Despite the advent of `md-autocomplete` for all our language selection dropdowns, we have been forcing upon them the old `<select>` behavior for technical and practical reasons (8f9e40b0ca18c5d89456ed369cd2ab4f64d235fc). This PR gets rid of these limitations to unleash the full power of autocompleting dropdowns for a better UX. :rocket: 

The limitations included:
1. The field was cleared and restored upon focus/blur, in order to make suggestions appear. → I found a (hackish) way to make suggestions appear without clearing the field.
2. The field was cleared and restored upon focus/blur, in order to make "Filter by language" blocks work (they reload the page upon language selection. → I kept that behavior only for these dropdowns.
3. The field was cleared and restored upon focus/blur, in order to make sure one cannot put the form in an invalid state. → If a language dropdown has invalid input, it now shows in red color and clicking the submit button does nothing.
4. Pressing the clear button (✘ button) was clearing the field but it was always restored to the last value. As a result, no matter you clicked on the ✘ or inside the field, it had the same effect. This makes ✘ useless. → Clearing the field, using backspace or the ✘ button, now leaves a placeholder showing what used to be a special option, such as "Any language" (except in case 2.)
5. Showing a 300+ languages list is a bit overwhelming and can [bring the users’ focus into spending dozens of seconds searching through the list rather than typing something](https://en.wiki.tatoeba.org/articles/show/ux-test-3#contributing). → Only profile languages are now showing first. If whatever was typed in returns no results, the "No language found." text now shows along with an additional (hackish!) "Show all" button. As for guests, they still see "last used languages" as suggestions. As a result, guests do not see any suggestions when they visit the website for the very first time; they *have* to type something in, only that time.
6. In the "Translate sentences" page, the "Not directly translated into" dropdown had two special options: "—" and "Any language". → I changed "—" into a checkbox. I also noticed a bug: selecting "—" produced the same results as "Any language". I fixed it.

As a bonus, I also included language icons within the suggestions because it looks so cool. :sparkles: I had to widen the box a little bit to make the icons fit. While I was at it, I decided to I widen it even more because it felt cramped in the first place.

shot

Currently set up on https://dev.tatoeba.org/.
Wall thread: https://tatoeba.org/wall/show_message/36512

### DIFF
--- a/src/Controller/ActivitiesController.php
+++ b/src/Controller/ActivitiesController.php
@@ -114,10 +114,10 @@ class ActivitiesController extends AppController
         $this->helpers[] = 'Languages';
         
         $langFrom = $this->request->getQuery('langFrom');
-        $langTo = $this->request->getQuery('langTo');
-        if ($langFrom && $langTo)
+        if ($langFrom)
         {
             $sort = $this->request->getQuery('sort', 'created');
+            $langTo = $this->request->getQuery('langTo');
 
             $this->Cookie->write(
                 'not_translated_into_lang',
@@ -126,16 +126,18 @@ class ActivitiesController extends AppController
                 '+1 month'
             );
 
+            $searchParams = array(
+                'from' => $langFrom,
+                'sort' => $sort
+            );
+            if ($this->request->getQuery('excludeLangTo') == 'yes') {
+                $searchParams['trans_filter'] = 'exclude';
+                $searchParams['trans_to'] = $langTo;
+            }
             $this->redirect(array(
                 'controller' => 'sentences',
                 'action' => 'search',
-                '?' => array(
-                    'from' => $langFrom,
-                    'to' => 'none',
-                    'trans_filter' => 'exclude',
-                    'trans_to' => $langTo,
-                    'sort' => $sort
-                )
+                '?' => $searchParams
             ));
         }
     }
--- a/src/Controller/SentencesController.php
+++ b/src/Controller/SentencesController.php
@@ -628,10 +628,6 @@ class SentencesController extends AppController
      */
     public function random($lang = null)
     {
-        if ($lang == null) {
-            $lang = $this->request->getSession()->read('random_lang_selected');
-        }
-
         $randomId = $this->Sentences->getRandomId($lang);
 
         if (is_null($randomId)) {
--- a/src/Controller/UsersController.php
+++ b/src/Controller/UsersController.php
@@ -306,7 +306,7 @@ class UsersController extends AppController
                 $this->loadModel('UsersLanguages');
                 // Save native language
                 $language = $this->request->getData('language');
-                if (!empty($language) && $language != 'none') {
+                if (!empty($language)) {
                     $userLanguage = $this->UsersLanguages->newEntity([
                         'of_user_id' => $newUser->id,
                         'by_user_id' => $newUser->id,
--- a/src/Form/SentencesSearchForm.php
+++ b/src/Form/SentencesSearchForm.php
@@ -23,8 +23,8 @@ class SentencesSearchForm extends Form
 
     private $defaultCriteria = [
         'query' => '',
-        'from' => 'und',
-        'to' => 'und',
+        'from' => '',
+        'to' => '',
         'tags' => '',
         'list' => '',
         'user' => '',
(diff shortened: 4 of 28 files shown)


Yorwba:
I like this change a lot. It even feels snappier (maybe because it doesn't need to render so many items in the dropdown.)

But here are some things I noticed while testing on dev:
- [Searching without specifying any languages](https://dev.tatoeba.org/eng/sentences/search?query=&from=&to=) displays the description "Sentences in unknown translated into unknown" instead of "All sentences"
- As a guest with only a few "last used languages", there's a lot of whitespace below the "Show all" button.
- If the input is invalid, clicking the submit button does nothing, but it's still "clicky" (has the expanding circle animation) I think there should be stronger feedback that it's disabled and why.

jiru:
@Yorwba Thank you so much for the feedback! :+1:

I fixed the page title of search results.

Regarding the extra whitespace in the suggestions box, I removed it. After doing this, I realized the "show all" button is also creating an extra space at the end of the list. I couldn’t find a proper way to get rid of it, so I moved that button down the "No language found" text. That is to say, the button now only appears when the user types a text that matches no language at all. I like it this way because users should rarely make use of that button, so it’s little use to display it all the time.

As for the "clicky" state of submit buttons when the language is invalid, I am still looking for a way to improve that. This is not easy because the form is actually temporarily getting invalid when the dropdown is showing suggestions. I think it would be much better to get the same behavior as the registration form: the button stays clickable, but clicking it brings you to the invalid field instead of submitting.

jiru:
I looked into the form validation. The problem is that the current implementation only allows form-level validation, not field-level (we cannot know which field in particular is wrong). However, getting field-level validation requires quite of a rewrite of the form submission code. I don’t wanna do that in this PR because the PR is quite big already and I think the red text is good enough as a temporary solution.

mramosch:
First thought this is a bug. After a few days I accidentally discovered that typing letters in the field list other languages beside the profile languages.

There is no ‘Show All’ button visible on an iPad like on the screenshot above...

jiru:
Yes that screenshot is outdated. Now, the "Show all" button only appears when whatever you type yields no result. The fact that you didn’t know you can list other languages by typing something is the very reason I changed the dropdown behavior this way. Typing in should be the way to find a non-profile language because it’s so much more efficient than going through a list of 382 items. Maybe there is a better way to make users realize that, but at least for people who are browsing Tatoeba for the first time I think it should be rather obvious because no language is suggested at first, so the field looks like a regular text field.

mramosch:
Well, if I get a drop down, I assume that I get everything listed that is available without using the keyboard.

Try these scenarios on an iPad:

1. press the text field ‘From’
2. write ‘qw’
3. select ‘show all’
4. press the screen somewhere outside the selection box

The cursor disappears, there is no way to get it back into the text field, the interface is stuck with the language selection box, the only way to opt out is to select a random language...

1. press the text field ‘From’
2. write ‘qw’
3. select ‘show all’
4. write ‘qw’ again -> ‘show all’ doesn’t appear as an option anymore
5. tap on ‘qw’ in the text field

The interface is stuck, no keyboard input possible, the x icon doesn’t clean out the text field nor dismiss the select box, the only way out is to tap the empty language selection box...

6. tap on ‘qw’ in the text field again

‘qw’ turns red now - what does that mean???

7. tap on ‘qw’ in the text field again

‘qw’ turns black again - what does that mean???

8. tap on ‘qw’ in the text field again

text field goes empty...

Doing the same thing three times yields 3 different results in the interface - I am lost...

jiru:
Could you open a new issue instead, including your browser name, browser version and iOS version?

mramosch:
On iOS the search (especially) in iPad Split-Mode is totally broken and unusable now.

A major culprit is the fact that touching any new language select box (with the new flags) already triggers on mouse-down instead of mouse-up like a normal tap/click. The problem is that you can’t scroll because the drag-gesture for scrolling is hijacked by the mouse-down. As there are many language select boxes in the search, chances are small not to hit one, especially in 30:70 split-mode.

Further problems are a stuck search window (x doesn’t close the box) and scroll issues when the two scroll views (grayed out Safari-main-view and overlapping search window) share scroll gestures. Sometimes tapping any of the text entry fields (e.g. owner etc.) and adding a letter and deleting it again solves the stuck window problem, but the long search box cannot be scrolled to the extreme top or bottom, depending on the position of the underlying overlapped grayed out main-view...

jiru:
@mramosch Sorry to repeat myself, but could you open a new issue instead, including your browser name, browser version and iOS version?

Issue #2619, opened 2021-02-10
Mahasu Pahari (bfz) 


RyckRichards:
* Language name: Mahasu Pahari
* Language info: https://en.wikipedia.org/wiki/Mahasu_Pahari
* ISO 639-3 code: BFZ
* Public List: https://tatoeba.org/eng/sentences_lists/show/169218
* Icon: https://imgur.com/a/66phzSy
* Requested by: Rajneesh

PR #2620, opened 2021-02-10
Added Mahasu Pahari

RyckRichards:
This PR solves #2619.
It adds [Mahasu Pahari (bfz)](https://en.wikipedia.org/wiki/Mahasu_Pahari).


jiru:
The PR looks good, thank you! :slightly_smiling_face:

Issue #2561, opened 2020-09-23
Can no longer use Enter key to navigate through search fields in right sidebar


alanfgh:
**To Reproduce**
Steps to reproduce the behavior, on Chrome on an Android device:
1. Do a search (ordinary or regular).
2. On the page with the search results, go to the search sidebar on the right, and put the cursor into the "Words" field at the top. 
3. Change the value in the "Words" field.
4. Press "Enter".

Prior to recent changes, this would take you to the Owner field. Three more presses of the Enter key would take you through Tags and Link Owner and finally your search would be submitted. Compared to the desktop browser, where a single press of the Enter key would submit your search, this was a pain. But at least it was workable.

Now that the UI has changed, hitting Enter wipes out whatever was in the Language field and makes it impossible to do anything until a language is selected. I often want to use the same language as before, so this is a huge step backward in usability for me. The fact that I often have my keyboard configured for Cyrillic while my UI is in English makes things even worse, since I can't type the language name into the field and am forced to scroll through the choices (difficult to do on my tablet). To make things still worse, the language list (which is very long) is mostly hidden by my on-screen keyboard. I have to choose between a bunch of time-consuming alternatives:  close the on-screen keyboard so I can choose a language, change the keyboard to English so I can type a language name, and so on. All this to change a language name which I generally don't even want to change!

This problem could be mitigated somewhat if the submit button (labeled "Search") were at the top rather than the bottom of the page. But I'm not sure whether there would be a key press that would take me from the "Words" field to the submit button, which is what I would really want.

What use case was the change in behavior meant to address, and how much demand was there for a change? If there are people who do want to change the language in search often, and don't mind the fact that Enter wipes out the current value, could there be some way (a configuration option?) that would let us choose the behavior?

trang:
This change was not exactly planned but is the result of unifying the language selectors on every page (#2328).

When you press "Enter" (which is actually "Next"), the browser will jump to the next field that accepts keyboard input. Prior to the changes, the language fields were just regular dropdown lists so there was no keyboard interaction with them, therefore they were skipped. 

With the latest changes, the language fields are using the new language selector which has an autocomplete input. So when you press "Next" from the "Words" field, the next field that accepts keyboard input is no longer "Owner" but "Language".

[...]

PR #2625, opened 2021-02-12
Upgrade AngularJS Material to 1.1.21

jiru:
This PR upgrades AngularJS Material to 1.1.21 in order to solve #2561.

In that very version, they improved `<md-autocomplete>` but they also did a change that got into our way. They enforced a non-empty value for the `name=""` attribute of the underlying `<input>` of `<md-autocomplete>`. We used to purposely not set any name so that the dropdown field is ignored and the form gets submitted without it. Instead, we used a hidden field to send the language code. This is because the dropdown field only contains the localized language name, which is no use to the controller.

So it turned out we cannot really rely on this hidden field trick any more. But here is what I did. I kept the hidden field and to set it with the same `name=""` attribute as the autocomplete input. This makes the form send the same parameter twice, but only the latest is "seen" by controllers as it "overwrites" the others. This solution is okay for POST forms, but GET forms still get a very visible extra parameter in the URL. That said, the only GET forms we have are the search and advanced search. For these, I rewrote the form submission code in Javascript, which happens to be a good thing in my opinion because it paves the way for improving the search experience.


jiru:
I set up this branch on https://dev.tatoeba.org/.

PR #2627, opened 2021-02-15
New user banner now includes QSG link

alanfgh:
New user banner now points to the Quick Start Guide, as suggested by Trang and others.


jiru:
Thank you Alan! Your pull request is all right. The red cross is displayed because I messed up with the code previously.

PR #2636, opened 2021-03-06
Allow custom exports including language pairs

### LINKED ISSUE #1609: Allow more export options on the download page

jiru:
This PR does two things:

1. Add a new section at the bottom of the downloads page to allow anyone to download custom exports. (Closes #1609.)
2. Using that new tool, allow downloading complete sentence exports for any given language pair.

I also removed the hidden /exports/index page because it was not very useful. The new section in the download page should be the main UI for exports.

Code-wise, I made it easier to add new kind of custom exports in the future. Although not yet possible, this PR also paves the way to allow custom fields (i.e. what columns you want in the file) and custom formats (such as spreadsheets, PDF…).

@alanfgh Could you have a look at the wording on https://dev.tatoeba.org/downloads ? In particular, I am not sure about how to express this idea of "language pair" (or "sentences pairs"?) in a clear and concise way.


alanfgh:
@jiru For reference, this is how things look now on the dev site:

custom_export

A question (as much for curiosity as anything else): Is the "List" option offered here exactly like downloading a list from its own page?

I tested the "Language pair" functionality (which is really awesome, by the way, and will please people who have been asking for it for a while). It looks like it only downloads sentences in language A that are translated into language B. This is not clear from the current wording. Maybe I would change the wording as follows:

_Download all sentences in language A along with all translations in a language B_
->
_Download all sentences in language A that are translated into language B, along with the translations_

I'll think about it some more and see whether I can come up with anything else.

You might want to think about putting this at the top of the page rather than the bottom. Also, we should reconsider the note at the upper right that says "The data you will find here will NOT be useful unless you are coding a language tool or processing data." It may no longer be true, depending on the interpretation of "processing data".

jiru:
Thank you for your feedback! :cat: 

> A question (as much for curiosity as anything else): Is the "List" option offered here exactly like downloading a list from its own page?

Yes, it is the same code under the hood that made it possible to lift the 100 sentences limitation. :slightly_smiling_face: 

About the wording, since it is quite long already, maybe we could come up with a shorter version for the button, and then display a longer and more descriptive sentence after clicking on it, right above the two fields "Sentence language" and "Translation language".

I also thought about rewording the various notes of the download page, but I eventually did not, because I think the custom export tool is not so user friendly yet at this stage. There isn’t quite much a normal user can do with these .tsv files, as awesome as they might look to us.

I will try to move the section to the top.

alanfgh:
If you want a shorter version for the button, and then a longer sentence, here's a possibility:

SHORT:
_Sentence pairs_
_Download all sentences in language A with translations in language B_

LONG [identical to what I proposed above]:
_Download all sentences in language A that are translated into language B, along with the translations_

jiru:
@alanfgh I reworded the text as you suggested, moved the custom export section to the top and adjusted the look of the page.

alanfgh:
It looks good. But I would replace this:

_Use this tool to generate and download customized exports on demand._

with this:

_Use these tools to generate and download customized exports on demand._

given that the caption appears above a pair of buttons rather than just one.

alanfgh:
My previous comment presumes that you leave the list button where it is. But I find that button confusing for several reasons:
(1) It's not clear whether that function differs from simply downloading a list from its own page (though I found out from you that it doesn't).
(2)  The box in the corner still says: "If you simply want sentences that you can use to learn a language, check out the _sentence lists_. You can build your own, or view the ones that others have created. The lists can be downloaded and printed." This compounds my confusion.

jiru:
@alanfgh Thank you for sharing your opinion. I agree with you that it is confusing. To address this problem I simply removed the list exports button from the custom exports section. About the phrasing "this tool" vs. "these tools", I phrased it like this because the way I see it there is just one "custom export" tool that can produce different kind of custom exports. Anyway, now that there is only one button left, only the singular makes sense, so I left it like this.

LBeaudoux:
Thanks @jiru  for taking the time to implement this feature. Tatoeba users will love it.

While testing, I noticed that it took more than three minutes to generate the file of English sentences translated into German.
By processing [weekly monolingual files](https://downloads.tatoeba.org/exports/per_language/) with [pandas](https://pandas.pydata.org/), I managed to generate the same file in less than 7 seconds, and this without overloading the website database. Besides, using pandas does not prevent from filtering the results to keep for example only sentences created in a certain time interval, owned by a native speaker, etc...

This performance gain seems quite significant to me. What do you think about it?

jiru:
You approach might be fast, but how much RAM does it use? There is typically a tradeoff between execution speed and memory consumption. Besides, since the database already holds most of the data in RAM, additionally loading the CSV into RAM feels like a waste of memory to me.

Also, we need to keep in mind that we perform the export on the same machine that runs the whole website, so one lenghty export could very well slow down other’s browsing experience. To avoid that, the export is purposely slowing down itself when the load on mysqld is too high. In addition, only one export can be performed at a time, so if someone else is exporting, you have to "wait for your turn". Finally, the export queue idles for up to 10 seconds before actually starting the export, but this should be removed if we upgrade to Cakephp 3.9. These are the reasons it took 3 minutes. Normally it would have taken no more than one minute. You can try on https://dev.tatoeba.org/ to compare with a close-to-idle system (even though there is a bit less data to export out there).

While any speedup would be a welcomed improvement, I don't think it so so much of a concern for the "average user" because the export typically a one-time operation. At the moment, I am more concerned about the lack of feedback about the progress of the export generation. One might think nothing's happening and just close the tab. (Then again, you might argue that there is no need for such feedback for a 7-seconds export. :wink:)

I am not declining your suggestion (it is even quite appealing), but I just think speed is one of the many things to consider. We have to consider all the advantages and drawbacks in terms of impact on the system, resource consumption, features, maintainability… Could you perform some RAM usage measurements on your tool?

LBeaudoux:
I admit that speed is not the main consideration here, especially if a progress bar notifies the user.

As you asked, I just profiled the memory consumption of my English to German export and it seems that the script consumes about 250 MiB of memory. This does not seem extravagant to me, but I admit that it is not insignificant either...

In addition to the speed, another plus is that this implementation would be more consistent with the 'standard downloads' on the page, and would allow to cache custom exports files as long as a newer weekly dump is not available.

jiru:
Well 250MB feels like something could be optimized but it is still reasonable I guess. Just to make be sure, could you also report the peak RAM usage using `/usr/bin/time -f "max RAM usage: %MKB" your_export_command`. Also is your code available on a public repository?

LBeaudoux:
For now, my script relies on the [tatoebatools](https://pypi.org/project/tatoebatools/) library:


#### **`export_parallel_corpus.py`**
```python
import csv

from tatoebatools import ParallelCorpus


def export_parallel_corpus(src_lang, tgt_lang, cols, out_dir):

    df = ParallelCorpus(src_lang, tgt_lang)._df.loc[:, cols]
    df.sort_values(by=["sentence_id"], inplace=True)
    df.to_csv(
        f"{out_dir}/{src_lang}-{tgt_lang}_parallel_corpus.tsv",
        sep="\t",
        header=False,
        index=False,
        quoting=csv.QUOTE_NONE,
    )


if __name__ == "__main__":

    src_lg = "eng"
    tgt_lg = "deu"

    cols = [
        "sentence_id",
        "text_sentence",
        "translation_id",
        "text_translation",
    ]

    out_dir = "/my/out/directory"

    export_parallel_corpus(src_lg, tgt_lg, cols, out_dir)
```

Here is what I get with your  RAM profiling tool:
```sh
$ /usr/bin/time -f "max RAM usage: %MKB" python export_parallel_corpus.py
max RAM usage: 480968KB
```

I previously used the [memory-profiler](https://pypi.org/project/memory-profiler/) library:
```sh
$ python -m memory_profiler export_parallel_corpus.py
Filename: export_parallel_corpus.py

Line #    Mem usage    Increment  Occurences   Line Contents
============================================================
     8     84.6 MiB     84.6 MiB           1   @profile
     9                                         def export_parallel_corpus(src_lang, tgt_lang, cols, out_dir):
    10                                         
    11    233.6 MiB    148.9 MiB           1       df = ParallelCorpus(src_lang, tgt_lang)._df.loc[:, cols]
    12    233.6 MiB      0.0 MiB           1       df.sort_values(by=["sentence_id"], inplace=True)
    13    233.7 MiB      0.1 MiB           2       df.to_csv(
    14    233.6 MiB      0.0 MiB           1           f"{out_dir}/{src_lang}-{tgt_lang}_parallel_corpus.tsv",
    15    233.6 MiB      0.0 MiB           1           sep="\t",
    16    233.6 MiB      0.0 MiB           1           header=False,
    17    233.6 MiB      0.0 MiB           1           index=False,
    18    233.6 MiB      0.0 MiB           1           quoting=csv.QUOTE_NONE,
    19                                             )

```
I wonder why the results vary by a factor of two.

Regarding memory optimization, `tatoebatools` already only loads useful rows.  In addition, only required columns could be loaded. Data chunking is also a possibility.

LBeaudoux:
I just took a look at the indexes on the `sentences_translations` table.

```sql
MariaDB [tatoeba]> SELECT index_name, seq_in_index, column_name FROM information_schema.statistics WHERE table_schema='tatoeba' AND table_name='sentences_translations'; 
+------------------+--------------+------------------+
| index_name       | seq_in_index | column_name      |
+------------------+--------------+------------------+
| PRIMARY          |            1 | id               |
| sentence_id      |            1 | sentence_id      |
| sentence_id      |            2 | translation_id   |
| translation_id   |            1 | translation_id   |
| sentence_lang    |            1 | sentence_lang    |
| translation_lang |            1 | translation_lang |
+------------------+--------------+------------------+
6 rows in set (0.001 sec)

```
It seems that there is no multiple column index on both `sentence_lang` and `translation_lang`. Maybe we could replace the single index on `sentence_lang` by a multiple index on `sentence_lang` and `translation_lang` to speed up queries on language pairs.

jiru:
Sure. Feel free to submit a PR along with some benchmarks.

Issue #2671, opened 2021-03-29
Tatodetect’s n-grams generation script is suboptimal


jiru:
The [n-grams generation script](https://github.com/Tatoeba/Tatodetect/blob/master/tools/generate.py) is executed every week. It consumes about 1.5 GB of RAM lately. While this causes no serious harm, it reduces cache for other processes, and RAM usage can only keep growing as more sentences and languages are being added.

**Ideas**
* Instead of calculating all stats in memory and write them all at once to the database, insert and increment the n-gram `hit` column directly in the sqlite database while reading the CSV file. Then, only after the CSV file is completely read, update the `percent` column by dividing the hit count by the total number of ngrams read.
* Calculate n-grams stats one language at a time instead of all languages at once, by reading from the per-language export files instead of sentences_detailed.csv.

LBeaudoux:
In my opinion,, incrementing the hits directly in the sqlite table would substantially slow down the script. Splitting the task by language seems more promising. 

Could I be assigned to this task ?

jiru:
Sure, go ahead 🙂

jiru:
> In my opinion,, incrementing the hits directly in the sqlite table would substantially slow down the script. Splitting the task by language seems more promising.

I don’t know about sqlite internals but maybe there is a way to have it insert data in memory, and only flush the database to disk when the amount of used RAM reaches a threshold or something.

Note that the main problem here is RAM usage, not speed. Since the script is executed in the background with `nice -n 19` once a week, it doesn’t really make a difference if it takes 10 minutes, 1 hour or even 2 hours. Incrementing the hits directly in the sqlite table makes RAM usage no longer bound to the total number of ngrams, which depends on the number and variety of sentences. This means no matter how big the corpus grows in the future, RAM usage should not increase significantly.

PR #2673, opened 2021-03-29
Avoid deletion of weekly export datafiles after zipping

LBeaudoux:
During the weekly exports, most of the generated data files are deleted once they are available in compressed form. It would be preferable to keep them, as they may be useful in the following use cases:
- [customized exports](https://github.com/Tatoeba/tatoeba2/pull/2636) from language-split data files
- [Todetect n-grams](https://github.com/Tatoeba/tatoeba2/issues/2671) update  by language


jiru:
Looks good, thank you. 🙂 Could you maybe explain the reason you are doing this in your commit message? Explain that we need to keep uncompressed versions for other processings such as in Tatodetect.

jiru:
@LBeaudoux Thank you for rewording the commit message. I suggest adding a multiline description in the commit message, see for example [this guide](https://chris.beams.io/posts/git-commit/) about how to write good commit messages.

I am having second thoughts on this PR. We are quite short on disk space on the production server and your PR increases the space used for exports by 2.1GB. Maybe I can free up some disk space here and there but basically we cannot really afford that at the moment. In addition, to know if it is really worth keeping uncompressed files, we’d have to measure the overhead of on-the-fly decompression when processing the files. Maybe it’s not that of a big deal.

LBeaudoux:
@jiru Originally, this PR was primarily intended to speed up the generation of custom exports. But since [this](https://github.com/Tatoeba/tatoeba2/pull/2681) other pull request might have a similar effect without consuming as much storage. I don't see any harm in dropping it for now..

Thanks for your reading tips, it helps me improve.

jiru:
In this case, I will close this PR for now. If the language pair export code ever gets rewritten to read data from the CSVs instead of the database, we can reconsider keeping uncompressed versions.

PR #2681, opened 2021-04-06
Speed up custom export of parallel corpus

LBeaudoux:
This pull request follows [this discussion](https://github.com/Tatoeba/tatoeba2/pull/2636#issuecomment-813416868) about the speed of custom exports.

As requested by @Jiru, I performed a benchmark based on the following query:
```sql
SELECT sentence_id, translation_id 
FROM sentences_translations 
WHERE sentence_lang='eng' and translation_lang='deu';
```
The results are quite impressive:

- with a single-column index on `sentence_lang`: `347969 rows in set (1 min 41.922 sec)`
- with a multiple-columns index on `sentence_lang` and `translation_lang`:  `347969 rows in set (1.208 sec)`

This PR provides:

- a SQL script that replaces the single-column index by the new composite index
- an updated definition of the `sentences_translations`table

Note that the former `sentence_lang` index is still mentioned in a few fixture files, but as I am not familiar with the inner workings of Tatoeba, I left them unchanged.


jiru:
Thanks! :+1:

We don’t use the scripts under `docs/database/updates/` any more. Please revert all the changes in `docs/database/` and use the command `cake migrations create` instead to generate a migration script (or just write it yourself). You can have a look at the file `config/Migrations/20210325195224_UserLangIndex.php` for an example.

jiru:
I did some measurements on dev.tatoeba.org using the English-German pair.

* The migration took 59 seconds to execute.
* The custom export generation time went from 45-50 seconds to 20-25 seconds.
* The underlining SQL query for the export now takes 6.79 seconds to execute when executed from the mysql console (`SELECT SQL_NO_CACHE Links.sentence_id AS Links__sentence_id, Links.translation_id AS Links__translation_id, Sentences.id AS Sentences__id, Sentences.text AS Sentences__text, Translations.id AS Translations__id, Translations.text AS Translations__text FROM sentences_translations Links INNER JOIN sentences Sentences ON Sentences.id = (Links.sentence_id) INNER JOIN sentences Translations ON Translations.id = (Links.translation_id) WHERE (Links.sentence_lang = 'eng' AND Links.translation_lang = 'deu');`).
* The index size of the `sentences_translations` grew by +314MB (+25%).

These are quite good results! Thanks again, @LBeaudoux! :+1:

Issue #2672, opened 2021-03-29
Adding Traditional Chinese (Taiwan) as website display language


albb0920:
I've already sent a request through transifex, but since it hasn't been processed yet, I figured to request it here too.

Traditional Chinese is a different script of Chinese, here in Taiwan we generally prefer to read Traditional Chinese, the ability to read Simplified Chinese varies person or person.

We also have different ways of saying certain things, although translation of vocabulary can be done by a giant list (incomplete example: https://github.com/tongwentang/tongwen-core/blob/master/dictionaries/s2t-phrase.json ), preferred sentence structure can't.

I'd like to help translate tatoeba to `zh-TW`, please considering adding this to transifex, thanks

RyckRichards:
Hi there! Good to meet you!

Since I'm in charge of localization on Tatoeba ("recruiting" people, handling requests, etc,), I'm here to help.

Tatoeba always follows the ISO639-3 code in order to accept it as a language request or not - https://iso639-3.sil.org/code_tables/639/data / // https://en.wikipedia.org/wiki/ISO_639-3. According to it, Taiwanese Mandarin has no ISO 639-3 code - https://en.wikipedia.org/wiki/Taiwanese_Mandarin

![image](https://user-images.githubusercontent.com/7658430/112791919-d3bed780-9038-11eb-99dd-485cabfccb82.png)

We can try to figure out how to handle that but as far SIL didn't established a ISO 639-3 code, I believe we can't do much.

albb0920:
Thanks you for your prompt response!

Since this issue is here to request a new UI language, at the technical level, I believe this project is using IETF language tag to store UI strings, which is very common and can support our written language as `zh-Hant-TW`, or `zh-TW`.
https://github.com/Tatoeba/tatoeba2/tree/dev/src/Locale

I personally find dedicated `ISO639-3` code as criteria for UI language quite absurd, as `ISO639-3` more or less only distinguish spoken language. There are various languages have different written language (eg. Mongolian).
IMHO, for UI language, written language are more relevant.

As for the language code for data in tatoeba, if there's no intention to use more satisfied code scheme, could assigning a code in qaa–qtz be an option?

Thanks

jiru:
[...]

PR #2689, opened 2021-04-19
Allow region-specific UI languages

jiru:
tl;dr: this PR changes the way UI language is encoded in the URL in order to solve #2672.

The existing UI language identifiers have been made to follow ISO-639-3 in order to keep the same language codes as sentences. While this approach sounds right because it makes both UI and corpus rely on the same list, it has limitations:
* we need to define the current UI language in a separate global config variable `Config.language` because it’s different from the actual locale, which is set with `I18n::setLocale()`
* we cannot provide region-specific UI language variants (whereas Transifex allows them, for good reasons), and there was a need for that recently with Taiwanese Chinese in #2672

In order to get rid of these limitations, this PR makes use of [BCP 47](https://en.wikipedia.org/wiki/IETF_language_tag) as UI language identifier. In particular, `cmn` gets changed to `zh-cn`, `por` to `pt-br` and `zh-tw` is added. All previous 3-letter codes are properly redirected, so old URLs work as excepted.

In addition, it looks like this PR solves the (unreported?) issue of font selection for Chinese UI. When Chinese is selected, the tag `<html>` currently has the attribute `lang="zh"` which doesn’t let the browser know if it should choose a font tailored for simplified or traditional characters. With this PR, it now sets it to `lang="zh-cn"` and the font look different if your system is configured with a specific font for "Chinese of mainland China".


jiru:
Note to self: we also need to change the wiki language identifiers from pt to pt-br and zh to zh-cn, and ideally have a redirection for these wikis.

Issue #2669, opened 2021-03-26
Yet another API attempt


githubshrek:
Author: Hugh
Tatoeba: https://tatoeba.org/eng/user/profile/tatoebashrek 

# Introduction
## Problem definition
I would like to build an API for Tatoeba to allow easy and direct use by other websites.

I am the owner of the language-learning website https://lingopolo.org/ which is a sort of cross between Anki (flashcard tool), Duolingo (language-learning tool) and Tatoeba. It builds language lessons automatically based on the added sentence content. 

Recently we have started to use content from Tatoeba, e.g. see the Tatoeba acknowledgement at the bottom of these pages:
* https://lingopolo.org/thai/word/why-didnt-you-call-me-last-week 
* https://lingopolo.org/thai/word/i-ran-across-my-old-teacher-on-the-street 
* https://lingopolo.org/dutch/word/tom-hasnt-eaten-enough 
* https://lingopolo.org/dutch/word/most-of-my-friends-are-still-single 

What we would like to do is automatically add sentences from Tatoeba, directly and programatically from the Lingopolo website. The Lingopolo website would regularly, via cron, make requests to Tatoeba of the type:
* give me the ids of 100 sentences in NL with an EN translation and audio
* for each of the ids, give me the full content of that particular sentence, such as:
  * id
  * EN text
  * NL text
  * copyright information
  * user
  * audio

## Proposed solution summary
This describes part of a possible Tatoeba API, which would allow users to directly via API (on a URL):
query the ids of sentences matching a particular tag
query an individual sentence

I see that there have been some attempts at building an API before:
* in 2014, https://github.com/Akirato/tatoeba-api 
* in 2015, https://github.com/Tatoeba/tatoeba-api 
* in 2020, https://github.com/Tatoeba/tatoeba2/pull/2341 

## Lingopolo
Lingopolo consists of 2 developers, myself and Alex. It is based on Drupal (which means it is written in PHP like Tatoeba). We would anticipate that we would do most of the development of this API.

## Useful links on API structure
There are many external APIs which it would be worth studying to get the best structure, but here are a few recent links I was looking at.

Wikidata API:
* https://www.wikidata.org/wiki/Wikidata:Data_access 

See how Wikidata offers two formats, one which shows HTML and one which shows Json, by adding the "format=json" parameter:
* https://www.wikidata.org/w/api.php?action=wbgetentities&ids=Q9128 
[...]

PR #2778, opened 2021-05-08
Yet another API attempt

githubshrek:
Add API for sentence and search. Addressing issue https://github.com/Tatoeba/tatoeba2/issues/2669

### DIFF
--- a/src/Controller/ApiV1Controller.php
+++ b/src/Controller/ApiV1Controller.php
@@ -1,0 +1,211 @@
+<?php
+/**
+ * Tatoeba Project, free collaborative creation of multilingual corpuses project
+ * Copyright (C) 2009-2010  HO Ngoc Phuong Trang <tranglich@gmail.com>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * PHP version 5
+ *
+ * @category PHP
+ * @package  Tatoeba
+ * @author   HO Ngoc Phuong Trang <tranglich@gmail.com>
+ * @license  Affero General Public License
+ * @link     https://tatoeba.org
+ */
+namespace App\Controller;
+
+use App\Controller\AppController;
+use App\Form\SentencesSearchForm;
+use App\Model\CurrentUser;
+use App\Model\Table\SentencesTable;
+use App\Lib\LanguagesLib;
+use App\Lib\SphinxClient;
+use App\Lib\Licenses;
+use Cake\Core\Configure;
+use Cake\Database\Expression\QueryExpression;
+use Cake\Event\Event;
+use Cake\Routing\Router;
+use Cake\Utility\Hash;
+use Cake\View\ViewBuilder;
+use Exception;
+
+/**
+ * Controller for Api.
+ *
+ * @category Sentences
+ * @package  Controllers
+ * @author   HO Ngoc Phuong Trang <tranglich@gmail.com>
+ * @license  Affero General Public License
+ * @link     https://tatoeba.org
+ */
+class ApiV1Controller extends AppController
+{
+    public $name = 'Sentences';
+    public $components = array (
+        'LanguageDetection',
+        'CommonSentence',
+        'Permissions',
+    );
+    public $helpers = array(
+        'Sentences',
+        'Menu',
+        'Lists',
+        'SentenceButtons',
+        'Html',
+        'Logs',
+        'Pagination',
+        'Comments',
+        'Languages',
+        'CommonModules'
+    );
+
+    public $uses = array(
+        'Audio',
+        'Sentence',
+        'SentencesSentencesLists',
+        'SentencesList',
+        'User',
+        'UsersLanguages',
+        'Tag',
+        'UsersSentences',
+        'Vocabulary'
+    );
+
+    public function initialize() {
+        parent::initialize();
+
+        $params = $this->request->params;
+        $noCsrfActions = ['edit_sentence', 'change_language'];
+        if (in_array($params['action'], $noCsrfActions)) {
+            $this->components()->unload('Csrf');
+        }
+    }
+
+    /**
+     * Show Json of specified sentence id.
+     * 
+     * @param mixed $id Id of the sentence.
+     * 
+     * @return void
+     */
+    public function sentence($id = null) {
+        // We will handle the output directly, so as not to display the header and footer
+        $this->disableAutoRender();
+
+        if (is_numeric($id)) {
+            // Retrieve the sentence
+            $sentence = $this->Sentences->getSentenceWith($id, [
+                'translations' => true,
+                'sentenceDetails' => true
+            ]);
+
+
+            return $this->getResponse()
+                ->withStringBody($sentence);
+        }
+    }
+
+    /**
+     * Show in Json format a list of sentence ids which matches a given search.
+     * 
+     * Note that this method is basically a copy/paste of the SentencesController.php 
+     * search() method. Only the minimum changes necessary to deliver an array of ids
+     * have been made. Therefore, there may be some redundant code below.
+     *
+     * @return void
+     */
+    public function search()
+    {
+        // We will handle the output directly, so as not to display the header and footer
+        $this->disableAutoRender();
+
+        if (!Configure::read('Search.enabled')) {
+            $this->render('search_disabled');
+            return;
+        }
+
+        /* Apply search criteria and sort */
+        $search = new SentencesSearchForm();
+        $search->setData($this->request->getQueryParams());
+
+        /* Control input */
+        if ($search->generateRandomSeedIfNeeded()) {
+            return $this->redirect(Router::url($search->getData()));
+        }
+        $search->checkUnwantedCombinations();
+
+        $limit = CurrentUser::getSetting('sentences_per_page');
+        $sphinx = $search->asSphinx();
+        $sphinx['page'] = $this->request->query('page');
+//        $sphinx['limit'] = $limit;
+
+        // TODO: improve on hard-coded limit to the number of API results
+        $sphinx['limit'] = 100; // Limit the API to returning 100 entries.
+
+        $model = 'Sentences';
+        $query = $this->Sentences
+            ->find('hideFields')
+            ->find('withSphinx')
+            ->find('nativeMarker')
+            ->find('filteredTranslations', [
+                'translationLang' => $search->getData('to'),
+            ])
+            ->select($this->Sentences->fields())
+            ->contain($this->Sentences->contain(['translations' => true]));
+
+        $this->paginate = [
+            'limit' => $limit,
+            'sphinx' => $sphinx,
+        ];
+
+        try {
+            $results = $this->paginate($query);
+            $real_total = $this->Sentences->getRealTotal();
+            $results = $this->Sentences->addHighlightMarkers($results);
+            $this->set(compact('results', 'real_total'));
+        } catch (Exception $e) {
+            $syntax_error = strpos($e->getMessage(), 'syntax error,') !== FALSE;
+            if ($syntax_error) {
+                $this->set('syntax_error', true);
+            } else {
+                $this->loadComponent('Error');
+                $error_code = $this->Error->traceError('Search error: ' . $e->getMessage());
+                $this->set('error_code', $error_code);
+            }
+        }
+
+        $strippedQuery = preg_replace('/"|=/', '', $search->getData('query'));
+        $this->loadModel('Vocabulary');
+        $vocabulary = $this->Vocabulary->findByText($strippedQuery);
+
+        $searchableLists = $search->getSearchableLists(CurrentUser::get('id'));
+        $ignored = $search->getIgnoredFields();
+
+        $this->set($search->getData());
+        $this->set(compact('ignored', 'searchableLists', 'vocabulary'));
+        $this->set(
+            'is_advanced_search',
+            !is_null($this->request->getQuery('trans_to'))
+        );
+
+        $ids = array();
+        foreach ($results as $sentence) {
+            $ids[] = $sentence->id;
+        }
+    
+        $ids_json = json_encode($ids);
+        echo '{ "ids" : ' . $ids_json . ' }';
+    }
+}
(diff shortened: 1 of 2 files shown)


jiru:
Thanks for the PR :slightly_smiling_face:
I’m going to review your code.

src/Controller/ApiV1Controller.php
```diff
@@ -0,0 +1,211 @@
+<?php
+/**
+ * Tatoeba Project, free collaborative creation of multilingual corpuses project
+ * Copyright (C) 2009-2010  HO Ngoc Phuong Trang <tranglich@gmail.com>
```
jiru, line 4: Please either use correct info here (year, name and email) or do not include a license header at all.

src/Controller/ApiV1Controller.php
```diff
@@ -0,0 +1,211 @@
...
+use App\Model\Table\SentencesTable;
+use App\Lib\LanguagesLib;
+use App\Lib\SphinxClient;
+use App\Lib\Licenses;
+use Cake\Core\Configure;
+use Cake\Database\Expression\QueryExpression;
+use Cake\Event\Event;
+use Cake\Routing\Router;
+use Cake\Utility\Hash;
+use Cake\View\ViewBuilder;
```
jiru, line 41: There are some unnecessary `use` here. For example, I don’t see any SentencesTable, LanguagesLib or SphinxClient in the rest of the file.

src/Controller/ApiV1Controller.php
```diff
@@ -0,0 +1,211 @@
...
+use Exception;
+
+/**
+ * Controller for Api.
+ *
+ * @category Sentences
+ * @package  Controllers
+ * @author   HO Ngoc Phuong Trang <tranglich@gmail.com>
```
jiru, line 49: This `@author` annotation is incorrect.

src/Controller/ApiV1Controller.php
```diff
@@ -0,0 +1,211 @@
...
+ * @link     https://tatoeba.org
+ */
+class ApiV1Controller extends AppController
+{
+    public $name = 'Sentences';
+    public $components = array (
+        'LanguageDetection',
+        'CommonSentence',
+        'Permissions',
+    );
```
jiru, line 60: It appears you are not using any of these components, so I think it is safe to remove them.

src/Controller/ApiV1Controller.php
```diff
@@ -0,0 +1,211 @@
...
+ *
+ * @category Sentences
+ * @package  Controllers
+ * @author   HO Ngoc Phuong Trang <tranglich@gmail.com>
+ * @license  Affero General Public License
+ * @link     https://tatoeba.org
+ */
+class ApiV1Controller extends AppController
+{
+    public $name = 'Sentences';
```
jiru, line 55: I think this is not the correct name for this controller. You can simply remove this line and CakePHP will derive the name from the class name.

src/Controller/ApiV1Controller.php
```diff
@@ -0,0 +1,211 @@
...
+        'Menu',
+        'Lists',
+        'SentenceButtons',
+        'Html',
+        'Logs',
+        'Pagination',
+        'Comments',
+        'Languages',
+        'CommonModules'
+    );
```
jiru, line 72: Since you are not using any view template, you shouldn’t need any of these helpers.

src/Controller/ApiV1Controller.php
```diff
@@ -0,0 +1,211 @@
...
+        'Audio',
+        'Sentence',
+        'SentencesSentencesLists',
+        'SentencesList',
+        'User',
+        'UsersLanguages',
+        'Tag',
+        'UsersSentences',
+        'Vocabulary'
+    );
```
jiru, line 84: This array is a relic from CakePHP 2 that can be safely removed. With this array we used to load all the models we needed for all actions, but with CakePHP 3 you need to call `$this->loadModel()` instead.

src/Controller/ApiV1Controller.php
```diff
@@ -0,0 +1,211 @@
...
+    public function initialize() {
+        parent::initialize();
+
+        $params = $this->request->params;
+        $noCsrfActions = ['edit_sentence', 'change_language'];
+        if (in_array($params['action'], $noCsrfActions)) {
+            $this->components()->unload('Csrf');
+        }
+    }
+
```
jiru, line 95: I don’t think this method is doing anything useful for you because you don’t have `edit_sentence` nor `change_language` actions.

src/Controller/ApiV1Controller.php
```diff
@@ -0,0 +1,211 @@
...
+    /**
+     * Show Json of specified sentence id.
+     * 
+     * @param mixed $id Id of the sentence.
+     * 
+     * @return void
+     */
+    public function sentence($id = null) {
```
jiru, line 103: By writing ` = null`, you are defining the `$id` parameter as optional. This will make the `/api_v1/sentence` a valid URL that will call this method with `$id` set to null. However, I think you want to enforce this parameter instead.

src/Controller/ApiV1Controller.php
githubshrek: OK, will do.

src/Controller/ApiV1Controller.php
githubshrek: Yeah, I just copy/pasted the SentencesController.php and changed as little as possible. I'll try and remove anything not being used.

src/Controller/ApiV1Controller.php
githubshrek: Hey, I get to credit myself, great! Will do!

src/Controller/ApiV1Controller.php
githubshrek: OK.

src/Controller/ApiV1Controller.php
githubshrek: OK.

src/Controller/ApiV1Controller.php
githubshrek: OK, I clear this up.

src/Controller/ApiV1Controller.php
githubshrek: OK, will clean up.

src/Controller/ApiV1Controller.php
githubshrek: Probably right, I'll try without it.

src/Controller/ApiV1Controller.php
githubshrek: OK

src/Controller/ApiV1Controller.php
```diff
@@ -0,0 +1,211 @@
...
+        // We will handle the output directly, so as not to display the header and footer
+        $this->disableAutoRender();
+
+        if (is_numeric($id)) {
```
jiru, line 107: I don’t think we really this `[is_numeric()](https://www.php.net/is_numeric)` check here. The method should return error when the sentence cannot be found, be it because `$id` is not numeric (e.g. `/api_v1/sentence/hello`) or because the provided number doesn’t match any sentence (e.g. `/api_v1/sentence/99999999999999999`). In both cases, `getSentenceWith()` will return null so all you need to do is to check against this returned value.

src/Controller/ApiV1Controller.php
jiru, line 103: I suggest renaming this method from `sentence` to `sentences`. Using plural in this situation is a well-known API naming convention.

src/Controller/ApiV1Controller.php
```diff
@@ -0,0 +1,211 @@
...
+                ->withStringBody($sentence);
+        }
+    }
+
+    /**
+     * Show in Json format a list of sentence ids which matches a given search.
+     * 
+     * Note that this method is basically a copy/paste of the SentencesController.php 
+     * search() method. Only the minimum changes necessary to deliver an array of ids
+     * have been made. Therefore, there may be some redundant code below.
```
jiru, line 125: Please remove the redundant code instead of writing a comment like this. Alternatively, you can remove the search method from your PR and submit it another time.

src/Controller/ApiV1Controller.php
```diff
@@ -0,0 +1,211 @@
...
+            !is_null($this->request->getQuery('trans_to'))
+        );
+
+        $ids = array();
+        foreach ($results as $sentence) {
+            $ids[] = $sentence->id;
+        }
+    
+        $ids_json = json_encode($ids);
+        echo '{ "ids" : ' . $ids_json . ' }';
```
jiru, line 209: Using `echo` here prevents this method from being properly tested with a unit test. You should do it the CakePHP way, for example like you did in the `sentence()` method.

jiru:
If you want to implement the search endpoint, I don’t recommend using `SentencesController::search()` as a base because it relies on `App\Form\SentencesSearchForm`, which is basically an HTML-form-aware wrapper of the actual search class `App\Model\Search`. `SentencesSearchForm` was built around the advanced search form, which may change as we see fit. In contrast, we want the parameters of the API search endpoint to be stable. And quite frankly, there is room for improvement in the way search criteria are provided. In the API, we don’t need them to match any kind of HTML form. That’s why I think we should use a different implementation than `SentencesSearchForm`, either in a separate class or directly using `App\Model\Search` from the controller.

src/Controller/ApiV1Controller.php
githubshrek: Heading changed to:
image

src/Controller/ApiV1Controller.php
githubshrek: Changed to:
image

src/Controller/ApiV1Controller.php
githubshrek: Gone:
image

src/Controller/ApiV1Controller.php
githubshrek: In theory that sounds like it makes sense.

In practice, commenting it out breaks everything:
image

with the referenced line 93:
image

image

Line 143 looks a bit weird:
image

I'm not sure why it breaks things commenting it out, but it works with it in.

src/Controller/ApiV1Controller.php
githubshrek: For my use case, https://github.com/Tatoeba/tatoeba2/issues/2669 I'm fine with just getting a blank back. What would you like returned in case of bad data?

src/Controller/ApiV1Controller.php
githubshrek: Meh, on this one I'm currently not so hot about changing it. 

https://nordicapis.com/10-best-practices-for-naming-api-endpoints/ says:
image

I think the "Admittedly this is a matter of preference" suggests it depends. I certainly think "sentence/5" (implying sentence number 5) rather than "sentences/5" (implying 5 sentences?) appeals more to me.

src/Controller/ApiV1Controller.php
githubshrek: Removed echo, changed to:
image

src/Controller/ApiV1Controller.php
jiru: Do you remember when you had to insert some `$this->loadModel()` after copy-pasting code from a different controller? It’s a similar problem here.

Until now, the controller was named "Sentences", so the model "Sentences" was automatically loaded by CakePHP. Therefore, your `sentence()` method could access the model with `$this->Sentences->...` directly. However, now that you renamed the controller, the "Sentences" is not automatically loaded any more, so you need to `loadModel()` it. If there was an "ApiV1" model, CakePHP would load if for you, but obviously there is no such model, so no model is automatically loaded.

src/Controller/ApiV1Controller.php
jiru: Yes. As scary as it may look, I think the "too few arguments" error is what we want here because `/api_v1/sentence/` is not a valid URI. Since we are building an API, the point is reply with an appropriate [HTTP status code](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes), and we don’t care about how it looks in the browser. Basically you should return the status 200 if there is a sentence, otherwise some 4xx or 5xx error. One way to check the actual return code is to use the "Developer Tools" of your browser (in the "Network" tab).

src/Controller/ApiV1Controller.php
jiru: The proper way for an API to handle this is to return a 4xx error, for example 400. To do that in CakePHP, just throw the exception `\Cake\Http\Exception\NotFoundException`. For an example, you can have a look at this code in ExportsController:
https://github.com/Tatoeba/tatoeba2/blob/dcbad1e3c9d29ce4c3ab3c08c9a774605c21be7d/src/Controller/ExportsController.php#L50-L73

src/Controller/ApiV1Controller.php
jiru: Yeah it’s probably a matter of perspective. The way I see "/sentences/5" is "number 5 inside the directory of sentences". I don’t think we need to argue on such detail at that point; just leave it as you want.

jiru:
By the way, you don’t need to include screenshots of all your code changes. I can see them all as long as you `git push` your changes. :cat:

jiru:
Oops, it looks like I overwrote your last comment with my answer, I don’t know how that happened... I restored your comment and here is my answer.

> You're probably right. I think for now though, this is as clean as my knowledge of the code can make it. The reason I have implemented like this is that I can start with the working /search/ and just copy/paste and make small adjustments.

I understand why you did this. It makes sense to get inspiration from other parts of the code that perform similar tasks. I think all developers copy-paste code to some extent. You are free to copy-paste around to locally experiment with the code. However, I don’t want to knowingly merge copy-pasted code into the main branch. The reason is that not only [copy-pasting](https://en.wikipedia.org/wiki/Copy-and-paste_programming) is a source or many current of future bugs, but it also generates a sort of maintenance hell known as [technical debt](https://en.wikipedia.org/wiki/Technical_debt). It is necessary to avoid it if we want to keep a clean code base for Tatoeba.

If your goal is to quickly get something working on tatoeba.org, maybe you can call the `SentenceController::search()` method from your `search()` method instead. Otherwise, please remove unnecessary code from it, or rewrite it with `App\Model\Search`, or remove it entirely from this PR (it could come later as a separate PR; I can also help you with the implementation :slightly_smiling_face:).

jiru:
By the way, why did you close the PR?

githubshrek:
I wanted to have a 2nd branch for the https://github.com/Tatoeba/tatoeba2/issues/2781 issue, so I renamed the "dev" branch, which I had been using for this "Yet another API attempt" issue:
image

I think that closed the pull request automatically, I guess since it didn't know anymore the associated branch.

src/Controller/ApiV1Controller.php
githubshrek: OK, I've removed the "public $name = 'Sentences';" line and added loadModel('Sentences') to both the sentence() and search() methods:
image

src/Controller/ApiV1Controller.php
githubshrek: It doesn't seem to find the "NotFoundException()". 
image

I'm not sure how or where to load it. I looked at ExportsController.php and did "use" of the same items there, but didn't fix things. For reference, my code currently:
image

PR #2784, opened 2021-05-13
Ensure fprintf is not called with arbitrary format strings

Yorwba:
The CLDRCountriesShell contained some `fprintf` calls with a dynamically constructed string passed as the format string, instead of as a `%s` parameter. This caused incorrect behavior when the string contained percentage signs.


jiru:
Thank you! :+1:

Issue #1638, opened 2018-08-09
New rephrasing of "My sentence logs" string


Guybrush88:
As suggested by Aiji on the Wall:

> When the mouse is on our username on the top-right corner, the item before the last is "My sentence logs". I think it would be better to call that "My contribution logs" (even the title of the page is "Logs of username's contributions")
> 
> Voilà voilà... a very important point, isn't it?

https://tatoeba.org/ita/wall/show_message/29569#message_29569

agrodet:
Actually, I think the problem is reversed and is due to the use of the word "contribution". This page lists only sentence-related CUD contributions. I think "My contribution logs" would still be better than the current "My sentence logs" as it would be consistent with the current usage of "contribution", although not perfect.

jiru:
I agree. Logs is a vague word that proved to be [hard to understand](https://en.wiki.tatoeba.org/articles/show/ux-test-1) for a new user.

agrodet:
What about simply using the title of the page? "My latest contributions"  
We will still have the problem of "contributions" meaning "written-sentence-related contributions", but it will be consistent with the new activity timeline. 

Maybe @ckjpn or @alanfgh will have other ideas?

ckjpn:
> Only the last 1,000 contributions are displayed.

This is an improvement, I think.

ckjpn:
Only the last 1,000 sentences are displayed.

This might make it more accurate since all of the following could also be considered contributions.

* comments with suggested improvements
* linking
* audio files
* wall posts
* corpus maintainers deleting sentences that need to be deleted.
* corpus maintainers editing sentences that need to be edited.
* tagging
* rating sentences
etc.

ckjpn:
Are you talking about this page?

https://tatoeba.org/eng/contributions/of_user/TRANG

If so, then this would be better.

Only the last 1,000 log entries are displayed.

If you say "contributions" I think you would also need to log all the following.

* comments added, edited and deleted
* wall posts added, edited and deleted
* links added, or removed
* tags added, or removed
* ratings added, edited or removed
* audio files
and perhaps some other things.

RyckRichards:
I'd like to try this one.

PR #2788, opened 2021-05-16
Replaced the word 'logs' by 'contributions' on the user menu and on Contributions/of_user

RyckRichards:
This PR solves partially #1638.

It replaces the word "logs" by "contribution on the user's menu.

image

However, it does not solve cases like this one, but maybe it wasn't the scope of this ticket:

https://tatoeba.org/eng/contributions/of_user/TRANG

image

### DIFF
--- a/src/Template/Contributions/of_user.ctp
+++ b/src/Template/Contributions/of_user.ctp
@@ -66,7 +66,7 @@ $this->set('title_for_layout', $this->Pages->formatTitle($title));
             ?>
             <div layout-padding>
                 <?= format(
-                    __('Only the last {n} log entries are displayed here.'),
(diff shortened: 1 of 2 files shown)


jiru:
@RyckRichards Thanks, that’s an improvement.

> However, it does not solve cases like this one, but maybe it wasn't the scope of this ticket:
> 
> https://tatoeba.org/eng/contributions/of_user/TRANG

If you are talking about changing the text "Only the last 1,000 log entries are displayed here.", you do so in the file `src/Template/Contributions/of_user.ctp`. Otherwise, I can merge the PR like this.

RyckRichards:
Thanks, @jiru! I changed ```src/Template/Contributions/of_user.ctp``` as well

image

PR #2794, opened 2021-05-20
Yet another API attempt (pull request take 2)

githubshrek:


githubshrek:
Original pull request was https://github.com/Tatoeba/tatoeba2/pull/2778 (which was automatically closed when the branch was renamed).

githubshrek:
There are a couple of outstanding points on this:

- there is not any error message if a user gives an invalid id or a text id. The current thing blocking this is described in the original pull request. https://github.com/Tatoeba/tatoeba2/pull/2778 
- the suggestion in https://github.com/Tatoeba/tatoeba2/pull/2778 was made to change the way in which the code is written to make it more robust to future changes. OK, I understand about robustness for future changes, but I have written a working solution to the problem, which works extremely well with very little coding, so I am rather "Meh!" about looking to rewrite it.

My preference is to put this into production as-is and to make the improvement of it a separate task, since it already delivers wonderful value to the external user (an API to search for sentences and get the data for an individual sentence).

trang:
Just FYI, @jiru is currently taking a break from Tatoeba so it'll take a bit longer to have your pull request merged. I'll try to check this in the next couple of days and we can try to have it on prod in one or two weeks 🙂 Sorry for the delay!

githubshrek:
OK, thanks for the update. It's OK, I know this is a volunteer-run open-source project, so things are running on a slower timetable than in a paid development.

I only had to see the fact that the idea of an API has been started a few times in the last 6 years during research for https://github.com/Tatoeba/tatoeba2/issues/2669 to make me see how slow things might move!

src/Controller/ApiV1Controller.php
```diff
@@ -0,0 +1,162 @@
...
+        $ids = array();
+        foreach ($results as $sentence) {
+            $ids[] = $sentence->id;
+        }
+    
+        $ids_json = json_encode($ids);
+        $ids_json = '{ "ids" : ' . $ids_json . ' }';
+        return $this->getResponse()
+            ->withStringBody($ids_json);
+    }
```
trang, line 161: This function has a lot of unnecessary code and I can't deny I would be sad to merge it as is. Below my take on how to trim this down to the essential. Please try it out and commit the code if that works fine for you :)

One thing I should note: I'm not exactly sure why you chose to return only the sentence ids rather than the sentence objects. My version of the code returns the whole object because I can't think of a practical use case where an app that consumes this API will need only the ids. Most likely, the API consumer will want the sentence data too, and if it only has the ids, it will be forced to do a whole bunch of `GET /sentence/{id}` after the `GET /search`. Better serve all this data in one call, in my opinion.

```
    public function search()
    {
        $this->loadModel('Sentences');

        $search = new SentencesSearchForm();
        $search->setData($this->request->getQueryParams());

        $limit = 100;
        $sphinx = $search->asSphinx();
        $sphinx['page'] = $this->request->query('page');
        $sphinx['limit'] = $limit;

        $query = $this->Sentences
            ->find('hideFields')
            ->find('withSphinx')
            ->find('filteredTranslations', [
                'translationLang' => $search->getData('to'),
            ])
            ->select($this->Sentences->fields())
            ->contain($this->Sentences->contain(['translations' => true]));

        $this->paginate = [
            'limit' => $limit,
            'sphinx' => $sphinx,
        ];

        try {
            $results = $this->paginate($query);
        } catch (Exception $e) {
            throw new \Cake\Http\Exception\BadRequestException();
        }
    
        $json = json_encode($results);
        return $this->response
            ->withType('application/json')
            ->withStringBody($json);
    }
```

src/Controller/ApiV1Controller.php
```diff
@@ -0,0 +1,162 @@
...
+        if (is_numeric($id)) {
+            // Retrieve the sentence
+            $sentence = $this->Sentences->getSentenceWith($id, [
+                'translations' => true,
+                'sentenceDetails' => true
+            ]);
+
+
+            return $this->getResponse()
+                ->withStringBody($sentence);
```
trang, line 65: I'd suggest to return this instead:
```
    return $this->response
            ->withType('application/json')
            ->withStringBody($sentence);
```

And with this you can normally remove `$this->disableAutoRender();` at the beginning.

src/Controller/ApiV1Controller.php
```diff
@@ -0,0 +1,162 @@
...
+/**
+ * Controller for Api.
+ *
+ * @category Sentences
+ * @package  Controllers
+ * @author   githubshrek
+ * @license  Affero General Public License
+ * @link     https://tatoeba.org
+ */
+class ApiV1Controller extends AppController
```
trang, line 41: Any objection to renaming this into `ApiBetaController` or `ApiV0Controller`? I'm just thinking that this is still at a quite experimental stage, and I'd rather keep `v1` for when we have something more stable :)

src/Controller/ApiV1Controller.php
githubshrek: That new version looks great! I think you're right that it is better to return the complete sentence objects, otherwise the return results will just create hundreds of other calls. I think my thinking was that if there are many ids, and each sentence had many translations, then getting every id and every sentence detail in one call would cause performance or memory problems, which would be avoided by getting smaller chunks. Nonetheless, I think on balance it is better to start off with getting the sentence for each id, and if we do find there are performance/memory problems then we can look into the alternative.

I can see the secret here is to write crappy code (not that I have to try), and someone who know the code base much better than me will rewrite it as it should be written. ;-)

src/Controller/ApiV1Controller.php
githubshrek: OK, looks great. I notice when making this change that the result seems to display in a user-friendly indented way.

image

src/Controller/ApiV1Controller.php
githubshrek: I'm OK to make this change. To be honest, I think we're just as likely to end up with v0 for the next 5 years as be making a grand upgrade to v1 in 6 months. I also think that the only user of the API for a long time will be me (or more precisely my site https://lingopolo.org/), so even if we tweek the returned results in a non backward-compatible way, I'm OK with that. Still, it seems a reasonable request, so I have made the required changes.

githubshrek:
To summarise (mainly for my own benefit) the changes made:

The auth_actions.php file has changed from v1 to v0:
image

- the Controller file is renamed to ApiV0Controller 
- the sentence() method:
-- has the return statement changed
-- has the $this->disableAutoRender(); statement removed
image

- the search() method is replaced with the @trang variant:
image

PR #2795, opened 2021-05-20
Change "Latest contributions" to use sentence text as click target

githubshrek:


githubshrek:
I see it's failed with "Notice (8): Undefined variable: sentenceId [APP/Template/Element/logs/log_text.ctp, line 46]". I _think_ what happened was that on my Imouto version of the code, I had both the API code and the "Latest contributions" code, and with that it worked fine. I see that with _only_ the "Latest contributions" code I get this error. Hmmm...

githubshrek:
I'm investigating this issue. I see I have to fix the passing of the sentence id correctly. OK, that's clear.

I have another issue in the code which is currently baffling me. It's this AssetCompress line which @jiru you added recently:
image

With it commented out as above, all seems to work fine:
image

With it included I get the following error:
image

Any thoughts as to what's going on?

trang:
@githubshrek I'll close this pull request for inactivity but feel free to reopen it, if someday you feel like resuming this work 🙂

PR #2796, opened 2021-05-20
Update README.tatovm.md

githubshrek:


trang:
@githubshrek I hope you don't mind, I'll close this pull request as I think we should remain as generic as possible and avoid instructions that are OS specific in the primary installation guidelines. We can provide more visual instructions for each OS as separate extra documentation. Thanks nonetheless for the proposal!

PR #2807, opened 2021-07-04
Update POT files

Yorwba:
The current POT files are out of sync with the strings used in the code, so some strings cannot be translated [which was reported on the wall](https://tatoeba.org/zh-cn/wall/show_message/37273#message_37273).

I just ran
```
tools/generate_pot.sh
git add src/Locale/*.pot
git commit -m "Update POT files"
```
without looking at the generated diff, but I assume it should be fine.

This is not covered [in the deployment handbook](https://github.com/Tatoeba/tatoeba2/wiki/Deployment), where I'd expected to find this info, but now that I think about it, the POT files on dev should ideally always be up-to-date, shouldn't they? So ideally this script should be run at least after every merge. Maybe even as a commit hook?


trang:
Thanks for this PR @Yorwba! Indeed I had forgotten to run the generate_pot.sh script for the last update. Normally, we would re-generate the POT files and push the changes a few days before deploy on prod, so that translators have time to translate before the new strings go live. That's why this task is not part of the deployment instructions. Having not done prod deployments for a while, I had forgotten that detail.

I agree that the POT files should ideally be updated after every merge. Ideally as well, they shouldn't be tracked and pushed into the git repo at all, since they are generated files. They should be sent directly to Transifex, although I'm not sure if that's a possibility that Transifex offers, I never looked into it. What Transifex offers for sure, and that we make use of, is that we can provide the URLs to our POT files, and it will take care of uploading the new POT files automatically by checking from our URLs. That's the reason why we commit these POT files in git: we provide to Transifex the URL to the raw files in GitHub.

PR #2812, opened 2021-07-18
Represent empty non-default parameters by "any" (fixes #2806)

### LINKED ISSUE #2806: "Is orphan: Any" option is reset to "Is orphan: No" when using random sort in advanced search
**To Reproduce**
Originally described by [espamatics on the wall](https://tatoeba.org/en/wall/show_message/37259#message_37259).

For convenience, I created a search template that shows the same problem: https://tatoeba.org/en/sentences/advanced_search?from=&has_audio=&native=&orphans=&query=aardvark&sort=random&sort_reverse=&tags=&to=&trans_filter=limit&trans_has_audio=&trans_link=&trans_orphan=&trans_to=&trans_unapproved=&trans_user=&unapproved=no&user=

Steps to reproduce:
1. Open the search template.
  Note that "Is orphan" in the "Sentences" block of the form is set to "Any" and the sort order is "Random".
2. Click on "SEARCH", ending up on [the search results page](https://tatoeba.org/en/sentences/search?query=aardvark&sort=random&trans_filter=limit&unapproved=no&rand_seed=UeDz).
  Note that "Is orphan" is now set to "No"

**Expected behavior**
The setting on the search results page should be the same as specified in the search form.

**Additional information**
When specifying the sort order as "Relevance" instead, [the corresponding search results page](https://tatoeba.org/en/sentences/search?from=&has_audio=&native=&orphans=&query=aardvark&sort=relevance&sort_reverse=&tags=&to=&trans_filter=limit&trans_has_audio=&trans_link=&trans_orphan=&trans_to=&trans_unapproved=&trans_user=&unapproved=no&user=) does correctly show "Is orphan" as "Any".

It looks like this bug is caused by the interaction of three separate things:
1. The "Any" option is represented by an empty `orphans` parameter, but when this parameter is absent, the default is "No".
2. The code for cleaning up the URL makes the reasonable, but incorrect, assumption that empty parameters are just clutter and can be removed without changing the search results.
3. For some reason, only URLs with random sort order are cleaned up.

Yorwba:
As mentioned in #2806, the issue is only triggered when the search is set to "random" sort. This is because of the [redirect using `Router::url()`](https://github.com/Tatoeba/tatoeba2/blob/2284e7caba68943bbf21191cae3bbbdc9c05467c/src/Controller/SentencesController.php#L497) when a random seed was generated for the search. Because `Router::url()` ignores empty parameters, the "Any" setting for "Is Orphan" and "Is Unapproved" should not be represented by an empty string.

### DIFF
--- a/src/Form/SentencesSearchForm.php
+++ b/src/Form/SentencesSearchForm.php
@@ -304,6 +304,14 @@ class SentencesSearchForm extends Form
             $keyCamel = Inflector::camelize($key);
             $setter = "setData$keyCamel";
             $this->_data[$key] = $this->$setter($value);
+
+            if(empty($this->_data[$key]) && !empty($this->defaultCriteria[$key])) {
+                /* Using Router::url() to reconstruct a URL for the given data
+                 * strips out empty parameters, which would lead to a non-empty
+                 * default being applied instead of the empty non-default value.
+                 * So represent them by "any" instead. */
+                $this->_data[$key] = 'any';
+            }
         }
     }
 
--- a/src/Template/Element/advanced_search_form.ctp
+++ b/src/Template/Element/advanced_search_form.ctp
@@ -118,7 +118,7 @@ echo $this->Form->create('AdvancedSearch', [
                         'label' => '',
                         'options' => [
                             /* @translators: dropdown option of "Is orphan" field in search form */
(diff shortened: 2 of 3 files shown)


trang:
Thanks @Yorwba! Sorry it took a bit of time to merge.

PR #2813, opened 2021-07-20
Update fixtures lists to ensure tests can be run independently

Yorwba:
I have an issue with my VM that causes `phpunit` to lose the database connection and crash if I execute it without arguments to run all tests at once. (Maybe I haven't provisioned enough memory? Not sure yet.)

Using `find tests/ -name '*Test.php' -exec phpunit '{}' ';'` as a replacement, some tests fail because they expect certain fixtures to be present that are only loaded by other tests.

This PR adds the missing fixtures in their respective lists, so each test can be run independently from its own file.

### DIFF
--- a/tests/TestCase/Controller/AngularTemplatesControllerTest.php
+++ b/tests/TestCase/Controller/AngularTemplatesControllerTest.php
@@ -11,6 +11,7 @@ class AngularTemplatesControllerTest extends TestCase
 
     public $fixtures = [
         'app.Users',
+        'app.users_languages',
     ];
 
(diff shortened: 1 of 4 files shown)


trang:
> I have an issue with my VM that causes phpunit to lose the database connection and crash if I execute it without arguments to run all tests at once. 

I can say that I've run into this problem too and I have no idea what it is either.

This problem doesn't seem to happen from a fresh installation of Tatoeba. I reinstalled Tatoeba the other day and I tried just right now to run `phpunit` and all tests were executed without losing database connection.

Issue #2198, opened 2020-03-13
Occasional blackholes when sending a PM


Guybrush88:
I just tried to reply to a PM I received and I just got the message that the request got blackholed, and the PM wasn't sent. I remember getting the same issue almost a couple of weeks ago.
EDIT = I noticed that sending PMs is pretty slow now, loading other Tatoeba pages is fine, though
NEW EDIT = now PMs are fine

agrodet:
Related : #1955, #1809
Especially concerning 1809, did you do something similar to what Yorwba described there? Taking a long time to write it, opening other tabs, or anything that you may think relevant?

Guybrush88:
well, it took no more than five minutes to write the message, I think, and I didn't send the message right away, but I took a couple of minutes to check what I wrote to avoid mistakes.
I had also other Tatoeba tabs open while sending the PM (such as the homepage and a couple of advanced search queries)

jiru:
Finally found a way to reproduce this!

### Steps to reproduce
1. Open a private browser window.
2. Go to http://localhost:8080/fr/user/profile/contributor
3. Click "Contact contributor" button. You’re being redirected to the login page.
4. Login. You’re being redirected to the "Write new message" page.
5. Ctrl+Click on the browser’s Back button. This will open a new tab that will briefly show "login" as title and directly redirect to the "Write new message" page.
6. Close that new tab and send your PM from the original tab.
7. The server returns a "Bad Request" error, which is what can be seen in place of the black hole error when `Config.debug` is enabled.

### Analysis
Relevant nginx logs (formatted as timedate, request line, server response code, response length, referer)
```
[28/May/2023:16:37:54 +0000] "POST /fr/users/check_login?redirect=%2Ffr%2Fprivate_messages%2Fwrite%2Fcontributor HTTP/1.1" 302 5 "http://localhost:8080/fr/users/login?redirect=%2Ffr%2Fprivate_messages%2Fwrite%2Fcontributor" 
[28/May/2023:16:37:54 +0000] "GET /fr/private_messages/write/contributor HTTP/1.1" 200 12172 "http://localhost:8080/fr/users/login?redirect=%2Ffr%2Fprivate_messages%2Fwrite%2Fcontributor"
[28/May/2023:16:37:54 +0000] "GET /css/private_messages/write.css?1601855983 HTTP/1.1" 200 863 "http://localhost:8080/fr/private_messages/write/contributor"
[...]

Issue #1955, opened 2019-09-12
Occasional black-holed requests while searching for specific tags


Guybrush88:
I just tried searching for any tag containing "country", and in one occasion I got this error message:

> The request has been black-holed
> 
> Errore: The requested address '/ita/tags/search' was not found on this server.

Other queries I tried, including the "country" one, worked fine, though.

Guybrush88:
apparently, the same issue happened also with comments to two other people, who reported it on the wall: https://tatoeba.org/ita/wall/show_message/32639#message_32639

Issue #1922, opened 2019-07-18
Safari on a Mac has trouble logging in with "Remember Me"


ckjpn:
## On Safari (on a Macintosh)

Over the past few days, or maybe a week or so, as I try to log in, as soon as I click the "remember me" box, instead of changing to a check mark, apparently the log in button gets clicked and I get logged in without the "remember me."  This has happened multiple times.

On experimentation, clicking the "remember me" first, before filling out the other info, everything appears to work OK, but then I get the "blackholed" error message.  This has happened repeatedly.

## On Firefox, Chrome, SeaMonkey and Opera (on a Macintosh)

This problem doesn't see to occur.

jiru:
Which version of Safari are you using?

jiru:
I made a tentative fix on dev.tatoeba.org, can you check if the problem still happens there?

ckjpn:
Safari Version 12.1.1 (13607.2.6.1.2)

It's likely the most recent, since I usually keep my computer up-to-date.

I tried it on dev and it still has the same problem.

jiru:
Okay so the problem does not come from the unclosed rememberMe input tag.

ckjpn:
Would it somehow be a CSS problem with having something overlapping something else?

jiru:
Somebody recently reported having this issue over email and I was able to capture more information.

Using Safari on iPad OS 17.7.10.

Relevant anonymized nginx logs:
```
Log format: [min:sec] "request protocol" http_status response_body_size "referrer"

[0:50] "GET /de/private_messages/write/TatoebaAdmins HTTP/2.0" 302 0 "https://en.wiki.tatoeba.org/"
[0:51] "GET /de/users/login?redirect=%2Fde%2Fprivate_messages%2Fwrite%2FTatoebaAdmins HTTP/2.0" 200 10757 "https://en.wiki.tatoeba.org/"
[0:53] "GET /cache_js/layout.js?1771065273 HTTP/2.0" 200 220094 "https://tatoeba.org/de/users/login?redirect=%2Fde%2Fprivate_messages%2Fwrite%2FTatoebaAdmins"
[1:02] "POST /de/users/check_login?redirect=%2Fde%2Fprivate_messages%2Fwrite%2FTatoebaAdmins HTTP/2.0" 302 0 "https://tatoeba.org/de/users/login?redirect=%2Fde%2Fprivate_messages%2Fwrite%2FTatoebaAdmins"
[...]

PR #2820, opened 2021-08-01
Disable Form Tampering Protection for comment forms (fixes #1809)

### LINKED ISSUE #1809: Error message when posting a comment on a sentence
I was trying to post a comment on a sentence, and, instead of the actual result (the comment being actually posted), I got this error message:

> The request has been black-holed
> 
> Errore: The requested address '/ita/sentence_comments/save' was not found on this server.

After the page was refreshed, the comment was properly posted, so I guess it's not a frequent error.

Yorwba:
CakePHP's `FormHelper` [uses a SHA1 HMAC signature](https://api.cakephp.org/3.8/source-class-Cake.View.Helper.SecureFieldTokenTrait.html#61) to protect hidden fields against modification by malicious users. The signature also covers the `session_id`, which means that if the `session_id` changes, all previously loaded forms break.

For the comment forms, the only hidden field is the `sentence_id`. Being able to modify it doesn't grant malicious users any interesting capabilities. If they want to post a comment on a different sentence, they can already do that.

<hr/>

I'm not sure whether this protection should be enabled for *any* form. There are a few other issues mentioning black-holed requests when submitting a form: #1922, #1955, #2198. I'd hope that in each case the form is validated on the backend and changing parameters on the frontend won't allow bypassing that even without the signature. But I guess it's nice to have in principle, as a kind of defense-in-depth?

### DIFF
--- a/src/Controller/SentenceCommentsController.php
+++ b/src/Controller/SentenceCommentsController.php
@@ -93,6 +93,13 @@ class SentenceCommentsController extends AppController
         $eventManager = $this->SentenceComments->getEventManager();
         $eventManager->attach(new NotificationListener());
 
+        // disable Form Tampering Protection for actions where it's no big deal
+        // (it was only protecting against changing the sentence_id)
+        $this->Security->setConfig('unlockedActions', [
+            'save',
+            'edit',
+        ]);
+
         return parent::beforeFilter($event);
     }
 


trang:
Just as a note, the form tampering projection also prevents a couple of other things aside from modification of hidden fields:

* Unknown fields cannot be added to the form.
* Fields cannot be removed from the form.

https://book.cakephp.org/3/en/controllers/components/security.html#form-tampering-prevention

> I'm not sure whether this protection should be enabled for any form. 
> ...
> But I guess it's nice to have in principle, as a kind of defense-in-depth?

Yea, I also can't think of a reason to enable this protection on any form, except to have an extra line of defense in case the backend is not properly doing its validation job.

Yorwba:
> Just as a note, the form tampering projection also prevents a couple of other things aside from modification of hidden fields:
> 
>     * Unknown fields cannot be added to the form.
> 
>     * Fields cannot be removed from the form.
> 
> 
> https://book.cakephp.org/3/en/controllers/components/security.html#form-tampering-prevention

Thanks for pointing this out. I guess this opens up potential security holes if unset parameters bypass the validation logic somehow, or if the received data is used generically, so that additional keys might end up e.g. in an SQL query. I've had a second look at the two endpoints affected here and I don't think they're vulnerable. 

I'd considered disabling this protection globally, but now I think it's better to check each endpoint individually to see whether it is safe to unlock.

PR #2828, opened 2021-08-12
Make Chinese Wu available in the UI

RyckRichards:
This PR enables Chinese Wu as a user interface language on Tatoeba.

image

https://www.transifex.com/tatoeba/tatoeba_website/translate/#wuu

### DIFF
--- a/config/app_local.php.template
+++ b/config/app_local.php.template
@@ -217,7 +217,7 @@ return [
             'cmn' => 'zh-cn',
             'zh-tw' => ['中文（臺灣正體）', 'zh-hant-tw', 'zh-hk', 'zh-hant-hk'],
             // Chinese (Wu)
-            'wuu' => [吳語], 'wuu' => 'wu'
+            'wuu' => ['吳語'],
             // Croatian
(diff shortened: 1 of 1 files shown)


config/app_local.php.template
```diff
@@ -216,6 +216,8 @@ return [
             'chi' => 'zh-cn',
             'cmn' => 'zh-cn',
             'zh-tw' => ['中文（臺灣正體）', 'zh-hant-tw', 'zh-hk', 'zh-hant-hk'],
+            // Chinese (Wu)
+            'wuu' => [吳語], 'wuu' => 'wu'
```
trang, line 220: You're missing quotes around `吳語` and you only `'wuu' => ['吳語']`.

The second part (`'wuu' => 'wu'`) should be removed because it would mean you're trying to redirect https://tatoeba.org/wuu to https://tatoeba.org/wu, which is not what we want.

RyckRichards: Thanks and sorry, Trang. Going to fix it right now,

trang:
@RyckRichards Wu Chinese is now available in production.

RyckRichards:
Thanks a lot, @trang!

PR #2842, opened 2021-09-08
Add links to Tatominer

trang:
Wall thread: https://tatoeba.org/en/wall/show_message/37536

### DIFF
--- a/src/Template/Activities/translate_sentences.ctp
+++ b/src/Template/Activities/translate_sentences.ctp
@@ -196,12 +196,16 @@ $langsTo = $this->Languages->profileLanguagesArray();
         </div>
 
         <md-subheader><?php echo __('Check Tatominer'); ?></md-subheader>
-        <div layout="column" layout-margin layout-padding layout-align="center center">
-            <?= __('Tatominer provides a list of the most searched words for which there are very few or no translations yet.'); ?>
+        <div layout="column" layout-margin layout-align="center center">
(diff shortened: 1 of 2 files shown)


RyckRichards:
@trang Perhaps it would be good to prefill the [Add sentence page](https://tatoeba.org/en/sentences/add) with the word we've clicked on Tatominer. Otherwise, it's easy to forget the word we're working on.

trang:
@RyckRichards I'm not entirely convinced it would help. I guess if you just look at the word and click on it without really thinking, then it's easy to forget. But I would imagine that you don't mindlessly click on a word. You would first try to come up with a sentence and only once you have an idea of the sentence you want to write, you could click on the word. In which case, normally, it shouldn't be that easy to forget about the word...

Perhaps it can help in the sense that you would have less characters to type if the word is already prefilled, but I feel it could also be an annoyance because you have to spend some time to move the cursor around the word.

RyckRichards:
I have thought on something like that:

1st - user click on the word "chuchu" in "https://tatominer.netlify.app/por.html". It'll open https://tatoeba.org/en/sentences/add#chuchu

2nd - that page would display something like this

trang:
I see. I would suggest that you open a separate ticket for this @RyckRichards :) It's not a bad idea, just not something that I think is high prio, so it doesn't have to be solved in the scope of this PR.

RyckRichards:
Sure, thank you!

PR #2859, opened 2021-10-22
Increase memory for VM build

Yorwba:
I rebuilt my VM today after upgrading to a new version of VirtualBox and ran into an issue with one of the C++ compilation steps failing with an out-of-memory error. The build succeeded once I doubled the available memory.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -27,7 +27,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     v.memory = 512 # in MB
 
     # This one is when we build a new VM with "BUILD=1 vagrant up"
-    v.memory = 1024 if ENV['BUILD'] == '1'
+    v.memory = 2048 if ENV['BUILD'] == '1'
   end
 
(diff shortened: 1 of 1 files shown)


jiru:
2GB feels like a lot; it would increase the RAM requirement of the host running the VM to 4GB or more. How about we try to optimize memory usage of the build process instead? Compilation can use a lot of memory but I think we should be able to get away with 1GB. It’s not like we are compiling super big projects.

Do you remember which step produced the error? Here are some suggestions:
* Maybe Tatoeba/Tatodetect@262d407b0e620c35c506f2fdf8eeb18b7ba01628 is causing the problem? (Sinoparserd also uses `-j 2` but it’s supposedly lightweight to compile.)
* What if you reorder tasks in `ansible/tatoeba.tasks.yml` so that the problematic compilation step happens first, before installing and starting up nginx, mysql and php?

Yorwba:
Closing in favor of Tatoeba/Tatodetect#3

Issue #2020, opened 2019-11-22
Possible Language Flag Improvement


ckjpn:
I wonder if it might not be a good idea to put the 3-letter code on the right side of all language flags.

This would serve as a reminder to members what the codes are and may also make it clearer that the flags represent languages and not countries.

trang:
I would rather find a way to get rid of the 3-letter code from all the language icons.

lovermann:
Sounds good, but people (in common I mean) DO confusing flags of countries with flags of languages. Moreover, country codes are more .. I would say FAMOUS, than languages flags.

ckjpn:
Since this is tagged "unclear," I'll try to explain this again.

* Standardizing the look of icons would likely make the website look better.
* Adding the 3-letter code would remind people what the codes are.
* Adding the 3-letter code would make it clearer that the flags represented the language and not the country
* Keeping the country flag as part of the flag would likely still be useful since in many cases this implies a language and can help those not yet familiar with language codes identify languages.
* Icons that are not only letters are good, too, since colors and shapes helps members more easily recognize the icons.

To view currently-used flags, go to https://tatoeba.org/eng/stats/sentences_by_language .

Here are some of the currently-used language icons with the language codes as part of the icon.

 ber, kab, nob, wuu, mhr, afr, cbk, dtp, ilo, kzj, war, nno, ceb, gla, nst, hsb, dsb, hoc, thv, ori, jav, lij, el, gsw, pan, rif, guj, asm, xho, afb, tir, zim, tpi, mrj, mww, bvy, moh, sme, hau, zul, mgm, enm, pus, ike, hil, sna, fij, ngu, bho, nch, hif, dng, umb, sco, mvv, aln, akl, ewe, kek, aoz, myv, ibo, roh, cay, ngt, nan, bam, brx, kam, cjy, mai, guc, fuc, mad, mdf, alt, bcl,tmw, gom, han, kxi, hnj, gaa, hak, gan, aym, tig, lug, cyo, urh

ArloJamesBarnes:
maybe my perspective as an outsider to Tatoeba is an uninformed one, but using a symbol for a country or ethnic group to stand in for a language when a purpose-built symbol (the language code) exists is a strange one, and clearly it can't stand on its own if it also needs the code as assistance.

https://tatoeba.org/img/flags/enm.svg and https://tatoeba.org/img/flags/eng.svg don't even use the same administrative level of flag.

ckjpn:
[...]

PR #2865, opened 2021-11-03
Rewrite the embedded PNG into SVG, for common flags

linkmauve:
The switch to SVG happened in 977786c2574ebf05acebd68d8df4c1bc29730f89
and its goal was to allow for hidpi versions of the flags, because they
currently get upscaled and that doesn’t look great.  The plan was to
then progressively convert each flag to a pure SVG version, but it
happened only for very few flags so far.

This reverts the revert in e7062ffe8ec5bb5d19b826fc2007f3efe14757db, but
that was done without any explanation so I’m not sure why it happened.

Some of the flags have slightly different colours, but I based my work
on Wikimedia Commons’ Flag_of_“country name”.svg, so those should be the
most correct versions.  For flags with a different aspect ratio than
3:2, I tried to adhere to the fixup of the current flag (either scale it
up in one direction, or crop it, or extend it), except for the
Macedonian flag where I felt like cropping was better than scaling as
the circle at the middle then became an ellipse otherwise.

I have tested the rendering on my phone and it indeed looks *much*
better.

I only picked the flags happening most often while reloading the front
page, which were simple enough to be smaller as SVG than as PNG.


jiru:
That’s cool, thank you. :sunglasses: Did you try to work with the @Tatoeba/language-team to see if they kept some of the flags in SVG?

> Some of the flags have slightly different colours, but I based my work on Wikimedia Commons’ Flag_of_“country name”.svg, so those should be the most correct versions.

It could be that the language team adapted the colors on purpose to better match with the website. For instance some flags use very bright colors that would stick out too much from the rest of the content.

linkmauve:
I didn’t get in touch with them, nor do I know how to contact them; could you redirect them to this pull request if you can?  If the flags were SVG originally that’d be indeed a good idea to replace the embedded PNGs with their original, but I also took care of reducing the size to the minimum possible in comparison to the Wikimedia Commons version, so care should be taken to avoid replacing them with a bigger version.

trang:
@linkmauve Thanks for this! We did start some work to replace the "fake" SVG with pure SVG but never completed it. 

Back in 2016, Ricardo (@RyckRichards) actually started to gather SVG files for as many languages as he could find, and uploaded them in some shared folder in Google Drive. But it didn't go further than that, as no one had the time (or knowledge) to adjust these SVG to fit into Tatoeba.

@RyckRichards I managed to find the folder by the way, it's here: https://drive.google.com/drive/folders/1ksNAilTmxROPLhIj3lerHx5n3W0XS6sv?usp=sharing

linkmauve:
This folder isn’t accessible publicly, if you can share a tarball or something I could try to merge it with this PR.

I’m still not sure what to do wrt. colours, whether to go with the official ones or with the modified ones from the PNGs, I’d tend towards the former as the differences are usually minimal (in at least all of the ones I’ve gathered so far).

trang:
> This folder isn’t accessible publicly, if you can share a tarball or something I could try to merge it with this PR.

Better yet if you could teach @RyckRichards how to edit SVGs so that he can do what you did in this PR, for the other languages 🙂  We can of course check and merge this PR in the meantime, we don't have to do it all at once. But the main reason why I'd like @RyckRichards to learn this is because he's the one who takes care of creating the SVG file for the language icons. It's just that he is not a technical person and learning on his own how to manipulate SVGs is not something he has been comfortable with. So that's why we have been stuck with the process of converting PNGs into "fake" SVGs for every new language we are adding. Anyway, I'll send to you both an email.

> I’m still not sure what to do wrt. colours, whether to go with the official ones or with the modified ones from the PNGs, I’d tend towards the former as the differences are usually minimal (in at least all of the ones I’ve gathered so far).

I would go with the official colors as well, but we'll need @sabretou to confirm this as he has the final call on the design of the language icons. I will deploy your changes on the dev website when I get the chance, and @sabretou can check over there to see if it looks fine for him, then we can merge.

RyckRichards:
Those files were on one of my old drivers :O

I was able to let everyone edit it -
https://drive.google.com/drive/folders/1ksNAilTmxROPLhIj3lerHx5n3W0XS6sv?usp=sharing

>

sabretou:
> I would go with the official colors as well, but we'll need @sabretou to confirm this as he has the final call on the design of the language icons. I will deploy your changes on the dev website when I get the chance, and @sabretou can check over there to see if it looks fine for him, then we can merge.

Yes, I am in favour of the official/original colours as well.

linkmauve:
The button “Download all” in Google fails after some time “preparing a zip archive”, do you have another way of sharing it which wouldn’t require me to download each file separately?

I have another question: since the goal I assume is to keep the file size low with better quality on hidpi displays, for very complex flags I was thinking about doing the basic features in normal SVG, while keeping the complex parts as embedded PNGs (like I’ve done for Spain here, see spa.svg). Any thought on that?

trang:
@linkmauve 

> do you have another way of sharing it which wouldn’t require me to download each file separately?

Here's the zip file: [SVG-20211218T080120Z-001.zip](https://github.com/Tatoeba/tatoeba2/files/7738953/SVG-20211218T080120Z-001.zip)

> for very complex flags I was thinking about doing the basic features in normal SVG, while keeping the complex parts as embedded PNGs

Sounds good to me.

> Also, could you deploy it on https://dev.tatoeba.org/ so that we can see it in action? :)

The changes of this pull request are deployed on the dev website. You can use the language stats page to compare the differences:

- dev: https://dev.tatoeba.org/en/stats/sentences_by_language
- prod: https://tatoeba.org/en/stats/sentences_by_language

@sabretou, you can find the list of languages which SVG has been updated from the "[Files changed](https://github.com/Tatoeba/tatoeba2/pull/2865/files)" tab of this pull request.

One thing I notice is that the ISO code is overflowing the box (for Kabyle, the 'B' letter isn't readable). Also, the font is a serif font on the pure SVG, while the font for the PNG-SVGs is sans serif.
image

sabretou:
The changed flags look very sharp, which is great, but as Trang pointed out, the flags that include the ISO code are overflowing. A lesser issue is that their font will apparently depend on the user's system, which is less-than-ideal in my opinion. I would prefer a uniform experience for all users. In Trang's screenshot I'm seeing the code in a serif font, while on my Firefox it's a Sans Serif font.

Other than that, I'm on board with the SVGs, they look great.

linkmauve:
Oops, this is a typo, I said `font-family="mono"` instead of `font-family="monospace"`.  This is fixed in the latest version of this PR, and although there could still be some slight variations depending on the installed fonts on the user’s computer they should all be monospace.

hugolpz:
To be honest, this directory should be a proper github repository `languages-flags`, `iso-flag` or something.

trang:
@hugolpz Thanks for your input and sorry for the delayed response!

> The idea of flag + iso = visual/icon representation of a language is elegant.

The ISO code indicator is actually a fallback design for when we are not creative enough to figure out a unique "avatar" for the language and end up using the same flag for several languages. The only way to distinguish between these languages is the ISO code.

It was suggested that we could add this ISO code next to all flags (#2020) but I'm not entirely convinced this is the way to go. The ISO code feels like an annotation to me rather than a design element.

> I saw various waves of migration from embeded to true svg. How many of your flags are "true" svg ?

There hasn't been any wave of migration to true SVG yet. I don't think we have any flag that are true SVG.

The migrations we had was from PNG to SVG (#1806). We had another mass replacement of the flags (#1998), but it was not to turn the embedded SVG into true SVG, just to have a visual update to make the language flags more consistent with each other. The updated flags were still embedded SVG.

> Are you in the process to switch them all to true svg ?

Not actively, but that is definitely a goal of ours. 

As I've mentionned, it is our member @RyckRichards who is currently in charge of creating these SVGs but he is not a technical person and creating pure SVG is still above his current skills. While he is willing to learn, that might still take some time.

> To be honest, this flags directory could be a proper github repository 

Definitely. If you feel like starting a git repo for this, please be my guest 🙂 

> On Wikimedia Commons, create [[Category:SVG language flag project]] then tagging all those flags' wikipage with it. 

This would indeed be a much better alternative to the Google Drive we have. I think anyway all of the flags collected in that Google Drive were from Wikimedia Commons.

> I could help on this Commons side if you see it as useful.

Yes, your help on this side would be very much appreciated. I don't think Ricardo has ever contributed to Wikipedia/Wikimedia and neither have I, so the whole process of editing these wiki pages are unfamiliar to us.

linkmauve:
All of the flags in this PR are from Wikimedia Commons as well, fyi.  I just cleaned them up a bit (using svgcleaner and manual XML editing) and resized them to 30×20 as it is the format used by Tatoeba.

ArloJamesBarnes:
The more neutral (and to my mind more sensible) approach I've seen taken in Wikimedia contexts is simply to decorate the language code to make it seem less 'text-like': https://commons.wikimedia.org/wiki/category:Blue_ISO_639_icons_with_double_parentheses

hugolpz:
I was reported by a contributor that Wikimedia had long talks and preferred to exclude flags by political caution and respect to minorities. So they headed toward non-flag, text based language avatar. It's a solid rational, but it's highly inelegant, and this question is in good part an UI question.

trang:
I'll merge this but will revert the SVGs where the language has an ISO code on the side because of the font difference reported by Sabretou in a comment above:

> A lesser issue is that their font will apparently depend on the user's system, which is less-than-ideal in my opinion. I would prefer a uniform experience for all users. In Trang's screenshot I'm seeing the code in a serif font, while on my Firefox it's a Sans Serif font.

If we can provide a uniform experience, that would be indeed better. Not just uniform for all users, but also uniform between the flags themselves.

In any case, thanks again, @linkmauve, for the pull request :)

trang:
I'll merge this but will revert the SVGs where the language has an ISO code on the side because of the font difference reported by Sabretou in a comment above:

> A lesser issue is that their font will apparently depend on the user's system, which is less-than-ideal in my opinion. I would prefer a uniform experience for all users. In Trang's screenshot I'm seeing the code in a serif font, while on my Firefox it's a Sans Serif font.

If we can provide a uniform experience, that would be indeed better. Not just uniform for all users, but also uniform between the flags themselves.

In any case, thanks again, @linkmauve, for the pull request :)

PR #2869, opened 2021-11-07
Use the same random sentence header for guests and members (fixes #2868)

### LINKED ISSUE #2868: No convenient way for logged-out users to refresh the random sentence on the home page
**Context**
A user pointed out to me that the random sentence on the homepage is different for logged-out and logged-in users, and only the latter get a button to "Show another". (It's still possible to get a new sentence by reloading the entire page, but that's slower and not as intuitive.) The user said that he likes repeatedly clicking that button to see different random sentences, and that he'd probably have become a regular user earlier if he'd known about the feature.

**Idea**
Use the same random sentence block for all users, authenticated or not. I'm not sure why that's not the case already, maybe concerns about mobile-friendliness again?

Yorwba:
With this PR, a language selector and refresh button are always included in the header of the random sentence block.

This required a little redesign to make everything fit on narrow screens:
900px
500px
300px

The screenshots are in Indonesian because the homepage was apparently getting cached somewhere and I couldn't figure out how to disable that to test my changes. (`bin/cake cache clear_all` didn't seem to have any effect.) On the bright side, this helped me notice that the "More tips" and "Advanced search" buttons below the search form can get quite wide next to each other in some languages, so I enabled wrapping for them, too.

### DIFF
--- a/src/Template/Element/random_sentence_header.ctp
+++ b/src/Template/Element/random_sentence_header.ctp
@@ -34,7 +34,7 @@ $langArray = $this->Languages->onlyLanguagesArray();
 
 <div ng-controller="RandomSentenceController as vm" ng-init="vm.init()">
 <md-toolbar class="md-hue-2">
-    <div class="md-toolbar-tools" layout-align="center center" layout-margin layout-wrap>
+    <div class="md-toolbar-tools" layout-align="center center" layout-wrap>
         <?php /* @translators: random sentence block header on the home page */ ?>
         <h2 flex><?= __('Random sentence') ?></h2>
 
--- a/src/Template/Element/random_sentence_header.ctp
+++ b/src/Template/Element/random_sentence_header.ctp
@@ -35,7 +35,7 @@ $langArray = $this->Languages->onlyLanguagesArray();
 <div ng-controller="RandomSentenceController as vm" ng-init="vm.init()">
 <md-toolbar class="md-hue-2">
     <div class="md-toolbar-tools">
-        <?php /* @translators: random sentence block header on the home page for members */ ?>
+        <?php /* @translators: random sentence block header on the home page */ ?>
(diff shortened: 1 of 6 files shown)


trang:
Thanks for this pull request, @Yorwba, and sorry for the late feedback. 

There's a couple of things we could improve:

1) When viewing on large screen, there seems to be more padding/margin than there should be. The padding/margin becomes normal again on medium screen. You can see it in action in the [Tatoeba Stream #5 video at around 59:30](https://www.youtube.com/watch?v=9CdWk6lnKZM&t=3570s).

2) I've also been considering replacing someday the "Show another" button by a [refresh icon](https://fonts.google.com/icons?selected=Material%20Icons%20Outlined%3Arefresh%3A) and have "Show another" as the tooltip. This could reduce the height of this header from 3 lines to 2 lines on small screens.

Yorwba:
Thanks for the feedback, and no worries about being slow to respond. I've been rather slow myself, after all :)

The change in margins is because I used the `layout-margin` attribute, which is used to provide Material Design containers and their children with copious amounts of margin. But it's [overwritten in `index.css`](https://github.com/Yorwba/tatoeba2/blob/f19951dd484f13cea77f794e5aa3ce9314944a0b/webroot/css/pages/index.css#L75) to set the margin to 0 for screens up to 960px wide. That's why the margin suddenly increases when the screen gets wider than that. I thought this was intentional, but apparently not.

Replacing the button text by an icon is a good idea. It doesn't completely prevent the possibility that three lines might be required, but the screen would have to be very narrow indeed (287px):
287px screen
350px screen
450px screen
959px screen
There's another discontinuity between a screen width of 959px (above) and 960px (below) where the entire sentence block suddenly gets smaller, but I don't think that's related to the changes in this PR, since it also happens on dev.
960px screen

PR #2877, opened 2021-11-15
Speedup expand/collapse buttons in new design

jiru:
My attempt to speedup the "show/hide _n_ more translations" and "expand/collapse sentence menu" buttons in the new design. This PR complementary to #2862. Basically, the buttons should be much more reactive, but at the cost of a longer initial page load. UX-wise, I think it’s a better compromise to have the user wait a bit more during page load rather than a slow interface.

That being said, results with ten sentences per page and 100+ translations per sentence get extremely long to load, like 20 seconds. Maybe we could have a higher "hard" limit of number of sentences to render during initial page load, and the rest would be rendered after clicking on the show/expand button, just like the current implementation.


Yorwba:
> results with ten sentences per page and 100+ translations per sentence get extremely long to load, like 20 seconds.

On my antique laptop, I timed such a search result page at 56 seconds. Usually, when I encounter a page that takes this long to load without a visible progress indicator, I tend to assume my adblocker is breaking something, disable it and try again. (Or I just leave.)

> Maybe we could have a higher "hard" limit of number of sentences to render during initial page load, and the rest would be rendered after clicking on the show/expand button, just like the current implementation.

How about a hybrid solution? Show only a few translations during the initial render to give the user *something* to look at as fast as possible, and then add the other translations with `ng-hide` in the background (I hope AngularJS supports enough asynchronicity to not lock up the interface this way). If the user clicks "show more translations" before the hidden translations have finished rendering, they'll still have to wait, but hopefully not as long.

<hr>

I was pretty sure your use of one-time bindings would break editing the sentence or its translations, but it turns out that the edit form is gated by `ng-if`s hiding everything else, which means that the values get refreshed when everything is rerendered after the form is closed. Of course this is slow and should probably use `ng-show` instead. (And *then* the one-time bindings break editing.)

jiru:
@Yorwba Thank you very much for your feedback. Good idea about the async load. About the one-time bindings, I must admit I sortof monkey-patched them wherever it made sense to me (that is to say data that is not really editable without a page reload), but I am not so familiar with all that. I will have a closer look at what happens if I change the `ng-if` of the form into a `ng-show`.

jiru:
> I was pretty sure your use of one-time bindings would break editing

@Yorwba I fixed the one-time bindings, and it helped me to uncover #2885.

> On my antique laptop, I timed such a search result page at 56 seconds.

I reverted to the previous behavior of not rendering all translations during page load.

> Maybe we could have a higher "hard" limit of number of sentences to render during initial page load, and the rest would be rendered after clicking on the show/expand button, just like the current implementation.

I abandoned that approach because I fear the initial page load, which is already rather long-ish, would just become too long, while the benefits (a more responsive expand button) are not that great considering the user could just not click on the button at all.

> How about a hybrid solution? Show only a few translations during the initial render to give the user something to look at as fast as possible, and then add the other translations with `ng-hide` in the background (I hope AngularJS supports enough asynchronicity to not lock up the interface this way). If the user clicks "show more translations" before the hidden translations have finished rendering, they'll still have to wait, but hopefully not as long.

@Yorwba I tried using [this approach](https://stackoverflow.com/questions/11953348). The result is that the UI simply becomes unresponsive when the "background" rendering happens, whether it's right after the page finishes loading (using a timeout of 0 seconds) or 10 seconds later. I guess this is because of the way AngularJS internally works, with the digest thing.

Maybe the are some other ways to optimize things, but I feel like I am are hitting hard the limits of AngularJS and I need to face the fact that AngularJS is just slow. We could consider switching to a different framework such as VueJS, [they say it’s faster](https://vuejs.org/v2/guide/comparison.html#Runtime-Performance-1). During my research I also stumbled upon this article about [people migrating to VueJS](https://www.holistics.io/blog/why-and-how-we-migrated-from-angularjs-to-vuejs/). It looks like these two frameworks are quite close, so migrating does not require a huge amount of work.

Yorwba:
Your new commits look good to me. 👍

> the UI simply becomes unresponsive when the "background" rendering happens, whether it's right after the page finishes loading (using a timeout of 0 seconds) or 10 seconds later. I guess this is because of the way AngularJS internally works, with the digest thing.

Yeah, I forgot that JavaScript in the browser is single-threaded, so you can't really do things "in the background"; any long-running task blocks everything else. I think it'd be possible to work around that by splitting the digest into many smaller ones,
each asynchronously scheduling the next before handling control back to the browser. But even if we make the digest as lightweight as possible (e.g. only rendering a single new sentence), we'd still have some per-digest overhead, which might make it take long enough to cause noticeable lag.

About changing frameworks, according to the latest [JS framework benchmark results](https://krausest.github.io/js-framework-benchmark/2022/table_chrome_98.0.4758.80.html) it looks like Vue takes only about 70%-80% of the time to create new elements compared to AngularJS. That is *a lot* faster, but if it takes several seconds with AngularJS, I guess it's also going to take several seconds with Vue, so we'd probably still need similar tricks with hidden elements and stuff.

I don't think switching to MaterializeCSS is likely to influence performance much, but it would be a good first step to decouple from AngularJS in preparation for switching to another framework.

PR #2893, opened 2022-01-15
Enable Icelandic in the UI

RyckRichards:
The user [nipbud](https://tatoeba.org/en/user/profile/nipbud) has started to translate the UI into Icelandic and would like to test in "real-time" how the translation is rendering. She even posted [this message on the Wall](https://tatoeba.org/en/user/profile/nipbud)

### DIFF
--- a/config/app_local.php.template
+++ b/config/app_local.php.template
@@ -259,6 +259,8 @@ return [
             // Interlingua
             'ia' => ['Interlingua'], 'ina' => 'ia',
             // Italian
+            'is' => ['íslenska', 'isl' => 'is',
+            // Icelandic
             'it' => ['Italiano'], 'ita' => 'it',
             // Japanese
             'ja' => ['日本語'], 'jpn' => 'ja',
--- a/config/app_local.php.template
+++ b/config/app_local.php.template
@@ -258,9 +258,9 @@ return [
             'id' => ['Bahasa Indonesia'], 'ind' => 'id',
             // Interlingua
             'ia' => ['Interlingua'], 'ina' => 'ia',
-            // Italian
-            'is' => ['íslenska'], 'isl' => 'is',
             // Icelandic
+            'is' => ['Íslenska'],
+            // Italian
             'it' => ['Italiano'], 'ita' => 'it',
             // Japanese
             'ja' => ['日本語'], 'jpn' => 'ja',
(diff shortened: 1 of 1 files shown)


Trendyne:
Already translated about 50%, but wan't to see it render before I touch anything declined

config/app_local.php.template
```diff
@@ -259,6 +259,8 @@ return [
             // Interlingua
             'ia' => ['Interlingua'], 'ina' => 'ia',
             // Italian
+            'is' => ['íslenska'], 'isl' => 'is',
```
trang, line 262: Would it be okay to change `íslenska` to `Íslenska` or is there a rule in Icelandic that language names should never have a capital letter? If there's no such rule, it would be more consistent to capitalize, as all the other language names start with a capital letter.

config/app_local.php.template
```diff
@@ -259,6 +259,8 @@ return [
...
+            // Icelandic
```
trang, line 263: The comments for Italian and Icelandic should be swapped.

In this line it should be `// Italian` and `// Icelandic` should be where `// Italian` is.

config/app_local.php.template
Trendyne: Languages are lowercase in Icelandic, but it's the same in Russian and that has it uppercased, so I'd follow that:
`'ru' => ['Русский'], 'rus' => 'ru',`

trang: Okay, thanks for the confirmation, @Trendyne 🙂 

@RyckRichards Please add the capital letter.

trang:
@Trendyne Even though this pull request is not merged yet, I have updated the dev website (https://dev.tatoeba.org/) to include Icelandic, so you can already check it out.

image

Trendyne:
Thank you!

Trendyne:
Works as I hoped (despite the warnings). Epic. Thanks again
image
image

config/app_local.php.template
jiru: For your information, there is no need to add `'isl' => 'is'` because this only helps keeping old URLs by redirecting `isl` to `is`. Since Icelandic `isl` has never been published on tatoeba.org, there shouldn’t be any `isl` URL in the wild. Keeping `'isl' => 'is'` won’t hurt nevertheless.

Trendyne: I think it's better to be consistent and languages are capitalised when they're the start of a sentence like any other word anyway, which is probably what you want here

Issue #2614, opened 2021-02-01
when I change my profile icon, I'm called a "previous member"


alanfgh:
In order to update my profile icon, I had to delete the old icon. After I did this, the UI started indicating that I was a "former member" even though I had not deleted my account:

![former_member_2](https://user-images.githubusercontent.com/1213027/106495061-3062a380-6489-11eb-8cc2-7bc655226390.png)

Later, the "former member" text stopped appearing. I think that this might have been after I uploaded a new icon.

This was on Firefox/Windows, though that may not be relevant.

ckjpn:
I checked this on dev.tatoeba.org

### 1. Before
<img width="415" alt="Screen Shot 2021-02-09 at 10 57 22" src="https://user-images.githubusercontent.com/7934273/107305233-ec455380-6ac5-11eb-8ade-544fc17955db.png">

### 2. After Deleting
<img width="258" alt="Screen Shot 2021-02-09 at 10 57 55" src="https://user-images.githubusercontent.com/7934273/107305273-05e69b00-6ac6-11eb-9543-ba26417dfa49.png">

### 3. After Uploading a New Image
<img width="220" alt="Screen Shot 2021-02-09 at 10 58 31" src="https://user-images.githubusercontent.com/7934273/107305284-097a2200-6ac6-11eb-84fb-478e074eafb4.png">

Issue #526, opened 2014-12-06
Invalid mail notification from comment on a deleted sentence


jiru:
More information on [tommy_san’s Wall post](http://tatoeba.org/jpn/wall/show_message/21130#message_21130).

I recieved a strange email notification when I posted a comment after deleting a sentence.

"Tatoeba - Comment on sentence : 削除されました

tommy_san has posted a comment on a sentence where you also
posted a comment.
http://tatoeba.org/sentence_com...75906#comments "

I usually don't get a notification about my comment.

Sorry, it seems that it happened not because the sentence has been deleted, but because I'd changed the email address.

By the way, it looks that it's technically possible to internationalize email notifications.
If we want to do that, we may need to let the users choose the language of notifications in the settings.

RyckRichards:
**updating**
I wasn't able to reproduce that. I've tried that from many different ways (changing my emails all the time, setting my own mailbox to don't get emails from Tatoeba, disabling emails notifications) but it works perfectly. I think we can close this issue.

jiru:
I am able to reproduce this bug. There are actually two bugs:
1. `$this->Auth->user('email')` isn’t updated after changing email address in your profile. That’s the reason why the email got sent, and also why it said "tommy_san has posted a comment on a sentence where you also posted a comment."
2. The subject of the notification email is wrong.

PR #2901, opened 2022-02-07
Improve consistency of current user's data

Yorwba:
This pull request addresses issue #2614, which has multiple causes:

1. The first reason the profile picture was replaced by the text "Former member" is that this was mistakenly used as the alt text for all users, because the username wasn't passed through to the image-tag generating function.

    To prevent similar problems from occurring in the future, I refactored the function to take the whole user object instead, so you can't accidentally pass it only the username or only the image file.

2. The second reason the profile picture was replaced by the text "Former member" is that the image file to use for the current user wasn't correctly changed after deleting the picture, causing the image to break.

    The underlying issue is that `CurrentUser` was using the data from `$this->Auth->user()`, which is stored in the session when authenticating and not reliably updated when the user data changes. This could also cause various other issues (e.g. what if a user's access gets restricted, but they're still logged into a session with higher privileges?) so I think the best solution is to always get the current data from the database.

### DIFF
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -124,7 +124,13 @@ class AppController extends Controller
 
         // So that we can access the current users info from models.
         // Important: needs to be done after RememberMe->check().
-        CurrentUser::store($this->Auth->user());
+        $user = $this->Auth->user();
+        if ($user) {
+            // Keep the info up to date
+            $this->loadModel('Users');
+            $user = $this->Users->getInformationOfCurrentUser($user['id'])->toArray();
+        }
+        CurrentUser::store($user);
 
         // Restore named parameters removed in CakePHP 3
         $this->request = Router::parseNamedParams($this->request);
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -150,6 +150,13 @@ class AppController extends Controller
      */
     public function beforeRender(Event $event)
     {
+        $current_user = CurrentUser::get('User');
+        $auth_user = $this->Auth->user();
+        if ($auth_user && $current_user && $auth_user != $current_user) {
+            // User data changed, tell the Auth component about it.
+            $this->Auth->setUser($current_user);
+        }
+
         // without these 3 lines, html sent by AJAX will have the whole layout
         if ($this->request->is('ajax')) {
             $this->viewBuilder()->setLayout('ajax');
--- a/src/Controller/UserController.php
+++ b/src/Controller/UserController.php
@@ -319,7 +319,6 @@ class UserController extends AppController
         $savedUser = $this->Users->save($user);        
 
         if ($savedUser) {
-            $this->Auth->setUser($savedUser->toArray());
             $this->Flash->set(
                 __('Profile saved.')
             );
@@ -541,7 +540,6 @@ class UserController extends AppController
             $this->Users->patchEntity($user, $data, ['fields' => $allowedFields]);
             $savedUser = $this->Users->save($user);
             if ($savedUser) {
-                $this->Auth->setUser($savedUser->toArray());
                 $flashMsg = __('Your settings have been saved.');
             } else {
                 $flashMsg = __(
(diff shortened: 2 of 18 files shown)


src/Controller/AppController.php
```diff
@@ -124,7 +124,13 @@ public function beforeFilter(Event $event)
...
         // So that we can access the current users info from models.
         // Important: needs to be done after RememberMe->check().
-        CurrentUser::store($this->Auth->user());
+        $user = $this->Auth->user();
+        if ($user) {
+            // Keep the info up to date
+            $this->loadModel('Users');
+            $user = $this->Users->getInformationOfCurrentUser($user['id'])->toArray();
+        }
+        CurrentUser::store($user);
```
jiru, line 133: While this definitely fixes this issue (as well as the issue of having to logout and login after a user status change), I am concerned with the design of pooling the users table on each and every request, most of the time needlessly. Pooling is also what we do to display the unread private message counter at the top-right, and it bloats the list of fixtures required to run most of our controller tests. I feel like it’s not a good design and I wonder if there is any better way to go.

As for the profile picture issue, you could as well update the session data after saving it. That’s what we do everywhere else, for example:
https://github.com/Tatoeba/tatoeba2/blob/9850fd42d3d70f6472c77e16668ecb4a7775e176/src/Controller/UserController.php#L318-L325

jiru: Well I’ve thought it over and I can’t think of a better way to go. The current implementation fails in a lot of ways, including when a user logs in from two different browsers simultaneously. We already pool the users_languages table in `CurrentUser::_setProfileLanguages()` anyway. You could then remove the call to `setUser()` that I mentioned, as well as other similar calls.

My only remaining concern is that I think (haven’t tested) that the Auth component won’t know about a user role update because you don’t run `$this->Auth->setUser()` with the latest user data. Because of this, the restrictions defined in `config/auth_actions.php` won’t apply match to the new role.

Yorwba: Aside: I think you mean "poll" instead of "pool".

> You could then remove the call to `setUser()` that I mentioned, as well as other similar calls.

Will do.

> the Auth component won’t know about a user role update because you don’t run `$this->Auth->setUser()` with the latest user data.

I was actually calling `$this->Auth->setUser()` at first, but then the request failed with an "auth error" (if I remember correctly), so I figured that the `Auth` component isn't supposed to be used like that.

> Because of this, the restrictions defined in `config/auth_actions.php` won’t apply match to the new role.

Thanks for pointing these out to me. I thought the permission checks all went through `CurrentUser::canEditSentenceOfUserId` and the like, so it would be enough to keep `CurrentUser` up-to-date. I'll look into updating the `Auth` user data as well.

Yorwba: > So I opted to update `CurrentUser` after calls to `Users->save`.

Scratch that part. I tested the wrong commit. Turns out `beforeRender` doesn't get called if there's nothing to render because the request is redirected. So `Auth->setUser` has to be called directly to make changes take effect immediately.

jiru: > I thought that would make the other `setUser` calls entirely redundant, but one of the tests didn't agree. When changing the user data, there would a delay of one request until the new data reaches `Auth->user`. So I opted to update `CurrentUser` after calls to `Users->save`.

Hah, I bet you’re talking about `testSaveBasic_changingEmailUpdatesAuthData()`! I wrote it long ago because of a similar bug #526. Of course the test fails because it only performs one request; it’s tied to that design of calling `Auth->setUser()` on user "write" rather than "read". Fell free to rewrite or remove that test.

Yorwba: I've rewritten the test to actually follow the redirect before checking that the Auth data was changed.

trang:
For what I've checked, looks good to me. 
Maybe @jiru you can confirm that this is okay for you too? 
In any case, if no objections I'll merge this pull request this weekend 🙂

PR #2902, opened 2022-02-10
Remove Google Analytics

Yorwba:
It seems like [using Google Analytics has been found to be illegal in France](https://www.cnil.fr/en/use-google-analytics-and-data-transfers-united-states-cnil-orders-website-manageroperator-comply).

This PR completely removes Google Analytics from the codebase.

As a faster measure, disabling Google Analytics in `config/app_local.php` should suffice as well.

# Replacements

As far as I know, Google Analytics has been used on Tatoeba to determine which browsers users use, which URLs are most frequently visited and the time when the fewest users are active. Those use cases should be possible to satisfy by parsing the server logs instead, e.g using [GoAccess](https://goaccess.io/)

Should something more featureful be required (but I don't know what for?), various other open-source analytics solutions were brought up in [this Hacker News thread last month](https://news.ycombinator.com/item?id=29888599).


jiru:
Neither are we asking for user consent with a cookie banner. We’ve been quite careless on this one. I agree we should get rid of GA. Thank you for bringing this up.

> parsing the server logs instead

We only keep the past two weeks of logs (because of limited disk space), which is probably not statistically very significant, so goaccess would not be very useful. I believe only @trang is really using the stats, so she’s probably in a better position to choose a replacement.

jiru:
Since you are changing the TOS, in theory that change should be `communicated to the persons already using the Website by displaying an announcement banner during their next visit` (quote from the TOS). See `src/Template/Element/announcement.ctp` if you want to go that route.

In addition, you should update the last line of the TOS that says:
> Les présentes Conditions Générales d’Utilisation sont entrées en vigueur le 20 janvier 2019. La <a href="https://en.wiki.tatoeba.org/articles/show/terms-of-use-v1">version précédente des Conditions Générales d’Utilisation</a> était en vigueur jusqu’au 19 janvier 2019.

Yorwba:
> We only keep the past two weeks of logs (because of limited disk space), which is probably not statistically very significant, so goaccess would not be very useful.

GoAccess supports [accumulating statistics from multiple logs with the `--persist` and `--restore` options](https://goaccess.io/man#incremental-processing). So it should be possible to handle logrotation.

> There is a reference to the `google_analytics.ctp` element in `src/Template/Layout/lists.ctp`. While we don’t use that layout as far as I know, it would be more consistent to remove that reference too.

Good catch. I removed the line.

> 

> Since you are changing the TOS, in theory that change should be `communicated to the persons already using the Website by displaying an announcement banner during their next visit` (quote from the TOS). See `src/Template/Element/announcement.ctp` if you want to go that route.

You're right. Seems like the current `new_terms_of_use` setting is already numeric `'1'`, so making it versioned should be easy enough.

> In addition, you should update the last line of the TOS that says:
> 
> > Les présentes Conditions Générales d’Utilisation sont entrées en vigueur le 20 janvier 2019. La [version précédente des Conditions Générales d’Utilisation](https://en.wiki.tatoeba.org/articles/show/terms-of-use-v1) était en vigueur jusqu’au 19 janvier 2019.

The current previous version links to the wiki, but I'm thinking of moving all `src/Locale/*/terms-of-use.html` to `terms-of-use-v2.html`, changing "version précédente" to "première version" and adding a sentence "La [deuxième version des Conditions Générales d’Utilisation](https://tatoeba.org/fr/terms_of_use_v2) était en vigueur depuis le 20 janvier 2019 jusqu’au xxx 2022." 

Not sure whether that makes sense, from a French perspective and otherwise. Should "xxx" be some future date when the new terms of use will actually come into force?

jiru:
> The current previous version links to the wiki, but I'm thinking of moving all `src/Locale/*/terms-of-use.html` to `terms-of-use-v2.html`,

Not sure if that’s what you meant, but considering that the French version is a translation source and the contents of all the non-French `terms-of-use.html` are mirrored from Transifex, I think we should keep `src/Locale/*/terms-of-use.html` as files always containing the latest version. This way, only that latest version will show up on Transifex and we never have to edit `.tx/config`. We can keep a copy of the previous versions as static files like `terms-of-use-v1.html` (matching the numbering of `new_terms_of_use` setting).

> changing "version précédente" to "première version" and adding a sentence "La [deuxième version des Conditions Générales d’Utilisation](https://tatoeba.org/fr/terms_of_use_v2) était en vigueur depuis le 20 janvier 2019 jusqu’au xxx 2022."
>
> Not sure whether that makes sense, from a French perspective and otherwise. Should "xxx" be some future date when the new terms of use will actually come into force?

About the phrasing, if we are to append a new sentence on every version, that paragraph is going to become a mess at some point. Maybe stick to mentionning the current and previous version only.

Yorwba:
> since this TOS thing is becoming a complex matter, maybe we should take care of it in a different PR instead so that we can move on.

OK, I removed the TOS change.

trang:
Thanks for bringing this up to our attention, @Yorwba!

No problem on my side for removing Google Analytics. I haven't been using analytics extensively. Finding a replacement can wait.

I'll merge this :)

PR #2913, opened 2022-03-20
Updating the Terms of Use for the removal of Google Analytics

Yorwba:
This PR creates a new configuration option `TermsOfUse.version` that will cause an announcement banner to be shown to users who have not accepted whatever the current version of the terms of use is. After deployment, the version number will have to be set to `'2'` in `app_local.php`.

I optimistically set the date for the new terms without Google Analytics to come into force to April 3, 2022 (in two weeks).

There was an old command to fix the `new_terms_of_use` setting for users who had it set to `false`, but I don't expect it to be required ever again, so I removed it and the associated tests.


config/app_local.php.template
```diff
@@ -342,6 +342,10 @@ return [
         'shutdownWarning' => false,
     ],
 
+    'TermsOfUse' => [
+        'version' => '2',
```
trang, line 346: I'm wondering if it makes sense to put the version number in the config. It's not really a config option to me. For instance we don't need to change the value based on the environment where it's deployed. And we shouldn't actually be able to change the value unless there has been actual changes in the ToS itself. So it would probably be better if the version number is hardcoded in a constant.

Yorwba: Yeah, I'm not sure what I was thinking there. What would be the best place to stuff a global constant? I'm thinking about adding it to `src/Application.php`, but maybe there's a better place to put it.

Yorwba: OK, done. I didn't update the date for the new terms to go into effect, so you'll have to do that once you know when you'll deploy it.

trang:
Thanks, @Yorwba! Will deploy this weekend :)

Yorwba:
Transifex gets its data from master, doesn't it? So if you merge into master this weekend and deploy immediately, it might happen that some users get the notification about a change in the terms of use even though the translated version in their language hasn't changed at all.

By the way, while looking at what's different between master and dev, I noticed that there's [a commit on master that isn't on dev](https://github.com/Tatoeba/tatoeba2/compare/master#diff-84177656b172486dce13056e8afc48ebc54400147289dd45b74e2be843896dd8) (probably accidentally committed to the wrong branch), so master might need to be merged back into dev.

trang:
> Transifex gets its data from master, doesn't it?

Transifex gets its data from dev but this doesn't include the Terms of Use. 
image

The Terms of Use are put in a separate project for which I haven't set up automatic update. We don't change the content often so I think it's okay if that is a manual step. 

I uploaded the new version a moment ago and ran the `./tools/update-translations.sh` script. I noticed that [the script will skip a language if it hasn't been 100% translated](https://github.com/Tatoeba/tatoeba2/blob/dev/tools/update-translations.sh#L45). Maybe we'll want to change that, but in the meantime I manually downloaded the HTML for the languages that previously had the Terms of Use 100% translated.

In any case, thanks for pointing this out!

> By the way, while looking at what's different between master and dev, I noticed that there's [a commit on master that isn't on dev](https://github.com/Tatoeba/tatoeba2/compare/master#diff-84177656b172486dce13056e8afc48ebc54400147289dd45b74e2be843896dd8) (probably accidentally committed to the wrong branch), so master might need to be merged back into dev.

Indeed. I've merged.

Yorwba:
> I manually downloaded the HTML for the languages that previously had the Terms of Use 100% translated.

Looks like you accidentally replaced the Turkish Terms of Use with the Kabyle ones in the process: https://github.com/Tatoeba/tatoeba2/commit/abea3ae994bc0bea1ed27fa2eb3eb37c6aad0552#diff-dac27b5122f41280d28d7444c417d3dfd38416ce4bbdf506ccebacf8f637c138

trang:
Indeed. Thanks!

PR #2918, opened 2022-04-13
Remove user info from hidden messages/comments

trang:
One of the community admins reported that it feels like a kind of "brandmarking" to have the username displayed along with the warning "this was against our rules" when a Wall post is hidden. See example below:

image

Keeping the username displayed while the content is hidden has indeed no real benefits, and it was suggested that we only make this information visible to the admins and to the user who posted the comment.

This pull request hides the username as well as the profile picture to anyone who is not an admin or the author of the comment/post. It also fixes a bug where the replies below a hidden Wall post were displayed in red even though the post was not hidden.

image

### DIFF
--- a/webroot/css/layouts/elements.css
+++ b/webroot/css/layouts/elements.css
@@ -266,12 +266,12 @@ html[dir="rtl"] .comment-sentence .text {
     text-align: right;
 }
 
-.comment.inappropriate .warning-info p {
+.comment .inappropriate .warning-info p {
     padding: 0 10px;
     margin: 0;
 }
 
-.comment.inappropriate > md-card-content > .content {
+.comment .inappropriate .content {
     color: #d50000;
     border-top: 1px #f1f1f1 solid;
     padding-top: 10px;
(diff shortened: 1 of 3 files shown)


webroot/css/layouts/elements.css
```diff
@@ -271,7 +271,7 @@ html[dir="rtl"] .comment-sentence .text {
     margin: 0;
 }
 
-.comment.inappropriate .content {
+.comment.inappropriate > md-card-content > .content {
```
Yorwba, line 274: I think it would make more sense to move the `inappropriate` class so that it only applies to the content instead of doing this direct child check.

There are also some other rules using `.inappropriate` [later in the file](https://github.com/Tatoeba/tatoeba2/blob/2f0d22913d48e3f61ba2be1108d0c6b676789e30/webroot/css/layouts/elements.css#L505), but I think those are obsolete. Might be best to delete them.

trang: Thanks for the review! I applied your suggestions.

Yorwba: Nice :)

PR #2929, opened 2022-05-09
Speedup background jobs startup

jiru:
This is a minor improvement I have been wanting to do for a while. It allows background jobs (currently audio import, list download and language pair export download) to start immediately when the worker is idle, instead of having to wait for up to 10 seconds that the worker pick up the newly-enqueued task. This brings a substantial speedup for the user when the job execution time is nearly instant (e.g. download of a small list, audio import of a few files, or language pair export involving a language with few sentences).

I had to update from CakePHP 3.7 to 3.8 because the dependency dereuromark/cakephp-queue requires it.

:warning: This PR could cause various regressions on Tatoeba because of the CakePHP update. Needs testing.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,7 @@
     "require": {
         "cakephp/cakephp": "3.8.*",
         "cakephp/plugin-installer": "^1.0",
-        "dereuromark/cakephp-queue": "^3.11",
+        "dereuromark/cakephp-queue": "dev-cake3",
         "markstory/asset_compress": "^3.5",
         "patchwork/jsqueeze": "^2.0",
         "natxet/cssmin": "^3.0",
--- a/composer.json
+++ b/composer.json
@@ -4,7 +4,7 @@
     "license": "AGPL-3.0-only",
     "description": "Code for tatoeba.org, a multilingual sentence and translation database",
     "require": {
-        "cakephp/cakephp": "3.7.*",
+        "cakephp/cakephp": "3.8.*",
         "cakephp/plugin-installer": "^1.0",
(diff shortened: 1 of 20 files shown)


trang:
Would it be too ambitious to try and upgrade to 3.10 instead of 3.8 in this PR?

We have this dependency warning: https://github.com/Tatoeba/tatoeba2/security/dependabot/2

jiru:
I’d rather upgrade to 3.10 in a separate PR. Step by step.

jiru:
By the way, if it works, you may not have to refresh the audio import page at all. This happens because the browser is doing a POST-then-GET request. So if the worker have enough time to complete the import before the GET request arrives, you can see the import result right away.

trang:
Thanks for the explanation @jiru! I confirm that I didn't need to refresh the page manually when trying to import audio.

I've tested this out and it all works fine :) 

Just one thing I noticed is that the worker isn't being waken up when [refreshing the list of sentences that can be switched to CC0](https://github.com/Tatoeba/tatoeba2/blob/92ac45e698a3beb146d58a7f921d809d05ac4a61/src/Model/Licensing.php#L36). Was it just forgotten or there was a reason to leave it out?

jiru:
> Just one thing I noticed is that the worker isn't being waken up when [refreshing the list of sentences that can be switched to CC0](https://github.com/Tatoeba/tatoeba2/blob/92ac45e698a3beb146d58a7f921d809d05ac4a61/src/Model/Licensing.php#L36). Was it just forgotten or there was a reason to leave it out?

I forgot about that.

PR #2937, opened 2022-05-21
Exclude "no offsite use" audio from weekly export

jiru:
This pull request removes audio which license is set to "no offsite use" from the weekly export file. It makes sense not to export them if we do not allow use outside of Tatoeba.


docs/database/scripts/weekly_exports.sql
```diff
@@ -94,6 +94,7 @@ INTO OUTFILE '/var/tmp/sentence_comments.csv';
 -- Sentences with audio
 SELECT a.sentence_id, u.username, u.audio_license, u.audio_attribution_url
 FROM audios a LEFT JOIN users u on u.id = a.user_id
+WHERE a.audio_license <> ''
```
trang, line 97: Should be `u.audio_license` I think.

jiru: Thank you!

Issue #2932, opened 2022-05-17
Rohingya (rhg)


trang:
* Language name: Rohingya
* Language info: https://en.wikipedia.org/wiki/Rohingya_language
* ISO 639-3 code: rhg
* Public List: https://tatoeba.org/en/sentences_lists/show/170261
* Icon: ![rhg](https://user-images.githubusercontent.com/221850/168808694-fd1bd941-2136-4594-afcf-0fe09b2bf4aa.png)
* Requested by: https://tatoeba.org/en/user/profile/lbdx, https://tatoeba.org/en/user/profile/TWB

soliloquist-tatoeba:
The Rohingya people have an ethnic flag. 

https://en.wikipedia.org/wiki/Rohingya_flag
https://www.facebook.com/RohingyaFC/photos/1187833701386024
https://www.youtube.com/watch?v=70cf704mFNw

As they have also recently suffered from [ethnic cleansing](https://en.wikipedia.org/wiki/Rohingya_genocide), I suggest adopting a similar policy as that of Uyghur and using it instead of a blank icon.

Also, if the text direction were set to auto, sentences written in [different scripts](https://en.wikipedia.org/wiki/Rohingya_language#Writing_systems) would all display correctly.

PR #2940, opened 2022-05-24
Add language Rohingya (rhg)

trang:
This pull request addresses #2932.

### DIFF
--- a/src/Shell/SphinxConfShell.php
+++ b/src/Shell/SphinxConfShell.php
@@ -290,7 +290,7 @@ class SphinxConfShell extends Shell {
         # Ol Chiki, for Santali (sat)
         'U+1c50..U+1c7d',
         # Hanifi Rohingya
-        'U+10D00..U+10D23', 'U+10D30..U+10D39'
+        'U+10D00..U+10D27', 'U+10D30..U+10D39'
     );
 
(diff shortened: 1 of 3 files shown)


src/Shell/SphinxConfShell.php
```diff
@@ -289,6 +289,8 @@ class SphinxConfShell extends Shell {
         'U+a800..U+a827',
         # Ol Chiki, for Santali (sat)
         'U+1c50..U+1c7d',
+        # Hanifi Rohingya
+        'U+10D00..U+10D23', 'U+10D30..U+10D39'
```
Yorwba, line 293: Are you sure the first range shouldn't go all the way to U+10D27? The characters you left out are diacritics, but that doesn't necessarily make them particularly ignorable. (Also, ignoring them this way treats them like spaces, which is probably as undesirable as splitting French words after every character with an accent mark.)

trang: No, I wasn't sure (hence why I left the branch unmerged). I didn't know exactly what effect it would have to include or exclude these characters.

I updated the charset range during the [last Tatoeba stream](https://youtu.be/3aRkxkbus_A?t=1415), following your and gillux's recommendation.

Issue #2903, opened 2022-02-11
Small screens


octavo:
Make it possible to contribute on small screens (a responsive/mobile site or phone app).

Yorwba:
There's already a [project to implement a responsive UI](https://github.com/Tatoeba/tatoeba2/projects/3).

So the good news is that we're working on it. The bad news is that it's been a few years and only some parts of the site are responsive.

You can slightly improve things by going to your settings and enabling the experimental new design.

E.g. this is what the old design looks like in Firefox for Android on my phone:
<img src="https://user-images.githubusercontent.com/6555947/153663488-90abfb76-6b1e-40f9-9fcd-c7b2d385be52.png" alt="old design" height="400">
And this is the new design:
<img src="https://user-images.githubusercontent.com/6555947/153663662-65c816a7-fb1e-4fa8-a8f6-13bd4de99153.png" alt="new design" height="400">

Better, right?

If there's some specific page you want to use that doesn't work well on mobile even with the new design, we might be able to prioritize it if you tell us about it in more detail.

octavo:
@Yorwba The random sentence page.

ckjpn:
It would be nice to be able to easily use the fully-functioning, non-experimental version on a smaller screen.

Perhaps, it's time to prioritize finishing the creation of a mobile-friendly website.

PR #2959, opened 2022-06-15
Make sentence page responsive

Yorwba:
This PR attempts to address #2903 by making the sentence page layout adjust appropriately on smaller screens.

I started out by just making everything wrap, but that led to a serious shortage of vertical space, so in the end I opted to hide some elements on very small screens.
- the search bar behaves the same as on the search page (fully-featured on wide-enough pages, single input box otherwise)
- I put the sidebar inside a "metadata" sidenav modeled after the advanced search options on the search page
- the sentence navigation hides the input box for the sentence number (actually, I'm wondering whether we couldn't just remove it altogether, considering you could just edit the number in the browser's address bar) and the labels for the navigation arrows

### My phone

With sidenav open:

### From the Firefox responsive design tools at various widths
400px:

800px:

1200px:

### DIFF
--- a/src/Template/Element/search_bar.ctp
+++ b/src/Template/Element/search_bar.ctp
@@ -115,7 +115,8 @@ echo $this->Form->create(
         </div>
     </div>
 
-    <div layout="row" layout-align="center end">
+    <div layout-gt-xs="row" layout-align-gt-xs="center end"
+         layout="column" layout-align="center center">
         <div layout="column">
             <?php /* @translators: search language field label in top search bar */ ?>
             <label for="SentenceFrom"><?= __('From') ?></label>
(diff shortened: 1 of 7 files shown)


trang:
Finally had the chance to review this PR. It looks pretty good!

I only found one issue: when viewing the sentence page with the old design, the audio button goes to a new line.

image

This also happened when I redesigned the search results page and to solve it I reduced the width of `.sentence .content.column` to 500px but for the search page only (https://github.com/Tatoeba/tatoeba2/commit/71f1d41c7e403f883dbc84b7ff4af0310030179d). It is otherwise [set at 520px in element.css](https://github.com/Tatoeba/tatoeba2/blob/dev/webroot/css/layouts/elements.css#L784).

Once this issue is solved, I think this can be merged 🙂 

As for:
> I'm wondering whether we couldn't just remove it altogether, considering you could just edit the number in the browser's address bar

I think we will still need to have a UI element where the user can select a sentence ID to navigate to because it won't be intuitive for everyone that they can edit the URL. 
The placement for this input is definitely not ideal though and, if I remember correctly, there has been suggestions to either have it integrated into the top menu or/and have it on the homepage instead.

Yorwba:
> I only found one issue: when viewing the sentence page with the old design, the audio button goes to a new line.

Good catch and thanks for pointing me to the solution!

Issue #2955, opened 2022-06-12
Create an all_sentences_with_audio.tar.bz2, or put all sentences with audio in sentences_with_audio.tar.bz2 .


ckjpn:
Since https://downloads.tatoeba.org/exports/sentences_with_audio.tar.bz2 no longer has all sentences with audio, could you please create another file each week that is similar to this, but that includes all sentences with audio files?

Maybe you could call it all_sentences_with_audio.tar.bz2, Or, perhaps you could just include all sentences with audio like the file used to be.

In addition to what I need, sharptoothed's page relies on this information.

https://tatoeba.j-langtools.com/userchart/chart2.php

jiru:
Next Saturday’s export will include all sentences with audio, like it used to be.

trang:
@ckjpn Did the audio get uploaded after all? I'm not seeing any audio to import on https://tatoeba.org/en/audio/import. 

FYI there's been a Tatoeba update today that impacts audio uploads among other things (#2929). When I deployed the changes, there was some issues at first that I didn't notice right away. I needed to run some database migration but wasn't aware. I found out later on and fixed the issues. Perhaps you tried to upload the audio files while it was not fixed yet.

ckjpn:
Why does it take a very long time sometimes and not so long other times?

Here is the one I'm waiting on now.
I notice that I have to be careful to make sure one import is done before
starting another one by a different "author."  Otherwise, files get
uploaded with the wrong username assigned.

Import recordingsAudio import status

Import task enqueued 15 minutes ago. Please refresh the page in a little
while to check if import is completed.
Files detected

A total of 22 files have been detected inside the import directory.
File nameSentence idLanguageExisting audioMay be imported
10903652.mp3 10903652 <https://tatoeba.org/en/sentences/show/10903652>
English None Yes
10902100.mp3 10902100 <https://tatoeba.org/en/sentences/show/10902100>
English None Yes
10908305.mp3 10908305 <https://tatoeba.org/en/sentences/show/10908305>
English None Yes
10908676.mp3 10908676 <https://tatoeba.org/en/sentences/show/10908676>
English None Yes
10910106.mp3 10910106 <https://tatoeba.org/en/sentences/show/10910106>
English None Yes
10910185.mp3 10910185 <https://tatoeba.org/en/sentences/show/10910185>
English None Yes
10903650.mp3 10903650 <https://tatoeba.org/en/sentences/show/10903650>
English None Yes
[...]

PR #2965, opened 2022-06-26
Increase Queue maxworkers to 2

trang:
CK reported that it sometimes took several minutes (up to 19 minutes) to import audio (see https://github.com/Tatoeba/tatoeba2/issues/2955#issuecomment-1163012139).

I noticed that when we run a worker, even if the `runworker` command is executed at 0 second, the `created` date is at 1 second and sometimes at 2 seconds.

Example:
```
MariaDB [tatoeba]> select * from queue_processes;
+-----+-------+---------------------+---------------------+-----------+--------+------------------------------------------+
| id  | pid   | created             | modified            | terminate | server | workerkey                                |
+-----+-------+---------------------+---------------------+-----------+--------+------------------------------------------+
| 199 | 18943 | 2022-06-26 17:00:01 | 2022-06-26 17:01:22 |         0 | sloth  | 5b689d57e599b30b2549d4361488a07592877172 |
+-----+-------+---------------------+---------------------+-----------+--------+------------------------------------------+
```

Because of this, when the worker is terminated (after 20 minutes in our case) and we try to run the next worker, it fails because the previous worker is still running and we end up with no workers at all.

As a result, I have updated the config on prod to increase the maximum number of workers to 2. Not sure if there's a better solution.

### DIFF
--- a/config/app_queue.php
+++ b/config/app_queue.php
@@ -23,7 +23,7 @@ return [
         'multiserver' => false,
 
         // set this to a limit that can work with your memory limits and alike, 0 => no limit
-        'maxworkers' => 1,
+        'maxworkers' => 2,
 
         // instruct a Workerprocess quit when there are no more tasks for it to execute (true = exit, false = keep running)
         'exitwhennothingtodo' => false,


jiru:
When I implemented the background job thing, I assumed that only one worker would run at the same time, so I am afraid having two workers will cause some race condition. We’d need to review the code to make sure it’s fine.

I think it would be safer, for now, to lower the value of `$config['Queue']['workermaxruntime']` a bit.

ckjpn:
Re: https://tatoeba.org/en/audio/import

My most-recent attempt to import files took over 8 minutes. 
Maybe much longer, they still haven't uploaded at the time of this message.

If the following now has an audio file, they did eventually get added.

https://tatoeba.org/en/sentences/show/11062081

22:35
Sunday, August 14, 2022
Coordinated Universal Time (UTC)

ckjpn:
Just to let you know that the previous import did finally complete.

ckjpn:
This problem had disappeared for a long time. Now it seems to be. permanently back.
Is there some kind of "reset" you can do to at least temporarily fix this problem.
Screen Shot 2022-08-16 at 9 59 52



00:59
Tuesday, August 16, 2022
Coordinated Universal Time (UTC)

trang:
@ckjpn The problem reappeared because we updated Tatoeba this weekend, and this erased the temporary fix that gillux did. I've re-applied the fix just now so it should be fine again.

Issue #2977, opened 2022-07-23
Ability to filter sentences in lists by language


Guybrush88:
**Story** or **Context**
Currently, when users want to browse a list with sentences in different languages, they have to open an advanced search to apply a filter and search for sentences in a specific language, and this might take longer than expected to see only a given language. I came up with this thought while browsing this list: https://tatoeba.org/it/sentences_lists/show/170280

**Idea**
To improve this page, there could be a dropdown menu above or below the "show translations" one, so users could be able to filter sentences by language directly from the page of the list.

ckjpn:
An additional advantage would be that advanced searches are limited to 1,000 results.  There are no such limits on lists.

@Guybrush88
You could do this kind of filtering offline using the weekly exported files if you know how.

vinkaks:
Implementing the language filter seems very straightforward and non-invasive. It will however require implementing some minor interface changes and creating related translation for the "action". I tucked away the filter by language option under "actions" since the title bar is already full of other buttons and the list title itself. Still need to add some margin between the two actions.

![tatoeba list filter 1](https://user-images.githubusercontent.com/21044267/183522335-389ce479-69da-4c50-8581-8a8a0d901492.png)

@jiru can I do a PR on this enhancement?  

BTW: If a language that does not exist in the list is chosen as a filter, this results in an empty result. In such a case maybe the default "List is empty" notice may be a little confusing? 

![tatoeba list filter 2](https://user-images.githubusercontent.com/21044267/183523879-3b31da94-875b-4d4d-af0e-388f61a09b30.png)

ckjpn:
Perhaps a message that says that the selected "filter by language" request gets no results would work better than "This list is empty."

Yorwba:
I think we should try to keep the interface consistent across different parts of the website.

Filtering sentences by language and choosing which translations to show is already possible on the page listing all sentences in a given language (e.g. [all sentences in Vietnamese](https://tatoeba.org/en/sentences/show_all_in/vie/none))
![sentences in Vietnamese](https://user-images.githubusercontent.com/6555947/183746079-155cced7-46d6-414c-8f38-2f78d3412699.png)

The page listing a user's sentences in a given language displays a message "`$user` does not own any sentences in `$lang`" when there are no results. (e.g. [I don't have any Bengali sentences](https://tatoeba.org/en/sentences/of_user/Yorwba/ben))
[...]

PR #2980, opened 2022-08-10
Issue 2977 - Added ability to filter sentences in lists by language

vinkaks:
This pull request addresses issue #2977 

Used a query Parameter in the controller to avoid messing with existing links to the view. i.e so links which do not have a $translationsLang can still work normally.

Made displaying "none" option, in the languagesArrayShowTranslationsIn() optional with defaulting === true. Makes changes to function  backward compatible.


jiru:
Hi @vinkaks, thank you for your pull request! 🙂 I am going to review your changes, although I cannot test out your code at the moment.

src/Controller/SentencesListsController.php
```diff
@@ -147,6 +147,7 @@ public function collaborative($filter = null)
      */
     public function show($id = null, $translationsLang = null)
     {
+        $lang = $this->request->getQuery('lang')??'und';
```
jiru, line 150: You can pass `'und'` as second parameter to `getQuery()` instead.

jiru:
I think it would be better to have this page work like `/sentences/show_all_in`, with sentence language filter as first parameter and translation language filter as second parameter. This would allow reusing `createFilterByLangMod()`, which is a good thing because the interface would be more consistent, and the code could be factorized. However, since the list page currently expects the translation language as first parameter, it is tricky because we could break many existing list page links by changing the URL parameters. But not impossible: you can make the action backward compatible by permanently redirecting `/sentences_lists/show/ID/LANG` to `/sentences_lists/show/ID/und/LANG`. This is what I would do.

src/Controller/SentencesListsController.php
```diff
@@ -169,13 +170,20 @@ public function show($id = null, $translationsLang = null)
...
         $this->loadModel('SentencesSentencesLists');
 
-        $query = $this->SentencesSentencesLists->find()
-            ->where(['sentences_list_id' => $id])
+        $query = $this->SentencesSentencesLists->find();
+        if ($lang!="und") {
+            if ($lang == 'unknown') {
+                $query->where(['lang IS' => null]);
+            } else {
+                $query->where(['lang' => $lang]);
```
jiru, line 178: If I remember correctly, I think CakePHP will automatically check for parameter nullity and add the `IS` keyword for you. Have you tried passing `['lang' => $lang]` with $lang being null?

jiru:
So we are currently maintaining two interfaces, that we call the "new" and "old" design. Some parts of the UI are displayed differently depending on a setting in your profile. The list page happens to be displayed differently in new and old design. We hope to get rid of the old design at some point but it is still there. Your changes only affect the new design, so you would have to do the same change for the [old design](https://github.com/Tatoeba/tatoeba2/blob/dev/src/Template/SentencesLists/show.ctp), too.

jiru:
I have concerns that this could run slowly on huge lists. Some contributors use automation to build lists containing [hundreds of thousands of sentences](https://tatoeba.org/en/sentences_lists/index?sort=numberOfSentences&direction=desc). When filtering a list by language, the resulting SQL query may (or may not) run slowly and hog resources. So it would be better to test that out before merging, either locally or on https://dev.tatoeba.org/.

src/Controller/SentencesListsController.php
vinkaks: The if else block is to catch the case where `lang` is `unknown` not `null`. So we can't directly make use of cakePHP's automatic null check during query generation. I could however use a ternary to convert `unknown` to `null` then use just one query. 

`$query->where(['lang IS' => $lang==='unknown'?null:$lang]);`

vinkaks:
Updated routing to make same as for sentences.. Users who try using two params assuming the schema is consistent  

`sentences_lists/show/:id/:filterLang/:translationLang`

and use

`sentences_lists/show/:id/:filterLang`

will be likely be surprised to end up at 

`sentences_lists/show/:id/und/:translationLang`

tests/TestCase/Controller/SentencesListsControllerTest.php
```diff
@@ -41,7 +41,7 @@ public function accessesProvider() {
             [ '/en/sentences_lists/show/', null, '/en/sentences_lists/index' ],
             [ '/en/sentences_lists/show/1', null, true ],
             [ '/en/sentences_lists/show/1', 'kazuki', true ],
-            [ '/en/sentences_lists/show/1/fra', 'kazuki', true ],
+            [ '/en/sentences_lists/show/1/und/fra', 'kazuki', true ],
```
jiru, line 44: By the way, feel free to add a new unit test testing for the new redirection code.

src/Controller/SentencesListsController.php
```diff
@@ -145,8 +145,14 @@ public function collaborative($filter = null)
      *
      * @return mixed
      */
-    public function show($id = null, $translationsLang = null)
+    public function show($id = null, $lang = null, $translationsLang = null)
     {
+        // this makes me cringe but it works
```
jiru, line 150: I suggest you change this comment into something that explains why the redirection is needed, i.e. backward compatibility with legacy URLs.

src/Controller/SentencesListsController.php
```diff
@@ -169,13 +175,16 @@ public function show($id = null, $translationsLang = null)
         $this->loadModel('Sentences');
         $this->loadModel('SentencesSentencesLists');
 
-        $query = $this->SentencesSentencesLists->find()
-            ->where(['sentences_list_id' => $id])
+        $query = $this->SentencesSentencesLists->find();
+        if ($lang!="und") {
+            $query->where(['lang IS' => $lang == 'unknown' ? null : $lang]);
```
jiru, line 180: I think you can remove the `IS` keyword now.

src/Template/SentencesLists/show_angular.ctp
```diff
@@ -19,6 +19,7 @@ $this->set('title_for_layout', $this->Pages->formatTitle($listName));
 ?>
 
 <div id="annexe_content">
+    <?php $this->Lists->displayFilterByLangDropdown($listId, $filterLanguage, $translationsLang); ?>
```
jiru, line 22: How about calling `CommonModules->createFilterByLangMod()` instead?

vinkaks: createFilterByLangMod seems to interfere with the `actions` dropdown, i would need to do some extra refactoring to make it work it seems. Is there a particular issue with current function? I found it used elsewhere in the codebase.

src/Controller/SentencesListsController.php
vinkaks: the docs for 3.8 seem to indicate it is necessary for automatic `IS NULL` creation.

https://book.cakephp.org/3/en/orm/query-builder.html#automatic-is-null-creation.

src/View/Helper/ListsHelper.php
```diff
@@ -309,14 +309,44 @@ public function displayTranslationsDropdown($listId, $translationsLang = null) {
                 'name' => 'translationLangChoice',
                 'languages' => $this->Languages->languagesArrayShowTranslationsIn(),
                 'initialSelection' => $translationsLang,
-                'onSelectedLanguageChange' => "window.location.pathname = '$path' + language.code",
+                'onSelectedLanguageChange' => "window.location.href = '$path' + '$filterLanguage' + '/'+language.code",
```
Yorwba, line 312: Note that by writing to `href`, you're overwriting the *entire* URL. Using `pathname` here is a deliberate choice, see commit 6781556405b8d30e778890fc0481475b3975d7ab.

src/View/Helper/ListsHelper.php
```diff
@@ -309,14 +309,44 @@ public function displayTranslationsDropdown($listId, $translationsLang = null) {
...
+            <h2><?php echo __('Sentences in:'); ?></h2>
+            <?php
+            $path = $this->Url->build(['action' => 'show', $listId]) . '/';
+            echo $this->_View->element(
+                'language_dropdown',
+                array(
+                    'name' => 'filterLanguageSelect',
+                    'languages' => $this->Languages->languagesArrayShowTranslationsIn(true, false),
+                    'initialSelection' => $filterLanguage,
+                    'onSelectedLanguageChange' => "window.location.href = '$path' +language.code +'/'+ '$translationsLang'",
```
Yorwba, line 341: `href` should probably be `pathname` here, too.

Yorwba:
You have some whitespace-only changes that I think should best be reverted. They clutter the diff, and when someone later tries to use `git blame` to find the origin of one such line, they'll end up at the commit where you changed the whitespace instead of the last meaningful change.

I guess you didn't change these manually and it's your editor that reformatted the code automatically. Maybe you should reconfigure it to be a bit less aggressive about reformatting things, or at least avoid committing those changes.

src/Template/SentencesLists/show_angular.ctp
jiru: You are welcome to do such refactoring if you want. You can as well display the "filter by lang" block as separate from the "actions" block, it's okay.

By the way, `displayFilterByLangDropdown()` is a function you introduced is this pull request. While there may not be any problem with it per se, the more we reuse existing code, the less chances we introduce new bugs and the easier it gets to maintain the code.

vinkaks: I agree, I think more of the helper functions can be made to be reusable across views. 

Sorry for bad phrasing in previous comment, I meant `displayFilterByLangDropdown()` is near identical to `Lists->displayTranslationsDropdown()`. My function is basically `Lists->displayTranslationsDropdown()` with just the path building, and options array changed.

It seems `CommonModules->createFilterByLangMod()` assumes the last parameter is the `lang` this makes it incompatible with the pattern I was using `/:id/:lang/:translationsLang` which is based on `sentences/show_all_in/:lang/:translationsLang`.

jiru: You're correct, we cannot reuse createFilterByLangMod here.

src/View/Helper/ListsHelper.php
```diff
@@ -297,24 +297,56 @@ public function displayDownloadLink($listId)
...
+        <div class="section md-whiteframe-1dp">
+            <h2><?php echo __('Sentences in:'); ?></h2>
+            <?php
+            $path = $this->Url->build(['action' => 'show', $listId]) . '/';
+            echo $this->_View->element(
+                'language_dropdown',
+                array(
+                    'name' => 'filterLanguageSelect',
+                    'languages' => $this->Languages->languagesArrayShowTranslationsIn(true, false),
+                    'initialSelection' => $filterLanguage,
```
jiru, line 342: I think there is a contradiction here. You are filling a sentence language filter with choices from a list of languages for translations (languagesArrayShowTranslationsIn()). I think you could use languagesArrayAlone() instead and not add this new parameter for disabling the option "none".

vinkaks: Thanks, had missed that function. Had initially found `onlyLanguagesArray()` . I think eventually some of these functions will need to be composed into fewer, more generalized functions. 

I imagine having a single `buildLanguageArray(params)` function that outputs a language list array used in views would be ideal.

src/Template/SentencesLists/show_angular.ctp
```diff
@@ -19,7 +19,31 @@ $this->set('title_for_layout', $this->Pages->formatTitle($listName));
...
+                echo '<p>';
+                $this->Lists->displayEditableByOptions($listId, $list['editable_by']);
+                echo '</p>';
+                ?>
+            </ul>
+        </div>
+        <?php
+    }
+    ?>
+    <!-- <?php $this->Lists->displayListsLinks(); ?> -->
```
Yorwba, line 46: This line is commented out, but shouldn't be.

Yorwba (changes requested):
Sorry for taking so long to review your latest changes.

Other than the commented-out line, there's still one thing left to do:

You'll need to change the 'This list is empty.' message to say something like 'This list does not contain any sentences in {language}.' if a filter language is defined.

vinkaks:
Given the context, I have lowered `$this->Lists->displayListsLinks();` in the aside menu because 

- in the SentenceList context, filters and information about the list takes a higher priority from a UX perspective than links navigating away from the list. 
- users can still easily navigate away from the sentence list using the main navigation

src/Template/SentencesLists/show_angular.ctp
```diff
@@ -186,37 +187,37 @@ $this->set('title_for_layout', $this->Pages->formatTitle($listName));
...
-                echo format(
-                    __(
-                        'You can create new sentences directly in this list by clicking on the '.
-                        '{addSentenceButton} icon in the header of this section.', true
-                    ),
-                    ['addSentenceButton' => '<md-icon>add</md-icon>']
-                );
+                    echo format(
+                        __(
+                            $filterLanguage=='und'?'This list is empty.':'This list does not contain any sentences in {language}'
```
Yorwba, line 197: Putting the `'und'` inside the `__()` makes xgettext think that `'und'` is a message that needs to be translated. Better have two separate calls to `__()` in the two branches of the ternary operator.

Also, I'd add a period after `{language}`, ending the sentence without a period looks weird to me.

src/Template/SentencesLists/show_angular.ctp
```diff
@@ -186,37 +187,37 @@ $this->set('title_for_layout', $this->Pages->formatTitle($listName));
...
+            </p>
+            <?php
+            if ($permissions['canAddSentences']){
                 ?>
-            </div>
+                <div class="hint">
+                    <?php
+                    echo format(
+                        __(
+                            'Click on the {addNewSentence} icon above to add a new sentence to the list. You can also add existing sentences to this list, from other pages, by clicking on '.
```
Yorwba, line 210: Note that most supported UI languages aren't actively maintained. Adding a new UI string for a new feature is okay, and it's unavoidable that it might not get translated for a long time, but I think it's best not to change existing UI strings unless there's something wrong with them, to protect existing translations from getting lost.

So I'd keep the two hints separate and use the same wording as before.

Yorwba (changes requested):

Yorwba (approved):

Yorwba:
@vinkaks Thanks a lot for you patience working on this pull request!

Issue #2969, opened 2022-07-09
Saraiki (skr)


trang:
* Language name: Saraiki 
* Language info: https://en.wikipedia.org/wiki/Saraiki_language
* ISO 639-3 code: skr
* Public List: https://tatoeba.org/en/sentences_lists/show/170360
* Icon: ![skr](https://user-images.githubusercontent.com/221850/178106899-ba2de088-4b1d-4bc2-ab34-ff10ced10409.png)
* Requested by: https://tatoeba.org/en/user/profile/Saraiki

trang:
Just for the record, the first icon design was with the language code on the right side (as we usually have): ![skr](https://user-images.githubusercontent.com/221850/178107095-29a6a48c-5e0b-4775-b783-f1642dc3d40b.png)

It was then suggested to have the "skr" code on the left side, which deviates a bit from our regular design pattern but it allows to keep the aspect ratio of the flag, which we think makes it look better.

PR #2981, opened 2022-08-14
Added language Saraiki

DJ-Saidez:
This PR solves #2969


Yorwba:
Thank you for the pull request!

In the ["Adding a new language" guide on the wiki](https://github.com/Tatoeba/tatoeba2/wiki/Adding-a-new-language), one of the steps is

> - If the language is a right to left language, also add it in `getLanguageDirection($lang)`.

Saraiki seems to be a RTL language, so that step is missing.

Other than that, looks good to me.

DJ-Saidez:
Thanks for letting me know!
It seems that SaudQahir in a [previous commit](https://github.com/SaudQadir/tatoeba2/commit/84e6b9c91389a22e8d7383d546870da9b7e78bf5) already did so. Do I have to do the same and then push another pull request?
If I were to do it myself, would I likewise just go to $rightToLeftLangs array and add its 639-1 code there, after I link it and its 639-3 code together on the map?
I ask because, since I can't host a virtual machine of Tatoeba on my Mac, any code I write has to be written directly to my branch on GitHub.

Yorwba:
> Do I have to do the same

Basically yes, but I'd also add a comma after the `"skr"`, so that you don't need to modify the line the next time you're adding something below.

> and then push another pull request?

Just make the change on the issue-2969 branch, then it will automatically get added to this pull request.

> I ask because, since I can't host a virtual machine of Tatoeba on my Mac, any code I write has to be written directly to my branch on GitHub.

You could probably at least install git so you can make changes locally without directly publishing them on GitHub, but without a virtual machine for testing, that would of course not help much.

Anyways, I think you have the right idea.

DJ-Saidez:
Sorry for the delay! I did that last part, and I'll get on the other language right now.
Now that I know how this works, I should be able to manage this quite faster now.

src/Lib/LanguagesLib.php
```diff
@@ -183,6 +183,7 @@ public static function get_Iso639_3_To_Iso639_1_Map() {
             'sag' => 'sg', // Sango
             'san' => 'sa', // Sanskrit
             'sin' => 'si', // Sinhala
+            'skr' => 'sr', // Saraiki
```
Yorwba, line 186: `sr` is the ISO 639-1 code for Serbian (you can find it under `srp` below). I don't think there's an ISO 639-1 code for Saraiki, so this change will have to be reverted.

DJ-Saidez: Done, thanks for checking for me!

DJ-Saidez: Is anything else still a problem?

mpQadir: Language is not added in site yet.

trang: @mpQadir The language will be added on the website when this pull request and merged and when the changes are deployed on production. For now this pull request is still open.

@DJ-Saidez I think everything's good now :)

Yorwba: I'll merge this now and deploy on Sunday. @DJ-Saidez Thank you for your work!

Yorwba: This PR is now live. 🎉

Issue #2964, opened 2022-06-25
Neapolitan (nap)


trang:
* Language name: Neapolitan 
* Language info: https://en.wikipedia.org/wiki/Neapolitan_language
* ISO 639-3 code: nap
* Public List: https://tatoeba.org/en/sentences_lists/show/170321
* Icon: https://en.wikipedia.org/wiki/Naples#/media/File:Flag_of_Naples.svg
* Requested by: https://tatoeba.org/en/user/profile/gouchi

Issue #2911, opened 2022-03-04
Silesian (szl)


trang:
* Language name: Silesian
* Language info: https://en.wikipedia.org/wiki/Silesian_language
* ISO 639-3 code: szl
* Public List: https://tatoeba.org/en/sentences_lists/show/170066
* Icon: https://en.wikipedia.org/wiki/Flag_of_Silesia_and_Lower_Silesia#/media/File:Flag_of_Silesia.svg
* Requested by: https://tatoeba.org/en/user/profile/vxern

PR #2987, opened 2022-08-29
Added language Neapolitan

DJ-Saidez:
This issue solves #2964 

I should be more efficient with these from now on.
Assuming everything is good, once the changes take place I'll let the requesters know.


webroot/img/flags/nap.svg
```diff
@@ -1,4 +1,4 @@
...
...ovL3d3dy53My5vcmcvMjAwMC9zdmciPjxkZWZzCiAgICAgaWQ9ImRlZnM3MDIiIC8+PHNvZGlwb2RpOm5hbWVkdmlldwogICAgIGlkPSJuYW1lZHZpZXc3MDAiCiAgICAgcGFnZWNvbG9yPSIjZmZmZmZmIgogICAgIGJvcmRlcmNvbG9yPSIjMDAwMDAwIgogICAgIGJvcmRlcm9wYWNpdHk9IjAuMjUiCiAgICAgaW5rc2NhcGU6c2hvd3BhZ2VzaGFkb3c9IjIiCiAgICAgaW5rc2NhcGU6cGFnZW9wYWNpdHk9IjAuMCIKICAgICBpbmtzY2FwZTpwYWdlY2hlY2tlcmJvYXJkPSIwIgogICAgIGlua3NjYXBlOmRlc2tjb2xvcj0iI2QxZDFkMSIKICAgICBzaG93Z3JpZD0iZmFsc2UiCiAgICAgaW5rc2NhcGU6em9vbT0iMC4xMzIxODc1IgogICAgIGlua3NjYXBlOmN4PSItMTA5My4xNDQyIgogICAgIGlua3NjYXBlOmN5PSI1NDguNDYzMzYiCiAgICAgaW5rc2NhcGU6d2luZG93LXdpZHRoPSIxOTIwIgogICAgIGlua3NjYXBlOndpbmRvdy1oZWlnaHQ9IjEwMDUiCiAgICAgaW5rc2NhcGU6d2luZG93LXg9Ii05IgogICAgIGlua3NjYXBlOndpbmRvdy15PSItOSIKICAgICBpbmtzY2FwZTp3aW5kb3ctbWF4aW1pemVkPSIxIgogICAgIGlua3NjYXBlOmN1cnJlbnQtbGF5ZXI9InN2ZzY5OCIgLz48cmVjdAogICAgIHdpZHRoPSIzMDAiCiAgICAgaGVpZ2h0PSI0MDAiCiAgICAgZmlsbD0iI0ZGREI3MiIKICAgICBpZD0icmVjdDY5NCIKICAgICBzdHlsZT0iZmlsbDojZmVkYTAwO2ZpbGwtb3BhY2l0eToxIiAvPjxyZWN0CiAgICAgeD0iMzAwIgogICAgIHdpZHRoPSIzMDAiCiAgICAgaGVpZ2h0PSI0MDAiCiAgICAgZmlsbD0iI0YwMTUxNiIKICAgICBpZD0icmVjdDY5NiIKICAgICBzdHlsZT0iZmlsbDojZTUwMDA1O2ZpbGwtb3BhY2l0eToxIiAvPjwvc3ZnPgo=" width="30" height="20"/>
```
Yorwba, line 3: Replacing the PNG with an SVG was a good idea, but the way you did it was a bit unorthodox.

We use SVG-with-PNG-embedded for many flags so that we can have a uniform SVG-based format even for images where only a PNG is available, and then slowly switch to pure SVG. (Most recently, a few flags were switched to SVG in PR #2865 .)

However, instead of turning SVG-with-PNG-embedded (which had 3154 bytes) into SVG-with-SVG-embedded (with 1958 bytes), you should have used the SVG you got from Wikipedia directly (which has 1310 bytes), only changing the size to 30×20. Changing the size is something you could do with an SVG editor like [Inkscape](https://inkscape.org/)

... in this case I had a look at the Wikipedia SVG and it still contains a lot of unnecessary fluff, so that it could be pared down to 230 bytes (without the newlines I added below for readability)
```svg
<?xml version="1.0"?>
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="30" height="20">
<rect width="15" height="20" fill="#FEDA00"/>
<rect width="15" height="20" fill="#E50005" x="15"/>
</svg>
```
... but this last optimization is not something I'd expect you to do.

DJ-Saidez: With the rect commands (for lack of a better term), you’re setting the dimensions, the color, and location of the rectangle, right?

I couldn’t tell what was fluff and what wasn’t, because when I downloaded directly from Wikipedia, it looked like the bottom half was white, but when I ran it through the Data URI converter it gave me only the flag as far as I could tell, without that empty space.

I didn’t realize to change the dimensions before converting it, I’ll do it next time.

Thanks for the pointers and the help, I’ll try it again later today!

DJ-Saidez: Since I'm already familiar with Photopea/Photoshop, I used it to scale it down to 30x20. I think that should be do it better!

Yorwba: > With the rect commands (for lack of a better term), you’re setting the dimensions, the color, and location of the rectangle, right?

Yes, exactly. For flags that are made entirely out of rectangular shapes, defining the image by hand like this is relatively straightforward.

> I couldn’t tell what was fluff and what wasn’t

Me neither :) Looking at your new version, I noticed that there was another big chunk that I could've removed. (Apparently Photoshop does some file size optimization automatically.)

> Since I'm already familiar with Photopea/Photoshop, I used it to scale it down to 30x20.

That did work, but you still embedded it as an SVG-in-SVG. For images already in SVG format, there should never be a need to use the Data URI converter.

DJ-Saidez: Then if I have a SVG file, how would that be turned into Data URI without the converter?

Yorwba: Just replace `nap.svg` by the SVG file you created. The Data URI is only required for non-SVG images.

webroot/img/flags/nap.svg
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   version="1.0"
+   width="600"
+   height="400"
+   id="svg698"
```
trang, line 6: You will have to remove this `id="svg698"` because it messes up with the build of the language icons asset.

I updated the wiki article about adding a new language and mentioned this issue in [step 2](https://github.com/Tatoeba/tatoeba2/wiki/Adding-a-new-language#2-add-the-language-icon):

> - Open the file with a text editor:
>   - Remove comments if there are any.
>   - Remove the "id" attribute from the <svg> tag if there is any.
>   - Not doing so will trigger an error with the asset build (see for example [issue #2911](https://github.com/Tatoeba/tatoeba2/issues/2911)).

webroot/img/flags/nap.svg
```diff
@@ -3,16 +3,13 @@
...
    height="400"
-   id="svg698"
    sodipodi:docname="Flag_of_Naples.svg"
    xml:space="preserve"
    inkscape:version="1.2 (dc2aedaf03, 2022-05-15)"
    xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
    xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
    xmlns="http://www.w3.org/2000/svg"
    xmlns:svg="http://www.w3.org/2000/svg"><defs
-     id="defs702" /><sodipodi:namedview
```
Yorwba: You deleted a bit too much here. (Or too little, depending on how one looks at it.)
Only the `id="defs702"` part is the id attribute. The `/>` closes the preceding `<defs` and then the following `<sodipodi:namedview` starts a new element.
So after deleting this line, all the attributes that were part of the `sodipodi:namedview` element are now attached to the `defs` element.

src/Lib/LanguagesLib.php
```diff
@@ -690,6 +690,7 @@ public static function languagesInTatoeba()
                 'swc' => __d('languages', 'Congo Swahili'),
                 'rhg' => __d('languages', 'Rohingya'),
                 'mfa' => __d('languages', 'Kelantan-Pattani Malay'),
+                'skr' => __d('languages', 'Saraiki'),
```
Yorwba, line 693: This did not actually fix the conflict. Although the content is the same, you're still modifying a line that was also modified in a parallel pull request. It's necessary to explicitly specify the intended result and mark the conflict as resolved.

GitHub has [specialized functionality for conflict resolution](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/addressing-merge-conflicts/resolving-a-merge-conflict-on-github). I think that should also be available to you as the author of the pull request. (If not, I can just resolve the conflict when merging.)

DJ-Saidez: I tried resolving it myself, let's see if it works.

webroot/img/flags/nap.svg
DJ-Saidez: I got rid of the trailing <defs I left alone, maybe that'll settle it? Or should I readd everything except id="svg698"?

DJ-Saidez:
If I fail the check, does that mean there's some code that will cause an error if it's actually run?

src/Lib/LanguagesLib.php
Yorwba: It did work (there's no conflict anymore), but it seems like you accidentally added a new empty line in the process.

webroot/img/flags/nap.svg
Yorwba: > Or should I readd everything except id="svg698"?

That would be the option least likely to cause breakage. Right now, the attributes that used to belong to `<sodipodi:namedview` are just awkwardly hanging around without a tag.

Yorwba:
> If I fail the check, does that mean there's some code that will cause an error if it's actually run?

Usually yes, though it could also be that the tester is broken.

You can click "Details" next to the failed check to see what went wrong. In this case, the error messages are about the namespace prefixes "sodipodi for docname" and "inkscape for version", referring to the attributes
```
sodipodi:docname="Flag_of_Naples.svg"
inkscape:version="1.2 (dc2aedaf03, 2022-05-15)"
```

You can try to remove those, but I think this might actually be a bug in our asset-building code. I'll try to have a closer look next weekend.

Yorwba:
> You can try to remove those, but I think this might actually be a bug in our asset-building code. I'll try to have a closer look next weekend.

As I suspected, our asset-building code removes the XML namespace definitions for `sodipodi` and `inkscape`, but not the attributes that rely on these namespaces. Fortunately, the attributes are only relevant for editing the file and unnecessary when displaying it, so we can simply remove them.

I've since discovered a tool that can do this removal and a bunch of other optimizations for you: [SVG Cleaner](https://svgcleaner.com/). It also removes id attributes. So if you take the original SVG from Wikipedia, open it with SVG Cleaner and download the result, things should just work.

DJ-Saidez (commented):
This was to combine it with other pulls that were done on main since I forked into issue-2964

src/Lib/LanguagesLib.php
DJ-Saidez: I don't see where the empty line is. If I leave it like that will it run a mistake?

DJ-Saidez:
I ran the original SVG through the cleaner, hopefully that should work now.

DJ-Saidez:
@Yorwba Could you verify everything is fine now?

src/Lib/LanguagesLib.php
Yorwba: The empty line is number 695: https://github.com/Tatoeba/tatoeba2/pull/2987/commits/94e3e8bc1e7ea5d558b36d75e9469ffeecb1a5ab#diff-2f26e6cbce81128867f0a6ba0d187c4ff3b76872c5af503c4780dc9f54bcf02aR695

But it's not a big deal, so I'll merge this now.

Issue #2984, opened 2022-08-22
Incorrectly formatted CSV files


CangarejoAzul:
All the files on the Downloads page are TSV files, and not CSV files, except for queries.csv. The file extensions should be changed.

Some fields in the user_languages.csv file contain newlines, which should not be allowed in a TSV file. Maybe they should be replaced with spaces. And some fields are empty, which maybe should also not be allowed.

Some fields in the queries.csv file contain commas, which should not be allowed, because that file is a CSV file. Maybe they should be replaced with spaces or maybe the file should be converted to TSV.

queries.csv is not encoded in UTF-8, although it should be.

vinkaks:
Seems it be resolved with a `gsub("[\r\n\t]", " ", $field)` in the export scripts.

For the following input 

```
this is a test
of spaces
x	this is a tab
```

We get

`this is a test \nof spaces \nx\tthis is a tab`

The original text can be reproduced from this output. I am hesitant to purge all special characters for that reason. Anyone who wants to get rid of the escaped `\n` `\t` etc can do so themselves? Will PR something. 

@CangarejoAzul where is queries.csv downloaded from?

CangarejoAzul:
https://downloads.tatoeba.org/stats/queries.csv.bz2

vinkaks:
Turns out some of the files are exported straight from SQL. So the output can just be escaped like

-- User skill level per language
`SELECT ul.language_code, ul.level, u.username, REPLACE(REPLACE(REPLACE(REPLACE(ul.details,' ',' '),'\t','\\t'),'\n','\\n'),'\r','\\r') as details FROM users_languages ul INNER JOIN users u ON ul.of_user_id = u.id ORDER BY ul.language_code ASC, ul.level DESC, u.username ASC;`

CangarejoAzul:
Out of curiosity, where can I download a copy of Tatoeba's database?

trang:
If by "copy" you mean to download the SQL database dump including the data, that is not something available.

If you would just like to have the database schema, I believe the only way at the moment would be to [install Tatoeba locally](https://github.com/Tatoeba/tatoeba2/blob/dev/README.tatovm.md), SSH into the TatoVM and run `mysqldump --no-data tatoeba > database.sql`.

Yorwba:
[...]

PR #2988, opened 2022-08-29
Issue 2984 - Incorrectly formatted CSV/TSV files

vinkaks:
Addresses #2984 .TSV files should not have newlines and tabs. Added SQL function which can be used on all text fields to escape any tabs and newlines before writing to export files. Tested function on `user_languages.csv` file.export. 

If approach acceptable, I can apply to all user generated text fields during export. 

If there is a better location to put the function where it is run only once I will move it.

Tested on TatoVM.


vinkaks:
Narrow minded approach. Closing in favour of warning users files may need extra escaping before being parsed.

PR #3005, opened 2022-10-27
Issue 3004 - Change input name for 'username' search to avoid signin auto-fill popup

### LINKED ISSUE #3004: Unexpected opening of SignIn ‘Suggest User/Password’ on iOS
This one has been ‘bugging’ me for years now.

Opening the ‘List of all members’ automatically invokes the SignIn Username/Password suggestion box on iOS.

https://user-images.githubusercontent.com/725445/197865350-675b1bbf-3929-4742-8ab9-a9cae27c350d.mov

vinkaks:
This PR is an attempt to fix #3004

**REQUIRES TESTING ON SAFARI BROWSER**
(I don't have a iOS device.)

Safari/iOS users face a disruptive sign-in suggestion pop-up when they land on or attempt to fill in the username search form on https://tatoeba.org/en/users/all. Same issue observed but not as disruptive on other browsers.

Attempt to solve by

- Avoiding common auth forms input names like `name`, `username` etc
- making the input name unique. `username` is also used on login and registration pages. This could be triggering autofill if password and username are saved and autofill active
- add `search` to input name to highlight it is a search field

### DIFF
--- a/src/Controller/UsersController.php
+++ b/src/Controller/UsersController.php
@@ -413,7 +413,8 @@ class UsersController extends AppController
      */
     public function search()
     {
-        $userId = $this->Users->getIdFromUsername($this->request->getData('username'));
+        $username = $this->request->getData('search_username');
+        $userId = $this->Users->getIdFromUsername($username);
 
         if ($userId != null) {
             return $this->redirect(array("action" => "show", $userId));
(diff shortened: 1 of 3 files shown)


Yorwba:
I don't have an iOS device to test this on, but the changes make sense to me and at least in Firefox the context menu stops suggesting to manage logins and allows setting a search keyword instead. Once this is deployed, mramosch can test whether this works in Safari as well.

Issue #2992, opened 2022-09-21
Latin script for Baluchi


trang:
We have been contacted by a couple of users requesting to support Latin script for Baluchi.

From what I see, there has been sentences added in Latin script recently:
- https://tatoeba.org/en/sentences/show/11122491
- https://tatoeba.org/en/sentences/show/11121847

We just have to ensure that the direction is set to 'auto' for sentences in Baluchi. When we initially added Baluchi (#2418), we set it as RTL only. 

For the record, this is what was sent to us at the team@tatoeba.org email:

(1)
> Tatorba is very good site for language translation works. I found   My mother language Baluchi here but in arabic script. But for coorect sounds and pronounciation we use latin script for Baluchi (Balóchi) and about equal numbers of Baluchi arabic script writers and readers using Baluchi latin script.

(2)
> The translations were in Arabic Script but the problem with Arabic script is that Balochi words are not pronunced correctly if the diacritical marks are not added. Because of this problem Balochi writers are using Latin script more and more nowadays. As Balochi is shifting from Arabic Script to Latin Script and will be in Latin Script in future, I suggest you add Balochi Latin in translation projects.

DJ-Saidez:
Once the Neapolitan addition is resolved, I'll taking care of this.

PR #3007, opened 2022-10-30
Reassigned Baluchi direction to "auto"

DJ-Saidez:
This PR solves #2992 
It removes Baluchi as a RTL language and adds it as an "auto" direction language.


Yorwba:
Thanks a lot!

PR #3009, opened 2022-11-01
Allow transcriptions edits

### LINKED ISSUE #3006: Can't edit transcription on sentences with audio
1. Go to https://tatoeba.org/en/sentences/show/11211585 which is a Japanese sentence with audio. The software-generated transcription is incorrect (おひゃくどせき should be おひゃくどいし), and I want to change it.

Screen Shot 2022-10-28 at 8 02 50 PM

2. Click on the pencil icon to edit the transcription.

Screen Shot 2022-10-28 at 8 03 12 PM

3. See error

Normally, when I click on the pencil icon, I'm able to edit or confirm the transcription (which doesn't affect the actual sentence text), even though I don't own the sentence. The fact that it has audio, logistically, shouldn't affect its ability to be modified to match the audio better.

I'm using Google Chrome on Mac 12.4.

vinkaks:
Attempt to fix  #3006

Minor Interface Fix. The backend seems to already have adequate validation to stop sentences with audio from being changed, but allowing transcriptions to be edited. 

The` edit sentence` form/action now has three states

1) Sentence has Audio and no transcriptions (`edit sentence` form disabled)
Screenshot from 2022-11-01 18-03-15

2) Sentence has no Audio  (edit sentence form displayed)
Screenshot from 2022-11-01 18-03-25

3) Sentence has Audio and transcriptions (language and sentence inputs disabled)
Screenshot from 2022-11-01 18-03-38

Issue 3006 - Allow editing of transcriptions for sentences with audio

### DIFF
--- a/src/Template/Element/sentences/sentence_and_translations.ctp
+++ b/src/Template/Element/sentences/sentence_and_translations.ctp
@@ -171,7 +171,7 @@ $sentenceUrl = $this->Url->build([
 
         echo $this->element('sentences/list_form');
 
-        echo $this->element('sentences/sentence_form');
+        echo $this->element('sentences/sentence_form',['hasAudio'=>isset($sentence->audios) && count($sentence->audios)]);
     }
     ?>
 
--- a/src/Template/Element/sentences/sentence_and_translations.ctp
+++ b/src/Template/Element/sentences/sentence_and_translations.ctp
@@ -171,7 +171,7 @@ $sentenceUrl = $this->Url->build([
 
         echo $this->element('sentences/list_form');
 
-        echo $this->element('sentences/sentence_form',['hasAudio'=>isset($sentence->audios) && count($sentence->audios)]);
+        echo $this->element('sentences/sentence_form');
     }
     ?>
 
--- a/src/Template/Element/sentences/sentence_form.ctp
+++ b/src/Template/Element/sentences/sentence_form.ctp
@@ -1,6 +1,6 @@
 <form layout="column" layout-margin ng-if="vm.visibility.sentence_form">
 
-<div ng-if="vm.sentence.audios && vm.sentence.audios.length > 0">
+<div ng-if="vm.sentence.audios && vm.sentence.audios.length > 0 && vm.sentence.transcriptions.length === 0">
     <p><?= __('You cannot edit this sentence because it has audio.'); ?>
 
     <div layout="row" layout-align="end center">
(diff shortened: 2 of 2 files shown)


src/Template/Element/sentences/sentence_form.ctp
```diff
@@ -10,13 +10,13 @@
...
 
-<div name="sentenceForm" layout="column" ng-if="!(vm.sentence.audios && vm.sentence.audios.length > 0)">
+<div name="sentenceForm" layout="column" ng-if="!(vm.sentence.audios && vm.sentence.audios.length > 0 && vm.sentence.transcriptions.length === 0)">
 
     <div layout="row" layout-align="start center" ng-if="vm.sentence.permissions.canEdit">
         <md-input-container flex="50">
             <?php /* @translators: language field label on sentence addition form */ ?>
             <label><?= __('Language') ?></label>
-            <md-select ng-model="vm.sentence.lang">
+            <md-select ng-model="vm.sentence.lang" ng-disabled="<?php echo $hasAudio ?>">
```
jiru, line 19: While your code works, it would be more consistent with the rest to use the AngularJS condition `vm.sentence.audios && vm.sentence.audios.length > 0` instead. Then, you would not have to pass on the `$hasAudio` variable.

jiru:
Thank you, @vinkaks!

Beside the code comment, I would like to point out a discrepancy between two scenarios:
1. As the author of the Japanese sentence having audio, I click edit and I am presented with two disabled fields and the enabled transcription field (your screenshot).
2. As an advanced contributor not being the author of the Japanese sentence having audio, I click edit and I am presented with the transcription field only.

In terms of UX, I think it would be better to show the 3 fields regardless of the scenario. We can merge the PR without that however.

Finally, it would be nice to port the code to the old design, but we can totally merge the PR without that.

vinkaks:
Thanks for reviewing the PR @jiru 

>Finally, it would be nice to port the code to the old design, but we can totally merge the PR without that.

In the old design, I think you can already edit transcriptions for sentences with audio, as long as you have the permissions.

jiru:
> In the old design, I think you can already edit transcriptions for sentences with audio, as long as you have the permissions.

Apparently that is true if you are the sentence owner, but advanced contributors cannot edit the transcription of a sentence having audio.

src/Template/Element/sentences/sentence_and_translations.ctp
```diff
@@ -171,7 +171,7 @@ $sentenceUrl = $this->Url->build([
 
         echo $this->element('sentences/list_form');
 
-        echo $this->element('sentences/sentence_form');
+        echo $this->element('sentences/sentence_form',['hasAudio'=>isset($sentence->audios) && count($sentence->audios)]);
```
jiru, line 174: I believe this change is no longer needed.

jiru:
Thanks again, @vinkaks!

PR #3014, opened 2022-11-08
Disable jump to top

### LINKED ISSUE #3011: Unwanted scroll when opening sidenav on sentence page
**Story**

From a mobile phone, when you go to the sentence page, scroll down a bit and click on "Metadata", the main content of the page scrolls up to the top, "which throws you out of context when you are working somewhere further below", as reported by mramosch in https://github.com/Tatoeba/tatoeba2/issues/3001#issuecomment-1290813154.

**Idea**

We should prevent scrolling when the sidenav opens.

Otherwise, we should figure out a way to place the "Metadata" button more towards the top of the page, so that the scrolling is barely noticeable (like for the "Narrow search" button on the search page).

vinkaks:
Attempt to fix #3011 

Steps to solve

- edit the controller so that it stops scrolling to top of `body`
- Edit `sidenav` so that it is fixed to viewport, not absolutely positioned relative to body

Tested on search `sentences/search` page and show `sentences sentences/show/:id` page

https://user-images.githubusercontent.com/21044267/200612482-ba0206fe-187e-47a8-bae6-c230edd33be0.mov

### DIFF
--- a/webroot/css/layouts/default.css
+++ b/webroot/css/layouts/default.css
@@ -580,11 +580,11 @@ html[dir="rtl"] .search-bar-extra a:before {
     }
 }
 
-.md-sidenav-right{
+.md-sidenav-right {
     position: fixed;
 }
 
-.md-sidenav-backdrop{
+.md-sidenav-backdrop {
     position: fixed;
 }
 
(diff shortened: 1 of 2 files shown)


jiru:
Thank you, @vinkaks! It works for me and the fix looks clean.

The only comment I have is about code style. In the CSS, you could add spaces before opening brackets so that the style matches with the rest of the file.

PR #3015, opened 2022-11-11
Add Santali as UI language

Prasanta-Hembram:
Hi, I have translated almost 50% of the Website in Santali language. Let's see how it looks in the dev website. :-)


jiru:
@Prasanta-Hembram Thank you! I have deployed Santali UI to dev. Sorry it took so long, and just to make it clear, you did not need to reach 50% before deploying to dev.

PR #3020, opened 2022-12-11
Reword "use new design" setting

### LINKED ISSUE #2972: Experimental options >> Display sentences with the new design has been set as the default for new accounts.
The Experimental option "Display sentences with the new design" has (accidentally?) been set as the default for new accounts.

This means that new members may not realize that they aren't getting the fully-functioning version, and may find it confusing when existing members try to help by explaining things about the UI.

I think it would be better to revert this to the way it was. That is, unless the displayed message is no longer true, in which case that option should be moved out of the experimental options.

"Display sentences with the new design. Note that some features are not yet implemented in this new design but are coming soon."

jiru:
Solves #2972.

I removed the part that says "Note that some features are not yet implemented in this new design..." because it is no longer true. Since we plan to remove the old design, this option will disappear at some point, so I left it in the "experimental" option, but reworded it so that "using the old design" is now considered as experimental.

image

### DIFF
--- a/src/Template/User/settings.ctp
+++ b/src/Template/User/settings.ctp
@@ -325,13 +325,13 @@ $this->set('title_for_layout', $this->Pages->formatTitle(__('Settings')));
                 <md-checkbox
                     ng-false-value="0"
                     ng-true-value="1"
-                    ng-model="useNewDesign"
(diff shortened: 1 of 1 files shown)


Yorwba:
Thanks! I'm looking forward to the day when we'll finally be able to get rid of the old design for real.

PR #3032, opened 2023-01-30
Add word count filter criterion to search

### LINKED ISSUE #1954: Possibility to filter sentences by word count
As suggested by MisterTrouser on the Wall:

> 
> Suggestion for a new filter: Word count.
> 
> Current situation:
> It's possible to sort by "fewest word count". However, that might give you the first 30 pages with sentences like "really?", "hurry!" or "silence!".
> 
> Suggestion:
> Add a new filter, where the word count can be set.
> 
> Reason / Good points:
> - Beginners could search short to mid-long sentences specifically through word count.
> 
> Regards

https://tatoeba.org/ita/wall/show_message/32611#message_32611

jiru:
Closes #1954. No reindexation required for this to work since the word count is already indexed by Manticore. Some large chunks of code changes in `tests/` are just code refactoring.

### DIFF
--- a/src/Form/SentencesSearchForm.php
+++ b/src/Form/SentencesSearchForm.php
@@ -320,7 +320,10 @@ class SentencesSearchForm extends Form
             $setter = "setData$keyCamel";
             $this->_data[$key] = $this->$setter($value);
 
-            if(empty($this->_data[$key]) && !empty($this->defaultCriteria[$key])) {
+            $strippedParam = ($this->_data[$key] === null
+                              || $this->_data[$key] === false
+                              || $this->_data[$key] === '');
+            if ($strippedParam && !empty($this->defaultCriteria[$key])) {
                 /* Using Router::url() to reconstruct a URL for the given data
                  * strips out empty parameters, which would lead to a non-empty
                  * default being applied instead of the empty non-default value.
--- a/src/Form/SentencesSearchForm.php
+++ b/src/Form/SentencesSearchForm.php
@@ -32,8 +32,8 @@ class SentencesSearchForm extends Form
         'unapproved' => 'no',
         'native' => '',
         'has_audio' => '',
-        'word_count' => '0',
-        'word_count_op' => 'ge',
+        'word_count_min' => '1',
+        'word_count_max' => '',
         'trans_to' => '',
         'trans_link' => '',
         'trans_user' => '',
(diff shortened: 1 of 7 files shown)


jiru:
@alanfgh or anyone else: feedback welcome on wording:

img
img

alanfmw:
The wording looks fine to me. But we could just as well start from 1 rather than 0, since we don't have any empty sentences (or shouldn't, anyway). I think that would raise fewer questions.

Thanks for asking, @jiru .

alanfgh:
The wording looks fine to me. But we could just as well start from 1 rather than 0, since we don't have any empty sentences (or shouldn't, anyway). I think that would raise fewer questions.

Thanks for asking, @jiru .

LBeaudoux:
@jiru thanks for adding this feature. 

Why not use a [range slider](https://material.angular.io/components/slider/overview#selecting-a-range)? This would be very helpful for those who prefer medium length sentences.

It would be also nice if you could set your favorite sentence length range on your settings page so you don't have to do it every time you search.

jiru:
@alanfgh Thanks for checking the wording. In theory, there are no 0-words sentences, but it actually sometimes happens as a result of some bug that needs to be fixed, so searching for 0-words sentences actually helps tracking such bugs. I agree it would be less confusing to default to 1 instead of 0.

@LBeaudoux About allowing both upper and lower limit, yes it would be a good improvement. About using the range slider, I am not sure of the benefits in terms of usability because the min/max numbers are not so visible. Besides, how to decide on the upper limit of the slider? By the way, Tatoeba is using [AngularJS Material](https://material.angularjs.org/), not Angular Material, so we cannot use what you linked.

About adding a new user setting, if we are to add a new setting for every little thing like that, it would become very complicated for both developers and users. I would rather try a different approach, such as improving the ranking algorithm (or adding a new one) that favor sentences that are not too long nor too short.

LBeaudoux:
> About allowing both upper and lower limit, yes it would be a good improvement.

What do you think of 2 custom inputs taking optional values and displaying "min." and "max." as placeholders?

Length:`min.`⊠ to `max.`⊠ word(s)

jiru:
@LBeaudoux Thanks for the suggestion, it totally makes sense.

I’ve updated this pull request, the form now looks like this:
image

ckjpn:
Note that the word "Length:" doesn't fit.

Screen Shot 2023-07-30 at 7 21 28

One idea that might solve this problem would  be to change "Length" to  "Length in words:" and eliminate the "word(s)" part.
This may give the same information, but properly word-wrap in the space provided.

jiru:
@ckjpn I updated the design on https://dev.tatoeba.org/, can you check if it still word-wraps the label on your end?

ckjpn:
It looks good on the several Macintosh browsers I tested it on.

ckjpn:
This gets unexpected results.

Length = 8, search Japanese

="何と読む"

https://dev.tatoeba.org/en/sentences/search?from=jpn&has_audio=&native=&orphans=no&query=%3D%22%E4%BD%95%E3%81%A8%E8%AA%AD%E3%82%80%22&sort=relevance&sort_reverse=&tags=&to=&trans_filter=limit&trans_has_audio=&trans_link=&trans_orphan=&trans_to=&trans_unapproved=&trans_user=&unapproved=no&user=&word_count_max=8&word_count_min=8

jiru:
This happens because `「` and `」` are treated as words. @ckjpn Can you open a new issue please?

Issue #3034, opened 2023-02-02
Sylheti (syl)


DJ-Saidez:
* Language name: Sylheti
* Language info: https://en.wikipedia.org/wiki/Sylheti_language
* ISO 639-3 code: syl
* Public List: https://tatoeba.org/en/sentences_lists/show/171139
* Icon: https://drive.google.com/file/d/1up85F9zLO4y9EMXZ-MBlGy6zcpV0OSgk/view
![output-onlinepngtools](https://user-images.githubusercontent.com/70669145/216437451-171de4bd-c08a-4f3e-989d-7e2b7d71e51b.png)
* Requested by: robingaspi

PR #3035, opened 2023-02-02
Added language Sylheti

DJ-Saidez:
This issue solves #3034 
Slightly concerned about the file size of the flag, I wasn't able to shrink it any more, but it appears to be functional.


jiru:
The flag is okay. Thank you! :+1:

Issue #2448, opened 2020-07-10
Some partial dates are not localized correctly


Yorwba:
**To Reproduce**
1. Set UI language to Japanese
2. Edit your profile.
3. Note that the dropdowns for year, month and day of the birthday have values 2017, 2016, etc., 1月, 2月, etc. and 1, 2, etc.
3. Set the year to some value, e.g. 2017, and leave the other values blank.
4. Save the changes.
5. Note that the year is displayed as 2017.

**Expected behavior**
It has been pointed out to me that the years should properly be formatted as 2017年 etc. and the days as 1日 etc. 

**Additional information**
For partial dates with more than one component, it's possible to fix this on Transifex by translating e.g. `{year}{month}` as `{year}年{month}`, but that doesn't work for the standalone values.

PR #3038, opened 2023-02-15
Months in date selector are not localized when editing profile

### LINKED ISSUE #2999: Months in date selector are not localized when editing profile
The months in the date selector are not localized when editing your own profile. They are always in English regardless of the set language.

**To Reproduce:**
1. Change the language on Tatoeba to something other than English.
2. Go to 'https://tatoeba.org/user/edit_profile'
3. Press the field used to select a month.
4. The names of the months appear in English rather than the selected language.

vinkaks:
Attempt to fix #2999. Switched input to a simple select. Tested in 

**English**
Screenshot from 2023-02-15 22-06-01

**& Japanese**
Screenshot from 2023-02-15 22-06-46

### DIFF
--- a/src/Template/User/edit_profile.ctp
+++ b/src/Template/User/edit_profile.ctp
@@ -137,10 +137,7 @@ $this->Languages->localizedAsort($countries);
         'minYear' => date('Y') - 100,
         'maxYear' => date('Y') - 3
     ]);
-    echo $this->Form->month('birthday', [
-        'empty' => true,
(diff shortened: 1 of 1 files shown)


Yorwba:
Thank you! Japanese might not be the best example to use here because it's still displayed incorrectly (years should have 年 and days 日 in addition to 月 for months, related to #2448) but it's still going to be an improvement for other languages, so I'll merge this.

Issue #2994, opened 2022-09-23
Strange formatting when displaying Aiji's LIst 170670


ckjpn:
https://tatoeba.org/en/sentences_lists/show/170670
This doesn't seem to be happening in other lists.

I wonder if we need to filter certain allowed characters in the titles.  Perhaps that's the cause.

<img width="1021" alt="Screen Shot 2022-09-23 at 15 44 26" src="https://user-images.githubusercontent.com/7934273/191906099-d8257cc7-9088-4985-bb5c-4b9e065805bf.png">

Yorwba:
This happens because the `edit-list-name` element [is displayed even when you only have permission to add sentences](https://github.com/Tatoeba/tatoeba2/blob/51060d800dcd36e5c45669dad077e302e3dba741/src/Template/SentencesLists/show.ctp#L156), but the contents are undefined if you don't actually have permission to edit the list.

So it's not related to special characters in the title (which would've been a security issue).

ckjpn:
This seems to be a problem with all collaborative lists.
I don't think this was the case before, so I think perhaps some code needs to be reverted.

https://tatoeba.org/en/sentences_lists/collaborative

Why does that character indicated by the "cause?" in the above image appear?

Yorwba:
This is now fixed.

PR #3039, opened 2023-02-26
Fix strange formatting in collaborative lists, when using old layout

vinkaks:
Fix attempting to solve #2994 
based on @Yorwba observation of the `edit-list-name` div being shown even if user cannot edit the list name. Tested locally.

**before**
image

**after**
Screenshot from 2023-02-26 20-01-15

### DIFF
--- a/src/Template/SentencesLists/show.ctp
+++ b/src/Template/SentencesLists/show.ctp
@@ -154,7 +154,9 @@ $this->set('title_for_layout', $this->Pages->formatTitle($listName));
     );
 
     if ($permissions['canAddSentences']) {
-        echo $this->Html->div('edit-list-name', $editImage);
+        if($permissions['canEdit']){
+            echo $this->Html->div('edit-list-name', $editImage);
+        }
         $this->Lists->displayAddSentenceForm($listId);
     }
     ?>


src/Template/SentencesLists/show.ctp
```diff
@@ -154,7 +154,9 @@ $this->set('title_for_layout', $this->Pages->formatTitle($listName));
     );
 
     if ($permissions['canAddSentences']) {
-        echo $this->Html->div('edit-list-name', $editImage);
+        if($permissions['canEdit']){
+            echo $this->Html->div('edit-list-name', $editImage);
+        }
```
Yorwba, line 159: I think all of this should probably happen outside the `canAddSentences` check. As is, setting "Who can add/remove sentences" to "No one (list inactive)" also disables changing the list name, which seems weird.

Also, maybe instead of checking the `canEdit` permission twice, only check whether `$editImage` is defined or not. Then we have less coupling and it's easier to see that we won't attempt to show an undefined image.

vinkaks: Seems when edit permissions are set to `no_one` the edit icon will be displayed but attempts at editing the name fail possibly due to the following permission checks
https://github.com/Tatoeba/tatoeba2/blob/03d2d6fc04074d076c33fa6865f89acf753e48e0/src/Model/Table/SentencesListsTable.php#L631-L640
https://github.com/Tatoeba/tatoeba2/blob/03d2d6fc04074d076c33fa6865f89acf753e48e0/src/Model/Entity/SentencesList.php#L30-L34

This made me believe `no_one` should be able to edit the name after a list in "No one (list inactive)". 

I am not 100% sure how the `isEditableBy` check is utilised throughout the rest of the project, so my solution had simply nested the permissions. I was trying not to touch the  controllers, entities or modelTables over what I think is a cosmetic QOL issue.

Yorwba: Okay, then I'll just merge this as-is. Thanks!

PR #3051, opened 2023-04-16
make tags lowercase before comparing them

### LINKED ISSUE #3050: Clearer message with redirection of tags in the advanced search
**Story** or **Context**

I wanted to translate some English sentences from the Tanaka Corpus into Italian using this query: https://tatoeba.org/en/sentences/search?from=eng&has_audio=&native=&orphans=no&query=&sort_reverse=&tags=tanaka+corpus&to=&trans_filter=exclude&trans_has_audio=&trans_link=&trans_orphan=&trans_to=ita&trans_unapproved=&trans_user=&unapproved=no&user=&sort=words and I put 'tanaka corpus' (without quotes and everything in lower case) as the tag contained in the sentences I wanted to translate.
After the query started, I got this error message:
 
> Warning: the following criteria have been ignored:
> 
>     “tagged as tanaka corpus”, because it's an invalid tag name

However, I still received the proper sentences since they contain the proper tag, that is 'Tanaka Corpus', so here the proper capital letters of the tag are included and the advanced search recognized my initial input of the tag

**Idea**

The error message should be reworked, since the original query is still done, and the tag isn't actually ignored.

vinkaks:
Attempt to fix #3050. Issue results in a false error.

Issue is due to the `applied tags` vs `user submitted tags` check being case sensitive even though the database search used to generate `applied tags` is case insentitive.

### DIFF
--- a/src/Form/SentencesSearchForm.php
+++ b/src/Form/SentencesSearchForm.php
@@ -187,8 +187,8 @@ class SentencesSearchForm extends Form
             $tags = implode(',', $appliedTags);
 
             $ignoredTags = array_diff(
-                    array_reduce($tagsArray, function($carry,$item){$carry[] = strtolower($item); return $carry; }), 
-                    array_reduce($appliedTags, function($carry,$item){$carry[] = strtolower($item); return $carry; }), 
+                    array_reduce($tagsArray, function($carry,$item){$carry[] = strtolower($item); return $carry; },[]), 
+                    array_reduce($appliedTags, function($carry,$item){$carry[] = strtolower($item); return $carry; },[]), 
                 );
             foreach ($ignoredTags as $tagName) {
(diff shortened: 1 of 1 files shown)


jiru:
Thanks for your PR! It appears the default collation makes varchar type case-insensitve by default, whereas the code assumes it is case-sensitve. Even though I didn’t know tags are case insensitive, we may want to keep it like this.

`\App\Model\Search::filterByTags()` is supposed to return the list of tags that have been successfully applied among the provided tags. Because that method is doing the SQL query, I think it would make more sense to have it handle the case thing, instead of its caller `App\Form\SentencesSearchForm::setDataTags()`. The caller shouldn’t have to deal with such implementation details. Can you try modifying `filterByTags()` in a way it works regardless of the case?

Issue #1069, opened 2016-03-17
It is no more possible to search favorites


trang:
The favorites used to be all displayed on one page. This allowed users to search for sentences containing a certain word in their favorites with their browser's search (ctrl+F). Now there is pagination (#970) and searching a sentence in the favorites becomes impossible for users who have a lot of favorites.

Problem reported by odexed on the [Wall](http://tatoeba.org/eng/wall/show_message/25900#message_25900):

> I've noticed that my favorite sentences now appear on several pages, but before they were just on one page. Why was it changed?

>Just to make things clear, now I have to go through 141 pages in order to look my favorite sentences over. What if I need to find some of them quickly? I used to press Ctrl+F and put a clue word and now I have no idea what to do with it. The search engine can find hundreds of sentences with that word but it's not what I need. I need just the sentences that I harvested because they help me to memorize new words the best way.

PR #3074, opened 2023-07-31
Allow searching through favorites

jiru:
This PR solves #1069 in a simple way.


jiru:
This PR uses SQL LIKE operator on sentence text, which is not supposed to perform very good, so I did some measurements on dev.tatoeba.org on a user having 450k+ favorites. It turns out the full scan only takes 0.8 to 0.9 seconds, so I think this is acceptable for a worst-case scenario. Among all members having favorites on Tatoeba now, 99.26% of them have less than 1000 favorites anyway.

Issue #3085, opened 2023-10-07
Gun (guw)


DJ-Saidez:
* Language name: Gun
* Language info: https://en.wikipedia.org/wiki/Gun_language 
* ISO 639-3 code: guw
* Public List: https://tatoeba.org/en/sentences_lists/show/171793
* Icon: default icon requested
* Requested by: https://tatoeba.org/en/user/profile/Jugunu

jiru:
Gun is now supported on tatoeba.org.

PR #3086, opened 2023-10-07
Add language Gun (guw)

DJ-Saidez:
This issue solves #3085


DJ-Saidez:
Requester suggested a better icon: Benin flag with ISO code on the side. I’ll get to it later, but the language itself can be added and the flag updated later on.

jiru:
Thanks! Feel free to open a new pull request when the new flag is ready.

Issue #2352, opened 2020-05-28
wording: "generated by a piece of software"


alanfgh:
On the dev site, when I look at sentence number 16491, I see two Japanese translations. The first has another version under it with a red triangle icon next to it. When I hover over that, I see the text "This alternative script may contain errors because it has been generated by a piece of software that is not perfect."

![automatic_not_perfect](https://user-images.githubusercontent.com/1213027/83200186-c8490a00-a110-11ea-982f-ac94834553ff.png)

When I hover over a second one, I see the text "This alternative script was generated by a piece of software."

![piece_of_software](https://user-images.githubusercontent.com/1213027/83200223-deef6100-a110-11ea-8065-b5ee8968a0e6.png)

There are a couple of issues with this.

(1) The first message feels a bit alarmist.

(2) The wording seems a little awkward.

(3) It's not clear to me what the difference is between the two sentences. Is it that one was not reviewed, while the other was?

I prefer this text:

"This transcription was generated automatically."

If you want to stress that the transcription has not been reviewed, you can say:

"This transcription was generated automatically and has not been reviewed."

trang:
Thanks for the wording suggestions, @alanfgh.

To summarize, we have the following cases:

1) The transcription was automatically generated and we know that the software is 100% accurate. It therefore doesn't need to be reviewed.
→ The tooltip would be `This transcription was generated automatically.` or `This alternative script was generated automatically.`

2) The transcription was automatically generated and we know that the software is not always accurate. It needs to be reviewed.
→ The tooltip would be `This transcription was generated automatically and has not been reviewed.` or `This alternative script was generated automatically and has not been reviewed.`

3) The transcription has been reviewed.
→ The tooltip would indicate who reviewed it and when, as it does now.

alanfgh:
That makes sense.

I think of "script" as referring to a system of writing and "transcription" as referring to (a) a process or (b) a single piece of text. (Case b is the relevant one here.) Therefore, I prefer "transcription" to "alternative script".

trang:
We actually have different wordings for alternative scripts and transcriptions.

For instance, pinyin will be referred as to "transcription". But simplified/traditional Chinese will be referred to as "alternative script".

The various tooltip texts are defined here:
https://github.com/Tatoeba/tatoeba2/blob/dev/src/Model/Entity/Transcription.php#L193

Unless we decide to merge the tooltips and refer to everything as "transcription", we would need to adapt the strings for both cases.

[...]

PR #3104, opened 2024-03-06
Clearer transcription tooltips

profesyonal:
#2352


profesyonal:
I did test this BTW, but I was having problems with merging a commit without including some other irrelevant changes I'd made, so I did the edit on Github.

Yorwba:
Thanks again! While testing, I noticed that the changes didn't affect the old design (which is supposed to go away at some point, but can still be selected by users in their settings if they prefer it). That's because some of this code is duplicated in two functions in [`src/View/Helper/TranscriptionsHelper.php`](https://github.com/profesyonal/tatoeba2/blob/3779388739d94307400aad338cc7e669a5035808/src/View/Helper/TranscriptionsHelper.php#L106).

You could replace the same strings in `TranscriptionsHelper.php` as well (easiest), or rewrite it to eliminate the duplication by using the transcription's `info_message` property, which you've already updated. (This second solution might be a bit more finicky.)

profesyonal:
Thank you!

I went with the second solution, but now I'm having trouble putting it on Github. Getting this:

`[username@MACHINE tatoeba2]$ sudo git push origin patch-3
error: src refspec patch-3 does not match any
error: failed to push some refs to 'https://github.com/profesyonal/tatoeba2.git'`

I couldn't find any helpful solutions to this error by googling.

Yorwba:
First of all, you shouldn't have to use `sudo` for git. If you're getting a "permission denied" error about some local file in `.git` when doing some operations (e.g. `error: insufficient permission for adding an object to repository database .git/objects`), you may have to fix ownership of git's internal files by running `sudo chown -R $USER: .git` so that they're no longer owned by the root user.

Secondly, the `git push origin patch-3` command means: "Take the commit at `patch-3` in my local repository and push it to the `patch-3` branch of the remote `origin` repository." The error message `error: src refspec patch-3 does not match any` means that at the _source_ (your local repository) there's nothing under the name `patch-3` (`refspec` basically means "a way to refer to commits", this includes branch names, commit hashes, stuff like `HEAD` etc.)

Most likely you do not have a `patch-3` branch in your local repository. If you have named your local branch differently (or you have not created a branch yet, which you can do with `git checkout -b transcription-tooltips-3104` for e.g.) you can push it with `git push --set-upstream origin $YOUR_LOCAL_BRANCH:patch-3`, which also remembers the mapping so that you can simply use `git push` afterwards.

profesyonal:
I pulled patch-3 onto my machine, re-did the edit, did git add -p, did git push, did git commit, and the changes never showed up.

I saw something about merging a commit on the interface, so it said to do that... and I ended up pulling all of Tatoeba dev into this pull request.

I don't know why things have to be so difficult.

profesyonal:
I'm probably going to need to undo this pull request somehow and start a new one.

trang:
> I pulled patch-3 onto my machine, re-did the edit, did git add -p, did git push, did git commit, and the changes never showed up.

You need to do the commit before the push (add → commit → push).
* git add = saves your changes into a "draft"
* git commit = creates a commit from the draft
* git push = sends the commit(s) to GitHub

If you did `git add` then `git push` right after, then there was no commit, so nothing was sent to GitHub.

> I saw something about merging a commit on the interface, so I tried that, thinking it was mine... and I ended up pulling all of Tatoeba dev into this pull request.

You don't have to worry about this. It didn't affect the content of the pull request. Note that whenever you push new commits, you can see what are the effective changes by looking at the "Files changed" tab of the pull request.

image

profesyonal:
Thanks! Let's see how this does.

Yorwba:
Thank you for your changes @profesyonal !

I don't know whether you noticed, but our automated tests [failed](https://jenkins.tatoeba.org/blue/organizations/jenkins/tatoeba2/detail/PR-3104/3/pipeline) because some test was generating notices about `Trying to get property 'info_message' of non-object`.

This notice is output when you attempt to use a PHP array as if it was an object, i.e. `$transcr->info_message` instead of `$transcr['info_message']`. The test producing these notices is `phpunit tests/TestCase/Controller/TranscriptionsControllerTest.php --filter testResetDoesResetTranscription`.

You can also see them when doing a reset in the old design:
Transcription reset

The underlying reason is that when the transcription is reset, it is sent back as a freshly-generated array instead of being fetched from the database, so you cannot use object property syntax with `->`.

While investigating this, I discovered that there are also other places in the codebase where this edge case is not handled properly.

Since the necessary change to your code is fairly small, I don't want to ask you to make it and then wait another few weeks until I have time to look at it again, so I'll merge this PR as-is and then apply the fixes for this edge case separately.

Yorwba:
This change is now live :tada:

PR #3116, opened 2024-04-04
Add additional step to audio import docs

cblanken:
As a new contributor, it wasn't immediately obvious to me that the audio import job wasn't configured to run automatically after being triggered from `http://localhost:8080/en/audio/import`, so I added a reminder to the audio import README to run the queued job manually.


jiru:
Thank you! 😄

Issue #2551, opened 2020-09-10
When there is no identified voice for an audio file, the display on the page looks like a bug.


ckjpn:
Perhaps instead of "Recorded by:", have "Recorded by: Unknown" or "uncertain" or "anonymouse" or "unidentified."

Maybe that would require creating a user with the username "Unknown."

<img width="469" alt="Screen Shot 2020-09-10 at 11 38 23" src="https://user-images.githubusercontent.com/7934273/92675638-d981f380-f35a-11ea-8b69-1fe07a3ce220.png">

jiru:
"Unknown author" is now displayed, see this [example](https://tatoeba.org/sentences/show/377978).

ckjpn:
Perhaps it might be a good idea to give the "unknown author" a number, so that audio files by the same person can be grouped together.  Not only could people see this on the page, but the data would also be reflected in the exported file ([sentences_with_audio.tar.bz2](https://downloads.tatoeba.org/exports/sentences_with_audio.tar.bz2)

In the meantime, I'll make those lists "public" so they show on the page with the audio file.

PR #3120, opened 2024-04-11
Add "unknown" audio contributor text when author is null

cblanken:
This PR solves issue #2551 
On the `sentences/show/<id>` page conditionally render "unknown" for the audio contributor field when an `audio` has no associated user.

## Before
image

## After
image

### DIFF
--- a/src/Template/Element/sentences/audio.ctp
+++ b/src/Template/Element/sentences/audio.ctp
@@ -94,7 +94,7 @@ $this->AngularTemplate->addTemplate(
                 ]) ?>
             </span>
             <span ng-if="audio.author == null" class="audio-author">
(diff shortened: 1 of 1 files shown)


jiru:
Great! I think it would look even better using an alternative text such as "unknown author" instead of "by unknown". "by unknown" feels like robot talk.

cblanken:
> Great! I think it would look even better using an alternative text such as "unknown author" instead of "by unknown". "by unknown" feels like robot talk.

@jiru do you mean I should remove the attribution text "by" as well like this?
image

or add "author" to the existing text like this?
image

I think it may also make sense to make the "unknown user" text red in contrast to the green text used for an existing user. Like this
image


Thoughts?

jiru:
Thank you :+1:

Issue #2944, opened 2022-05-30
Audio contributed by ___  - Perhaps needs to be updated


ckjpn:
https://tatoeba.org/en/audio/of/CVAF

Perhaps this is not a bug. However, at first glace only 5 audio buttons are shown, but the title says "(total 8)"

Clicking the audio play button does properly cycle through only CVAF's audio files and doesn't play CK's audio.
Maybe inserting a number near the play button "(4)" might clearly indicate to the user that there are 4 audio files and not just one for a given sentence by a given username.

<img width="764" alt="Screen Shot 2022-05-30 at 11 40 51" src="https://user-images.githubusercontent.com/7934273/170907654-63e568d4-d8dd-4fbe-adef-c8dfd73a284e.png">

PR #3121, opened 2024-04-12
Add audio source count per sentence for user audio contributions page

cblanken:
This PR solves issue #2944

So far I've only tested on a couple audio sources as shown in the screenshot, but this solution should work for any number of audio sources. I'll do some more testing when I get a script working to import audio files automatically.
image

### DIFF
--- a/src/View/Helper/SentenceButtonsHelper.php
+++ b/src/View/Helper/SentenceButtonsHelper.php
@@ -180,6 +180,8 @@ class SentenceButtonsHelper extends AppHelper
         $total = count($sentenceAudios);
         if ($total) {
             $startIn = rand(0, $total-1);
+            $audioCount = 1;
+            $audioLinks = "";
             foreach ($sentenceAudios as $audio) {
                 $author = $this->_View->safeForAngular($audio->author);
                 if (empty($author)) {
@@ -193,8 +195,9 @@ class SentenceButtonsHelper extends AppHelper
                 $class = 'audioButton audioAvailable';
                 if ($startIn == 0) {
                     $class .= ' nextAudioToPlay';
+                    $audioCount = $this->Html->div("audioButtonCount", $total);
                 }
-                echo $this->Html->Link(
+                $audioLink = $this->Html->Link(
                     null,
                     $this->Url->build([
                         'controller' => 'audio',
@@ -207,8 +210,12 @@ class SentenceButtonsHelper extends AppHelper
                         'onclick' => 'return false',
                     )
                 );
+
+                $audioLinks .= $audioLink;
                 $startIn--;
             }
+
+            echo $this->Html->div("audioButtonWrapper", $audioLinks . $audioCount);
         } else {
             echo $this->Html->Link(
                 null,
--- a/src/View/Helper/SentencesHelper.php
+++ b/src/View/Helper/SentencesHelper.php
@@ -531,18 +531,18 @@ class SentencesHelper extends AppHelper
             );
         }
 
-        // Copy
-	echo '<div class="copy column">';
-	$this->SentenceButtons->displayCopyButton($sentence->text);
-	echo '</div>';
-
         // Sentence
         $hasAudio = isset($sentence->audios) && count($sentence->audios);
         $canEdit = $isEditable && !$hasAudio;
         $this->displaySentenceContent($sentence, $canEdit);
         echo '</div>';
 
-        // audio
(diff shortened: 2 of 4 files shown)


jiru:
Thanks! :slightly_smiling_face: I have a few comments.

From my understanding, the code you changed also affects the display of sentences everywhere on the website, including search results etc. I wonder if this is a necessary or wanted change.

Since the majority of sentences with audio only have one audio, I suggest not displaying anything in that case to reduce cognitive load.

cblanken:
> From my understanding, the code you changed also affects the display of sentences everywhere on the website, including search results etc. I wonder if this is a necessary or wanted change.

I can't quite pin down why, but there does appear to be a different component for the regular sentence search. The changes I've made only seem to affect the search by audio page (`/audio/index`) and the individual contributor audio page (`/audio/of/<contributor>`).

In the long run, I think it would be nice to have a complete dropdown menu to select available audios and replay them at will instead of cycling through them one at a time especially if any features to mark audios as particular dialects/accents are ever implemented. A number indicator telling how many different audios exist for a sentence is a good stopgap though. 

## `/audio/index`
image

## `/audio/of/contributor`
image

## Main Sentence Search
image

jiru:
Thank you for making the change I suggested.

> I can't quite pin down why, but there does appear to be a different component for the regular sentence search. The changes I've made only seem to affect the search by audio page (/audio/index) and the individual contributor audio page (/audio/of/<contributor>).

Sorry I wasn’t very clear. Sentences can be displayed using two different designs, because we "in the process" of migrating from the "old design" to the "new design". There is a setting in user preferences that allows you to switch between both, it says:
```
Display sentences with the old design. Note that the old design will be removed at some point.
```
You can see the difference between both designs on the search results page for example. Using the old design, your change does affect search results (and pretty much all the other pages where sentences are displayed).

As for the `/audio/index` and `/audio/of/<contributor>` pages, they both happen to use the old design no matter what the setting says, because they have not been migrated yet.

As for the impact of your change, the only concern I have is that your change reduces the available space for sentence text, especially in search results where the space is already reduced compared to other places, because of the right pane. Looking back, at some point we added the audio icon, then the copy-paste button, which makes the whole thing more and more crowded, I think we want to avoid adding yet another item.

I think that while it totally makes sense to make the "number of audios" prominently visible on /audio pages, maybe not so much in other pages where audio isn’t the main focus. Maybe the display can be adjusted with a CSS specific to /audio pages.

cblanken:
Ah, I see now. Thank you. I was really confused what was going on :sweat_smile: 

> As for the impact of your change, the only concern I have is that your change reduces the available space for sentence text, especially in search results where the space is already reduced compared to other places, because of the right pane. Looking back, at some point we added the audio icon, then the copy-paste button, which makes the whole thing more and more crowded, I think we want to avoid adding yet another item.

If space is the main concern, I think the audio count could be minimized a bit, so it only takes up about 1 character's width at most. Like this:

## `/audio/index`
image

## `/sentences/search`
image

This solution just adds an extra 3/4 character width in margin on the right to make space for up to 2 digits in the audio count (the audio count font size is 60% of the main sentence font size here).

> I think that while it totally makes sense to make the "number of audios" prominently visible on /audio pages, maybe not so much in other pages where audio isn’t the main focus. Maybe the display can be adjusted with a CSS specific to /audio pages.

@jiru I think it would be better for the audio playback interface to be consistent between pages whether that means leaving out the audio count entirely or adding it, but I can look into applying it specifically to the `/audio` pages if you think that's the way to go?

jiru:
> If space is the main concern

My concern is also about cognitive load. More things displayed means it takes longer for the eyes and brain to process the information.

> @jiru I think it would be better for the audio playback interface to be consistent between pages whether that means leaving out the audio count entirely or adding it, but I can look into applying it specifically to the /audio pages if you really think that's the way to go?

I understand the point of having a consistent interface, but striking for consistency is also what made us to build the current sentences+translations block, which is somewhat crowded. This block has all the functionalities, you can do anything everywhere, but in terms of design I think it is rather disconnected from the context of the page it is displayed in. Sorry for digressing. :laughing: 

As showed in issue #2944 there are good reasons to add the number of audio in audio pages. Let’s add it to the audio pages only for now, and if we figure out there is a need somewhere else, or everywhere, we can later add it there too.

jiru:
Also it seems your changes break the display of sentences with transcriptions:
image

PR #3124, opened 2024-04-20
Remove footer link to Google group

cblanken:
The Google group link was removed from the README a while ago in this commit: b67af4e, so it probably makes sense to remove it from the site footer as well.


trang:
Indeed, the link to the Google group no longer makes sense. Thanks for this! 😄 

In fact you reminded me that we should also remove the link to the Facebook group at some point... It's another communication channel that we no longer really use.

Issue #2159, opened 2020-02-22
Only show original sentences in the advanced search


RyckRichards:
[Reported by brauchinet on the Wall](https://tatoeba.org/eng/wall/show_message/34246#!#message_34246)

> Now there is an option "only show original sentences" on the user's sentence page. That's great!

> Wouldn't it be nice if this option were also available in the advanced search? So one could filter out original sentences in a specific language.

ckjpn:
Making this a 3-choice item like the current "audio" setting it would be better.

Is original: 
any - yes - no

Issue #1643, opened 2018-08-11
Review info text about original/derivative sentences


trang:
This information currently is displayed in the "Logs" section of the sentence page, which is maybe not the best place.  As mentioned in https://github.com/Tatoeba/tatoeba2/pull/1623#issuecomment-408595821: 

> The other options I can think of:
>
> 1. Somewhere still in the sidebar but in a separate block. We already have many blocks in the sidebar so I'm not too fond of adding yet another one.
> 2. Next to the sentence id (above the sentence block). We would have to find shorter wordings, or use icons with tooltips.
> 3. Below the sentence id and above the sentence block. Maybe not an optimal use of vertical space. The information is not *that* important to be above the main subject of the page.
> 4. Below the sentence block. I'm afraid some people might think that the information refers to the last translation in the sentence block. Maybe not a big deal.
> 5. In the sentence block itself. It's a quite difficult one, but there's surely a way to fit the information in there. We could then mark the "original" sentence even when we're not on the page of that sentence.

The info text we currently display (there are 3 cases):
- `This sentence is original and was not derived from translation.`
- `This sentence was initially added as a translation of sentence #{n}.`
- `We cannot determine yet whether this sentence was initially derived from translation or not.`

These strings need to be reviewed too.

alanfgh:
My initial suggestions:

- This sentence was not added as a translation.
- This sentence was added as a translation of sentence #{n}.  [that is, use the text you already suggested]
- This sentence may have been added as a translation.

Is there a reason why we need to display the third sentence? Would blank text be an option? Also, why do you suggest "yet"? Is it that we need to perform further analysis before we can make the determination? If so, is it that (a) we don't know how to do the analysis yet but expect to be able to do it in the future, or (b) we don't have the data yet but expect to have it in the future, or (c) we haven't had enough time to run through all the sentences yet or (d) some other possibility?

jiru:
> Is there a reason why we need to display the third sentence? Would blank text be an option?

No particular reason. Yes, I think we could as well have no text at all, considering this information may not be very relevant for most visitors.

> Also, why do you suggest "yet"?

The "yet" means that we need to perform further analysis before we can make 
the determination (reason a). However, while this is true for most of the sentences, there are a few sentences (about 14000) which lack a creation log, so I’m afraid we will actually never be able to find out their origin.

PR #3136, opened 2024-08-13
Search by original or translation

jiru:
Solves #2159.
search1
search2


jiru:
Related: #1643.

LBeaudoux:
It seems a bit clunky to me. Wouldn't a "Not derived from another sentence (original)"l checkbox at the bottom of the form be more appropriate?

alanfgh:
Rather than "Not as translation (original)", I would suggest one of these:

- Not as a translation
- Original

I agree that using both of them together sounds clunky.

While "Original" is concise and natural, I'm guessing that the objection to it is that people can add sentences that they take from elsewhere as long as they are in the public domain, so they're not original in the sense of having been authored by the owner. In the context of a drop-down list where one of the other alternatives is "Not as a translation", this interpretation is less likely, but still possible, I suppose. So my vote is for "Not as a translation" (note that I do prefer the presence of "a" before "translation"). However, I could also live with "Original".

_> Wouldn't it be more appropriate to have a checkbox at the bottom of the form that says "Not derived from another sentence (original)"?_
My guess is that this is offered as an alternative to the drop-down list, and assumes that people are not going to want to search explicitly for "Unknown" or "As a translation of another sentence". However, I think it's better to have this added flexibility, given that it's on an advanced search dialog that already offers this level of granularity for other choices.

LBeaudoux:
> My guess is that this is offered as an alternative to the drop-down list, and assumes that people are not going to want to search explicitly for "Unknown" or "As a translation of another sentence". 

To be honest, I can't think of a use case in which a user would specifically search for derived sentences...  Also, I find the label `not a translation` misleading because an original sentence is the back-translation of a sentence that is derived from it.
How about a simple checkbox labelled `in the original language`, as for movies?

jiru:
Thank you for your feedback. You both have solid points and it’s difficult to make a decision. One could argue that the other, existing filter "Owned by a self-identified native" does not offer such level of granularity (there is no "not owned by a self-identified native" filter, even though I can see a use case). I generally think that we should avoid unnecessary complex UIs, and the advanced search is already quite complex. My original aim was to only show original sentences, so I am going to stick with a simple checkbox until there is need for other filters.

jiru:
Alright this is what I came up with:
* simpler "Is original" labelled checkbox
* infobox with text appearing on mouseover/tap: Original sentences are the ones initially added as examples in that language only, as opposed to the ones initially added as translations of another sentence. Such sentences are likely to feature more traits specific to that language.

Feedback welcome!

shot

LBeaudoux:
@jiru thank you for these changes.

Here's what I think could be improved:
 - Move the new checkbox below the "added by a self-identified native" checkbox, which has a similar purpose, appearance and priority.
 - Shorten the text of the infobox. I suggest  _The sentences in the original language are not added by translating other sentences. They often sound more natural_.

LBeaudoux:
> Personally, I will rarely use these two checkboxes, but I filter by list very often.

What if we introduced even more changes? I suggest:
- move the two checkboxes just below `Length`.   
- move `Tags` and `List` below `Owner`

jiru:
Well I never filter by list. I think I just opened a can of worms :smile:

We might as well try to improve fields order in another ticket.

jiru:
@alanfgh I simplified the search option design, could you review the English used in the [infobox text](#issuecomment-2290842738) please?

alanfgh:
Are you talking about this text?

> Original sentences are the ones initially added as examples in that language only, as opposed to the ones initially added as translations of another sentence. Such sentences are likely to feature more traits specific to that language.

I agree with @LBeaudoux that it should be shorter. In addition, I don't see why it's necessary to say anything about the relative worth of original sentences vs. non-original. I don't even believe that original sentences are better; they could just as easily be added by someone who's a poor non-native speaker or a sloppy native speaker. In any case, the whole reason for supplying this option is that we think people want it. We don't need to try to convince people to search for these sentences.

With that in mind, I suggest simply:

_Original sentences are ones that were not added as translations of other sentences._

jiru:
Thank you @alanfgh for your insightful comment, I totally agree. I updated the infobox text.

jiru:
Note: we will need to do a full reindexation of manticore indexes during deployment, like this:

```sh
# 5.
cd /var
./create-new-prod-folder.sh --no-dry-run

# full reindexation
screen -S reindex
crontab -e # disable reindexation tasks refresh_delta.sh and merge_indexes.sh
SHA=...
sed -i "s/www-prod/www-prod-$SHA/" /etc/manticoresearch/manticore.conf
www-prod-$SHA/bin/cake sphinx_indexes update main && www-prod-$SHA/bin/cake cake sphinx_indexes update delta -f
sed -i "s/www-prod-$SHA/www-prod/" /etc/manticoresearch/manticore.conf

# 6.
./set-prod-folder.sh {SHA}

# restore automatic reindexation
crontab -e # enable reindexation tasks refresh_delta.sh and merge_indexes.sh

# 7.
systemctl reload php7.3-fpm
```

PR #3152, opened 2025-01-12
Update the Silesian flag to use the Upper Silesian flag associated with the language.

vxern:
The flag used for Silesia is the one that was used for [the general Silesian region and/or Lower Silesia](https://en.wikipedia.org/wiki/Flag_of_Silesia_and_Lower_Silesia). The modern Silesian language is spoken in [Upper Silesia](https://en.wikipedia.org/wiki/Upper_Silesia), and the Upper Silesian flag is used pretty much predominantly used to represent Silesian these days. I'm correcting my mistake.

The updated flag also fixes sizing. The previous flag had dimensions 600x400, which breaks certain pages:
image

### DIFF
--- a/src/Lib/LanguagesLib.php
+++ b/src/Lib/LanguagesLib.php
@@ -683,7 +683,7 @@ class LanguagesLib
                 'bom' => __d('languages', 'Berom'),
                 'sat' => __d('languages', 'Santali'),
                 'mik' => __d('languages', 'Hitchiti'),
-                'szl' => __d('languages', 'Silesian'),
+                'szl' => __d('languages', 'Silesian (Ślabikŏrz)'),
                 'igs' => __d('languages', 'Interglossa'),
                 'knc' => __d('languages', 'Central Kanuri'),
                 'nnb' => __d('languages', 'Nande'),
--- a/src/Lib/LanguagesLib.php
+++ b/src/Lib/LanguagesLib.php
@@ -683,7 +683,7 @@ class LanguagesLib
                 'bom' => __d('languages', 'Berom'),
                 'sat' => __d('languages', 'Santali'),
                 'mik' => __d('languages', 'Hitchiti'),
-                'szl' => __d('languages', 'Silesian (Ślabikŏrz)'),
+                'szl' => __d('languages', 'Silesian'),
                 'igs' => __d('languages', 'Interglossa'),
                 'knc' => __d('languages', 'Central Kanuri'),
                 'nnb' => __d('languages', 'Nande'),
--- a/webroot/img/flags/szl.svg
+++ b/webroot/img/flags/szl.svg
@@ -1,7 +1,5 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<svg xmlns:svg="http://www.w3.org/2000/svg" xmlns="http://www.w3.org/2000/svg" version="1.0" width="600" height="400" viewBox="0 0 600 400" xml:space="preserve"><defs id="defs8"/>
-<text transform="translate(21.6978,28.5215)" style="font-size:12px;fill:#1a161b;font-family:Myriad-Roman" id="text3">Flagge erstellt von: David S. Liuzzo 2006</text>
-
-<rect width="600" height="200" x="0" y="200" style="fill:#ffce08" id="Unten"/>
-<rect width="600" height="200" x="0" y="0" style="fill:#ffffff" id="Oben"/>
+<?xml version="1.0"?>
+<svg xmlns="http://www.w3.org/2000/svg" width="30" height="20">
+  <rect width="30" height="20" fill="#007cc2"/>
+  <rect width="30" height="10" fill="#f8ce00"/>
 </svg>


jiru:
Thank you for this pull request, @vxern :slightly_smiling_face: 

The technical adjustments to the SVG file are welcomed.

As for replacing the "Silesia and Lower Silesia" flag with the "Upper Silesia only" flag, I cannot tell if it’s a good move or not. Could you provide some more hindsight and references on the matter?

As for adding the orthographic variant to the language name:
* I am afraid of the potential clutter because the language name is used in many places in the UI.
* Is Silesian orthography on Tatoeba confusing to a point that we need to display the chosen orthography all over the place? In my opinion, this information could be either guessed by looking at existing sentences, or explicitly told as part of welcome messages or interaction with other contributors.
* You said Ślabikŏrz is the de facto standard. To me, it means pretty much everyone uses it by default. If so, then why do we need to add it to the language name?

vxern:
Regarding the orthography - I suppose you're right, since it's the de facto, it wouldn't need to be specified that it's Ślabikorz that we're using. Just wanted to prevent potential misunderstandings in the future. Someone could submit sentences in another orthography and then wonder why they're not fine to do so, given it's not written that it's the orthography used.

As for the flag - `szl` is the language code specifically for Upper Silesian. Thus, we ought to use the Upper Silesian flag to represent it, rather than any other flag. For reference, here are some places where Upper Silesian is represented by the blue-yellow colours: https://youtu.be/IUJLLtMrmCg?si=jwtyHRQUYsh3o3ex https://youtu.be/Nhdlz9MeTr4?si=wKbRYpwuddIW5NtZ https://tvpworld.com/77207969/polish-parliament-recognizes-silesian-as-official-regional-language

jiru:
About the flag change, thank you for clarifying, the change looks good.

About the orthography, don’t take it personally but it looks to me that you are pushing for Ślabikorz while the other orthography is still in use. I can see a lot of [citation needed] around the "de facto" status of Ślabikorz on the [Wikipedia article about Silesian orthography](https://en.wikipedia.org/wiki/Silesian_orthography). So may I ask why is that only Ślabikorz should be used on Tatoeba in the first place? What if a new Silesian contributor shows up and wants to contribute a lot of good sentences, but not using Ślabikorz? Just trying to understand the situation.

vxern:
My intent is not at all that Ślabikŏrz be the only orthography used for Silesian. In reality, it's the opposite - It would be nice to see Steuer as well, even though I personally do not use it. This is why I wanted to have 'Silesian (Ślabikŏrz)', which would make room for other orthographies in the future if someone wanted to added, say, 'Silesian (Steuer)' or otherwise. The goal with this change would have been to prevent misunderstandings and to have clear separation between orthographies, the same way 'Norwegian Bokmal' and 'Norwegian Nynorsk' do. This would prevent the sentence database getting mixed up with various orthographies.

vxern:
Ślabikŏrzowy szrajbōnek is considered to be 'de facto' because various modern online sources, games (for example Minecraft, Euro Truck Simulator 2, Age of History all use Ślabikŏrz), YouTube channels (Grzegorz Kulik, Niklaus Pierōn), Wiktionary (lists Ślabikŏrz as the primary entries), Wikipedia (accepts both Ślabikŏrz and Steuer) all use it. Even meritable dictionaries such as Słownik Gōrnoślōnskij Gŏdki (SGG for short) use it too. It's unfortunate that there is no citation, we need one, but yeah, I mention this just to back up that claim.

jiru:
Thank you for clarifying your point. You are suggesting to have "Silesian (Ślabikŏrz)" for now and to let anyone create another "Silesian (Steuer)" "language" if needed, but I am afraid your proposed solution won’t work, allow me to explain why. We rely on the ISO 639-3 standard for adding new languages on Tatoeba. That standard is managed by the SIL, who updates it every now and then. Currently they have assigned `szl` for Silesian, but they will never assign another language code for, say, Silesian (Steuer), for two reasons:
1. `szl` is currently not assigned for Silesian in Ślabikŏrz, but for Silesian in general
2. In general, the creation of a new language code must be backed by scientific research that "proves" it’s a distinct language. However, we are not talking about distinct languages but distinct orthographies.

I suppose that the SIL created distinct codes for Norwegian Bokmal and Nynorsk on [historical and social grounds](https://en.wikipedia.org/wiki/Norwegian_language_conflict), but the situation is clearly different for Silesian.

By the way, the main reason we strictly adhere to the ISO 639-3 standard is to avoid having to decide what is a language, since it’s an almost impossible thing to do and we don’t have the resources for that anyway.

Even if we were to ignore the ISO 639-3 standard, imagine for a second what it would be to have separate languages for Ślabikŏrz and Steuer. If I want to search for Silesian sentences containing a word, now I have to perform two searches in two languages *just to be sure that I am not missing out potential results in the "other" language*.

Let me suggest an alternative way of dealing with the problem of having multiple orthographies in Silesian. The differences between Steuer and Ślabikŏrz seem closer to ones of a writing system than of a separate language. On Tatoeba we have a functionality that allows having different scripts in a single language. For example, [Uzbek](https://tatoeba.org/fr/sentences/show_all_in/uzb/none) may be written either in Cyrillic or Latin alphabet. Contributors can write sentences in either script, and a piece of software automatically converts their sentence into the other script and add it as an alternative writing. Either script can be used to perform searches.

A quick search for _ślabikŏrz steuera converter_ brings me [a converter](https://silling.org/konwerter-ze-steuera-na-slabikorz/), so it looks technically feasible. What do you think?

vxern:
I think that would be brilliant, yeah, it would solve the issue of misunderstandings arising, so I'm all for it.

vxern:
I assume converters have to be written from scratch - If that's the case, I'd be interested in writing one myself and coming back once that's complete. Let me know.

vxern:
I assume converters have to be written from scratch - If that's the case, I'd be interested in writing one myself and coming back once that's complete. Let me know.

jiru:
If you want to contribute a converter, please have a look and the existing ones. They are defined in `src/Model/Entity/Transcription.php` and their implementation is in `src/Lib/Autotranscription.php`. I can also provide help if you need.

For now, do you mind reverting the language name change? Then I can merge your pull request with the flag update only, and you can open a new 'enhancement' issue for the converter.

vxern:
Yes, that's fine, let's isolate this PR to a flag change.

Issue #3154, opened 2025-01-16
Interslavic (ISV)


profesyonal:
* Language name: Interslavic
* Language info: https://en.wikipedia.org/wiki/Interslavic
* ISO 639-3 code: ISV
* Public List: https://tatoeba.org/en/sentences_lists/show/173464
* Icon: https://commons.wikimedia.org/wiki/File:Flag_of_Interslavic.svg
* Requested by: [rul](https://tatoeba.org/en/user/profile/rul)

PR #3156, opened 2025-01-26
Add language Interslavic (isv)

DJ-Saidez:
This issue solves #3154


jiru:
The flag doesn’t look right: there are two transparent bands on the top and bottom. The height/width ratio should be 2/3 (for example, if scaled down so that the height is 20 pixels, then the width should be 30 pixels).

image

DJ-Saidez:
Is this showing properly now?

jiru:
Yes :+1:

Issue #3153, opened 2025-01-13
West-Central Oromo (GAZ)


Guybrush88:
* Language name:  West-Central Oromo
* Language info: https://en.wikipedia.org/wiki/Oromo_language
* ISO 639-3 code: GAZ
* Public List: https://tatoeba.org/en/sentences_lists/show/173460
* Icon: https://upload.wikimedia.org/wikipedia/commons/3/3c/Flag_of_the_Oromia_Region.svg
* Requested by: [abiniz](https://tatoeba.org/it/user/profile/abiniz)

profesyonal:
So the flag needs to be uploaded and the language needs to be added to LanguagesLib.php? Is there anything else?

DJ-Saidez:
@profesyonal That's basically it. You can follow https://github.com/Tatoeba/tatoeba2/pull/3086 as a reference if you wish. I'll see if I can get it done now though.

PR #3157, opened 2025-01-26
Adds language West-Central Oromo (gaz)

DJ-Saidez:
This issue solves #3153


DJ-Saidez:
Check fails due to #3156, #3157 and #3158 being based off the same dev.

jiru:
The flag is alright now, thank you!

jiru:
Actually, the flag still needs some fixing.

1. The ratio is currently 2:1 (1200×600) but it should be 3:2 instead.
2. The SVG height/width should be set to 30×20.

I can make these changes if you want.

DJ-Saidez:
That would be great, thank you! That way I can see how to do it next time from the data I have here.

jiru:
@DJ-Saidez The commit above does the necessary changes.

Changing the SVG height and width is just a matter of changing the height and width attributes of the `<svg>` root tag. It is the default size at which the browser should rasterize the image.

Changing the ratio however is a bit more complex, but the easiest way is to change the `viewBox` attribute. The viewbox is like a picture frame. Its values consist of the coordinates of a rectangle (top left corner x and y, width and height). After drawing the SVG elements, whatever part of the image that happens to be inside that rectangle will be visible in the final result; everything out of the rectangle is discarded. So, adjusting the viewbox lets you cut off some part of the image, effectively changing the height/width ratio. If you use a SVG editor to open the file, it will let you see the parts outside the view box:
gaz inkscape

PR #2900, opened 2022-01-31
Rename Toki Pona code

jiru:
Solves #2898.

Deployment instructions (as root):

1. Disable reindexation in crontab (both merge and delta update)
    ```
    crontab -e
    ```
2. Make sure latest indexation finished
    ```
    [ -f /tmp/SphinxIndexesShell.php.lock ] && echo "Wait" || echo "Good to go"
    ```
3. Turn on maintenance mode
    ```
    vim /etc/nginx/sites-enabled/tatoeba.org.conf
    systemctl reload nginx
    ```
4. [Deploy](https://github.com/Tatoeba/tatoeba2/wiki/Deployment) (`set-prod-folder.sh` should run the migration script)
5. Generate new Toki Pona indexes
    ```
    cake sphinx_indexes update main tok
    cake sphinx_indexes update delta tok
    ```
6. Turn off maintenance mode
    ```
    vim /etc/nginx/sites-enabled/tatoeba.org.conf
    systemctl reload nginx
    ```
7. Enable reindexation in crontab (both merge and delta update)
    ```
    crontab -e
    ```
8. Cleanup old Toki Pona index files
    ```
    rm /var/manticore/indices/toki_*
    ```

Issue #3140, opened 2024-08-18
Replace Old Tupi [tpw] with Tupinambá [tpn]


LBeaudoux:
In January 2023,  the`tpw` ISO 639-3 language code was deprecated and merged into `tpn`.

More details [here](https://iso639-3.sil.org/code/tpw).

DJ-Saidez:
Tupinambá was not yet a language designation in Tatoeba, so this should be a simple case of renaming Old Tupi to Tupinambá and updating the ISO code.

PR #3158, opened 2025-01-26
Updates deprecated Old Tupi (tpw) to Tupinambá (tpn)

DJ-Saidez:
This issue seeks to solve #3140 
This was almost entirely based off jiru's fix #2900


config/Migrations/20250125_OldTupiCodeRename
```diff
@@ -30,6 +31,10 @@ class OldTupiCodeRename extends AbstractMigration
                      ->execute();
             }
         }
+        $audioBasePath = Configure::read('Recordings.path');
+        if (is_dir($audioBasePath.DS.$from)) {
+            rename($audioBasePath.DS.$from, $audioBasePath.DS.$to);
+        }
```
jiru, line 37: There is no need to rename audio directories anymore since #2880 was merged.

config/Migrations/20250125_OldTupiCodeRename
```diff
@@ -0,0 +1,42 @@
+<?php
+
+use Migrations\AbstractMigration;
+
+class OldTupiCodeRename extends AbstractMigration
+{
+    private $langColumns = [
```
jiru, line 8: In theory you want to add the new column
```
'audios' => ['sentence_lang'],
```
let’s say, for the sake of completeness. But in practice it’s not really necessary because we don’t have any audio in Old Tupi.

DJ-Saidez:
Would you like me to make those changes, or are you just mentioning those for posterity?

jiru:
Well it’s your pull request so you have to make the suggested changes (unless you ticked "allow maintainters to modify this PR" setting when you created it). My initial thought was that you take care of the flags and language name changes and I handle the code part, but since you added a migration script to your pull request, I take it that you want to contribute some code too, which is great by the way, so I am reviewing it.

The "column" I mentioned is a column in the SQL table that contains data about audio recordings.
The audios table goes like this in SQL terms:
```sql
CREATE TABLE `audios` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `sentence_id` int(11) NOT NULL,
  `sentence_lang` varchar(4) DEFAULT NULL,
  `user_id` int(11) DEFAULT NULL,
  `external` varchar(500) DEFAULT NULL,
  `created` datetime NOT NULL,
  `modified` datetime NOT NULL,
  PRIMARY KEY (`id`),
  KEY `sentence_id` (`sentence_id`)
);
```
So there are 7 columns, one of them `sentence_lang` contains the language of the sentence associated to that audio. The migration script takes care of going through all the records of all the tables in Tatoeba’s database and renames language codes wherever they are. The script has to know what columns in what tables contain language codes. We tell it by setting the `$langColumns` attribute to the various locations and `'audios' => ['sentence_lang'],` is an additional one, it basically means "column sentence_lang in the table audio".

DJ-Saidez:
> I take it that you want to contribute some code too
My programming capabilities are basic at best, I took one semester of Python, so maybe not much that I would be able to do 😅
I do appreciate the reviews and comments though, this makes it very helpful both for myself and any other future contributors!

config/Migrations/20250125_OldTupiCodeRename
```diff
@@ -0,0 +1,45 @@
...
+use Cake\Core\Configure;
+use Migrations\AbstractMigration;
+
+class OldTupiCodeRename extends AbstractMigration
+{
+    private $langColumns = [
+        'contributions' => ['sentence_lang', 'translation_lang'],
+        'contributions_stats' => ['lang'],
+        'languages' => ['code'],
+        'audios' => = ['sentence_lang'],
```
jiru, line 12: There is a typo here. It should be 
```
'audios' => ['sentence_lang'],
```
without the extra equal sign.

jiru:
And one last thing, the migration file should be named with a .php extension, e.g. `20250125_OldTupiCodeRename.php`.

jiru:
There is also a problem with the name of your migration script file. The prefix number 20250125 is used to compare against other migration scripts in `config/Migrations/` and see which one is older or newer in order to apply migrations in correct order. Since 20250125 is a lower number than all the other 14-digit numbers (such as 20190401052133), it is considered to be the oldest, and that confuses the program that applies migrations. Normally the creation of migration files is automated with the help of `[cake bake](https://book.cakephp.org/migrations/3/en/index.html#creating-migrations)` so you don’t have to care about such details.

jiru:
@DJ-Saidez I did the above-mentioned changes myself eventually because I wanted to get this merged. Thank you for contributing this PR!

DJ-Saidez:
Sorry, I didn't catch this notification, thank you!

PR #3182, opened 2025-04-27
SVG updates

boracasli98:
I've updated the SVG flag files for lots of languages.


jiru:
Hello, thank you for your PR.

1. Can you explain what is being updated in the icons and why you think these changes are required?
2. We have set a soft limit of 4k per flag in order to keep the icons size-optimized. Because many of your icons are over 4k, the checks are showing red. Please try to optimize the icons for size, and if some icon really cannot be kept under 4k, we could add some exceptions or raise the limit.

jiru:
Just to clarify what is needed to merge this PR:

1. Justify the graphical changes made to some of the flags.
2. Set width="30px" height="20px" in the `<svg>` tag instead of custom values.
3. Optimize the SVGs for size, try to keep them under 4KB.

I would also like to point out that updating so many flags at once requires a lot of work, not only to get it done correctly, but also to get it reviewed. Instead, I suggest that you split this PR into separate PRs that each change only a few flags, or focus on a single type of change applied to all flags. To do so, just close this PR and open some new ones.

Issue #3169, opened 2025-04-07
Mandar (mdr)


trang:
* Language name: Mandar
* Language info: https://en.wikipedia.org/wiki/Mandar_language
* ISO 639-3 code: [mdr](https://iso639-3.sil.org/code/mdr)
* Public List: https://tatoeba.org/en/sentences_lists/show/173571
* Icon: https://upload.wikimedia.org/wikipedia/commons/9/95/Flag_of_West_Sulawesi.svg
* Requested by: https://tatoeba.org/en/user/profile/Rusydy

PR #3190, opened 2025-06-04
Add mdr lang

Rusydy:
This PR is trying to solve this Issue; https://github.com/Tatoeba/tatoeba2/issues/3169


src/Lib/LanguagesLib.php
```diff
@@ -543,6 +543,7 @@ public static function languagesInTatoeba()
                 'urh' => __d('languages', 'Urhobo'),
                 'mvv' => __d('languages', 'Tagal Murut'),
                 'mdf' => __d('languages', 'Moksha'),
+                'mdr' => __d('languages', 'Mandar'),
```
jiru, line 546: Can you please add the language at the very end of the list, instead of the middle? This is what we have been doing so far and I don’t want to unnecessarily change this habit.

Rusydy: okay

jiru:
Thank you @Rusydy! The next step is to wait for the next update of tatoeba.org, and then Mandar will be available as a new language on https://tatoeba.org. It will happen whenever I have time to do it.

PR #3191, opened 2025-06-04
Fix unit test

Rusydy:


jiru:
Sorry for messing up with the tests back then, but I have already fixed the problem in acb40ee0c42b036f6a5ac407cf22a7c2e92bf537. This PR does not fix anything any more, instead it removes most of the new tests I added, so I am going to close it.

PR #3198, opened 2025-08-04
Issue 3192 - Replace CakePHP Error Messages with Translatable Strings

SrulyP:
CakePHP's default error messages were not translatable, making the pages display 400 and 500 errors in English, even if the page is in a different language. This PR converts the errors to use the translation system so they can be displayed in different languages.

Screenshot 2025-08-04 110214

### DIFF
--- a/src/Template/Error/error400.ctp
+++ b/src/Template/Error/error400.ctp
@@ -31,8 +31,12 @@ endif;
 $this->end();
 endif;
 ?>
-<h2><?= h($message) ?></h2>
-<p class="error">
-    <strong><?= __d('cake', 'Error') ?>: </strong>
-    <?= __d('cake', 'The requested address {0} was not found on this server.', "<strong>'{$url}'</strong>") ?>
-</p>
+<h2><?= h(__($message)) ?></h2>
+<?php if ($code == 404): ?>
+    <h2><?= format(
+        __('The requested address {0} was not found on this server.'),
+        "<strong>'{$url}'</strong>"
+    ) ?></h2>
+<?php else: ?>
+    <h2><?= h(__("Client error")) ?></h2>
+<?php endif; ?>
--- a/src/Template/Error/error500.ctp
+++ b/src/Template/Error/error500.ctp
@@ -36,8 +36,9 @@ if (Configure::read('debug')) :
     $this->end();
 endif;
 ?>
-<h2><?= __d('cake', 'An Internal Error Has Occurred') ?></h2>
-<p class="error">
-    <strong><?= __d('cake', 'Error') ?>: </strong>
-    <?= h($message) ?>
-</p>
+<?php if ($code >= 500): ?>
+    <h2><?= h(__('Server error')) ?></h2>
+    <p><?= h(__('An internal error has occurred.')) ?></p>
+<?php else: ?>
+    <h2><?= h(__('Unexpected error')) ?></h2>
+<?php endif; ?>


jiru:
Thank you for this PR, @SrulyP! :slightly_smiling_face: 

There are a few minor problems with your code.

You committed changes that uses CRLF [end-of-line](https://en.wikipedia.org/wiki/Newline#Representation) (DOS style) while we use CR only (UNIX style). It makes it hard to see what lines you changed because every single line of the files you touched are marked as "changed". This could be caused by your text editor or git. If you are using Git for Windows, there is a [setting described here](https://github.com/Tatoeba/tatoeba2/blob/dev/README.tatovm.md#installing-a-local-instance). Please [convert back the two files to CR end-of-line](https://en.wikipedia.org/wiki/Newline#Conversion_between_newline_formats) and commit them again.

src/Template/Error/error400.ctp
```diff
@@ -1,38 +1,41 @@
...
+<?= $this->element('auto_table_warning') ?>
+<?php
+if (extension_loaded('xdebug')) :
+    xdebug_print_function_stack();
+endif;
+
+$this->end();
+endif;
+?>
+<h2><?= __($message) ?></h2>
```
jiru: This line won’t work because the content between `__()` needs to be a raw string. Otherwise, the string won’t be picked up by gettext script `./tools/generate_pot.sh`. In other words, we can’t really translate it, or we need to translate it beforehand in the calling code (the controller). However the code passing on this variable is CakePHP itself in `vendor/cakephp/cakephp/src/Error/ExceptionRenderer.php`, so it’s not easy to change.

I suggest that you use the `$code` view variable to display an appropriate title instead, such as
```php
<?php if ($code == 404): ?>
<h2><= h(__("Page not found")) ?></h2>
<?php else: ?>
<h2><= h(__("Client error")) ?></h2>
<?php endif; ?>
```
`$code` contains the [HTTP status code](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes), which could be any number from 400 to 499.

Also you need to keep `h()` to escape content (e.g. if the translation contains a `>`, it will be converted to `&gt;`, keeping the overall HTML valid).

src/Template/Error/error500.ctp
```diff
@@ -36,8 +36,8 @@ if (Configure::read('debug')) :
     $this->end();
 endif;
 ?>
-<h2><?= __d('cake', 'An Internal Error Has Occurred') ?></h2>
+<h2><?= format(__('An Internal Error Has Occurred')) ?></h2>
 <p class="error">
-    <strong><?= __d('cake', 'Error') ?>: </strong>
-    <?= h($message) ?>
+    <strong><?= format(__('Error: ')) ?> </strong>
```
jiru, line 41: You should not need any `format()` here, because there is no placeholder parameter in the string.

SrulyP: Without the format() it would not keep the url, and would just display the {0}

src/Template/Error/error400.ctp
jiru: Yes! But remove the first line; we only want a single `<h2>`.

SrulyP: Thank you for your patience and help, I really appreciate it!

src/Template/Error/error500.ctp
```diff
@@ -36,8 +36,8 @@ if (Configure::read('debug')) :
...
+    <?= __($message) ?>
```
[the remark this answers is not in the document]
SrulyP: For some reason the error500 page does not want to load, and instead I get this every time:
image

I have tried putting foobar(); within the block and before the block; both show the same page.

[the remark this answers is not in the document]
jiru: Weird. Try to add this line inside `beforeFilter()` of `src/Controller/AppController.php`.
```php
throw new \Exception();
```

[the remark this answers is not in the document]
jiru: Thank you! I am not a native English speaker, but it looks good. Feel free to commit the whole thing and we can ask @alanfgh to review the wording.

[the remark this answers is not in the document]
SrulyP: Thank you! Again, I really appreciate all your help and patience, and for letting me contribute to this cool project :) 
I just pushed all my changes.

PR #3199, opened 2025-08-19
Limit paging on long listings of sentences

jiru:
This PR limits the total number of results on listings of sentences that go over 1000.

We already have such limitation on "show all sentences in {language}" and "latest contributions by {user}" (#1353). This PR additionally limits the following listings, that can easily go over thousands of pages, thanks to the contributions of prolific members:

* `/audio/index`
* `/audio/of/{user}`
* `/sentences_lists/show/{n}`
* `/tags/show_sentences_with_tag/{n}`

The rationale for this PR is that we are seeing increasingly aggressive and numerous crawlers going through all these listings. The higher the page number, the more burden for the server. The current workaround is to use nginx to limit the number of requests on high-numbered pages. This workaround is only enabled on tatoeba.org and produces "429 Too Many Requests" error pages. This PR makes these limits more integrated, configurable and user-friendly.

## Notable changes
* On "sentences with tag" and "sentence list" listings, sorts that are not listed in the menu can no longer be used.
* Audio listing is slightly different (corrected). For sentences having multiple audios, the oldest audio was used to rank the sentence. Now, the newest audio is used. This means uploading a new audio or enabling an existing audio to a sentence will immediately push it on the top of the listing.
* Audio listing is further optimized and should load faster

## Audio
<a href="https://dev.tatoeba.org/en/audio/index">
image
</a>
<a href="https://dev.tatoeba.org/en/audio/index/eng">
image
</a>
<a href="https://dev.tatoeba.org/en/audio/of/CK">
image
</a>

## Sentence list
<a href="https://dev.tatoeba.org/en/sentences_lists/show/907">
image
</a>

## Sentences with tag
<a href="https://dev.tatoeba.org/en/tags/show_sentences_with_tag/4970">
image
</a>

### DIFF
--- a/src/Controller/AudioController.php
+++ b/src/Controller/AudioController.php
@@ -98,7 +98,11 @@ class AudioController extends AppController
         }
         $total = $total->count();
 
-        $sentencesWithAudio = $this->paginate($this->Audios, compact('finder'));
+        try {
+            $sentencesWithAudio = $this->paginate($this->Audios, compact('finder'));
+        } catch (\Cake\Http\Exception\NotFoundException $e) {
+            return $this->redirectPaginationToLastPage();
+        }
 
         $this->set(compact('sentencesWithAudio', 'totalLimit', 'total'));
         
--- a/src/Controller/AudioController.php
+++ b/src/Controller/AudioController.php
@@ -86,14 +86,21 @@ class AudioController extends AppController
     public function index($lang = null) {
         $this->loadModel('Audios');
 
-        $finder = ['sentences' => []];
+        $totalLimit = $this::PAGINATION_DEFAULT_TOTAL_LIMIT;
+        $finder = ['sentences' => [
+            'maxResults' => $totalLimit,
+        ]];
+        $total = $this->Audios->find()->select(['sentence_id'])->distinct();
         if (LanguagesLib::languageExists($lang)) {
-            $finder['sentences'] = compact('lang');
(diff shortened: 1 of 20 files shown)


jiru:
@ckjpn Thank you for taking the time to leave a feedback.

> * Complete Lists can still be used within advanced search results
> 
> * Compete list data is in the weekly export.
> 
> * (and the same for tags)

I confirm that the above is true, these things won’t change.

> Perhaps you could first try adding this limit only for non-logged in users.
> 
> Or, if that's still too "open", further use this limit for anyone with less than "advanced contributor" status.
> 
> This would likely deal with the problem of "aggressive and numerous crawlers going through all these listings" and wouldn't affect those members who might want to browse through a complete list online for study or to add translations.

I won’t add this limit only for non-logged in users because it is not very practical nor sustainable. By "practical", I mean that even legitimate browsing by logged-in users may result in very long loading times (like 20 or 30 seconds), or even reach the timeout error of 60 seconds. By "sustainable", I mean having a production server that can stand a relatively high pressure, otherwise every now and then it goes down and somebody needs to take action to counter ddos-like avalanche of requests.

In the future, we can consider raising the limit from 1000 to a higher value because I am guessing that browsing around the 1000th to 10000th result could still be computationally reasonable, but this requires more practical testing and I am not sure anybody needs it in the first place. For those who really need to browse thousands of results, there are alternative methods such as the [exports](https://tatoeba.org/downloads) or the [new API](https://api.tatoeba.org/).

PR #3217, opened 2025-10-16
Modify per-language user list page to make it easier to find active native speakers

### LINKED ISSUE #3075: offer an efficient way to find speakers who are active and native-level (or as close to both as possible)
There are several situations when I want to find speakers who have been active as recently as possible, and who have self-reported native-level proficiency in a given language (or as close to it as possible):

(1) when I want to find help with sentences in a language that is not my native one
(2) when I want someone to comment on the correctness/naturalness of a sentence in a language that is not my native one (sometimes when I'm acting as an admin/corpus maintainer, sometimes not)

I haven't found a good way to do this. The closest I've come is an advanced search. Here's an example for Bulgarian:

- Browse/Browse by language: Bulgarian
- Check "Owned by a self-identified native"
- Set Sort/Order to either "Last created first" or "Last modified first" (both have their strengths and drawbacks; in this example, I chose "Last created first")

Then I need to plow through the sentences. This is not ideal because what I really want is a list of speakers, perhaps sorted by proficiency in Bulgarian, perhaps sorted by most recent activity, perhaps sorted by one and then the other. But instead I get six sentences by "dobri" (who has not visited the site for almost five months), followed by at least 50 by "iu" (who has not visited the site for almost six months). I eventually managed to find another contributor -- but this contributor is not a native speaker.

I tried using @ckjpn's [lists of native speakers with recent contributions](https://tatoeba.org/en/wall/show_message/40063#!#message_40063) (not located on Tatoeba). But that was also inefficient because it contained many rows where people had not made any contributions in the last X months. Also, I could only sort by one column at a time, not two. I was able to combine a one-column sort with a search for the language code, but this was not ideal.  

I would expect "Community/List of all members", "Community/Languages of members", or "Community/Native speakers" to allow me to find this information. However: 
- the "List of all members" is in the wrong order for that (it's sorted by role and then by date of joining Tatoeba)
- "Community/Languages of members" is also in the wrong order (the list starts with many members with zero contributions)
- "Community/Native speakers" doesn't provide any links, only statistics

I would love it if a new submenu were added to the "Community" menu. This submenu would make it easy to find as-recent-as-possible as-proficient-as-possible members.

franklindyer:
fixes #3075 

- adds an icon to users on the per-language user list page (at the endpoints `/users/for_language/XXX`) to users who have been active within the past week
- also sorts users on the per-language user list page by the time of their most recent sentence/translation

This should make it easier to find the most active / likely-to-respond users in a given language on Tatoeba. See [this former PR](https://github.com/Tatoeba/tatoeba2/pull/3212) for further discussion. 

It would be nice to get some feedback from @alanfgh on whether this would meet his need. Alternatively, perhaps this could get rolled out to the dev instance of Tatoeba and we could ask alanfgh to try it out.

### DIFF
--- a/tests/TestCase/Controller/SentencesControllerTest.php
+++ b/tests/TestCase/Controller/SentencesControllerTest.php
@@ -439,29 +439,4 @@ class SentencesControllerTest extends IntegrationTestCase {
         $this->get("/en/sentences/of_user/$user?page=9999999");
         $this->assertRedirect("/en/sentences/of_user/$user?page=$lastPage");
     }
-
-    public function testNewSentence_UpdatesLastContributionField() {
-        $users = TableRegistry::get('Users');
-
-        $this->logInAs('contributor');
-
-        $this->post('/en/sentences/add_an_other_sentence', [
-            'value' => 'This is my new English sentence.', 
-            'selectedLang' => 'eng', 
-        ]);
-        $this->assertResponseOk();
-        $user = $users->find()->where(['username' => 'contributor'])->first();
-        $previousLastContribution = $user->last_contribution;
-
-        sleep(1);
-        $this->post('/en/sentences/add_an_other_sentence', [
-            'value' => 'This is my newer English sentence.', 
-            'selectedLang' => 'eng', 
-        ]);
-        $this->assertResponseOk();
-        $user = $users->find()->where(['username' => 'contributor'])->first();
-        $newLastContribution = $user->last_contribution;
-
-        $this->assertGreaterThan($previousLastContribution, $newLastContribution);
-    }
 }


franklindyer:
It looks like the CI is failing because the `users` table that is being used for Jenkins testing lacks the new `last_contribution` column. Is there something that needs to be changed in the repo to make the tests go through, or is there some database used for testing on a machine external to the repo?

Fyi, this PR includes a script `docs/database/updates/2025-10-13.sql` for adding the new column to the `users` table.

jiru:
To fix the CI, you can edit the [fixture](https://book.cakephp.org/3/en/development/testing.html#creating-fixtures) file `tests/Fixture/UsersFixture.php`. You need to edit both the schema definition and the table data. You can edit the file manually, or use the cake console with `cake bake fixture` (from within the VM, from the Tatoeba repository), or a combination of both.

To run the tests locally, you can run `phpunit`. This will help you confirm the CI should pass without having to `git push` your changes.

To add the new column, we used to create migration scripts in `docs/database/updates/`, but this is no longer the way we do it now. Instead, you should [create a migration script](https://book.cakephp.org/migrations/3/en/index.html) with `cake bake migration`, and then execute it with `cake migrations migrate`. This method allows you to also rollback with `cake migrations rollback`, as well as execute custom code, typically to populate the new field for existing users.

The table definition in `docs/database/tables/users.sql` is what the table looked like *before* we started using migration scripts, so you need not to edit it. Sorry this is all very confusing!

src/Controller/SentencesController.php
```diff
@@ -294,6 +305,7 @@ public function add_an_other_sentence()
         if ($userLevel < 0) {
             return;
         }
+        $this->update_user_last_contribution($userId);
```
jiru: CakePHP offers a mechanism to perform some action on every creation or update of a model: the afterSave callback. Have a look at the `afterSave()` method in `UsersTable.php` or `SentencesTable.php`. You could call `update_user_last_contribution()` from there instead. The advantage is that:
* you call it *only once*, instead of calling it from every controller action potentially involved in sentence creation/update
* it is easier to write unit tests

franklindyer:
Thanks for the pointers! The unit tests are now passing, and I moved `update_user_last_contribution` to a similarly named method in `UsersTable.php` which gets called in the `afterSave` method of `SentencesTable.php`.

alanfgh:
> It would be nice to get some feedback from @alanfgh on whether this would meet his need. Alternatively, perhaps this could get rolled out to the dev instance of Tatoeba and we could ask alanfgh to try it out.

It does sound like this approach would accomplish what I was looking for. In any case, if you move your changes over to dev, let me know, and I'll check it out there. Thanks!

docs/database/updates/2025-10-13.sql
jiru: This file can be removed now.

src/Model/Table/UsersTable.php
```diff
@@ -592,4 +592,13 @@ public function updatePasswordVersion($userId, $plainTextPassword)
             $this->save($user);
         }
     }
+
+    public function updateLastContribution($userId)
+    {
+        if ($this->exists(["Users.id" => $userId])) {
```
jiru: You can use the following pattern instead:
```php
use Cake\Datasource\Exception\RecordNotFoundException;

try {
    $user = $this->get($userId);
} catch (RecordNotFoundException $e) {
}
```
The benefit is that only one SQL SELECT request is performed (at `get()`) instead of two (both at `exists()` and `get()`).

config/Migrations/20251016065152_AddUsersLastContributionColumn.php
```diff
@@ -0,0 +1,23 @@
...
+     * Change Method.
+     *
+     * More information on this method is available here:
+     * https://book.cakephp.org/phinx/0/en/migrations.html#the-change-method
+     * @return void
+     */
+    public function change()
+    {
+        $table = $this->table('users');
+        $table->addColumn('last_contribution', 'integer', [
```
jiru: That’s really not a big deal, but the position of the `last_contribution` column is different here than what you specified in the fixtures. The migration script adds the column as the last position, unless specified with the option `after`, e.g. `'after' => 'last_time_active'`.

You can use `cake migrations rollback` and `cake migrations migrate` to re-run the migration, and check how the table looks with `desc users;` in the mysql shell.

src/Template/Users/for_language.ctp
```diff
@@ -126,6 +127,9 @@ $this->set('title_for_layout', $this->Pages->formatTitle($title));
                         $username
                     )
                 );
+                if ($timeSinceLastActivity < 604800) { /* Add a calendar icon for users who have been active within the past week */
+                    echo '<md-icon class="material-icons user-recently-active-icon" aria-label="comment" title="Active this week">comment</md-icon>';
```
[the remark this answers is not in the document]
franklindyer: Got it. I re-ran `generate_pot.sh` as well to update the localization files... let me know if you see any issues with this.

franklindyer:
Okay, I've taken care of the following:

- removed the legacy migration script and tweaked cakephp migration script
- made cakephp migration script populate default `last_contribution` values from `contributions` table
- changed `last_contribution` field to datetime type
- changed use of `exists` to use of `get` with try-except block
- replaced `title` attribute with `md-tooltip` element
- improved tooltip text and made it localizable

I'm currently struggling with the unit tests but haven't made much progress. I'll keep playing around with them, but if you want to merge and add the unit tests yourself I won't complain. :-)

jiru:
> Did you have something more sophisticated than that in mind?

Thank you, it’s good and I can merge it like this already. Two possible improvements if you want:
* Make it a _model test_ instead, because it really is happening in the model, not the controller.
* Instead of creating two sentences and sleeping in between, you can mock the "now" time into some hard-coded value using `Time::setTestNow()`. You need to call it once to mock (with the time you want as parameter), and once again to disable mocking (without any parameter). Example: https://github.com/Tatoeba/tatoeba2/blob/4d71b5bf44e5d6af49501319f264c865cfdab8a4/tests/TestCase/Model/Table/ExportsTableTest.php#L259-L276

jiru:
Oh and you should also test that `users.last_contribution` is updated on sentence update. The `afterSave()` callback is called both during creation and update. This includes sentence text edition and sentence adoption.

jiru:
By the way, sentence adoption/unadoption is not logged in the `contributions` table (which is a pain, I think). If we decide that sentence adoption should update `users.last_contributions`, while that field is not populated for existing users in a way that takes sentence adoption into account, then it means we introduce a discrepancy between both. It’s not a big deal though.

src/Model/Table/UsersTable.php
```diff
@@ -592,4 +594,15 @@ public function updatePasswordVersion($userId, $plainTextPassword)
...
         }
     }
+
+    public function updateLastContribution($userId)
+    {
+        try {
+            $user = $this->get($userId);
+        } catch (RecordNotFoundException $e) {    
+        }
+        $user = $this->get($userId);
```
jiru: This line is a duplicate of what’s in the `try {}` block.

tests/TestCase/Model/Table/SentencesTableTest.php
```diff
@@ -1512,4 +1513,40 @@ function testGetSeveralRandomIds() {
...
+        $this->Sentence->saveNewSentence('This is my new English sentence.', 'eng', 1);
+        $user = $this->Sentence->Users->get(1);
+        $previousLastContribution = $user->last_contribution;
+
+        Time::setTestNow(new Time('2019-02-02 00:00:00')); 
+        $this->Sentence->saveNewSentence('This is my newer English sentence.', 'eng', 1);
+        $user = $this->Sentence->Users->get(1);
+        $newLastContribution = $user->last_contribution;
+
+        $this->assertGreaterThan($previousLastContribution, $newLastContribution);
```
jiru: This test works but it feels overkill. I think you can simplify it by only saving one sentence and checking that `users.last_contribution` field equals the time you set in `setTestNow()`.

jiru: Also please call `Time::setTestNow()` without any parameter at the end of the test, otherwise the "now" time will be set in all subsequent tests, too.

tests/TestCase/Model/Table/SentencesTableTest.php
```diff
@@ -1512,4 +1513,40 @@ function testGetSeveralRandomIds() {
...
+        $previousLastContribution = $user->last_contribution;
+
+        Time::setTestNow(new Time('2019-02-02 00:00:00')); 
+        $before = $this->Sentence->get(7);
+        $data = ['id' => '7', 'text' => 'This is the new text of sentence #7.'];
+        $after = $this->Sentence->editSentence($data);
+        $user = $this->Sentence->Users->get(7);
+        $newLastContribution = $user->last_contribution;
+
+        $this->assertGreaterThan($previousLastContribution, $newLastContribution);
```
jiru: Same comments as above.

jiru:
> Got it. Maybe another issue should be opened to include sentence adoption/unadoption in both the contributions/last_contributions table, and the user last_contribution field?

Yes, I agree. In the mean time, it would be best if the last_contribution field was *not* updated on sentence adoption, or on any other modification that is not what the migration script takes into account. In other words, `updateLastContribution()` should behave the same as the contribution logger, which is by the way implemented in `src/Event/ContributionListener.php`.

src/Model/Table/SentencesTable.php
```diff
@@ -278,6 +278,10 @@ public function afterSave($event, $entity, $options = array())
         ));
         $this->getEventManager()->dispatch($event);
 
+        if (!is_null($entity->user_id)) {
+            $this->Users->updateLastContribution($entity->user_id);
```
jiru: There is a logic problem here. If a corpus maintainer or an admin edits someone else’s sentence, it’s gonna update the last_contribution time of the sentence owner instead of the actual sentence editor.

jiru:
Thank you for keeping up with this PR, Franklin. I think we are good now, I will merge the PR once the minor points I mentioned are fixed.

PR #3244, opened 2026-02-09
Add some troubleshooting to README.tatovm.md

kumakyoo42:
I had some trouble installing Tatoeba locally with Debian "trixie" this morning and thought it might be a good idea to add the solution to the README file.


jiru:
Thank you!

PR #3245, opened 2026-02-10
Add translation statistics per language

kumakyoo42:
When I visited [sentences by language](https://tatoeba.org/de/stats/sentences_by_language) for the first time and clicked on a language, I expected some statistics for that language (you get a list of all sentences of that language instead). As this sort of statistics is not available I thought it might be a good idea to add it.

From my experience as a PR reviewer I know it's better to keep PRs small. Hence, this is only part of the work - I did not yet add what will happen when no language is given, or when the language given does not exist. I plan to add this in future PRs. (If you prefere larger all at once PRs, please tell me!)

As this change isn't complete, I didn't change the above mentioned links yet, so for testing you need to point your browser to .../stats/statistics_for_language/eng. 

There is one issue about translations, I couldn't solve yet. Maybe you can help me out here: The sentence above the table "There are 9 sentences in Deutsch, 7 of them have at least one translation." contains two numbers which would change the sentence when being the value 1. As there are two numbers, I can't use `__n()` here or at least I'm not aware of how to do this. Any ideas?

And one more thing: My editor automatically removes trailing spaces and I don't know, how to stop it from doing so. Hence there are a few unrelated minor changes in this PR. Hope this doesn't hurt.

### DIFF
--- a/src/Controller/StatsController.php
+++ b/src/Controller/StatsController.php
@@ -27,6 +27,7 @@
 namespace App\Controller;
 
 use App\Controller\AppController;
+use App\Lib\LanguagesLib;
 use Cake\Event\Event;
 
 /**
@@ -52,6 +53,34 @@ class StatsController extends AppController
         $this->set('stats', $stats);
         $this->set('audioStats', $audioStats);
         $this->set('totalSentences', $totalSentences);
+    }
+
+
+    /**
+     * Show translation statistics for a specified language.
+     *
+     * @param mixed $lang Language for which statistics should be displayed.
+     *
+     * @return void
+     */
+    function translations_by_language($lang = null)
+    {
+        if ($lang !== null && !array_key_exists($lang, LanguagesLib::languagesInTatoeba())) {
+            throw new \Cake\Http\Exception\NotFoundException();
+        }
+
+        $this->set('lang', $lang);
+
+        $this->loadModel('Sentences');
+        $stats = $this->Sentences->getTranslationStatistics($lang);
+        $this->set('stats', $stats);
+
+        $total = $this->Sentences->getTotalTranslations($lang);
+        $this->set('totalTranslations', $total);
+
+        $this->loadModel('Languages');
+        $total = $this->Languages->getTotalSentencesNumber($lang);
+        $this->set('total', $total);
     }
 
 
--- a/src/Model/Table/SentencesTable.php
+++ b/src/Model/Table/SentencesTable.php
@@ -843,7 +843,7 @@ class SentencesTable extends Table
      *
      * @return array Array of translations (direct and indirect).
      */
-    public function getTranslationsOf($id,$lang = null)
+    public function getTranslationsOf($id, $lang = null)
     {
         if ( ! is_numeric($id) ) {
             return array();
@@ -1392,5 +1392,58 @@ class SentencesTable extends Table
             ->all()
             ->extract('lang')
             ->toArray();
+    }
+
+    /**
+     * Computes a table of translation counts for a given language.
+     *
+     * @param string $lang Language for which to calculate the statistics.
+     *                     If null calculate the table for all languages.
+     *
+     * @return query object
+     */
+    public function getTranslationStatistics($lang = null)
+    {
+        $results = $this->Links->query()
+            ->select([
+                'code' => 'translation_lang',
+                'translations' => 'COUNT(*)',
+            ])
+            ->group(['code'])
+            ->order(['translations' => 'DESC',
+                     'translation_lang' => 'ASC']);
+        
+        if ($lang !== null)
+          $results = $results->where(['sentence_lang' => $lang]);
+        
+        return $results->toList();
+    }
+
+    /**
+     * Counts the number of sentences with translations for a language.
+     *
+     * @param string $lang Language for which to calculate the statistics.
+     *                     If null calculate the number of all sentences
+     *                     with a translation.
+     *
+     * @return int
+     */
+    public function getTotalTranslations($lang = null)
+    {
+        $results = $this->find()
+            ->join([
+                'table' => 'sentences_translations',
+                'alias' => 'T',
+                'type' => 'RIGHT',
+                'conditions' => 'Sentences.id = T.sentence_id',
+            ])
+            ->select([
+                'count' => 'COUNT(DISTINCT(T.sentence_id))',
+            ]);
+        
+        if ($lang !== null)
+          $results = $results->where(['Sentences.lang' => $lang]);
+            
+        return $results->firstOrFail()->count;
     }
 }
--- a/src/Template/Stats/translations_by_language.ctp
+++ b/src/Template/Stats/translations_by_language.ctp
@@ -1,0 +1,159 @@
+<?php
+/**
+ * Tatoeba Project, free collaborative creation of multilingual corpuses project
+ * Copyright (C) 2026 Bernhard Seckinger <kumakyoo@kumakyoo.de>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * PHP version 5
+ *
+ * @category PHP
+ * @package  Tatoeba
+ * @author   Bernhard Seckinger <kumakyoo@kumakyoo.de>
+ * @license  Affero General Public License
+ * @link     https://tatoeba.org
+ */
+
+if (empty($lang)) {
+    $title = __('Translation statistics for all languages');
+} else {
+    $title = format(
+        __('Translation statistics for {language}'),
+        array('language' => $this->Languages->codeToNameToFormat($lang))
+    );
+}
+
+$this->set('title_for_layout', $this->Pages->formatTitle($title));
+
+?>
+<div id="annexe_content">
+    <?php $this->CommonModules->createFilterByLangMod(); ?>
+</div>
+
+<div id="main_content">
+<div class="section md-whiteframe-1dp">
+    <h2><?= $title; ?></h2>
+
+    <p>
+    <?php
+        if (empty($lang)) {
+            echo '<p>'.format(
+                __n('There is one sentence in the corpus.',
+                'There are {total}&nbsp;sentences in the corpus.',
+                $total),
+                array('total' => $total)
+            );
+        }
+        else {
+            echo '<p>'.format(
+                __n('There is one sentence in {language}.',
+                'There are {total}&nbsp;sentences in {language}.',
+                $total),
+                array('language' => $this->Languages->codeToNameToFormat($lang),
+                      'total' => $total)
+            );
+       }
+    ?>                          
+
+    <?php 
+    if (empty($stats)) {
+        echo __('There are no sentences with translations.');
+        return;
+    }
+    echo format(
+        __n('One sentence has at least one translation.',
+            '{total}&nbsp;sentences have at least one translation.',
+            $totalTranslations),
+        array('total' => $totalTranslations)
+    );
+    ?>
+
+    <table class="language-stats">
+    <tr>
+        <th colspan="3">
+        <?php /* @translators: table header text in "Translation statistics for language" page */ ?>
+        <th><?php echo __('Language'); ?></th>
+        <?php /* @translators: table header text in "Translation statistics for language" page */ ?>
+        <th><?php echo __('Translations'); ?></th>
+    </tr>
+
+    <?php
+    $rank = 1;
+    foreach ($stats as $language) {
+        $langCode = $language['code'];
+        $numSentences = $language['translations'];
+        
+        if (!isset($max)) {
+            // table is sorted by $numSentences, thus the first entry is the maximum
+            $max = $numSentences;
+        }
+        
+        if ($max == 0) {
+            $percent = 0;
+        } else {
+            $percent = ($numSentences / $max) * 100;
+        }
+        $numSentencesDiv  = '<div class="bar" style="width:'.$percent.'%"></div>';
+        $numSentencesDiv .= $this->Number->format($numSentences);
+
+        $languageIcon = $this->Languages->icon($langCode);
+
+        if (empty($lang)) {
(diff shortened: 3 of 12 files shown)


src/Controller/StatsController.php
```diff
@@ -55,6 +56,37 @@ function sentences_by_language() {
...
 
 
+    /**
+     * Show statistics for a specified language.
+     *
+     * @param mixed $lang Language for which statistics should be displayed.
+     *
+     * @return void
+     */
+    function statistics_for_language($lang = null)
```
jiru: Please try to use an action name that makes a [nice URL](https://en.wikipedia.org/wiki/Clean_URL). The URL `http://localhost:8080/fr/stats/statistics_for_language/eng` includes both words stats and statistics, which is redundant. You could change the action name here, or alternatively reuse the existing `sentences_by_language` action, so it becomes `/stats/sentences_by_language/eng`.

src/Model/Table/SentencesTable.php
```diff
@@ -1393,4 +1393,55 @@ public function languagesHavingSentences()
...
+                'conditions' => 'Sentences.id = T.sentence_id',
+            ])
+            ->select([
+                'code' => 'T.translation_lang',
+                'translations' => 'COUNT(*)',
+            ])
+            ->where(['Sentences.lang' => $lang])
+            ->group(['code'])
+            ->order(['translations' => 'DESC'])
+            ->toList();
```
jiru: This request is too expensive to be executed just like that. To give you an idea, I tried to execute it for English on https://dev.tatoeba.org/, on which `sentences_translations` contains 15,437,391 records, and it took 33 seconds. By the way, https://tatoeba.org/ ’s `sentences_translations` table contains 27,688,900 records at the moment.

We can consider executing it as a cronjob, or caching the result of the query.

Also I think you don’t need to do any JOIN here. You can simply use `sentences_translations.sentence_lang` instead of joining with `sentences` to get `sentences.lang`.

src/Model/Table/SentencesTable.php
```diff
@@ -1393,4 +1393,55 @@ public function languagesHavingSentences()
...
+                'table' => 'sentences_translations',
+                'alias' => 'T',
+                'type' => 'RIGHT',
+                'conditions' => 'Sentences.id = T.sentence_id',
+            ])
+            ->select([
+                'count' => 'COUNT(DISTINCT(T.sentence_id))',
+            ])
+            ->where(['Sentences.lang' => $lang])
+            ->first()->count;
```
jiru: Same remarks as for `getTranslationStatistics()`.

In addition, consider using `firstOrFail()`. Alternatively, directly call `->count()`, which makes CakePHP wrap your query into a `COUNT()`.

src/Template/Stats/statistics_for_language.ctp
```diff
@@ -0,0 +1,125 @@
...
+    ?>
+
+    <table class="language-stats">
+    <tr>
+        <th></th>
+        <th></th>
+        <th></th>
+        <?php /* @translators: table header text in "Statistics for language" page */ ?>
+        <th><?php echo __('Language'); ?></th>
+        <th><?php echo __('Translations'); ?></th>
```
jiru: Consider adding a `@translators:` note above `__('Translations')`, too. The notes need to be on the line immediately preceding the `__()` call.

tests/TestCase/Model/Table/LanguagesTableTest.php
```diff
@@ -74,6 +74,10 @@ function testGetTotalSentencesNumber() {
         $expected = 56;
         $n = $this->Languages->getTotalSentencesNumber();
         $this->assertEquals($expected, $n);
+        
+        $expected = 21;
+        $n = $this->Languages->getTotalSentencesNumber('eng');
+        $this->assertEquals($expected, $n);
```
jiru: Can you add a test case for a invalid language code?

webroot/css/stats/statistics_for_language.css
```diff
@@ -0,0 +1,48 @@
...
+
+.language-stats img {
+    display: block;
+}
+
+.language-stats .status-warning {
+    color: #AA0000;
+    margin-left: 10px;
+    float: right;
+}
```
jiru: I don’t think the `status-warning` class appears anywhere in the statistics_for_language view.

kumakyoo42: I wasn't sure if this might creep in from other code I do not know about, so I kept it (I copy-paste from `sentences_by_language.css`). I'll remove it.

kumakyoo42:
I think I addressed every one of your comments now. 

I copied the database to my local version to did some speed tests. Unfortunately querying the English corpus is still too slow (20 seconds on my computer). I learned that InnoDB cannot index `count(*)` and therefore the query can't be speed up. So, in my opinion, we have to resort to something else, like cronjob or caching as you mentioned. (Alternativley we could add a translations table similar to the languages table, but keeping this up to date would not be worth it, I think.)

Suggestion: I'll fill in the "Not yet implemented" parts first, before we dicide on how to continue.

I also want to do the cherry picking for removing the unrelated trailing-whitespace lines from this PR. I didn't do that yet, but I'd like to learn how this works.

src/Controller/StatsController.php
```diff
@@ -55,6 +56,35 @@ function sentences_by_language() {
...
+     * @return void
+     */
+    function translations_by_language($lang = null)
+    {
+        if (!array_key_exists($lang, LanguagesLib::languagesInTatoeba())) {
+            $lang = null;
+        }
+
+        $this->set('lang', $lang);
+        if ($lang === null) return;
```
jiru: I think these two ifs could be simplified:
1. Move the return statement into the first if:
```php
if (!array_key_exists($lang, LanguagesLib::languagesInTatoeba())) {
    return;
}
```
2. In the view, replace `$lang === null` with `isset($lang)`

tests/TestCase/Model/Table/LanguagesTableTest.php
```diff
@@ -74,6 +74,13 @@ function testGetTotalSentencesNumber() {
         $expected = 56;
         $n = $this->Languages->getTotalSentencesNumber();
         $this->assertEquals($expected, $n);
+
+        $expected = 21;
+        $n = $this->Languages->getTotalSentencesNumber('eng');
+        $this->assertEquals($expected, $n);
+
+        $n = $this->Languages->getTotalSentencesNumber('und');
+        $this->assertNull($n);
```
jiru, line 85: Could you also test `$this->Languages->getTotalSentencesNumber(null)`? Alternatively, add a string type constraint on the `$lang` parameter of `getTotalSentencesNumber()`.

jiru:
Here are some stats of running your SQL requests on tatoeba.org:

* getTranslationStatistics (without join):
  * eng: 4 to 6s
  * epo: 1 to 1.5s
  * jpn: 0.3 to 0.5s

* getTotalTranslations (with join, which is currently faster than without):
  * eng: 12 to 15s
  * epo: 4 to 6s
  * jpn: 1.2 to 1.8s

I think this is reasonable if we compute the stats for all languages with a daily cronjob. You can create a simple Cake shell that calculates all the stats and write them to a cache store (see CakePHP Cache class) and retrieve this data on page load.

There is one issue with this approach though: we clear all cache when updating the website, so after a site update, the stats would be unavailable until the next cronjob execution. I think this can be solved by setting up a special kind of persistent cache (in `config/app.php`) that survives website updates and server reboot.

> Suggestion: I'll fill in the "Not yet implemented" parts first, before we dicide on how to continue.

Sure.

> I also want to do the cherry picking for removing the unrelated trailing-whitespace lines from this PR. I didn't do that yet, but I'd like to learn how this works.

* To stage individual blocks to your diff, before commiting: `git add --patch`
* To unstage  individual blocks from your diff: `git restore --staged --patch`

src/Model/Table/SentencesTable.php
```diff
@@ -1393,4 +1393,49 @@ public function languagesHavingSentences()
...
+    public function getTranslationStatistics($lang)
+    {
+        $results = $this->Links->query()
+            ->select([
+                'code' => 'translation_lang',
+                'translations' => 'COUNT(*)',
+            ])
+            ->where(['sentence_lang' => $lang])
+            ->group(['code'])
+            ->order(['translations' => 'DESC'])
```
jiru: When two languages have the same number of translations, their order is undefined so their ranking could be swapped randomly between different runs. You could add a second ordering criterion to make the ordering consistent.

jiru:
After removing the JOIN clause, on tatoeba.org and dev.tatoeba.org the stats are different, some of the numbers got slightly bigger.

I guess this must be caused by dangling links (records in `sentences_translations` that do not refer to existing sentences), or mismatches between `sentence.lang` and `sentences_translations.sentence_lang`.

The different is not so big.

Maybe this can be considered as a different issue we may separately solve.

kumakyoo42:
- I added a global table (when there is no language given) showing for every language the number of sentences with at least one translation.
- I added language selection in the side bar.
- I added a link from the homepage statistics to this page. (I changed my mind here, changing the links on the language statistics page seems not apropriate to me any more.)
- I replaced the links in the table to point to an (advanced) search showing all sentences which are counted in this row. I couldn't test that though, because search doesn't work locally on my computer. Don't know, what I have to do here.
- I addressed all your concerns above.

If that's all fine, the only thing, that misses now (from my point of view), is the use of a cronjob and cached data.

jiru:
It looks very good, thank you.

> * I added a link from the homepage statistics to this page. (I changed my mind here, changing the links on the language statistics page seems not apropriate to me any more.)

It think it is reasonable to add the link on the homepage, but we now have a weird blank space on the left block:

image

I am not sure what’s the best approach. We could add more text to the left part to reduce the void, or move each stats link into the text, for example "25,881 [sentences with translations](http://localhost:8080/en/stats/translations_by_language)".

> search doesn't work locally on my computer

It works for me. You can check `/var/manticore/log/searchd.log`.

src/Controller/StatsController.php
```diff
@@ -55,6 +56,34 @@ function sentences_by_language() {
...
+     * Show translation statistics for a specified language.
+     *
+     * @param mixed $lang Language for which statistics should be displayed.
+     *
+     * @return void
+     */
+    function translations_by_language($lang = null)
+    {
+        if (!array_key_exists($lang, LanguagesLib::languagesInTatoeba())) {
+            $lang = null;
```
jiru: If the language is provided but invalid, I think we should show a 404 error.

With the current implementation, opening http://localhost:8080/en/stats/translations_by_language/foobar shows "Translation statistics for all languages".

To show a 404 error, simply throw a Cake\Http\Exception\NotFoundException.

src/Template/Stats/translations_by_language.ctp
```diff
@@ -0,0 +1,149 @@
...
+<div id="main_content">
+<div class="section md-whiteframe-1dp">
+    <h2><?= $title; ?></h2>
+
+    <p>
+    <?php
+        if (empty($lang)) {
+            echo '<p>'.format(
+                __n('There is one sentence in the database.',
+                'There are {total}&nbsp;sentences in the database.',
```
jiru: I’d like to suggest using the word "corpus" instead of "database". I feel like "database" is unnecessarily technical.

src/Template/Stats/translations_by_language.ctp
```diff
@@ -0,0 +1,149 @@
...
+
+        $langName = $this->Languages->codeToNameAlone($langCode);
+        if (empty($langCode)) {
+            $langCode = 'unknown';
+        }
+        $languageLink = $this->Html->link(
+            $langName,
+            array(
+                "controller" => "sentences",
+                "action" => "search?from=".htmlspecialchars($from)
```
jiru: * You can have `link()` build the query string with:
  ```php
  [
    'controller' => 'sentences',
    'action' => 'search',
    '?' => [
      'from' => $from,
      /* etc. */
    ]
  ]
  ```
* You should add `to=$from` in addition to `trans_to=$from`. `trans_to=` is to limit to sentences having translations, while `to=` is to limit what translations are showed.

src/Template/Stats/translations_by_language.ctp
```diff
@@ -0,0 +1,149 @@
...
+            )
+        );
+
+        echo '<tr>';
+
+        echo $this->Html->tag('td', $this->Number->format($rank));
+        echo $this->Html->tag('td', $languageIcon);
+        echo $this->Html->tag('td', $langCode);
+        echo $this->Html->tag('td', $languageLink);
+        echo $this->Html->tag('td', $numSentencesDiv, array('class' => 'num-sentences'));
```
jiru, line 150: It looks like this `num-sentences` class is not used. It is used in `sentences_by_language.css`, which is not loaded here.

src/Template/Stats/translations_by_language.ctp
```diff
@@ -0,0 +1,149 @@
...
+            $from = $langCode;
+            $to = '';
+        } else {
+            $from = $lang;
+            $to = $langCode;
+        }
+
+        $langName = $this->Languages->codeToNameAlone($langCode);
+        if (empty($langCode)) {
+            $langCode = 'unknown';
```
jiru, line 122: Clicking on the "unknown language" link in the table brings to a search page that does not show sentences in "unknown language" or having translations in "unknown language". Currently, the search engine does not index sentences in "unknown" language, so you cannot display such sentences. So I think you should not show any link in that case.

kumakyoo42:
> It think it is reasonable to add the link on the homepage, but we now have a weird blank space on the left block:

Sorry, I was only aware of that block being used in the right side column.

Maybe we could add a link to https://tatoeba.org/de/donate at the left box?

kumakyoo42:
> It works for me. You can check /var/manticore/log/searchd.log.

See issue #3258 why it didn't work for me.

kumakyoo42:
`createFilterByLangMod` lists the languages of the profile of the user, but it should list the languages in the table. Do you know an easy way to get the correct list displayed?

jiru:
> Maybe we could add a link to https://tatoeba.org/de/donate at the left box?

Good idea. I can imagine a text saying something like "If you like what we do, consider making a [donation](https://tatoeba.org/donate)."

> `createFilterByLangMod` lists the languages of the profile of the user, but it should list the languages in the table. Do you know an easy way to get the correct list displayed?

You could allow the list of languages to be passed as a new parameter to `createFilterByLangMod`. Just to clarify, the profile languages show up in the dropdown when opening it, but any language can be selected by typing something.

kumakyoo42:
> Just to clarify, the profile languages show up in the dropdown when opening it, but any language can be selected by typing something.

Ah, I didn't know that. So no change is needed here.

kumakyoo42:
I still need to implement the caching stuff. But all else should be fine now.

kumakyoo42:
I meanwhile read a lot in the CakePHP documentation about shell and caching, but I still have no clue where to start at all with a cached cron job.

In my opinion CakePHP complicates things a lot and I realize that I start loosing interest in contributing at all.

Sorry, but I will not continue here. If someone want's to take over, feel free to do so.

PR #3253, opened 2026-02-17
The "permanently delete" function for email massages gets a NOT FOUND message

### LINKED ISSUE #2621: The "permanently delete" function for email massages gets a NOT FOUND message.
1. It doesn't work when you put one message into the trash, and then click the "permanently delete" icon, at the top of the resulting page, showing that one message.

https://dev.tatoeba.org/eng/private_messages/delete_forever/1218151
```
Not Found
Error: The requested address '/eng/private_messages/delete_forever/1218151' was not found on this server.
```

2. However, it does work when you click that icon when  looking at all the messages in your trash.

https://dev.tatoeba.org/eng/private_messages/folder/Trash

kumakyoo42:
This pull request addresses issue #{2621}.

Seems like someone confused the `delete_forever` icon with the `delete` action. This action is smart enough to distinguish between messages in the trash and elsewhere, so there is no need for a `delete_forever` action.

### DIFF
--- a/src/View/Helper/PrivateMessagesHelper.php
+++ b/src/View/Helper/PrivateMessagesHelper.php
@@ -80,7 +80,7 @@ class PrivateMessagesHelper extends AppHelper
                 'text' => __('Permanently delete'),
                 'icon' => 'delete_forever',
                 'url' => array(
-                    'action' => 'delete_forever',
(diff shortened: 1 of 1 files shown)


jiru:
Thank you! :heart:

PR #3254, opened 2026-02-17
Improve layout of users stats sidebar

### LINKED ISSUE #3113: User languages stats sidebar overflows
When a the user sets the browser font size to something larger than default (16px) the language stats sidebar can overflow. See examples below.

**To Reproduce**
1. Set browser font size greater than 16px ("Large" or "Very Large" on Chromium)
1. Go to https://tatoeba.org/en/users/for_language/deu

## Chromium font size "Large"
image

## Chromium font size "Very Large"
image

The effect is similar on other language stats pages.
- https://tatoeba.org/en/users/for_language/spa
- https://tatoeba.org/en/users/for_language/fra
etc.

**Additional information (if relevant)**
- This seems similar to issue #3097 which also shows text overflow for non-default browser font sizes.
- The provided screenshots were taken on a Linux system running the Chromium Browser (Version 123.0.6312.58).

kumakyoo42:
This pull request addresses issue #{3113}.

### DIFF
--- a/webroot/css/users/for_language.css
+++ b/webroot/css/users/for_language.css
@@ -49,6 +49,14 @@
     vertical-align: middle;
 }
 
+.usersLanguagesStats td:nth-child(2) {
+    overflow-wrap: anywhere;
+}
+
+.usersLanguagesStats td:nth-child(3) {
+    text-align: right;
+}
+
 .usersLanguagesStats .selected {
     background: #EAF2E6;
     font-weight: bold;


webroot/css/users/for_language.css
```diff
@@ -40,15 +40,36 @@
...
+
+.usersLanguagesStats td:nth-child(2) {
+    padding: 2px 5px;
+    width: auto;
+}
+
+.usersLanguagesStats td:nth-child(3) {
+    padding: 2px 0px 2px 5px;
+    text-align: right;
+    width: 3.5em;
```
kumakyoo42: I couldn't come up with a better solution here. (Currently the number never exceed 5 digits, so this should be enough, independend of font and size. But if we've got 100,000 members of one language, the number might be split into two lines. I also do not know, if numbers might occupy more space in some other languages.)

jiru:
I think you can achieve a similar result by simply letting the middle column wrap anywhere:
```css
.usersLanguagesStats td:nth-child(2) {
    overflow-wrap: anywhere;
}
```

I think this would be better because it does not rely on the assumption that Noto font is used, or that the numbers are less than 6 digits.

kumakyoo42:
Never heard of `overflow-wrap` before. Seems to be a better solution, indeed. I kept the right-align of the 3rd column though, as numbers should be right-aligned anyway.

Issue #3145, opened 2024-11-10
Unable to validate furigana


jiru:
**To Reproduce**
1. Go to https://tatoeba.org/en/sentences/show/224073. This particular sentence is Japanese with text "この２ヶ月間、私はとても忙しかった。" and unconfirmed furigana "この２｛｝ヶ月｛かげつ｝間｛かん｝、私｛わたし｝はとても忙｛いそが｝しかった。"
2. Click on Edit sentence.
3. Edit furigana to "この２｛に｝ヶ月｛かげつ｝間｛かん｝、私｛わたし｝はとても忙｛いそが｝しかった。"
4. Furigana is edited, however かげつ spans over 月 which is incorrect.
5. Click Edit again and click the "undo" button to revert furigana
6. Try to edit again furigana to "この２｛に｝ヶ｛か｝月｛げつ｝間｛かん｝、私｛わたし｝はとても忙｛いそが｝しかった。"
7. See error: The provided sentence differs from the original one near “[”.

**Expected behavior**
The first attempt to fix furigana on ２ should produce correct furigana around ヶ月.

DJ-Saidez:
It appears to be resolved from my end. Do you still see it incorrectly?

jiru:
Yes, it is still incorrect. What I see currently is that the furigana is not validated (i.e. it’s the autogenerated version with the warning sign). This is not correct because it lacks furigana above ２ (it should be <ruby>２<rt>に</rt></ruby>). Lacking furigana over a character as simple as ２ is certainly not a big deal, but the point is that it cannot be corrected in a correct way.

ckjpn:
Verified Problem: I tried it to, and it didn't work for me either.

crtmh.GetHubIsGreat,#0

Pasting that sentence into the following, shows the same furigana as tatoeba.org.

ckjpn:
Perhaps the code tatoeba.org is using doesn't allow furigana to be placed above kana.
That would be one thing to check.

kumakyoo42:
The example can be simplified:

- Enter sentence "ヶ" (language japanese)
- Edit sentence, change transcript to "ヶ｛か｝"

kumakyoo42:
> Perhaps the code tatoeba.org is using doesn't allow furigana to be placed above kana. That would be one thing to check.

[...]

PR #3256, opened 2026-02-20
Allow furigana above a single kana

kumakyoo42:
This pull request addresses issue #3145 

In the current implementation furigana can only be placed above non-kana symbols. But there are some cases where a furigana above a kana makes sense, see https://en.wiktionary.org/wiki/%E3%83%B6 for an explanation. I extended the regex to accept either a non-empty series of non-kana symbols or a single kana symbol.

I excluded the empty sequence, because it will make for a better error message when the user tries to add a furigana to a punctuation symbol (_furigana sentences differs from the original_ vs. _unexpected [ symbol_).

### DIFF
--- a/webroot/js/directives/sentence-and-translations.dir.js
+++ b/webroot/js/directives/sentence-and-translations.dir.js
@@ -590,7 +590,7 @@
                 var punct = '　-〄〇-〠・'; // 。、「」etc.
                 punct += '！＂＃＇（），．／：；？［＼］＾｀～｟｠'; // fullwitdh forms
                 punct += ' '; // space
-                var regex = '([^｝' + hiragana + katakana + punct + ']*)｛([^｝]*)｝';
+                var regex = '([^｝' + hiragana + katakana + punct + ']{1,}|[ヶ])｛([^｝]*)｝';
                 text = text.replace(uniRegExp(regex, 'g'), '[$1|$2]');
                 text = text.replace(uniRegExp('｜', 'g'),  '|');
             }


jiru:
Thank you! But I feel like ヶ should be treated as a special case. I am not a specialist, but it does not seem quite right to allow furigana on any kana.

kumakyoo42:
I am not a specialist here either. I think I saw some more when I tried to find out, what this is about, but I can't find it anymore.

But even, if furigana is only useful above this single kana, would it do much harm to allow it on others too?

jiru:
> I am not a specialist here either.

I can ask around.

> I think I saw some more when I tried to find out, what this is about, but I can't find it anymore.

I think we should keep the implementation based on verified sources. Right now ヶ needs to have furigana in some context, so we can allow it. If later we figure out some other kanas need furigana, we can allow them too, and if eventually we figure out all kanas should be allowed, we can allow them all.

> would it do much harm

One could put furigana on kanas by mistake. One could start inserting irrelevant furigana like <ruby>私<rt>わたし</rt></ruby><ruby>は<rt>わ</rt></ruby> thinking it’s helpful. I think that strong validation rules are key to data consistency.

kumakyoo42:
Accepted. ;-)

jiru:
Thank you for your understanding. :-)

(It looks like the CI breaks because of GitHub issues, you can ignore that for now.)

Your PR only seems to partially solve the original issue #3145 though. Looking at the steps to reproduce, the problem mentioned at step 4 (かげつ spans over 月 only) still exists, but it can now be worked around and step 7 works. Do you think you can make it work directly at step 3?

jiru:
By the way, your PR only solves the issue for the new design. The relevant code for the old design is in `webroot/js/transcriptions.js`. I know it’s a pain to have code duplicated around, but it would be great if you can consider fixing the old design too.

kumakyoo42:
> Your PR only seems to partially solve the original issue https://github.com/Tatoeba/tatoeba2/issues/3145 though. Looking at the steps to reproduce, the problem mentioned at step 4 (かげつ spans over 月 only) still exists, but it can now be worked around and step 7 works. Do you think you can make it work directly at step 3?

Would be nice, yes. But it probably will not be me, who is doing this. By the way, I found sentence [#1630368](https://tatoeba.org/de/sentences/show/1630368), where the furigana is correct. I have not the rights to check what happend there, but it might be useful to know.

> By the way, your PR only solves the issue for the new design. The relevant code for the old design is in webroot/js/transcriptions.js. I know it’s a pain to have code duplicated around, but it would be great if you can consider fixing the old design too.

No sorry, but I will not fix anything concerning the old design. In my opinion the new design is much better and, as far as I know, has less bugs. Therefore I'd say it's time to remove the old design instead of wasting time to fix bugs there.

jiru:
No problem. I will merge your PR as-is.

> By the way, I found sentence [#1630368](https://tatoeba.org/de/sentences/show/1630368), where the furigana is correct.

Thanks, I was curious so I had a look. The furigana in the database is `[２|に][ヶ月|か|げつ][間|かん][雨|あめ]がない。` and it was added by Yorwba. Yorwba is also a developer and contributed code in the past, so it could be that he used his skills to bypass the frontend and replay the POST request with a correct furigana syntax.

Issue #3264, opened 2026-03-07
New sentences should be automatically checked for some typical mistakes


kumakyoo42:
Some mistakes can easily be detected algorithmically, for example missing punctuation. Such algorithms could be used in several places. As a first usecase I'd suggest to use them to issue a warning, whenever one of the algorithms triggers.

See [wall post about stopping new users from doing harm](https://tatoeba.org/de/wall/show_message/41704#!#message_41704) for reference.

kumakyoo42:
This is a first such algorithm. It checks the last character of a sentence to be usual for the language of that sentence. This check is not perfect yet and may be refined in the future, but it should do as a start.

I went though all sentences of all languages with at least 10,000 sentences and counted the last symbols. Sentences in languages with less than 10,000 sentences (as of 7th of March 2026) are not checked at all.

```
function isLastCharacterUsual($s, $lang)
{
    $last = mb_substr($s,mb_strlen($s)-1);
    
    if (in_array($lang,['eng','rus','ita','epo','kab','deu','tur','fra',
                        'ber','por','spa','hun','heb','nld','ukr','fin',
                        'lit','pol','ces','mar','tat','mkd','tgl','tok',
                        'dan','swe','lat','srp','ina','tlh','ron','lfn',
                        'vie','slk','ind','bul','oci','swc','shi','hau',
                        'nds','nob','lvs','kor','bel','ido','nnb','isl',
                        'aze','ile','gos','nno','cat','kmr'])) {
        return mb_strpos(".?!\"”':…)»“",$last)!==false;
    }
 
    if (in_array($lang,['jpn','cmn','yue'])) {
        return mb_strpos("。？！」…）：",$last)!==false;
    }
 
    if (in_array($lang,['ara','pes','ckb'])) {
        return mb_strpos(".؟!\")",$last)!==false;
    }
    
    if (in_array($lang,['hin','ben','asm'])) {
        return mb_strpos("।?!.\"",$last)!==false;
    }
    
    if ($lang=='tig') {
        return mb_strpos("።:?.!፧፥፡",$last)!==false;
    }
 
    if ($lang=='hye') {
        return mb_strpos("։:",$last)!==false;
    }
 
    if ($lang=='ell') {
        return mb_strpos(".;!\"'”“»:…)",$last)!==false;
    }
 
    if ($lang=='yid') {
        return mb_strpos(".?!״”‟\":",$last)!==false;
    }
 
    if ($lang=='zgh') {
[...]

PR #3265, opened 2026-03-07
Add function isCorrectLastCharacter

kumakyoo42:
This is a start on issue #3264.

I thought it might be a good idea to add this one already, although it is not used yet. I plan to add some more of such functions (checking the first symbol, checking all symbols of a sentence, checking some typical mistakes like space before comma and so on). Eventually they can be used to implement the real check.

I changed the function a little bit compared to the function given in the issue: It now returns the offending character instead of false. This might be useful in warning messages.

I'm unsure if the place I've choosen (`Model/Table/SentenceTable`) is a good place for these functions. If you prefere some other place, please just tell me.

### DIFF
--- a/src/Model/Entity/Sentence.php
+++ b/src/Model/Entity/Sentence.php
@@ -155,4 +155,82 @@ class Sentence extends Entity
     {
         return $this->correctness == \App\Model\Table\SentencesTable::MIN_CORRECTNESS;
     }
+    
+    /**
+     * Checks if the last character of a sentence is a typical character in
+     * that sentence's language. Assumes that the sentence is not empty.
+     * 
+     * @param string $sentence The sentence to be checked.
+     * @param string $lang Three-letter-code of the language of the sentence.
+     * 
+     * @return true|string Returns true, if the last character is judged to be OK,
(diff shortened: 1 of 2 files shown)


jiru:
Thank you :slightly_smiling_face: 

I would like to suggest moving the check method to `src/Model/Entity/Sentence.php`.

You could do something like
```php
class Sentence extends Entity
{
    public function isCorrectLastCharacter() {
        $sentence = $this->sentence;
        $lang = $this->lang;
        // do your thing... 
    }
}
```

And then
```php
class SentenceTest extends TestCase
{
    public function testIsCorrectLastCharacter() {
        $entity = new Sentence(['lang' => 'eng', 'text' => 'Test this']);

        $result = $entity->isCorrectLastCharacter();

        $this->assertFalse($result);
    }
}
```

And later we can make it a model validation rule [like this](https://book.cakephp.org/3.x/orm/validation.html#using-entity-methods-as-rules).

kumakyoo42:
I didn't understand that part about the model validation rule, but that can be kept for later...

jiru:
Merged as 96df45eae84a5cb4ab21f42c12591e868b6795af

Issue #3267, opened 2026-03-08
If a link appears twice in a wallpost or a comment the html output is damaged


kumakyoo42:
Enter 'A http://example.com and another http://example.com.' at the wall.
The HTML now looks like this:  `A <a href="http://example.com" target="_blank" rel="nofollow"><a href="http://example.com" target="_blank" rel="nofollow">http://example.com</a></a> and another <a href="http://example.com" target="_blank" rel="nofollow"><a href="http://example.com" target="_blank" rel="nofollow">http://example.com</a></a>.`

Expected: `A <a href="http://example.com" target="_blank" rel="nofollow">http://example.com</a> and another <a href="http://example.com" target="_blank" rel="nofollow">http://example.com</a>.`

Issue #2430, opened 2020-07-05
Links ending with tabs are not clickable


jiru:
1. Go to the Wall.
2. Type of copy a link.
3. Insert a tab character right after the link (easiest way is to  copy-paste it from a text editor or a terminal)
4. Post the message: https://dev.tatoeba.org/fra/wall/show_message/31708#!#message_31708

**Expected behavior**
The link should be turned into a clickable link after posting.

**Screenshots**
![image](https://user-images.githubusercontent.com/5107734/86526438-afc5cf80-bec6-11ea-9b18-316363641257.png)

Issue #1161, opened 2016-05-23
Links on the Wall, and perhaps other places, that include ) are wrong.


ckjpn:
##The Problem

The trailing ) gets included as part of the URL, resulting in a NOT FOUND error when clidked.
(https://dev.tatoeba.org/eng/), 

https://tatoeba.org/eng/wall/index#message_26466

![screen shot 2016-05-23 at 09 03 44](https://cloud.githubusercontent.com/assets/7934273/15457514/aeb1d7ea-20c5-11e6-8d58-0e7c727c77c5.png)

ckjpn:
Perhaps this is related, or at least could be corrected at the same time.

https://dev.tatoeba.org/eng

This URL when put on the Wall, doesn't show up correctly on the "home" page.

This is from the "dev" site.
![screen shot 2016-05-23 at 09 18 01](https://cloud.githubusercontent.com/assets/7934273/15457632/dfca71fa-20c7-11e6-8006-d94477a403df.png)

trang:
Side effect of #1103.

I think it's time we find a library for this. Some possible candidates here:
http://stackoverflow.com/questions/37684/how-to-replace-plain-urls-with-links

jiru:
I wonder if parenthesis should be considered as part of links or not. I often use and see Wikipedia links ending with a closing parenthesis, such as [https://en.wikipedia.org/wiki/Mercury_(planet)](https://en.wikipedia.org/wiki/Mercury_(planet)).

ckjpn:
Note that this bug still happens.
Here's a wall post from today.

https://tatoeba.org/eng/wall/show_message/31490#message_31490

ckjpn:
This still is a problem, at least in private messages.

Someone sent me this screenshot and reported this to me today.

- URL is malformed when you don't put space next to parenthesis
<img width="269" alt="image" src="https://user-images.githubusercontent.com/7934273/146617433-632eb4b0-dd07-4103-bc41-bf51e4d2018f.png">

Issue #956, opened 2015-12-24
Sometimes the "Latest Wall Message" on the "Home" page show clickable links, and sometimes they don't.


ckjpn:
https://tatoeba.org/eng

![screen shot 2015-12-24 at 09 44 08](https://cloud.githubusercontent.com/assets/7934273/11987467/57801040-aa23-11e5-9cba-b9e56daa74b1.png)

ckjpn:
The # in front of a sentence number link doesn't work either.

![screen shot 2015-12-25 at 14 01 35](https://cloud.githubusercontent.com/assets/7934273/12000014/5ca53ea4-ab10-11e5-8a1f-b5e4c6949326.png)

ckjpn:
Here's another example that may help you figure out why this happens.

![screen shot 2016-01-04 at 11 21 39](https://cloud.githubusercontent.com/assets/7934273/12082424/dfcf3894-b2d5-11e5-877f-22e166907b0d.png)

ckjpn:
It's happened again.

![screen shot 2016-01-23 at 23 52 28](https://cloud.githubusercontent.com/assets/7934273/12530772/b8bafd40-c22c-11e5-85f6-23b7fc46a555.png)

jiru:
Note: direct link to the mentioned messages:
* Not clickable: https://tatoeba.org/jpn/wall/show_message/25283#message_25283
* Clickable: https://tatoeba.org/jpn/wall/show_message/25275#message_25275

ckjpn:
I tested it again with the same URL, and it worked.  (I've since deleted the 2nd post, since it wasn't really needed, so it's not there now.  However, here's a screenshot.

![screen shot 2016-01-24 at 09 04 03](https://cloud.githubusercontent.com/assets/7934273/12533637/fc8265b6-c279-11e5-88df-8ede94bc1e42.png)

ckjpn:
It's happened again with this message.  https://tatoeba.org/eng/wall/index#message_25309

![screen shot 2016-01-28 at 19 27 36](https://cloud.githubusercontent.com/assets/7934273/12641875/9494007a-c5f5-11e5-8d21-518092601030.png)

ckjpn:
Yet another example.  Perhaps having multiple examples of the problem will help you figure out this bug.

https://tatoeba.org/eng/wall/index#message_25454

[...]

PR #3269, opened 2026-03-12
Add new version of URL checking

kumakyoo42:
Solves issues #3267, #1161 and #2430. It also touches issue #956, but does not solve this one.

I checked the links given by @trang in issue #1161 and several other url parsers in the net, but none was convincing for me. So I wrote yet another parser. I based it on the formal syntax of an URL given in [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986), but decided to simplify it a lot. For example it allows for URLs starting with http:// or https:// only, it does not allow for addition of a username, it does not check ? and # in the last part of the URL.

The old version tried to check for a correct top level domain, but that check was broken anyway and IMHO it is nowadays impossible to add such a check, so I discarded this. But I made sure that at least one subdomain is present (the old version tried to do this too but actually failed).

Links will only be recognized if opening and closing parenthesis do match; this ensures that links in parenthesis are threated correctly.

I had to rewrite two tests, because they guarded agains unwanted behavior, namly including a closing parenthesis in a link without a corresponding opening parenthesis.


tests/TestCase/View/Helper/MessagesHelperTest.php
```diff
@@ -46,13 +46,33 @@ public function formatContentProvider() {
...
+            [ 'https://a-b',
+              'https://a-b' ],
+            [ 'https://a-b.com',
+              '<a href="https://a-b.com" target="_blank" rel="nofollow">https://a-b.com</a>' ],
+            [ 'https://example.com/abc%20(def%20()))',
+              '<a href="https://example.com/abc%20(def%20())" target="_blank" rel="nofollow">https://example.com/abc%20(def%20())</a>)' ],
+            // The following url contains a white space early ending it and thus
+            // contains unbalanced parenthesis; it therefore should not be recognized
+            [ 'https://example.com/abc( )',
+              'https://example.com/abc( )' ],
```
jiru, line 67: Shouldn’t it be recognized as
```html
<a href="https://example.com/">https://example.com/</a>( )
```
?
I imagine one could write something like `According to this definition https://en.wiktionary.org/wiki/example(scroll down to Verb section)`.

src/View/Helper/ClickableLinksHelper.php
```diff
@@ -69,54 +103,69 @@ public function clickableURL($text)
...
+            
+            // the parsed URL is stored in two pieces: $url and $tail
+            // when parsing is complete, the symbols in $tail can be removed
+            // while the symbols in $url are fixed. This prevents the trailing
+            // ; from an entity to be removed, while being able to remove
+            // other trailing ;
+            //
+            // $par counts the nummer of opened paranthesis to ensure they
+            // are balanced
+
```
jiru, line 128: Feel free to name your variable in a way that makes this comment obsolete, for example `$nbOpenedParenthesis`.

jiru:
We should be using some linter, but for now could you make sure that:
* comparison operators (== < > <= >= etc.) are surrounded by spaces
* commas are followed by space or line return
* if statements span on at least two lines ("if-condition line" and "instruction line" separated)

tests/TestCase/View/Helper/MessagesHelperTest.php
kumakyoo42: First of all, I think people would normally add a space between "example" and the paranthesis in which case everything works.

If you try to accept this type of link, the parser would get much more complicated, because recognition of the link happens much later - at the space character and there could be other things in between, which might complicate thing much more, think of `https://en.wiktionary.org/wiki/example(somelongwordwith&nbsp;(entieties)&nbsp;and additional paranthesis)` In this case the parser would end at the space after the "and" and would need to backtrack to the correct opening paranthesis.

So I would prefere to leave this as is; its a really rare case anyway.

kumakyoo42:
I already tried to do that, but obviously I overlooked some places. (My personal style differs much from the one used here.) Anyway, I didn't know about the spaces around comparison operaters.

src/View/Helper/ClickableLinksHelper.php
```diff
@@ -44,22 +44,58 @@ class ClickableLinksHelper extends AppHelper
 {
     public $helpers = array('Html', 'Pages');
 
-    const URL_PATTERN = '/((ht|f)tps?:\/\/([\w\.]+\.)?[\w-]+(\.[a-zA-Z]{2,4})?[^\s\r\n"\'<]+)/siu';
     const SENTENCE_ID_PATTERN = '/([\p{Ps}：\s]|^)(#([1-9]\d*))/';
+    const START_OF_URL = '|^https?://[[:alnum:]~_-]+\.[[:alnum:]~_-]+|';
```
jiru, line 48: Could you add back support for FTP links? While usage of FTP links almost disappeared now, it was still a thing back in the days, and used in [some](https://tatoeba.org/fr/sentences/show/485311#comment-21226) [old](https://tatoeba.org/fr/sentences/show/485360#comment-23202) [comments](https://tatoeba.org/fr/sentences/show/485351#comment-23168).

tests/TestCase/View/Helper/MessagesHelperTest.php
jiru: I agree that it’s a reasonable compromise.

For the record, here are a few instances of such links, which will stop working:
* https://tatoeba.org/fr/sentences/show/701012#comment-120948
* https://tatoeba.org/fr/sentences/show/1801094#comment-195756
* https://tatoeba.org/fr/sentences/show/2001361#comment-217572
* https://tatoeba.org/fr/sentences/show/2558814#comment-270839
* https://tatoeba.org/fr/sentences/show/7104618#comment-1522361
* https://tatoeba.org/fr/sentences/show/13030811#comment-1523396

src/View/Helper/ClickableLinksHelper.php
```diff
@@ -153,7 +208,7 @@ function ($m) use ($self, $model) {
     public function hasClickableLink($text)
     {
         $patterns = array(
-            $this::URL_PATTERN,
+            $this::START_OF_URL,
```
jiru, line 211: I think you are introducing a small regression here, because the mere presence of START_OF_URL pattern does not always mean the text contains a link that will be linkified eventually. (It could be that some parenthesis is preventing linkification.)

jiru:
I have checked your new implementation against a subset of existing comments, and I observed that it solved many closing-parenthesis issues. Thank you very much!

I realized there is one minor regression: your implementation does not support [international domain names (IDN)](https://www.w3.org/2003/Talks/0425-duerst-idniri/slide12-0.html) (see [examples of IDNs](https://www.w3.org/2003/Talks/0425-duerst-idniri/slide12-0.html)). So for example if I paste `https://räksmörgås.josefsson.org/` in a comment, the current implementation makes it a usable link that points to `https://räksmörgås.josefsson.org/`, but your implementation does not make it a link. The easy workaround is to directly use punycode instead (that is to say http://xn--rksmrgs-5wao1o.josefsson.org/), and it looks like that’s what Tatoebians have been [doing](https://tatoeba.org/fr/sentences/show/3493570#comment-394088) so far anyway.

I think the PR is good to merge.

src/View/Helper/ClickableLinksHelper.php
```diff
@@ -44,22 +44,58 @@ class ClickableLinksHelper extends AppHelper
...
+        }
+    
+        $last = $url[strlen($url)-1];
+        
+        if ($last == ')' && $par < 0) {
+            return $this->checkTail(substr($url, 0, strlen($url)-1), $par+1, ')'.$tail);
+        }
+        
+        if (in_array($last, ['!', '.', ',', ';', ':'])) {
+            return $this->checkTail(substr($url, 0, strlen($url)-1), $par, $last.$tail);
```
jiru:
```php
substr($foo, 0, strlen($foo)-1)
```
can be simplified to
```php
substr($foo, 0, -1)
```
here and in other places of this PR

tests/TestCase/View/Helper/MessagesHelperTest.php
kumakyoo42: Also for the record: The first four of them also do not work with the current implementation. They are linked, but the link points to the wrong place. I can't check the last two, but I guess it's the same there.

src/View/Helper/ClickableLinksHelper.php
kumakyoo42: I thought about this too. My reasoning to omit FTP was, that these links do high likely point to a place that does not exist anymore so a clickable link would be of no help. If I try for the links in the comments you gave, my browser only shows a blank page, but other browsers might still work.

But I don't care much about this. If you insist, I'll add FTP.

src/View/Helper/ClickableLinksHelper.php
kumakyoo42: I'm aware of that. This function is only used at one place, namely in the function which produces a short version of the wall post on the starting page of Tatoeba. This function is broken anyway, see issue #956 and the change will just be, that breaking occurs at a slightly different place. Fixing #956, which probably makes this function obsolete, might be a better way to handle this (might be one of my next projects, but I don't know for sure yet).

kumakyoo42:
> I realized there is one minor regression: your implementation does not support international domain names (IDN)...

And it does not support IPv6 and no user names and no ports. It's a question of how much effort we want to put into separating valid URLs from invalid ones. If we want to accept IDNs, I think we are down to: Starts with http(s)://, ends at the next whitespace (or shortly before that) and contains at least one dot. Not sure, if that is better.

src/View/Helper/ClickableLinksHelper.php
kumakyoo42: Ha, PHP meanwhile supports negative indices. I didn't know that. :-)

PR #3271, opened 2026-03-13
Add function isCorrectFirstCharacter

kumakyoo42:
This is a continuation on PR #3265, this time for the first character of a sentence.

As this would cause terrible merge conflicts when merged after PR #3265 I didn't base this against `dev` but against PR #3265.

In several languages a decimal digit will not be accepted, although it might be correct at the beginning of a sentence. The reason, why I excluded them, is just, that I could not verify if it is correct or not, because there where to few examples in the corpus of that language. If I add them and they should not be added, it is very unlikely that this will be noticed, while it is much more likely the other way round.


kumakyoo42:
I changed the numbers to hex numbers because that's more convenient when looking them up in unicode tables.

kumakyoo42:
While preparing the third part I found some mistakes here, which I corrected.

src/Model/Entity/Sentence.php
```diff
@@ -155,4 +155,205 @@ protected function _getIsUnapproved()
...
+     * 
+     * @return true|string Returns true, if the last character is judged to be OK,
+     *                     and the offending character in all other cases.
+     */
+    public function isCorrectLastCharacter() 
+    {
+        $sentence = $this->sentence;
+        $lang = $this->lang;
+        
+        $last = mb_substr($sentence, mb_strlen($sentence)-1);
```
jiru, line 174: can be simplified to
```php
$last = mb_substr($sentence, -1);
```

jiru:
Thank you, this is gonna be super useful.

About the return value. Your methods return the offending character in case it is suspicious, but that’s not really useful because it can be easily computed by the caller (just check what’s the first or last character). I think your methods should return the list of expected first or last characters instead, so that the caller can display a useful message like "English sentences are expected to finish with "?", "!" or ".", but your sentence does not. Are you sure it’s not a mistake?"

> I changed the numbers to hex numbers because that's more convenient when looking them up in unicode tables.

Hex is convenient to lookup, but inconvenient to read and makes it difficult to understand what characters are used in the code. Could we somehow keep both the string form and the hex form? For example, use the string form in the code instruction, and add the hex form as a side comment?

kumakyoo42:
> About the return value. Your methods return the offending character in case it is suspicious, but that’s not really useful because it can be easily computed by the caller (just check what’s the first or last character).

In earlier versions I returned just a boolean value. In case of this method (and the one about the last character) it doesn't matter much. But with the other two methods (the fourth one will contain some regexes which should not match like "/\s,/"), knowing, which characters/substrings are the problem might be important and much more difficult to be recalculated by the calling function. I just wanted to have the same signature for all of these functions. (But actually the third one returns an array of wrong characters, while these only return one character. Maybe they should also return an array?) And yet another idea would be to add the file position.

Maybe it would be an idea to have a separate function, which returns the set of correct characters for each function? Moving these long arrays to class level might be a good idea anyway. This way, the calling function would have the possibility to get both informations.

What do you think?

kumakyoo42:
> Hex is convenient to lookup, but inconvenient to read and makes it difficult to understand what characters are used in the code. Could we somehow keep both the string form and the hex form? For example, use the string form in the code instruction, and add the hex form as a side comment?

It's risky, because comment and code might diverge in the future, when someone changes the one and forgets about the other. While some ranges might be clearly understandable, others might not, for example the part about hiragana would look like this:  `'あ', 'い', 'う', 'え', ['お', 'ぢ'], ['つ', 'も'], 'や', 'ゆ', ['よ', 'ろ'], ['わ', 'を']`. Could you tell from that, if 'ん' is allowed as first character or not? And if some Japanese people tell us, that there are cases where 'ん' is a correct first character of a sentence, do you know where to add it in this list?

An other problem will be the various spaces (they are not relevant for the first character but for the next two functions they are). You might not be able to understand the constants in the code anymore, because they are not visible, and there it's high likely that someone in the future accidentially replaces one without recognizing.

So maybe I could add some comments defining the set, e.g. for hiragana it would be something like: "All normal sized hiragana, with the exception of ん."?

jiru:
> So maybe I could add some comments defining the set, e.g. for hiragana it would be something like: "All normal sized hiragana, with the exception of ん."?

I understand, that’s a good compromise. :+1:

> > About the return value. Your methods return the offending character in case it is suspicious, but that’s not really useful because it can be easily computed by the caller (just check what’s the first or last character).
> 
> In earlier versions I returned just a boolean value. In case of this method (and the one about the last character) it doesn't matter much. But with the other two methods (the fourth one will contain some regexes which should not match like "/\s,/"), knowing, which characters/substrings are the problem might be important and much more difficult to be recalculated by the calling function. I just wanted to have the same signature for all of these functions. (But actually the third one returns an array of wrong characters, while these only return one character. Maybe they should also return an array?) And yet another idea would be to add the file position.
> 
> Maybe it would be an idea to have a separate function, which returns the set of correct characters for each function? Moving these long arrays to class level might be a good idea anyway. This way, the calling function would have the possibility to get both informations.
> 
> What do you think?

I think the most practical choice is to stick with signatures similar to CakePHP’s validation classes, because in the end we are going to use them to validate and display errors. CakePHP’s validation classes basically check validity using a callback that returns a boolean to indicate if the field is valid or not, and optionally a localized error message. A single field can have one or more errors.

This means we might as well have your validation function directly produce/return a human-readable error message. This error message could be formatted from other "intermediary" data (such as the position of the error or the set of allowed characters), but the point is that we need it as a string, and we need it to be clear and easy to understand for the user.

Issue #3246, opened 2026-02-10
Normal contributors should be able to unlink links they created themselves


kumakyoo42:
When contributing to Tatoeba it's normal to make mistakes from time to time. Let's take [sentence 6157455](https://tatoeba.org/de/sentences/show/6157455) as an example. This sentence is a question - the greek question mark looks like a semicolon. But assume, you overlooked this and translated the sentence to "You don't have school today.", which would be the correct translation, if the sentence ended with a dot. There are several things, that might happen:

1. "You don't have school today." is not yet in the database. In this case a completely new sentence is created. You could correct it, if you know the correct translation, because it doesn't have any other translation but the original greek sentence. If you realise at this time, you do not know the correct translation you are at a loss. You can't delete this sentence, because it has a translation and of course you can't unlink. All you can do is to ask someone to remove the sentence or to remove the link. Or (which I did in such situations), replace the sentence with 'xxx', which indirectly is a deletion, because this sentence is typically removed quite soon.
2. "You don't have school today." is in the database and you own it yourself. This is similar to the case above, but typically you can't correct the sentence, because it has already translations which would get wrong, when changing the sentence. Once again, all you can do is to ask for help or to replace this sentence by "xxx" accepting that some more translations will be broken, when this sentence is deleted.
3. "You don't have school today." is in the database and you do not own it. This is the worst case, because you can't change anything. Just beg for help.

When this happend to me the first time, I actually thought of leaving Tatoeba altogether, because I felt quite bad; it took two days until I came up with the "xxx" workaround which is OK, but still does not feel good.

All in all, it feels like you have to type some text and you need to ask for permission, whenever you need to press backspace. 

**Idea**
Well, all these problems would vanish, if normal contributors were able to undo the linking which they did themselves. This is not to be confused with unlinking links of others (which, as far as I understand, advanced contributors can do). This of course needs some trust in the user. But undoing what you did yourself can't do any harm, does it?

kumakyoo42:
> Because I assume new users are likely to think of removing their sentence as a way to undo their mistake.

Worse: They will "correct" the sentence to make it fit the one translation, overlooking that they break other translations while doing so. A warning might be a good idea in such cases.

I meanwhile had a look into the source code on unlinking. I think it's rather complicated to implement my idea, because right management does not know about translations. Might also slow down everything - not sure about this, though.

[...]

PR #3272, opened 2026-03-14
Normal contributors should be able to unlink links they created themselves

kumakyoo42:
Adresses issue #3246.

This works now, but I still have to add some tests.


kumakyoo42:
I decided to stop contributing to Tatoebas codebase. Meanwhile I tried several times to add some tests, but I have to give up here too - I don't understand the code, I've got no idea where to start, and this is systematic, because it repeats over and over again. I'm sorry, but I can't continue.

jiru:
Thank you for your valuable contributions, I can take over your opened PRs. It looks like you have been struggling a lot with Tatoeba’s code and I feel a bit sorry for that. Maybe this particular PR is a bit too hard, because it is highly tied to CakePHP, which I know is not your cup of tea. :wink:

Issue #3067, opened 2023-07-10
Update to CakePHP 4 series (or later)


Guybrush88:
**Story** or **Context**
I was browsing the CakePHP repository to take a look, and I stumbled upon this page: https://github.com/cakephp/cakephp/releases/tag/3.10.5

Basically speaking, the CakePHP 3 series (so the current one that is used by Tatoeba) won't be maintained anymore by its developers, so maybe it could be taken into consideration, for Tatoeba, to upgrade to the 4 series (or later), so it will be also maintained on upstream, and maybe, it could be more attractive for new developers for Tatoeba itself.

jiru:
CakePHP 5 is out.

2colours:
I think the main aim here should be to move to the latest PHP version. Basically there is no way forward without doing that.

cblanken:
Based on the [version map](https://github.com/cakephp/cakephp/wiki) it probably makes sense to upgrade to CakePHP 4.x first before trying to get PHP up to date since 3.x doesn't support anything after PHP 7.4 anyway.
![image](https://github.com/Tatoeba/tatoeba2/assets/19908880/248d556c-28c0-4335-be1d-1668d6923690)
 

Just for reference, I ran the codebase through [PHP Compatibility](https://github.com/PHPCompatibility/PHPCompatibility) and here are the results for checks with PHP 8.0 and 8.4
- [PHP 8.0](https://gist.github.com/cblanken/80d2490659d5d82665317bba6aafd475)
- [PHP 8.4](https://gist.github.com/cblanken/6fb5bd33debc243e89dfd875c616de81)

PR #3284, opened 2026-04-28
Upgrade CakePHP to 4.4

jiru:
Solves #3067 and adds some more unit tests. Also fixes 4.x deprecations so that updating to 5.x should be much easier when the time comes.

Bonus changes we get for free:
* Validation error messages set in our models automatically end up as HTML form validation popup message, for example in Users we do `->notEmptyString('email', __('Field required'))` and the form will have something like `<input type="email" name="email required="required" ...>`. If the field is left empty, the browser used to display a built-in popup (e.g. Firefox says "Please fill out this field."), but now it displays the localized "Field required" message there instead.

Things to do during deployment:
- [ ] Update POT files link to new location on https://www.transifex.com/
- [ ] Add cookie key to `app_local.php`
- [ ] Changes references to `src/Locale` into `resources/locales` in https://github.com/Tatoeba/tatoeba2/wiki/Deployment


jiru:
Note to self: rename `tests/TestCase/Model/Exporter/PairsExporter.php` after this is merged.

PR #3318, opened 2026-07-07
Add language badge to sentence page. Fixes #2611

### LINKED ISSUE #2611: When seeing an unknown flag, it’s hard to tell the language
**Story**
When I look at a group of sentences (main sentence + translations), I often see flags that I don’t know. I sometimes ask myself "I wonder what’s that language". From a computer, I can mouse-hover the flag to display the language name. That’s fine but I think it is rather difficult to guess for a newcomer. What’s more, I cannot do that from a phone. So the most obvious thing to do is to click on the (i) button to go to the sentence page. However, from there, again it is quite hard to find what’s the language: I have to look at the page title.

**Ideas**
1. On all sentence pages, clearly display the name of the language in which the sentence is.
2. On all flags, use modern tooltip that works by tapping on phones.

Davd-B:
image
Adding language name badge in sentence page.

Fixes #2611

### DIFF
--- a/templates/element/sentences/license.php
+++ b/templates/element/sentences/license.php
@@ -31,6 +31,11 @@ use App\Model\CurrentUser;
     <h2><?php echo __('Sentence text') ?></h2>
 
 <?php
+/* Language name of the sentence text */
+echo '<div style="margin-bottom: 12px;">' . format(
+    __('Language') . ': {}',
+    $languageName
+) . '</div>';
 echo format(
     /* @translators: placeholder is the name of the license
        of the sentence text, could be CC BY 2.0 FR or CC0 */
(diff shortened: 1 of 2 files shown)


templates/element/sentences/license.php
```diff
@@ -31,6 +31,11 @@
     <h2><?php echo __('Sentence text') ?></h2>
 
 <?php
+/* Language name of the sentence text */
+echo '<div style="margin-bottom: 12px;">' . format(
+    __('Language') . ': {}',
```
jiru, line 36: Please make the colon translatable too. Some languages may use a different format such adding a space before the colon. You could use `__('Language:')` since it is already used in `templates/element/advanced_search_form.php`.

templates/element/sentences/license.php
```diff
@@ -31,6 +31,11 @@
     <h2><?php echo __('Sentence text') ?></h2>
 
 <?php
+/* Language name of the sentence text */
+echo '<div style="margin-bottom: 12px;">' . format(
```
jiru, line 35: This HTML markup is not consistent because you wrap the language name into a `<div>` here but the `__('License: {}')` text below is not wrapped at all. How about wrapping them both with `<ul>` and `<li>`?

PR #3320, opened 2026-07-10
Adding 'Delete my account' section in user profile settings. Fixes #3278

### LINKED ISSUE #3278: Make it Easy to Find the "Delete Account" link
Add a link to https://en.wiki.tatoeba.org/articles/show/delete-account in settings or another logical place a person who wants to delete his/her account can easily find the link.

From a person who recently asked to have his account deleted.

"Furthermore, the fact that the unsubscribe section is so hidden isn't a good thing. Could you please let all the admins know? "

Davd-B:
image

Fixes #3278

### DIFF
--- a/templates/User/settings.php
+++ b/templates/User/settings.php
@@ -443,18 +443,18 @@ $this->set('title_for_layout', $this->Pages->formatTitle(__('Settings')));
             </div>
         </md-toolbar>
 
-        <div layout-padding style="font-size: 0.95em; line-height: 1.5;">
-            <p style="color: #c62828; font-weight: bold; margin-bottom: 16px;">
-                <?php
-                echo __('Before asking to delete your account, make sure to remove all personal information from your profile. Please note that your contributions on the site will still be visible.');
-                ?>
-            </p>
+        <div layout-padding>
+            <?php
+            echo $this->Html->div('experimental-info',
+                __('Before asking to delete your account, make sure to remove all personal information from your profile. Please note that your contributions on the site will still be visible.')
+            );
+            ?>
 
-            <p style="margin-bottom: 16px;">
+            <p>
                 <?php
                 $adminsUrl = $this->Url->build('/private_messages/write/TatoebaAdmins');
                 echo format(
-                    __('To delete your account, send a private message to <a href="{url}" target="_blank">TatoebaAdmins</a> with the title "Account deletion".'),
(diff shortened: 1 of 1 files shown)


templates/User/settings.php
```diff
@@ -435,4 +435,40 @@ class="md-primary">
...
     </section>
+
+    <section class="md-whiteframe-1dp">
+        <md-toolbar class="md-hue-2">
+            <div class="md-toolbar-tools">
+                <h2><?php echo __('Delete my account'); ?></h2>
+            </div>
+        </md-toolbar>
+
+        <div layout-padding style="font-size: 0.95em; line-height: 1.5;">
```
jiru: Can I ask why is this style needed here?

templates/User/settings.php
```diff
@@ -435,4 +435,40 @@ class="md-primary">
...
+            <p style="color: #c62828; font-weight: bold; margin-bottom: 16px;">
```
jiru: Consider using `class="warning"` instead to keep the design consistent with other parts of the website displaying warnings.

templates/User/settings.php
```diff
@@ -435,4 +435,40 @@ class="md-primary">
...
+                <?php
+                echo __('Before asking to delete your account, make sure to remove all personal information from your profile. Please note that your contributions on the site will still be visible.');
+                ?>
+            </p>
+
+            <p style="margin-bottom: 16px;">
+                <?php
+                $adminsUrl = $this->Url->build('/private_messages/write/TatoebaAdmins');
+                echo format(
+                    __('To delete your account, send a private message to <a href="{url}" target="_blank">TatoebaAdmins</a> with the title "Account deletion".'),
```
jiru: I don’t think we need `target="_blank"` here and below.

templates/User/settings.php
```diff
@@ -435,4 +435,40 @@ class="md-primary">
...
+        <div layout-padding>
+            <?php
+            echo $this->Html->div('experimental-info',
+                __('Before asking to delete your account, make sure to remove all personal information from your profile. Please note that your contributions on the site will still be visible.')
+            );
+            ?>
+
+            <p>
+                <?php
+                $adminsUrl = $this->Url->build('/private_messages/write/TatoebaAdmins');
```
jiru, line 455: Please use an array to build the URL, not a string ([example](
https://github.com/Tatoeba/tatoeba2/blob/0fbb4506ce071827f85375bd4c8c8f9c7cf23d41/templates/element/vocabulary/menu.php)).

PR #3323, opened 2026-07-10
Creative commons logo in SVG - Fixes #3323

### LINKED ISSUE #3319: Creative commons logo in SVG
The Creative Commons logo at the bottom of the mainpage appears pixelated on my hand.

I suggest replacing it with this SVG version

https://commons.wikimedia.org/wiki/File:Cc.logo.circle.svg

Davd-B:
Switched Creative Commons logo to its SVG version

Before
image
After
image

Fixes #3319


templates/element/foot.php
```diff
@@ -201,7 +201,7 @@
     </div>
 
     <div layout="row" layout-align="center center" class="license">
-        <img alt="Creative Commons License" src="/img/cc-logo.png" />
+        <img alt="Creative Commons License" src="/img/cc-logo.svg" width="36" height="36" style="width: 36px; height: 36px;" />
```
jiru: I don’t think we need this extra CSS, the `width` and `height` attributes should be enough.

PR #3327, opened 2026-07-12
New profile design

Davd-B:
New profile design (redesign)

image

image

Fixes #3326


templates/User/profile.php
```diff
@@ -260,10 +146,131 @@
...
+                            'action' => 'write',
+                            $username,
+                        ]);
+                    ?>
+                    <a href="<?= $url ?>" class="contact-link">
+                        <md-icon>
+                            email
+                        </md-icon>
+                        <div class="contact-text">
+                            <?= format(__('Message user')) ?>
```
jiru: No need to call `format()` here.

templates/User/profile.php
```diff
@@ -260,10 +146,131 @@
...
+                    <div class="settings-summary" flex>
+                        
+                        <!-- Email notifications -->
+                        <div class="settings-item">
+                             <div class="settings-item-text">
+                                <?php
+                                if ($notificationsEnabled) {
+                                    echo format(__('Email notifications: {}.'), '<strong class="setting-highlight">' . __('enabled') . '</strong>');
+                                } else {
+                                    echo format(__('Email notifications: {}.'), '<strong class="setting-highlight">' . __('disabled') . '</strong>');
```
jiru: You can make this code more [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself) with something like:
```php
$status = $this->Html->tag(
    'strong',
    $notificationsEnabled ? __('enabled') : __('disabled'),
    ['class' => 'setting-highlight']
);
echo format(__('Email notifications: {}.'), $status);
```

templates/User/profile.php
```diff
@@ -260,10 +146,131 @@
...
+                        </div>
+
+                        <!-- Profile visibility -->
+                        <div class="settings-item">
+                            <div class="settings-item-text">
+                                <?php
+                                if ($isPublic) {
+                                    echo format(__('Visibility: {}.'), '<strong class="setting-highlight">' . __('public') . '</strong>');
+                                } else {
+                                    echo format(__('Visibility: {}.'), '<strong class="setting-highlight">' . __('other members only') . '</strong>');
```
jiru: Same comment here about DRY code.

templates/User/profile.php
```diff
@@ -260,10 +146,131 @@
...
+                    <?php
+                    if ($username == $currentMember):
+                        $editSettingsUrl = $this->Url->build(array(
+                            'controller' => 'user',
+                            'action' => 'settings'
+                        ));
+                    ?>
+                        <div class="settings-action" flex="none">
+                            <a href="<?= $editSettingsUrl ?>" class="change-choices-link">
+                                <?= __('change choices') ?>
```
jiru: Maybe "edit settings" is clearer than "change choices".

templates/User/profile.php
```diff
@@ -313,17 +324,24 @@
             ?>
             <md-divider></md-divider>
             <?php
+            $tagOptions = array(
+                'class' => 'profileDescription',
+                'escape' => false
+            );
+
+            if ($isEmptyDescription) {
+                $tagOptions['class'] .= ' empty';
```
jiru: This code can be simplified. You can move this line up to the `if` above like this:
```php
        if (!empty($userDescription)) {
            /* ... */
        } else {
            /* ... */
            $tagOptions['class'] .= ' empty';
        }
```

and remove `$isEmptyDescription` entirely.

templates/User/profile.php
```diff
@@ -298,13 +306,16 @@
         ?>
 
         <?php
+        // this code changes description alignement (centered if it's empty, left if it contains informations)
```
jiru: This comment is not accurate because it does not describe the entire block; only what you are adding in this PR. Please either remove the comment or make it describe the whole block accurately.

webroot/css/user/profile.css
```diff
@@ -36,12 +36,182 @@
...
+    overflow: hidden;
+    border-radius: 15px;
+    border: 1px solid #e3e3e3;
+}
+
+#profile .profile-picture img {
+    width: 100%;
+    height: 100%;
+    object-fit: cover;
+    transform: scale(1.05);
```
jiru, line 52: I see you are scaling up the picture by 5%, can you elaborate on this change? I am concerned that it is actually hiding part of the image.

templates/User/profile.php
```diff
@@ -260,10 +146,135 @@
                     ?>
                 </div>
 
+                <!-- member since + contact -->
+                <div class="info-group">
+                    <div class="member-since">
+                        <strong><?= __('Member since') ?></strong> <?= $userSince ?>
```
jiru, line 152: `<strong>` means the content is important. I don’t think we need `<strong>` here.

webroot/css/user/profile.css
```diff
@@ -36,12 +36,182 @@
...
+    margin-right: 0;
+    margin-left: 8px;
+}
+
+html[dir="rtl"] #profile .actions .button-text {
+    padding-right: 0;
+    padding-left: 12px;
+}
+
+#profile .change-choices-link {
```
jiru, line 134: Now that the text changed, could you rename this CSS class too?

webroot/css/user/profile.css
```diff
@@ -36,12 +36,182 @@
...
+    color: #4CAF50;
+    text-decoration: none;
+    font-weight: 500;
+    cursor: pointer;
+}
```
jiru, line 139: This whole block looks unnecessary because these rules are already set to these values.

webroot/css/user/profile.css
```diff
@@ -36,12 +36,182 @@
...
+
+#profile .change-choices-link:hover {
+    text-decoration: underline;
```
jiru, line 142: This rule makes this particular link inconsistent with every other link on Tatoeba (on hovering, text color changes to orange and there is not text decoration).

templates/User/profile.php
```diff
@@ -117,133 +103,33 @@
...
-                            <?php /* @translators: profile edition button on profile page (verb) */ ?>
-                            <?= __('Edit') ?>
-                        </md-button>
-                        <?php
-                    }
-                    ?>
-                </div>
-
+            <!-- Personal info -->
+            <div class="info" flex>
```
jiru, line 128: Here (and elsewhere in that page) you are both using the `flex` directive (HTML attribute) and `display: flex` in the CSS. This is redundant. On each tag, please either use a CSS rule or a directive.

The directive is [something provided by AngularJS Material](https://material.angularjs.org/latest/layout/introduction) that dynamically adds the CSS for you.

templates/User/profile.php
```diff
@@ -260,10 +146,135 @@
...
+                                ?>
+                            </div>
+                        </div>
+
+                        <!-- Language settings (Optional) -->
+                        <?php if (!empty($languagesSettings)): ?>
+                            <div class="settings-item">
+                                <md-icon class="material-icons-outlined">
+                                    translate
+                                </md-icon>
```
jiru, line 250: I think this icon is rather confusing. I know the current design does not provide any label or icon, and this setting should be reworked anyway. My point is that this icon suggests something like "translation" or "multilanguage", which are (in my opinion) misleading compared to the real meaning. I think no icon is better than a misleading icon. Alternatively, we could have a label such as "Only seeing translations in:"

webroot/css/user/profile.css
```diff
@@ -36,12 +36,182 @@
...
+#profile .settings-item-text {
+    padding-top: 1px;
+}
+
+#profile .settings-item-text.languages-text {
+    color: rgba(0, 0, 0, 0.54);
+}
+
+#profile .setting-highlight {
+    color: #387dff;
```
jiru, line 192: This color feels unnecessary to me.

templates/User/profile.php
```diff
@@ -260,10 +146,135 @@
...
+                        <div class="settings-item">
+                            <div class="settings-item-text">
+                                <?php
+                                $status = $this->Html->tag(
+                                    'strong',
+                                    $isPublic ? __('public') : __('other members only'),
+                                    ['class' => 'setting-highlight']
+                                );
+                                
+                                echo format(__('Visibility: {}.'), $status);
```
jiru, line 240: I think the wording of the label "Visibility" could be improved. It is not obvious what visibility we are talking about. I suggest: "Profile visibility".

